PR #1, opened 2026-07-22
mutability-conditionals 1/7: data/compute kinds, conditional collections (coproducts), Case-arm joins, kind inference

sortalongo:
## Data/compute function kinds, conditional collections as dependent sums, Case-arm joins, and kind inference

PR 1/7 of the mutability-conditionals stack. Lays the **type-system foundation** for
conditional return values, conditional mutation, and conditional feeds, and makes the
data/compute function kind an inferred solver variable. Nothing user-visible compiles
differently yet; this PR makes the type system able to *describe* a control-flow join of
collections without silently losing data.

### Type-system foundation

- **`FunKind` on `Type::Fun`** — every function type records whether it is a *compute*
  function (`⇒`, a capability; lossy meet is fine) or a *data* function (`⤇`, an extent;
  joins must be lossless). `Fun{Data} <: Fun{Compute}` is the safe upcast; `Compute <: Data`
  is the rejection.
- **Conditional collections are real dependent sums (`Type::Sigma`).** A control-flow join
  of data functions with distinct extents forms the dependent sum `Σ{Dᵢ} ⤇ V`
  (`Type::conditional_collection`) — one shared body over a witness, whose candidate domains
  are the branch extents (contribution order, a tested contract), sharing the joined
  codomain. The witness is genuinely load-bearing: it *is* the runtime branch discriminant,
  referenced anonymously in the body's domain position (nullary `Type::Witness`) and
  **projected at elimination** (the value-`Case` fan-out). The witness taxonomy follows the
  classification hierarchy **values : types : kinds** and is built to cover every collection
  form up front (no future refactor), with machinery only for the conditional case:
  `SigmaType { witness: Witness, body }`, `Witness::Type(Kind)` (type-witness, anonymous) |
  `Witness::Value { ty, binder }` (value-witness `List`/`Map`, named — reserved), and
  `Kind::Enumerated(Vec<Type>)` (conditional, wired today) | `Kind::Any` (`Collection[T]`'s
  universe — reserved). Constrain rules: two are genuine subtyping — `Fun{Data} <: Σ`
  **injection** (a branch is-a sum; matches a candidate domain by value) and `Σ <: Σ`
  **width** (a smaller sum is-a bigger sum, by value) — and one is **elimination**, not
  subsumption: `Σ <: Fun` (a Σ value is one branch, not a function total on the union;
  consuming it distributes the consumer over the witness — the discriminated union of
  candidate extents, a fresh consumer domain absorbing it, a concrete narrower extent
  rejected, so a conditional collection never silently narrows). It rides the `<:` solver
  because that is the plumbing for consumption, but the docs and comments keep it distinct
  from the two real subtyping rules. Coalesce forms the Sigma from ≥2 surviving data-domain
  alternatives (`union_extents`), collapsing to a plain data fun at one survivor
  (idempotence).
- **Case-arm lattice joins** — `emit_case` types arms with a fresh var + `require_sub` (was
  strict `require_eq`), so `[1,2] if c else [1,2,3]` types as `Σ (w ∈ {[0,1],[0,2]}). w ⤇ Int`
  and `{a,b} if c else {a,c}` joins to the common-field intersection `{a: Int}` instead of
  failing.

### Kind inference (`FunKind` as a solver-resolved variable)

- **`FunKind::Var`** — a third kind, a solver cell resolved at coalesce. Construction sites
  mint a fresh kind var instead of hand-marking; an unconstrained var resolves from its
  domain (a map/finite-extent domain → `Data`, else `Compute`), so the map-carries-Data rule
  replaces the ~30-site introduce-time obligation.
- **`Compute <: Data` rejection** — surfaced at coalesce: a var left `forced_data` over a
  provably-scalar domain (or `forced_compute ∧ forced_data`) is the same violation as the
  concrete case, reported as `ComputeWhereDataRequired` (e.g. `sum(f)` for a plain
  `Int ⇒ Int` lambda `f`). `freshen_kind` mirrors a kind var's `<:` links onto each
  instantiation, so a use-site force still reaches a linked sibling.
- **Live `⤇` display** — a data collection renders `⤇`, a capability (and an unresolved
  kind var) `⇒`. One residual quirk: a comprehension body over a let-bound source can render
  `⇒` though it denotes a collection (its extent domain is reconstructed only post-coalesce).

### Consumption rejected cleanly
The `Σ <: Fun` elimination rule types a conditional collection *at the type level*, so it
flows down the pipeline, where the value-`Case` (not eliminable until 2/7) is rejected at
**lambda elimination** — a returned compile error, never a panic and never a silent
miscompile.

### Docs
`src/ccl/design/type-inference.md` §4.6 (conditional collections as dependent sums);
`CLAUDE.md` symbolic vocabulary + kind rendering. Green under `./ci.sh`.

Continues cambra-dev/cambra-old#301 (review history there).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,7 +109,8 @@ Vocabulary, matching `src/ccl/symbolic.rs`:
 
 Types (from `Display for Type` in `src/ccl/mod.rs`):
 
-- **Function**: `(T ⇒ U)`.
+- **Function**: `(T ⇒ U)` for a compute function (capability domain); `(T ⤇ U)` (plain-text `|=>`) for a data function (extent domain — the domain is the data map, joins must be lossless). Kind is inferred by the solver (`FunKind::Var`, resolved at coalesce — see `src/ccl/design/type-inference.md` §4.6) and `Display` renders it live: a data collection shows `⤇`, a capability (and an unresolved kind var) `⇒`. One benign quirk — a comprehension *body* over a **let-bound** source (its extent domain is only reconstructed after coalesce, so the kind var resolves to the conservative `⇒`) can display `⇒` though it denotes a collection: `let x = [1,2,3] in [y + 10 for y in x]` renders `⇒`, while the same comprehension over a literal source renders `⤇`. The denotation is identical; only the displayed arrow differs (aggregate uses still force `Data`, so nothing miscompiles).
+- **Sigma** (in-flight, same stack): `Σ{D0, D1} ⤇ V` — a data function whose extent is exactly one of the listed choices; with a live witness binder, `(Σ n ∈ {D0, D1}. n ⤇ V)`.
 - **Refinement**: `{T | predicate}` — predicate is rendered via `symbolic`.
 - **UIntRange**: `[0, N]`, or `∅` when empty.
 - **Hole**: `_`.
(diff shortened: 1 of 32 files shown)


sortalongo:

PR #2, opened 2026-07-22
mutability-conditionals 2/7: value-Case compilation via literal union-of-restricts

sortalongo:
## Value-Case compilation via literal union-of-restricts

PR 2/7 of the mutability-conditionals stack. With the type system able to describe a
conditional collection as a real dependent sum (the Σ from 1/7), this PR **compiles**
value-selecting conditionals. The unifying idea: a total value selection is a genuine union
of restricted arms `⧺ᵢ (eᵢ ↾ π̂ᵢ)` (first-match predicates `π̂ᵢ = gᵢ ∧ ¬⋁ⱼ<ᵢ gⱼ`), not a
fused select builtin — lazy arm evaluation falls out for free and the union carries its own
extent.

### What changes

- **Scalar value-Case C-form** — a scalar/compute-typed `x = a if p else b` compiles to
  gated scalar lifts over a `Units(1)` driver plus `final_or_default`; the extraction is
  inserted by the rewrite itself. One new operator-level primitive: a constant-predicate
  `Restrict` (dedicated test).
- **Data-typed value-Case gate fan-out** — `xs if c else ys` compiles to the gated union
  `⧺ᵢ (eᵢ ↾ π̂ᵢ)`, whose structural type is a data function over a discriminated-union domain
  `Variant([{Dᵢ | π̂ᵢ}]) ⤇ V`. The fan-out is **Σ-agnostic**: it stamps the union with the
  type inference gave the `Case` (a Σ for distinct extents, or its same-extent collapse to a
  plain data function), so lambda-elim stays type-preserving (Σ in, Σ out).
- **Source-less conditional feeds** — a guard-only source-less feed Case channelizes to gated
  scalar-lift channels rather than dying with `PartialFeedCaseUnsupported`; a scrutinee/pattern
  feed stays rejected.
- **Comprehension conditionals** — comprehension *over* a conditional collection, a conditional
  *element* in a comprehension, and a conditional *between* comprehensions, plus a clear error
  for the deferred per-element case-float case that is not yet wired.
- **if/else (both arms feed) in a pure-feed loop** — the generator lowering emits the `else` as
  the trailing `true`-guarded arm; the loop-source fan-out refines the *source* domain per arm
  into disjoint refined-source channels with a shared codomain.
- **Off-path laziness** — the scalar C-form never evaluates an off-path arm: `MapResultToConst`
  skips *pulling* the constant when all driver rows are deleted, so a partial off-path
  expression (e.g. `total // x` under `x != 0`) never faults. Pinned by a dedicated test.

### How the compiled form types against the Σ

The reconciliation between the compiled fan-out and the Σ is **not** an ad-hoc "bridge rule" —
it is the finite-Σ = gated-coproduct isomorphism, `Σ(w∈{Dᵢ}). w⤇V ≅ ⊕ᵢ (Dᵢ⤇V) ≅
Variant([{Dᵢ|π̂ᵢ}]) ⤇ V`. A function over a disjoint-union domain is normally a *product* of
fibers (total on all), the wrong type; the exclusive+exhaustive gates zero out all-but-one leg
(empty = unit), collapsing the product to the coproduct. So the wall check is **Σ-introduction**
(covariant — an introduction, not the function-domain contravariance): the compiled gated
partition realizes the whole Σ.

- **distinct extents** — the gated partition subtypes the Σ (`Fun(Data) <: Σ`), sound iff the
  legs' base extents, *as a set*, equal the Σ candidates (every fiber realized, nothing extra).
  This corrects a positional leg↔candidate bijection an earlier draft used, which rejected a
  valid `Case` whose branches *share* an extent (e.g. `[1] if a else [2] if b else [3,3]` — two
  length-1 legs, one Σ candidate). A dedicated repeated-extent test locks it.
- **same-extent collapse** — the Σ collapses to a plain data function `D ⤇ V`, and the gated
  partition subtypes that directly (`is_index_partition_of`).
- **consumption** — a conditional collection consumed as a plain collection is the `Σ <: Fun`
  elimination rule from 1/7.

The gates' exclusivity+exhaustiveness — the precondition the isomorphism rests on — is a
`lambda_elim` construction invariant (asserted at the fan-out boundary: a non-exhaustive
value-`Case` would realize an empty collection on the uncovered path, a silent miscompile),
not re-proven in the solver. Op-conversion compiles the union straight to a `UnionOperator`
from its operands (concrete extents), never consulting the Σ, so no Σ reaches a runtime
operator.

### Docs
`src/ccl/design/mutability.md` (value selection graduates out of "Not yet implemented"; `⧺` as
both spec and operational form); `src/interpreter/design-operators.md` (constant-predicate
Restrict); `src/ccl/design/type-inference.md` §4.6 (Σ-introduction).

Green under `./ci.sh`. Conditional *induction writes* (3/7) and path-based `with begin():`
conditionals (4/7) build on this.

Continues cambra-dev/cambra-old#303 (review history there).

### DIFF
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -4,7 +4,8 @@ use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
 
 use crate::ccl::{
-    BaseType, Builtin, Expr, Lit, Name, PredicateId, Refinement, Type, TypedExprNode,
+    BaseType, BinOpKind, Branch, Builtin, Expr, Lit, LogicKind, Name, PredicateId, Refinement,
+    Type, TypedExprNode, UnaryOpKind,
 };
 
 /// Returns `true` if `expr` directly references the given built-in primitive.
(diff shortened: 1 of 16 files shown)


sortalongo:

PR #3, opened 2026-07-22
mutability-conditionals 3/7: conditional induction writes

sortalongo:
## Conditional induction writes

PR 3/7 of the mutability-conditionals stack, on value-Case compilation (2/7). Delivers
conditional mutation inside an induction loop — `for x in xs: if p: total += x` — the
`mutability.md` motivating example. The model: a conditional write becomes **one
carry-complete writer** over a `Tile::Store` changelog (the write arm on `{r | π̂}`, the
complement arm carrying the snapshot forward), folded to a single commit-gated decision;
the union appears only in history slots, where the existing guarded-slot grammar already
admits it.

### What changes

- **Lower if-in-loop to a statement-Case** (`lower/loops.rs`), with `collect_mutation_loop_vars`
  recursing into branches (and correctly excluding inner `for` / `with begin()`).
- **Induction accumulators as a `Tile::Store` changelog** — induction routes through the
  changelog `InductionStore` and recognition folds the conditional write to one
  commit-gated writer (multi-writer dispatch is never born).
- **Value-Case inside a writer lambda** — the C-form from 2/7 applied inside a decision
  body, evaluated lazily (the off-path arm is not forced).
- **Uniform carry-complete write decision** — every path (write or carry) produces a full
  decision record, so trailing reads see one merged view.
- **`Tile::Store`: separate terminality from the watermark** — a store's "no more writes"
  signal is decoupled from its position watermark, so a conditional (sparse) writer bounds
  correctly and a loop whose tail carries reads its true final accumulator.
- **Commit unconditional writes beside a conditional** — an unconditional write *before* a
  conditional (`x := x+1; if p: x := x+10`) now commits on every position (the commit gate
  compares the carried expression against the entering snapshot at compile time; a
  differing carry forces commit), fixing a silent-drop where the unconditional increment
  was lost.

### Docs
`src/ccl/design/mutability.md` (conditional induction writes graduate to the changelog
realization); `src/interpreter/design-operators.md`; the induction-unification design plan
and the unordered-mutability design commitment; `docs/plan.md`.

Green under `./ci.sh`.

Continues cambra-dev/cambra-old#304 (review history there).

### DIFF
--- a/src/ccl/design/lowering.md
+++ b/src/ccl/design/lowering.md
@@ -70,17 +70,15 @@ The design of record for the recurrence construction and feed routing is
 implementation lives in `src/ccl/mut_elim.rs` (the `For`/`MutWrite` → `LetRec`
 → `Transact` via loop planning) and `src/ccl/channelize.rs` (feed routing).
 
-**Conditionals in loop bodies (designed, not yet implemented).** Today a mutation
-nested inside an `if` in a loop body is detected (`find_nested_mutation_var`) and
-**rejected** at lowering. The in-flight conditionals stack *will* add a
-`ChlStmt::If` arm to the direct-mirror loop-body lowering: each branch body will
-lower through the same recursive statement-chain helper and be emitted as a
+**Conditionals in loop bodies.** A mutation nested inside an `if` in a loop body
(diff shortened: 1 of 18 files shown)


sortalongo:

PR #4, opened 2026-07-22
mutability-conditionals 4/7: path-based with begin(): conditionals

sortalongo:
## Path-based `with begin():` conditionals

PR 4/7 of the mutability-conditionals stack. Extends transactional mutability from a single
bare deny guard to full conditional structure inside a `with begin():` block —
`if`/`elif`/`else` routing across keys — while keeping **one decision record per
transaction** (the soundness verdict from 2/7: txn path predicates read the commit-tick
snapshot, so per-path transaction *sites* would violate one-record atomicity).

### What changes

- **Path-based routing** (`transact_phase`) — the path walk threads `(path, env)` with
  read-your-writes, resolves guards against the incoming env, and rejoins each written key
  as a carry-forward Case (`env[k] = Case[a₁ → env₁[k]; …; true → snapshot]`). `commit`
  is the disjunction of the paths that wrote. These merge Cases compile via 2/7's C-form.
  Lowering simply stops rejecting elif/else/multi-if.
- **Carry conditionally-written keys via read-set finalization** — `collect_footprint`
  threads an `in_case` flag and adds any key written inside a statement-`Case` arm to the
  transaction's read set at collection time. Because all `Var` reads accumulate
  unconditionally regardless of path, a key read on only one path cannot be missed, and
  α-unique names mean a binder can never shadow a store key. A carried key is then seeded
  into the incoming env and reads its prior committed value correctly.
- **Per-tap fire field** — a routed feed carries a `to_<defer>__fire: Bool` gate so a tap
  fires only on its own path; the engine omits non-fired tap keys from the commit delta
  (no garbage in the durable log). Omitted when path == commit, preserving today's shape.
- **Conditional feed in a read-only begin() block** — a conditional reply inside a
  read-only block routes to the channelize feed fan-out; the both-arms `if`/`else` case
  works because the pure-feed fan-out (2/7) handles the trailing `true`-guarded arm.
  Two tests pin the single-arm and if/else forms.

### Semantic change (documented)
The deny guard is no longer transaction-scoped — a spine write beside an `if` commits
unconditionally (`commit = p ∨ true`). Audited: no existing test regresses; a new test
pins the new meaning. `with t = begin():` (handle binding), nested begin, loops in a
block, and guarded induction writes in a block stay rejected via a user-facing pre-check.

### Docs
`src/ccl/design/mutability.md` ("General in-transaction conditionals" graduates into the
main model; path-scoped deny semantics; fire-field mechanism; write-mask future work);
`src/interpreter/design-operators.md` (CommitWriter tap-fire decode). The per-tap fire field
and read-set finalization are documented in `src/ccl/design/mutability.md`.

Green under `./ci.sh`.

Continues cambra-dev/cambra-old#305 (review history there).


sortalongo:

PR #6, opened 2026-07-22
mutability-conditionals 6/7: variant elimination — VariantProject/VariantWrap sum algebra

sortalongo:
## Variant elimination: the point-free sum algebra

PR 6/7 of the mutability-conditionals stack. Adds the `VariantProject`/`VariantWrap` tile
operators and compiles a scrutinee-`Case` over a `Type::Variant` to a point-free union of
tag-restricts — the prerequisite sum algebra the decision-variant rework (7/7) builds on.

### VariantProject op + canonical Unit column

Add the `VariantProject` tile operator, the read-dual of `VariantWrap` and the
elimination primitive for a scrutinee-Case over a `Type::Variant`. Given a
`Scalar(Union)` scrutinee it narrows to arm `i`'s tag-restricted sub-domain and
yields that arm's inner payload column, keyed by the original UInt position.

Restrict and project are one operation: `ColumnValue::Union` already stores each
arm's payloads densely (in appearance order), so reading `variants[i]` IS the
tag-`i` restriction. There is no separate boolean Restrict step and no
`Predicate::Union` — a domain-Restrict could not express tag discrimination
anyway, since the tag lives in the scrutinee's codomain, not its domain.

Also canonicalize `ColumnValue::single(Unit)` to `Units(1)` (was the Variants
catch-all): `from_values`/`empty_tile` already produce `Units` for a Unit extent, so
the old singleton representation collided ("mismatched variants" in
append/merge) when a Unit-payload variant value (`.Abort()`) was fanned through a
Memo during elimination.

### Scrutinee-Case as union of tag-restricts

Compile a scrutinee-Case over a variant (`lambda x -> match x { Ci(wi) -> ei }`)
to the point-free union of tag-restricts, the sum-elimination dual of
`VariantCtor` construction:

  `match d { Ci(wi) -> ei }  ~>  (+)i ( d ▷ variant_project(i) ▷ (wi -> ei) )`

The branch tags partition the scrutinee's domain, so this is the same fan-out
the value-Case emits, keyed on tag rather than a boolean first-match gate. Each
arm narrows the fed scrutinee stream to arm `i`'s tag partition (`variant_project`),
binds the payload, and maps `ei`; the flat `UnionOperator` re-totals the disjoint
partitions. Exhaustiveness (one arm per scrutinee tag, enforced by inference's
width-subtyping) makes the union total, so no `final_or_default` collapse is
needed — the fan-out shape, not the C-form.

- `Builtin::VariantProject(usize)`: minted post-inference (carries no scheme; the
  CHECK-mode typecheck trusts its stamped type), composed as a bare builtin onto
  the fed scrutinee. op-conversion reads the union extents off the fed input's
  `Scalar(Union)` tiling and builds the `VariantProject` tile op.
- `lambda_elim` scrutinee-Case arm beside the value-Case fan-out.
- `simplify`: `try_const_reduce` refuses to collapse past a `variant_project`-led
  compose head (as it already does for `filter_values`) — the tag restriction is
  a runtime domain narrowing carrying no refinement to re-materialise, so
  dropping it for a const arm would apply the constant at every position.

### Arm bodies may close over the outer binder (zip)

Extend the scrutinee-Case compilation so an arm body can read the outer binder
as well as its payload — the Phase B per-key-view shape
`λ __c → match __c.decision { Commit(w) → {time: __c.time, write: w.i} }`, whose
arm reads both the record's sibling field (`__c.time`) and the Commit payload (`w`).

`lambda_elim` detects the dependence structurally (the arm body has the outer
binder free beyond the payload binder) and, only then, zips the whole element
alongside the projected payload:

  `⧺ᵢ ( ⟨id, x.f ▷ variant_project(i)⟩ ▷ zip ≫ (λ (x, wᵢ) → eᵢ) )`

The two binders merge into one pair (`x ↦ pair.0`, `wᵢ ↦ pair.1`), reusing the
nested-lambda rewrite. The payload-only path is unchanged. For the outer `id` arm
and the tag-restricted payload to co-iterate, `VariantProject` now preserves the
scrutinee's real domain keys for a union *stream* (`SealedFunction{D ⇒ Scalar(Union)}`),
and the zip's `FanIn` inner-joins the full outer element with the tag-restricted
payload on shared keys.

### VariantWrap point-free constructor (dual of VariantProject)

Complete the point-free sum algebra: a `VariantCtor` inside a lambda now elaborates
to a composable morphism `param_ty ⇒ Union`, so it can sit as the RHS of a `≫` — the
shape the value-Case-in-lambda fan-out `⧺ᵢ (filter_values(π̂ᵢ) ≫ eᵢ)` needs for a
writer decision `if p: acc := .Commit(…) else: acc := .Abort(unit)` (used by 7/7).

`Builtin::VariantWrap(usize)` is the exact dual of `Builtin::VariantProject(usize)`:
`elim_lambda`'s `VariantCtor` arm rewrites `.Cᵢ(eᵢ(p))` ⟹ `eᵢ ≫ variant_wrap(i)`;
op-conversion consumes the fed payload stream and builds the `VariantWrap` tile,
which now also accepts a `SealedFunction{D ⇒ Scalar(payload)}` input (the scalar
`Scalar(Union)` path is unchanged). `ColumnValue::repeat` now handles a singleton
Union (broadcasting a constant `.Abort(unit)` over a stream). No `simplify` guard is
needed: `variant_wrap` is a total element-wise map, so const-reduce dropping it would
be sound.

### Tests
Op-level (both arms project correctly; key-preservation + zip alignment over a stream
keyed by arbitrary UInts) and full-pipeline (channelize → lambda_elim → op-conversion →
drive): two-arm match firing each arm, the one-arm read-side shape, the `{time, write}`
per-key-view over a two-row stream, and the in-lambda writer-decision shape
`λ p → Case[p == 0 → .Commit(p + 1); true → .Abort(unit)]` reading back Commit(1) at
position 0 and Abort(unit) at position 1.

Continues cambra-dev/cambra-old#307 (review history there).


sortalongo:

PR #7, opened 2026-07-22
mutability-conditionals 7/7: writer decision as Commit|Abort variant; retire F_COMMIT

sortalongo:
## Writer decision as Commit|Abort variant (dense); retire F_COMMIT

PR 7/7 of the mutability-conditionals stack. Replace the Boolean-tagged decision product
`{commit, writes, to_<defer>*}` with the variant `[.Commit(⟨writes, taps⟩) | .Abort]`,
end-to-end, using the point-free `VariantWrap` constructor and `VariantProject` eliminator
(6/7). The whole-transaction grant/deny is now the variant tag, so the illegal
"denied yet real writes" state is unrepresentable.

The Commit payload is dense: an unwritten key on a committing path carries
its snapshot value, and a routed reply keeps its `to_<defer>_k__fire` gate
inside the payload (partial writes and `__fire`-folding are deferred — see the
`mutability.md` Future-work note).

- Construct (both phases funnel through `ccl_utils::wrap_decision_variant`):
  `transact_phase::build_writer` and `mut_elim::transform_loop` wrap the
  assembled `{commit, writes, taps}` record into the variant Case.
- Typing: `emit_transact_writer`'s decision codomain becomes the variant;
  `writer_tap_fields` / op-conversion's `body_tap_fields` peel the Commit payload.
- Reads eliminate the variant one-arm: the txn `per_key_view` is a point-free
  zip of `⟨time, decision ≫ variant_project(Commit) ≫ .writes.idx⟩`; the
  induction writes/acc/feed views insert `variant_project(Commit)` before the
  field read.
- Recognition (`planning/loops.rs`) peels the Commit payload for feed fields
  and matches the `variant_project` read prefix.
- Runtime `body_decision_at` decodes the `Value::Union` tag — `.Abort` => deny
  (carry, no proposal), `.Commit` => the payload's writes + taps.
- `letrec` causal-slot grammar accepts a zip of causal slots (the `per_key_view`).
- `F_COMMIT` retired from the decision codomain and every decode/recognition
  read; it survives only as the phase-internal `COMMIT_SELECTOR` field of the
  pre-variant intermediate record (`ccl_utils`). `F_FIRE_SUFFIX` kept (deferred).

### Docs
`mutability.md` / `design-operators.md`: present the writer decision as
`[.Commit(dense writes, taps) | .Abort]`; the illegal-state cost is now
unrepresentable and the whole-transaction commit flag is retired. Correct the
stale notes on the earlier eager `Select` (retired in 2/7) and the "partial op
faults off-path" NYI (fixed by the lazy `filter_values` union-of-restricts, 2/7).
Add a Future-work note: partial writes and folding `__fire` into payload presence
are deferred (both need a dense per-key presence encoding — a naive sparse column
is dropped by the record zip's inner-join).

### Tests
- **Off-path safety** (`if d != 0: acc := acc // d`): proves no off-path divide-by-zero
  fault. Off-path safety depends on the guard reading the value it protects — the lazy
  `filter_values` union never evaluates the off-path arm, so a guard that reads the
  protected register (`d != 0`) suppresses the fault. (An item-only guard that does not
  read the divisor offers no protection and correctly still faults; item-only guards
  otherwise commit and deny normally.)
- **Interleaved writers** (§9(a)): two writer sites on one register interleave
  Commit/Abort through the commit-Store; `get_prev_txn` reads back the conserved value.

Continues cambra-dev/cambra-old#308 (review history there).


sortalongo:

PR #8, opened 2026-07-22
Compound (tuple/record) mutable registers (induction + transactions)

sortalongo:
## Compound (tuple / record) mutable registers

A follow-up on top of the mutability-conditionals stack (based on 7/7, #308). Makes
**tuple- and record-valued mutable registers** work end-to-end, on both the induction
(`for` loop) and transactional (`with begin():`) paths. Found while reviewing the stack:
a compound accumulator previously **panicked** at op-conversion.

```python
acc := (0, 0)
for i in [1, 2, 3]:
    acc := (acc[0] + i, acc[1] + i)
acc                       # was: panic  →  now: (6, 16)

p: Mut[(int, int), Txn] := (100, 0)
for r in [10, 20]:
    with begin():
        p := (p[0] - r, p[1] + r)
# ... trailing read           →  (70, 30)
```

### Root cause

A mutable register holds one `Value`, so a compound accumulator is stored **boxed** — a
`Scalar(Record)` codomain (one column of record values) — while a tuple/record *literal*
compiles to a **struct-of-arrays** `Record` tiling (a column per field). The two
representations met *unreconciled* at the register boundaries, so the runtime tripped an
assert (a `debug_assert` in `ExtractLast`; hard asserts elsewhere).

The interpreter already has both converters — `scalar_tile_to_column_value`
(`Tile::Record` → boxed `ColumnValue::Records`) and its inverse `column_value_to_tile`
(unbox to a declared tiling shape). They just weren't applied at these seams.

### The fix

Four representation seams (each a small application of the existing helpers):

- **`ExtractLast::new`** — the assert compared tiling *structure*, but `get_impl` already
  boxes both source codomain and default via `scalar_tile_to_column_value`; the real
  invariant is equal *extent*, so match on that.
- **`read_initial_scalar`** (the changelog / commit-store init seed) — only matched
  `Tile::Scalar`; box a `Tile::Record` init into one record `Value`. Shared by the
  induction and commit stores.
- **`flat_merge`** (the conditional-write decision-value merge) — box each arm's codomain
  in, then rebuild the merged codomain to the operator's **declared** tiling shape via
  `column_value_to_tile` (a scalar field stays `Scalar`; a compound field unboxes to
  struct-of-arrays). This also fixes `if`/`else`-both-write, which otherwise emitted a
  boxed codomain against a struct-of-arrays declared tiling.

Plus two **type-annotation** forms in `lower_type_annotation`, the only thing blocking the
transactional variant (`Mut[(int, int), Txn]` / `Mut[{x: int}, Txn]`) — the store
machinery was already compound-ready:

- `ChlExpr::Tuple` → `Type::Tuple`
- `ChlExpr::Record` → `Type::Record`

### What works now (all tested)

Induction *and* transactions, over: unconditional writes, single-arm / `if`-`else` /
`elif` conditional writes (incl. deny), read-your-writes across fields, field projection
off the final value, heterogeneous tuples `(int, str)`, N-tuples, **nested** tuples
`((i, i), i)`, trailing-carry conditionals, named records, 3-field records, and mixed
scalar + compound multi-key stores (atomic per commit).

18 tests: 12 induction cases in `mutability.rs`, 6 transactional cases in `transactions.rs`.

### Still not supported (graceful, not a panic)

`for i in []` (empty-loop) yields a `Incompatible lower bounds` type error — but this is
**identical for a scalar register**, so it's an orthogonal empty-list inference issue, not
compound-related.

Green under `./ci.sh` (both clippy passes + doc + full suite).

Continues cambra-dev/cambra-old#309 (review history there).

### DIFF
--- a/src/ccl/lower/comprehension.rs
+++ b/src/ccl/lower/comprehension.rs
@@ -275,9 +275,9 @@ fn float_comp_source_case(source: Expr, iter_var: &str, body: &Expr) -> Expr {
     // chain Phase 5 builds. The compose form is equivalent (a map applies the
     // body to each element) but `emit_compose` stamps it with the *source's*
     // data kind (a map over a collection is itself a collection). That data kind
-    // is what lets two such arms of a conditional source `sigma_join` into the Σ
-    // (compiled by the gate fan-out) rather than colliding as compute-kinded
-    // lambdas whose distinct index extents would meet.
+    // is what lets two such arms of a conditional source join into the
+    // coproduct (`union_extents`, compiled by the gate fan-out) rather than
+    // colliding as compute-kinded lambdas whose distinct index extents would meet.
     Expr::compose(vec![
         source,
         Expr::lambda(iter_var, Type::Hole, body.clone()),
--- a/src/ccl/lower/stmts.rs
+++ b/src/ccl/lower/stmts.rs
@@ -898,6 +898,22 @@ pub(super) fn lower_type_annotation(annotation: &Spanned<ChlExpr>) -> Result<Typ
             )),
         },
         ChlExpr::Lit(ChlLit::None) => Ok(Type::Base(BaseType::Unit)),
+        // A tuple type annotation `(A, B, …)` — e.g. a compound register value
+        // type `Mut[(int, int), Txn]`. Each element lowered recursively.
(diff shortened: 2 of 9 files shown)


sortalongo:

PR #10, opened 2026-07-23
mutability-conditionals 1/7: data/compute function kinds and kind inference

dpmills:
Introduces `FunKind` (`Compute` / `Data`) — whether a `Type::Fun`'s domain is a *capability* or a collection's *data* — and states the two rules that distinction exists to buy: **a data function's domain is invariant**, and **the join of two data functions is a Σ over their candidate domains**.

`FunKind` is a **provenance** property, not a function of the domain: the same domain can back a collection or a capability (`Map(Color, V)` vs `Color ⇒ V`), so it is stamped at introduction — list literals, `++`, registered sources, comprehensions, `groupby`, recurrence carriers and every `History` erasure are `Data`; builtins and ordinary lambdas are `Compute`. The previous domain-shape heuristic is deleted. A `FunKind::Var` is minted only where the kind is genuinely inferred (a parameter, a freshened scheme), and the solver resolves it at coalesce over the two-point lattice `Data ⊑ Compute`, defaulting to `Compute` when unconstrained.

## Data domains are invariant

A compute function's domain is contravariant, and rightly so: it is a parameter, nothing in the language can ask a capability which inputs it accepts, and accepting more than demanded only under-promises. A data function's domain is not like that, and the reason is not a slogan about the domain "being the data" — it is that the domain is **observable**, in two ways a record's absent field is not.

- It is **the loop bound the emitted program runs**. Op-conversion builds an iteration source as `IterateExtent::new(extent_of(D))` from the *static* domain. Handing an 11-row collection to a slot declared `[0,5] ⤇ V` does not forget rows the way `{a: Int, b: Int} <: {a: Int}` forgets `b`; it emits a program that reads six of them and reports the result as the collection's.
- It is **reproduced in eliminator results**. A comprehension has the shape `D ⤇ A ⇒ D ⤇ B`, so `D` occurs covariantly as well as contravariantly at application, and a variable in both positions is invariant by the ordinary variance calculus. A compute domain occurs contravariantly only, because nothing enumerates it — which is the entire content of the `Data`/`Compute` split, stated so it can be checked rather than asserted.

Invariance is spelled the only order-independent way available: **both domain edges, unconditionally**, when both kinds are concretely `Data`. Any rule conditioned on what a domain looks like *at the moment the edge fires* — whether it is still a variable, whether its refinement has arrived — would make typing depend on constraint emission order. Invariance is a property of two types, not of when they are compared.

Both directions is also all it takes. `Type::UIntRange` relating only by equality already rejects both base directions, and refinement *drop* is already rejected one step less obviously, since behind a contravariant domain it demands `D <: {D | p}`. What the reverse edge adds is the one case that inversion left admitted: refinement *acquisition*, `D ⤇ V <: {D | p} ⤇ V`, an unfiltered collection standing where a filtered domain is declared.

The rule is stable under a surface data domain, which matters because `Array(n, T)` is planned and any argument resting on "a data domain is unwritable today" would rot. Both things a contravariant reading would buy are better bought elsewhere, and the syntax is what makes them cheap: "works for any length" is *quantification* (`∀n. Array(n, T) ⇒ …`, an ordinary scheme the solver already freshens per use), and a deliberate prefix is a *term*, visible where it happens at a use-site-chosen extent, rather than a coercion hidden in a declaration at whatever width that declaration names.

## Their join is a Σ

Two collections over different domains are incomparable, so their least upper bound is a different element of the lattice: the dependent sum `Σ (w ∈ {Dᵢ}). w ⤇ V` over the candidate domains, whose witness is the runtime branch discriminant and which is eliminated by distributing the consumer over it. The lattice is **incomplete** without Σ, and that incompleteness — not a semantic rejection — is what a domain join currently runs into. Subtyping, joining, and the collapse case where the candidates turn out to be one domain (`xs if c else xs`) are then one model rather than three rules.

Σ is not yet representable, so the join has no answer to produce and **diagnoses** instead. That is acceptable only because it is loud: the alternative to a Σ is an error, not a silent miscompile.

The same fact surfaces at two phases, because a join can be forced from either. When a consumer imposes a concrete demand the domains meet at a `Fun`/`Fun` edge and the invariant domain rule reports it there (`ConstrainError::DataDomainMismatch`, carrying the two domains, since the codomain is never what disagrees). When nothing forces it — the joined collection is the program's own value, or is only let-bound — the candidates ride to coalesce as accumulated alternatives and are reported there (`CoalesceError::DomainJoinConflict`). Both are live from source, neither is redundant, and both render the same "collection domain conflict" wording because they are one fact.

That two-phase structure is forced. At the compact merge a positive `Data ⊔ Data` accumulates its domain **alternatives** (`union_domains` — union and dedup, never a meet); it cannot decide there whether they are really two domains, because a compact domain still carries inference-variable identity that `simplify_type` may merge afterwards. Coalesce materializes each to a `Type` and deduplicates again, and that second comparison is the one that decides.

The consequence for the collections work is recorded in the doc rather than left to be discovered: **candidate domains arrive at a domain variable, not only at a `Case` result**, so Σ has to satisfy both sites. A domain position is not privileged — Σ is the join wherever it arises.

## Kind-aware subtyping

The `Fun`-vs-`Fun` arm carries a kind edge over `Data ⊑ Compute`: `Data <: Compute` upcasts, a concrete `Compute <: Data` is rejected (`ConstrainError::ComputeWhereDataRequired`), and a kind var picks up `forced_compute`/`forced_data` flags whose disagreement is the coalesce-time face of the same error. A capability demanded as data — `sum(λ x → x + 1)` — is caught right at the edge with no domain inspection, which is why the old domain-shape guess is gone. The rejection is emission-only: the post-inference check runs a kind-blind `ConstrainCache`, because elimination preserves denotation but not kind representation and a kind-aware re-check would false-reject.

## Documentation

The model is `src/ccl/design/type-inference.md`, "4.6 Data vs compute functions", in two sections — "Data domains are invariant" and "The domain join is a Σ" — plus "Deliberately incomplete here", which records the three real gaps: Σ itself, at both of the sites above; `KindMerge::Conflict` being covered only by hand-constructed compact graphs, with the source-level route unknown and the constraint-level face of the same rejection reachable; and the heterogeneous-scalar union, which stays an `IncompatibleBounds` error until strict scalar consumers can impose concrete bounds.

### DIFF
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -575,16 +575,17 @@ The pipeline passes downstream of inference treat function types structurally an
 
 ## 4.6 Data vs compute functions
 
-> **Status: implemented.** The `FunKind` marker, kind inference, and kind-aware
-> subtyping (the `Compute <: Data` rejection) are all live, as is the rejection of
-> a data join that would narrow a domain. Data-domain invariance holds for the
-> base and is unenforced for a domain *refinement* — see the callout below.
+> **Status: implemented, minus Σ.** The `FunKind` marker, kind inference,
+> kind-aware subtyping (the `Compute <: Data` rejection) and the invariant data
+> domain are all live. What is missing is the type the model says a domain join
+> *produces* — the Σ — so a join over distinct domains is diagnosed rather than
+> typed. See the callout below.
 
 The unresolved domain-join corner of §4.5 (O1/O4 — two collections meeting at one
 join point) is resolved by making a missing distinction explicit. The distinction
 does not by itself make every such join *typeable*; what it does is make the
 untypeable ones an **error** rather than a silently short collection. See
-[The domain-join rejection](#the-domain-join-rejection).
+[The domain join is a Σ](#the-domain-join-is-a-σ).
 
 **The distinction.** A function's domain can mean two things. A **compute
 function** `α ⇒ β` treats it as a *capability* — the inputs accepted; no data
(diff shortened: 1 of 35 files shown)


dpmills:

- #10 ◀
    - #11
        - #45
        - #12
            - #13
                - #14
                    - #15
                        - #53
                            - #54
                            - #16
                                - #17

src/ccl/design/type-inference.md
```diff
@@ -569,10 +569,248 @@ The expected binder is **always globally fresh** (proposal §5.2 verbatim; the 
...
+occurrence-typing language could keep the arms correlated through the path
+condition `flag` itself, with no tag; Cambra does not narrow on opaque `Bool`s, and
+the tagged variant is the substitute.
+
+### Deferred: data-domain invariance
+
+The `Fun`/`Fun` domain edge is contravariant for **both** kinds. For a *compute*
+function that is right — the domain is a parameter. For a *data* function it is not:
+the domain **is** the data, so a `[0,10] ⤇ 𝑉` standing where a `[0,5] ⤇ 𝑉` is declared
+is silent row addition, and the reverse is silent row loss. Neither should typecheck.
```
sortalongo: I disagree. I think data functions should be contravariant. It's totally fine for a `[0,10]|=> V` to stand in for a `[0,5]|=> V`. IMO, this section of the doc (and any supporting code) should be deleted.

sortalongo (approved):
Just one comment. Otherwise, LGTM.

src/ccl/design/type-inference.md
dpmills: We really do need data functions to be invariant on their domains. I edited the design to make the argument cleaner (and cleaned up some related code along the way), but the short reason is that allowing the subtyping edge makes it easy to drop data in real programs.  If a longer array is a subtype of a shorter array, then something like '[0] if x else [0, 1]' will type unconditionally as an array of length one because that's the LUB of the two result types.  We most likely want to allow a cast to refine the domain of a data function, but it can't happen implicitly as a subtyping relation

PR #11, opened 2026-07-23
mutability-conditionals 2/7: value-Case compilation via literal union-of-restricts

dpmills:
## The problem: conditionals type-check but do not compile

A conditional used as a *value* — a ternary, an `if`/`elif`/`else` that produces something — gets all the way through inference and then dies. `1 if c else 2` has a perfectly good type (the arms join; `tests/type_check.rs` has covered this for a while), and then `lambda_elim` hits it and returns

```
unsupported node kind in lambda elimination: Case { .. }
```

Only the *filter* `Case` compiles today: the two-arm `[𝑔 → action; true → unit]` shape a comprehension `if`-clause lowers to, which becomes a `Restrict`. Every other `Case` is a hole. The pre-change design doc says so in as many words:

> **Blocker:** both need a *value-selecting* `Case` (`x = if 𝑝: 𝑒₁ else: 𝑒₂`) as a compiled construct, but only the *filter* `Case` (`[𝑔 → action; true → unit]` → `Restrict`) compiles today — a value `Case` errors at `lambda_elim`.

Concretely, everything in this list is rejected before this PR, at three different layers:

- `1 if c else 2`, `10 if a < b else 20`, `a * 2 if c else a - 1`, `elif` chains, nested ternaries, string/bool arms — `lambda_elim`.
- `xs if c else ys` as a collection, and consuming one (`sum([1, 2] if c else [3, 4])`) — `lambda_elim`.
- `[f(x) for x in (xs if c else ys)]` and `[a if g(x) else b for x in xs]` — `lambda_elim`.
- `if c: o << 1 else: o << 2` (a source-less conditional feed) — `channelize`, with `DeferError::PartialFeedCaseUnsupported`.
- `if`/`else` inside a generator for-loop body — `lowering`, with *"if/else inside generator for-loop body is not supported; use a plain if-guard (no else branch)"*.

## Why now

This is the load-bearing prerequisite for the next two PRs in the stack, not an isolated feature.

A conditional induction write (3/7, `if 𝑝: total += x`) compiles to a write leg over `{𝑟 : 𝐷 | π̂}` plus a carry-forward leg over the complement — and the carry-forward body is *a value selection over the snapshot*. Path-based `with begin():` conditionals (4/7) are the same shape one level up: the per-key merge that gives a branch which does not write a key its snapshot value **is** a value-selecting `Case`. Neither can be built while a value `Case` is a compile error, which is exactly what the blocker note above was recording.

It also sits directly on top of 1/7 (data/compute function kinds): kind inference is what makes `([x for x in xs]) if c else ([y for y in ys])` join as two collections rather than collide as two compute-kinded lambdas over their index domains. That case is tested here because 1/7 is what unblocked it.

Standing on its own merits: a ternary is table-stakes surface syntax, and the surprising part of the status quo is not that the fancier forms fail but that `1 if c else 2` does.

## Why selection becomes data instead of a branch

The obvious implementation is a branch: an operator that evaluates the guard and picks an arm. Rejected, because a branch instruction is foreign to the substrate. CCL's compiled algebra is restricts and unions over collections, and the runtime is a pull-based tile pipeline with no control-flow edge — introducing one means threading a new concept through op-conversion, planning, and the operator protocol, and it puts a construct in the IR that none of the existing rewrites know how to see through.

So instead the selection becomes *data*: each arm is gated by its first-match predicate `π̂ᵢ` (synthesized from the original guards as `𝑔ᵢ ∧ ¬𝑔₀ ∧ … ∧ ¬𝑔ᵢ₋₁`, `synthesize_arm_predicate` in `ccl_utils`), and the gated arms are unioned. The gates are exclusive and exhaustive, so exactly one arm contributes at runtime. There is no branch instruction anywhere in the result — the selection is a restrict, and every downstream pass already understands restricts and unions.

The one thing a branch gives you for free is that the untaken arm is not evaluated, and that has to be earned here rather than assumed. It is: see the laziness note below.

## The two forms

By what the arms produce.

**The C-form**, for scalar / compute arms (`build_value_case_cform`): each arm becomes a gated one-shot lift over a `UIntRange(1)` driver, and the union is extracted by `final_or_default`.

```
Case{[g₀ → e₀; …; gₙ → eₙ]}  ⟹  (⧺ᵢ const(eᵢ) : {UIntRange(1) | π̂ᵢ} ⤇ V,  eₙ) ▷ final_or_default
```

The gate is constant in the driver element, so the one-element domain either survives whole or is emptied; exhaustiveness (the trailing `true` arm) means the union has exactly one element and the outer type is the `Case`'s own `V`, unchanged. This is what makes ternaries, `elif` chains, nesting, and string/bool arms compile.

**The fan-out**, for collection arms (`build_value_case_fanout`): each arm's whole collection restricted by its gate, unioned as `Fun{Data, Variant([{𝐷 | π̂ᵢ}]), 𝑐}`. The arms share one domain 𝐷 — a join at distinct domains is rejected at inference — so every leg's payload is that one domain under a different gate.

**Off-path arms are not evaluated.** `0 if b == 0 else 10 // b` must not fault when `b == 0`, and the fan-out gets this structurally (an arm gated false restricts to an empty extent, and an empty extent is never iterated). The scalar C-form needed one runtime fix to match: `MapResultToConst` now returns a terminal-empty tile when every driver row is deleted, *without pulling the constant* — previously it pulled first and `10 // b` panicked with "attempt to divide by zero". The decidedness of the input's domain predicate is carried onto the empty tile so a decided false-gate arm reads terminal-empty rather than blocking the union forever.

## What comes with it

**One subtyping rule** (`is_index_partition_of`, in the solver's `constrain`). The fan-out's structural `Variant`-domain type has to reconcile against the plain data function the type system gave the `Case`. Each leg `{𝐷 | π̂ᵢ} <: 𝐷` relates by refinement width, *covariantly* — this is an introduction, not the function-domain contravariance. It is guarded on the legs refining the same *concrete* target, so a genuine heterogeneous `++` flowing into a fresh-var domain still takes the ordinary contravariant arm.

That guard is what makes the rule recognize by *shape* rather than by counting: three `elif` arms over one domain give a **three**-leg partition over **one** domain. A positional leg↔domain bijection would reject it — and that is precisely the shape an `if`/`elif` accumulator write produces in 3/7, so it is tested here.

**Two comprehension interactions**, which go in opposite directions because a value `Case` has no driver of its own and therefore must gate the *iteration source*:

- A conditional *source* (`[f(x) for x in (xs if c else ys)]`) floats the `Case` out of the map — `Case{gᵢ → [f(x) for x in srcᵢ]}` — with each arm built as a `Compose` so `emit_compose` stamps it with the source's `Data` kind. Without that the arms collide as compute-kinded lambdas over their index domains.
- A conditional *element* (`[a if g(x) else b for x in xs]`) fans the *source* out by each arm's element-dependent gate: `⧺ᵢ [eᵢ for x in xs if π̂ᵢ]`, filtered maps over one source whose gates partition it and whose union recombines them by position (`fan_out_element_case`).

**Conditional feeds** fall out of the same gated-lift machinery with no iteration source, over the `Unit` driver — one channel per feeding arm, replacing `PartialFeedCaseUnsupported` for guard-only `Case`s. And `channelize` plus the `map` tile operator carry the union through: a tagged-`Variant`-domain data function compiles straight to a `UnionOperator` over operands whose domains are concrete domains.

## Scope

Deliberately left rejected, each with a diagnostic rather than a miscompile:

- **Conditional induction writes** (`if 𝑝: total += x`) — 3/7. Designed in `design/mutability.md`, still rejected at lowering.
- **A per-element conditional at a site with no visible iteration source** — a UDF body, or a comprehension carrying an extra `if`-filter beside the element `Case`. Same source fan-out, but no source in hand at that site; the new `lambda_elim` arm names the shape in its error.
- **Scrutinee / pattern `Case`s** — a pattern match cannot be gated by a boolean, so `PartialFeedCaseUnsupported` stays for those.

## Tests

`tests/compilation_pipeline/conditionals.rs`, end-to-end through the pipeline: the C-form across guards / `elif` / nesting / partial off-path arms, the collection fan-out at one domain including the three-legs-one-domain case, conditional sources, conditional elements, conditionals between standalone comprehensions, and conditional feeds (including off-path laziness). Plus the lowering test that previously asserted `if/else` in a generator body was *rejected*, now asserting it lowers.

### DIFF
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -6,7 +6,8 @@ use std::rc::Rc;
 
 use crate::ccl::scope::{ScopedItem, for_each_scoped_item};
 use crate::ccl::{
-    BaseType, Builtin, Expr, Lit, Name, PredicateId, Refinement, Type, TypedExprNode,
+    BaseType, BinOpKind, Branch, Builtin, Expr, Lit, LogicKind, Name, PredicateId, Refinement,
+    Type, TypedExprNode, UnaryOpKind,
 };
 
 /// Returns `true` if `expr` directly references the given built-in primitive.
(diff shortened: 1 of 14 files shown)


dpmills:

- #10
    - #11 ◀
        - #45
        - #12
            - #13
                - #14
                    - #15
                        - #53
                            - #54
                            - #16
                                - #17

sortalongo:
I don't understand why this PR is needed. It just describes a new representation for Cases without mentioning the previous state or motivating the change.

dpmills:
> I don't understand why this PR is needed. It just describes a new representation for Cases without mentioning the previous state or motivating the change.

I've updated the description to explain why this is needed

sortalongo (approved):

PR #12, opened 2026-07-23
mutability-conditionals 3/7: conditional induction writes

dpmills:
## Conditional induction writes

PR 3/7 of the mutability-conditionals stack, on value-Case compilation (2/7). Delivers
conditional mutation inside an induction loop — `for x in xs: if p: total += x` — the
`mutability.md` motivating example. The model: a conditional write becomes **one
carry-complete writer** over a `Tile::Store` changelog (the write arm on `{r | π̂}`, the
complement arm carrying the snapshot forward), folded to a single commit-gated decision;
the union appears only in history slots, where the existing guarded-slot grammar already
admits it.

### What changes

- **Lower if-in-loop to a statement-Case** (`lower/loops.rs`), with `collect_mutation_loop_vars`
  recursing into branches (and correctly excluding inner `for` / `with begin()`).
- **Induction accumulators as a `Tile::Store` changelog** — induction routes through the
  changelog `InductionStore` and recognition folds the conditional write to one
  commit-gated writer (multi-writer dispatch is never born).
- **Value-Case inside a writer lambda** — the C-form from 2/7 applied inside a decision
  body, evaluated lazily (the off-path arm is not forced).
- **Uniform carry-complete write decision** — every path (write or carry) produces a full
  decision record, so trailing reads see one merged view.
- **`Tile::Store`: separate terminality from the watermark** — a store's "no more writes"
  signal is decoupled from its position watermark, so a conditional (sparse) writer bounds
  correctly and a loop whose tail carries reads its true final accumulator.
- **Commit unconditional writes beside a conditional** — an unconditional write *before* a
  conditional (`x := x+1; if p: x := x+10`) now commits on every position (the commit gate
  compares the carried expression against the entering snapshot at compile time; a
  differing carry forces commit), fixing a silent-drop where the unconditional increment
  was lost.

### Docs
`src/ccl/design/mutability.md` (conditional induction writes graduate to the changelog
realization); `src/interpreter/design-operators.md`; the induction-unification design plan
and the unordered-mutability design commitment; `docs/plan.md`.

Green under `./ci.sh`.

Continues cambra-dev/cambra-old#304 (review history there).

### DIFF
--- a/tests/compilation_pipeline/conditionals.rs
+++ b/tests/compilation_pipeline/conditionals.rs
@@ -523,3 +523,122 @@ fn test_conditional_element_result() {
     let result = run_pipeline("[(x * 10 if x > 2 else 0) for x in [1, 2, 3, 4]]");
     assert_eq!(codomain_ints(&result), vec![0, 0, 30, 40]);
 }
+
+// ---------------------------------------------------------------------------
+// Conditional induction writes — `if p: acc += e` in a mutation loop.
+//
+// `lower_loop_body_chain` lowers the `if` to a statement-`Case`; `transform_chain`
+// merges its branches into one writer decision — a single conditional write is
+// `{commit: ĝ, writes: prev+e}` (a `!commit` position carries in the changelog).
+// One writer over the full source → a single-writer `Transact` → the changelog
+// `InductionStore`, read densely and reduced to the final accumulator.
+// ---------------------------------------------------------------------------
+
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+// Guard fires for 3, 4 → total = 3 + 4 = 7; positions 0, 1 carry.
+#[case(
+    "total := 0\nfor x in [1, 2, 3, 4]:\n    if x > 2:\n        total += x\ntotal",
+    7
+)]
+// Guard fires for none → total stays at its init.
+#[case(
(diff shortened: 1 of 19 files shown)


dpmills:

- #10
    - #11
        - #12 ◀
            - #13
                - #14
                    - #15
                        - #53
                            - #54
                            - #16
                                - #17

tests/compilation_pipeline/conditionals.rs
```diff
@@ -335,3 +335,122 @@ fn test_conditional_element_result() {
...
+// merges its branches into one writer decision — a single conditional write is
+// `{commit: ĝ, writes: prev+e}` (a `!commit` position carries in the changelog).
+// One writer over the full source → a single-writer `Transact` → the changelog
+// `InductionStore`, read densely and reduced to the final accumulator.
+// ---------------------------------------------------------------------------
+
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+// Guard fires for 3, 4 → total = 3 + 4 = 7; positions 0, 1 carry.
+#[case(
```
sortalongo, line 540: Please us raw strings to make these test cases readable.

sortalongo (approved):

tests/compilation_pipeline/conditionals.rs
dpmills: Done here and elsewhere in the stack

PR #13, opened 2026-07-23
mutability-conditionals 4/7: path-based with begin(): conditionals

dpmills:
## Path-based `with begin():` conditionals

PR 4/7 of the mutability-conditionals stack. Extends transactional mutability from a single
bare deny guard to full conditional structure inside a `with begin():` block —
`if`/`elif`/`else` routing across keys — while keeping **one decision record per
transaction** (the soundness verdict from 2/7: txn path predicates read the commit-tick
snapshot, so per-path transaction *sites* would violate one-record atomicity).

### What changes

- **Path-based routing** (`transact_phase`) — the path walk threads `(path, env)` with
  read-your-writes, resolves guards against the incoming env, and rejoins each written key
  as a carry-forward Case (`env[k] = Case[a₁ → env₁[k]; …; true → snapshot]`). `commit`
  is the disjunction of the paths that wrote. These merge Cases compile via 2/7's C-form.
  Lowering simply stops rejecting elif/else/multi-if.
- **Carry conditionally-written keys via read-set finalization** — `collect_footprint`
  threads an `in_case` flag and adds any key written inside a statement-`Case` arm to the
  transaction's read set at collection time. Because all `Var` reads accumulate
  unconditionally regardless of path, a key read on only one path cannot be missed, and
  α-unique names mean a binder can never shadow a store key. A carried key is then seeded
  into the incoming env and reads its prior committed value correctly.
- **Per-tap fire field** — a routed feed carries a `to_<defer>__fire: Bool` gate so a tap
  fires only on its own path; the engine omits non-fired tap keys from the commit delta
  (no garbage in the durable log). Omitted when path == commit, preserving today's shape.
- **Conditional feed in a read-only begin() block** — a conditional reply inside a
  read-only block routes to the channelize feed fan-out; the both-arms `if`/`else` case
  works because the pure-feed fan-out (2/7) handles the trailing `true`-guarded arm.
  Two tests pin the single-arm and if/else forms.

### Semantic change (documented)
The deny guard is no longer transaction-scoped — a spine write beside an `if` commits
unconditionally (`commit = p ∨ true`). Audited: no existing test regresses; a new test
pins the new meaning. `with t = begin():` (handle binding), nested begin, loops in a
block, and guarded induction writes in a block stay rejected via a user-facing pre-check.

### Docs
`src/ccl/design/mutability.md` ("General in-transaction conditionals" graduates into the
main model; path-scoped deny semantics; fire-field mechanism; write-mask future work);
`src/interpreter/design-operators.md` (CommitWriter tap-fire decode). The per-tap fire field
and read-set finalization are documented in `src/ccl/design/mutability.md`.

Green under `./ci.sh`.

Continues cambra-dev/cambra-old#305 (review history there).


dpmills:

- #10
    - #11
        - #12
            - #13 ◀
                - #14
                    - #15
                        - #54
                        - #67
                            - #17

src/ccl/design/mutability.md
```diff
@@ -860,24 +860,52 @@ The value-`Case` positions ride the same union-of-restricts:
...
-(`if 𝑝: x := 𝑒` → one write at path `𝑝` → `commit = 𝑝`), and admits value-selection, cross-key
-routing, `elif`, nesting (conjunction), and sequencing. The per-key merges are value-selecting
-`Case`s over the snapshot, compiled by the scalar form above; the path walk itself is the next PR
-of the conditionals stack. One shape is settled by analysis: the block stays **one decision record
-per transaction** —
-per-path writer sites are unsound (a path predicate reads the snapshot, which exists only at the
-commit tick; one `begin()` is one serialization point; multi-key read-your-writes needs one
-snapshot and one write-set).
+A `with begin():` block admits `if`/`elif`/`else` and multiple sibling `if` guards, compiled by a
+uniform **path-based** walk (`transact_phase::walk_block`/`walk_case`). Walking a block threads a
```
sortalongo: I don't think there's a clear definition of what a path is in the prose. Please add a definition/description.

src/interpreter/design-operators.md
```diff
@@ -192,7 +192,7 @@ The transaction engine that backs a `Type::Txn` [`Transact`](../ccl/design/ir.md
...
...ent` advances on the commit-ack `release`, so the next pull takes the next item), a **deny** (`current` advances here with no commit — invisible in the store frontier, so a frontier-growth signal alone would miss it), and a **not-ready** decision (`current` unadvanced, the pending body-input row reused so a re-push does not duplicate a buffer position against the body's `Memo`). It is the writer's re-arm — not any reader — that converges the store: the wakeup fans through the cyclic `FanOut` notify closure to every store branch, re-pulling the `AsOf` / `StoreValueStream` readers as commits land. A writer **drained but live** (`current >= n_items`, source not yet complete) does *not* re-arm, so an idle live server does not busy-poll — a future arrival wakes it through its source-forwarding consumer. This retires the readers' former producer-side drive-to-fixpoint (`drive_store_to_fixpoint`, deleted). Drain order rotates for fairness (round-robin `drain_start`) and retained state is bounded (superseded proposals dropped via `drop_superseded`). The proposal stream is an offset window: released (committed) prefixes are compacted away, keeping writer state bounded on a long-lived store.
```
sortalongo: This description is so detailed and meandering that it seems essentially useless. It makes references to code that are not accessible in this document, and ends up obfuscating itself into noise. I suggest limiting these descriptions to 50-100 words, and only having a high-level description of what the operators do.

tests/compilation_pipeline/transactions.rs
```diff
@@ -903,52 +1077,95 @@ fn nested_transactions_rejected() {
...
             a: Mut(Int, Txn) := 0
             b: Mut(Int, Txn) := 0
+            out = defer()
             for x in [1, 2, 3]:
                 with begin():
                     if x >= 2:
                         a := a + x
                     else:
                         b := b + x
+            with begin():
```
sortalongo, line 1122: These tests with transactions after a loop are taking a contract on relative ordering between them which I thought we'd agreed was not a semantic guarantee. Please mark them explicitly so we don't start drifting on the intended semantics until we actually have concurrency.

sortalongo (approved):

tests/compilation_pipeline/transactions.rs
dpmills: Added inline notes for every instance

src/interpreter/design-operators.md
dpmills: I simplified the table entries, dropped excess refs to code, and added new sections outside the table with extra content that still felt useful

src/ccl/design/mutability.md
dpmills: Added

PR #14, opened 2026-07-23
mutability-conditionals 5/7: one changelog realization for every loop (retire dense Recurse)

dpmills:
## Induction: one changelog realization for every loop

PR 5/7 of the mutability-conditionals stack (variant elimination, 6/7, and the
decision-variant rework, 7/7, sit above it). Induction previously had *two* operators for
one concept: the changelog `InductionStore` (finite, single-writer, tap-free) and a dense
`Recurse` fallback (async, or reply-tap loops). That was a realization incompleteness, not
a semantic distinction — the model treats a finite domain as a terminating stream. This PR
collapses induction onto the single changelog realization and retires the dense `Recurse`
induction path entirely, so every combination of induction mutability, transactions,
feeds, and conditionals composes on one operator.

### What changes

- **One changelog realization for finite and async loops** — every single-writer,
  tap-free `mut` loop, finite (list) *or* async (`DataSource`), compiles to the changelog
  `InductionStore`. The drive decodes the source by absolute domain position (async domains
  arrive unordered and compact on release), drives contiguously from `processed`, and
  reclaims the consumed prefix incrementally; `StoreDenseRead` sorts positions so a
  scalar-final `ExtractLast` reads the true final accumulator (fixing an async-tap scramble).
- **Carry reply taps on the changelog store; retire the dense `Recurse`** — a feed riding
  an accumulator loop appends `to_<defer>` taps to the changelog decision write-set (as the
  commit store's writer does); the `InductionStore` applies the decision's `tap_fired` gate
  and a `to_<defer>` read is a non-carry `StoreDenseRead` that emits only fired positions.
  The dense induction `Recurse` (and `recurse.rs`) is deleted. This also unblocks a
  conditional feed / write sharing a loop with a feed, which previously panicked at lowering.
- **Keep-latest changelog GC (never-terminating bound)** — `StoreDenseRead` forwards its
  consumed loop-position prefix as a tick-prefix release; the producer GCs the
  FanOut-intersected prefix via `CommitEngine::gc_released_prefix`, keeping each key's
  latest write. Retention is O(keys) + slowest-reader lag, independent of positions
  processed, and keep-latest keeps the drive's own `read_as_of` recurrence sound.

### Shared abstractions (dedup)
- **Dedup writer-decision assembly** across the transaction and induction phases into one
  `ccl_utils::writer_decision_record` (+ shared commit-gate `disjoin`), so a decision-record
  fix lands once for both.
- **Share the changelog per-key fold** across both readers (`fold_changelog_key`): carry
  reads take the running value, tap reads take the delta, off one implementation.

### Docs
`src/interpreter/design-operators.md` (§*Induction stores as a changelog*; the `Recurse`
operator row and the sibling-loop `Recurse` references retired); `src/ccl/design/mutability.md`;
`CLAUDE.md` and `src/ccl/expr.rs` (op-conversion dispatch of a concrete iteration extent now
names `InductionStore`, not `Recurse`); `docs/plan.md`.

Green under `./ci.sh`.

Continues cambra-dev/cambra-old#306 (review history there).


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14 ◀
                    - #15
                        - #54
                        - #67
                            - #17

src/ccl/design/mutability.md
```diff
@@ -801,25 +806,23 @@ branches into **one uniform, carry-complete writer decision over the full loop s
...
 Both a **finite** loop and an **async** (streaming) source drive an induction accumulator — the
-model treats a finite domain as a stream that terminates (§Liveness). At the realization level this
-is *one* concept but currently *two* operators: the changelog `InductionStore` for the finite
-single-writer case, a dense `Recurse` fallback for async/tap loops. Collapsing them onto the single
-changelog realization (so the finite/async distinction disappears from the code as it has from the
-model) is a mapped interpreter cleanup — see *Planned: one changelog realization for every loop* in
-`../../interpreter/design-operators.md`.
+model treats a finite domain as a stream that terminates (§Liveness) — and every induction accumulator now
+uses *one* realization: the changelog `InductionStore`. Plain, conditional, and feed-carrying loops
+over finite or async extents all route through it; the dense `Recurse` fallback is fully retired. The
```
sortalongo: rm references to prior versions

src/interpreter/tile_operators/extract_last.rs
sortalongo, line 1: I don't like the "last" terminology, since it's easy to confuse between "previous" and "final". Can you kick off a renaming at top of stack?

CLAUDE.md
```diff
@@ -123,7 +123,7 @@ Vocabulary, matching `src/ccl/symbolic.rs`:
...
+- **Transact** (the domain-parameterized recurrence carrier `plan_loops` produces from a `LetRec`): `transact (k₁ = init₁, …) { [reads]⇒[writes] over source do body; … }` — the store keys with their seeds, then one writer per site (its read/write footprint, iteration source, and decision body). `Transact` is born by loop planning (`plan_loops`, in `planning/loops.rs`) *after* `lambda_elim` (its writer bodies are already point-free); op-conversion dispatches on the domain: a concrete iteration extent → the `InductionStore` changelog, `Txn` → the commit operator. (The former `Loop` carrier and the dense `Recurse` induction realization are retired — do not render them.)
```
sortalongo: rm references to previous versions

sortalongo (approved):

src/ccl/design/mutability.md
dpmills: Done

src/interpreter/tile_operators/extract_last.rs
dpmills: Started that going

CLAUDE.md
dpmills: Done

PR #15, opened 2026-07-23
mutability-conditionals 6/7: variant elimination — the point-free sum algebra

dpmills:
Adds the `VariantProject` / `VariantWrap` tile operators and compiles a scrutinee-`Case` over a `Type::Variant` to a point-free union of tag-restricts — the prerequisite sum algebra the decision-variant rework (7/7) builds on.

## Arms are keyed by tag

A tagged sum is represented three times over — the per-arm extents of a union, its per-arm predicates, and the payload columns of a union value — and those three have to agree about which arm is which. They agree by **tag**, via a `TagMap<T>`: a `FieldKey`-keyed sequence in canonical (name) order, carrying the key-set operations the union/extent/predicate paths would otherwise hand-roll (`same_tags`, `zip_matching`, `zip_same_tags`, `get_or_insert_with`).

Two of those pair arms across sums, and the difference is which discipline the caller needs. `zip_matching` visits the intersection and skips what only one side carries; `zip_same_tags` lifts a binary operation *pointwise* over one arm set, so a missing arm is a caller bug rather than a case to absorb — there is no arm to combine with, and no conservative answer a combination could give. `Predicate`'s `intersect` / `minus` / `union` are the latter and say so once, instead of each asserting and then carrying a fallback the assert makes unreachable; `subsumes` is a *query* over two sums, so it keeps answering conservatively when the arm sets disagree.

A tag's *position* cannot key an arm, because a position is relative to its own sum's key set and so is not stable under width subtyping: ``{`b} <: {`a | `b}`` is a legal instance of the variant width rule, but `b` sits at position 0 in the subtype and position 1 in the supertype. Since an arm set is part of a union column's runtime layout, a positional encoding would need a renumbering coercion on every subsumption — and "renumber" means rebuilding the value. Keying on the tag removes it: an arm is found by name, and a key the map does not carry is simply *absent*, which is exactly what width subtyping means.

Canonical order is still maintained, because arms are also consumed positionally where that pairing is load-bearing (per-variant release pairs a predicate arm with its source sub-extent). Sorted order makes structural equality agree with key-set equality; it just no longer decides identity.

`FieldKey::Index` keys make an anonymous positional sum (`a ++ b`, and the domain of the union-of-restricts fan-out) the degenerate case of the same structure, so named and positional sums share one representation — mirroring `Type::Variant`, which already keys on `FieldKey` for both.

Runtime variant width subtyping is correspondingly *stated in one place*, `Extent::includes`, matched by tag rather than by position; and a projection of a tag the scrutinee does not carry yields an **empty** projection rather than an error, which is what leaves a width-subtype scrutinee well-formed.

## The row→payload partition

A union column *is* a partition of the row space, and the row→payload correspondence is the whole content of that partition, so `UnionArm { rows, values }` stores it: ascending row positions parallel to a dense payload column. Without it, five call sites reconstruct that correspondence — `index_at` by an O(n) scan (making element access linear and iteration quadratic), `select_indices`, `retain` and `drain_to_value_iter` by running counters, and `VariantProject` by a per-tag walk. With it, all five go away and `index_at` is a binary search per arm.

The two vectors are private, because "ascending, and parallel to `values`" is precisely what makes a row's position in `rows` *be* its payload slot — the property every read depends on. `slot_of` (the binary search), `row_slots` (the `(row, slot)` pairs), `gather` (the shared tail of `select_indices` / `retain`) and `extend_shifted` (the arm-level half of a concatenation) are that invariant expressed as operations rather than as a convention each call site has to honour.

## `VariantProject` — restrict and project are one operation

`VariantProject` is the elimination primitive for a scrutinee-`Case` over a `Type::Variant`. Given a union scrutinee it narrows to arm `c`'s tag-restricted sub-domain and yields that arm's inner payload column, keyed by the original position.

There is no separate boolean `Restrict` step and no `Predicate` for tag discrimination, because `UnionArm` already stores each arm's payloads densely against the rows that carry them: reading the arm *is* the tag restriction. A domain-`Restrict` could not express tag discrimination anyway — the tag lives in the scrutinee's codomain, not its domain.

Also canonicalizes `ColumnValue::single(Unit)` to `Units(1)`: `from_values` / `empty_tile` already produce `Units` for a `Unit` extent, so a singleton `Unit` encoded as a one-element arm column collides ("mismatched variants" in append/merge) when a `Unit`-payload variant value (`` `abort(unit) ``) is fanned through a `Memo` during elimination.

## Scrutinee-`Case` as a union of tag-restricts

Compile a scrutinee-`Case` over a variant to the point-free union of tag-restricts, the sum-elimination dual of `VariantCtor` construction:

    match d { `cᵢ(wᵢ) → eᵢ }   ⟹   ⧺ᵢ ( d ≫ variant_project(`cᵢ) ≫ (wᵢ → eᵢ) )

The branch tags partition the scrutinee's domain, so this is the same fan-out the value-`Case` emits, keyed on tag rather than on a boolean first-match gate. Each arm narrows the fed scrutinee stream to arm `cᵢ`'s tag partition, binds the payload, and maps `eᵢ`; the flat `UnionOperator` re-totals the disjoint partitions. Exhaustiveness over the scrutinee's tags makes the union total, so no `final_or_default` collapse is needed — this is the fan-out shape, not the C-form.

- `Builtin::VariantProject(FieldKey)`: minted post-inference (it carries no scheme; the CHECK-mode typecheck trusts its stamped type), composed as a bare builtin onto the fed scrutinee. Op-conversion reads the union extents off the fed input's tiling and builds the `VariantProject` tile op.
- `lambda_elim` gains a scrutinee-`Case` arm beside the value-`Case` fan-out.
- `simplify`: `try_const_reduce` refuses to collapse past a compose element that **narrows the domain at runtime, leaving no refinement behind**. There are two — `filter_values(p)` and `variant_project(c)` — and what unites them, and separates both from `restrict`, is that once the element is gone nothing downstream can reconstruct which positions survived; collapsing past one would apply the constant at *every* position, and for a tag fan-out that makes the arms overlap. They are one named predicate (`narrows_domain_irrecoverably`) rather than a guard per element, and both halves are now exercised — they are spelled differently (an `Apply` of its predicate vs. a bare builtin), so covering one left the other's pattern untested.

### Arm bodies may close over the outer binder

An arm body can read the outer binder as well as its payload — the Phase B per-key-view shape ``λ __c → match __c.decision { `commit(w) → (time: __c.time, write: w.i) }``, whose arm reads both the record's sibling field and the `Commit` payload. `lambda_elim` detects the dependence structurally (the arm body has the outer binder free beyond the payload binder) and, only then, zips the whole element alongside the projected payload:

    ⧺ᵢ ( ⟨id, x.f ≫ variant_project(`cᵢ)⟩ ▷ zip ≫ (λ (x, wᵢ) → eᵢ) )

The two binders merge into one pair (`x ↦ pair.0`, `wᵢ ↦ pair.1`), reusing the nested-lambda rewrite. The payload-only path is unchanged. For the outer `id` arm and the tag-restricted payload to co-iterate, `VariantProject` preserves the scrutinee's real domain keys for a union *stream* (`SealedFunction {D ⇒ Scalar(Union)}`), and the zip's `FanIn` inner-joins the full outer element with the tag-restricted payload on shared keys.

## `VariantWrap` — the point-free constructor

The exact dual of `VariantProject`. A `VariantCtor` inside a lambda elaborates to a composable morphism `param_ty ⇒ Union`, so it can sit as the RHS of a `≫` — the shape the value-`Case`-in-lambda fan-out `⧺ᵢ (filter_values(π̂ᵢ) ≫ eᵢ)` needs for a writer decision ``if p: acc := `commit(…) else: acc := `abort(unit)`` (used by 7/7). `elim_lambda`'s `VariantCtor` arm rewrites `` `cᵢ(eᵢ(p)) `` ⟹ `eᵢ ≫ variant_wrap(cᵢ)`; op-conversion consumes the fed payload stream and builds the `VariantWrap` tile, which also accepts a `SealedFunction {D ⇒ Scalar(payload)}` input (the scalar `Scalar(Union)` path is unchanged). `ColumnValue::repeat` handles a singleton union, broadcasting a constant `` `abort(unit) `` over a stream.

No `simplify` guard is needed for the dual: `variant_wrap` is a total element-wise map, so const-reduce dropping it would be sound.

## A merged stream declares the merged tag set

Keying arms by tag is also what lets a *merged* variant stream describe itself honestly. A conditional whose arms carry **different** tags — ``  `pos(n) if n > 0 else `neg(n) ``, or a `match` whose arms construct different tags — panicked at compile time:

    ExtractLast default tiling must match source codomain tiling
      left: Scalar({`neg{Int}})  right: Scalar({`0{{`pos{Int}}} | `1{{`neg{Int}}}})

The tile the union operator *builds* was right: its codomain column is the tag-merged sum ``{`neg(rows: []) | `pos(rows: [0], 5)}`` — both tags present, the arm that did not occur empty. Only the **declared tiling** was wrong. `UnionOperator::new` kept the arms' codomains apart as an anonymous positional sum whenever they were not literally equal, describing a value shaped like "arm 0's variant *or* arm 1's variant", which is not what any row holds. A tile-conformance check caught the mismatch, and `ExtractLast` — whose default is one of those arms, so width-narrower than the whole — hit it first.

So declare what the operator already produces. A union's **domain** keeps every arm apart, because which arm a row came from is exactly what `final_or_default` dispatches on. Its **codomain** does the opposite: the arms are alternative values at one row, so it is their **join**. `Extent::join` is that operation, and for two variants it merges tag maps — a tag they share joining its payloads, a tag only one carries taken as-is. Where the arms have no join the positional sum stays, which is right for a concatenation: there the arm a row came from *is* part of a row's identity.

Two consequences of declaring only what is produced. Differing arms merge into one *column*, which is also the only codomain shape `UnionProducer` can build, so a non-`Scalar` arm tiling is rejected where the shape is known rather than flattened into a declared "scalar column of functions" the producer would panic on at the first `get`. And the join is **symmetric**: `Extent::includes` relates a variant and a bare payload *both* ways, so an inclusion-based join would have let argument order pick the answer — and in one order claim an untagged value inhabits a tagged space. A mixed pair has no join.

`ExtractLast` then needs two things. Its `default` must be *representable* in the output value space rather than identical to it — `Extent::includes` already states variant width subtyping at the runtime boundary, so that is the check. And every emission is built at the **declared** extent via `ColumnValue::from_values`, not from the extracted value alone: a variant value carries only its own tag, so a column built from it is narrower than the tiling whenever the collapsed alternatives carry more tags between them. Building at the declared extent keeps every arm present with the ones that did not occur empty, which is the shape a downstream merge or append requires.

That second half is also what makes a **variant-valued mutable register** work, which otherwise failed in every spelling — same-tag writes as `Mismatched ColumnValue variants in append`, a `` `none `` seed with `` `some `` writes as the same tiling mismatch, and a conditional write likewise. A register's seed and its writes are alternatives in exactly the same sense as a conditional's arms.

## Rendering: CHL's surface syntax for a sum

A tagged sum renders the way the language spells it, on both sides of the compile/runtime boundary — braces around the arms, each arm introduced by a backtick with its stored type in braces of its own:

    {`commit{Int} | `abort}      Type::Variant
    {`abort | `commit{Int}}      Extent::Union
    {`pair{a: Int, b: Int}}      a record payload reuses its own braces
    {`outer{`x | `y{Int}}}       so does a nested sum
    `commit(7)                   Value::Union — a value renders as the constructor that built it

`Type` and `Extent` share one arm formatter (`fmt_variant_arms`) rather than each spelling the rule: the two are printed *side by side* in a tiling-mismatch panic, which is precisely where two spellings of one sum would mislead. Three rules live there. A `Unit` payload is the **nullary** constructor and renders bare, because "stores nothing" is the whole content of that arm and `` `abort{Unit} `` would spell an absence as a presence. A payload that already renders brace-delimited **reuses** those braces — unambiguous, since an arm always begins with a backtick, so a brace body that does not is a payload and never an arm list. And a single-arm sum is not special-cased; it is the general form with one arm.

**Arm names are lowercase** (`` `commit ``, `` `abort ``): an arm is a *constructor*, and `Caps` in CHL means a type — the same convention `BaseType::keyword` already documents.

A **term** always writes its payload, the nullary constructor's included (`` `abort(unit) ``), because the `Unit`-valued expression sitting there is a real node — unlike an arm's *type*, where storing nothing is the whole content. Destructuring mirrors construction, so a `match` arm is `` `commit(w) → … ``, and the point-free builtins keep the backtick on their tag: ``variant_project(`commit)``.

An **anonymous positional** sum — what `++`/`CollectionUnion` produces — is not a tagged variant and keeps its flat `A | B` rendering: its `Index` keys carry no user-meaningful information, and dressing them as surface arms nobody wrote would claim more than the type knows.

## Also here

`flat_merge`'s disjointness precondition is asserted rather than assumed. A flat merge is correct only if the arms' runtime domains are disjoint; two arms claiming one position put two values at one domain key, which is a monotonic-merge violation however it is resolved — and a *silent* one, since the sort just leaves the duplicates adjacent and a column one row too long flows on downstream. The scan is free beside the sort already happening, so it is a hard assert.

`VariantProject` now carries the projected arm's declared `payload_extent` rather than deriving it from its whole output tiling. The empty column an absent arm yields was being built from the `SealedFunction` extent, so it came out as the untyped `Variants` catch-all instead of (say) `Ints` — the same "mismatched variants in append" class of failure described above, reached from the width-subtyping direction.

The scrutinee-`Case` elimination is written as a `≫`-**composition**, and the comments now say so: every element of an arm is a morphism out of the eliminated binder, so `⧺ᵢ ( scrut ≫ variant_project(`cᵢ) ≫ (λ wᵢ → eᵢ) )` is a compose chain and not an application chain. In the outer-binder form `⧺ᵢ ( ⟨id, scrut ≫ variant_project(`cᵢ)⟩ ▷ zip ≫ … )`, `id` sits *beside* the projection rather than inside it because both components are morphisms out of the **outer** binder — `scrut ≫ ⟨id, variant_project(`cᵢ)⟩` would pair the scrutinee with its own payload and lose the sibling fields the arm body reads.

Every `Builtin` variant's doc opens with its type, written `name : 𝑇`; five had drifted (`permute_domain`, `flatten_domain`, `binop`, `sum`, `max`) and the convention is now stated on the enum, including which built-ins additionally carry an `OperatorSchemes` scheme.

`design-operators.md`'s `VariantWrap` / `VariantProject` rows describe the tag-keyed representation and say why a position cannot key an arm.

`mutability.md` gains *Tagged sums in a decision, and in a register*: why a decision is a sum rather than a `commit: Bool` beside a `writes` field that is meaningless when it is false, which half of this algebra each direction of that needs, and why a variant-valued register follows from the same join law with nothing register-specific.

## Tests

Op-level (both arms project correctly; key-preservation and zip alignment over a stream keyed by arbitrary UInts) and full-pipeline (channelize → lambda_elim → op-conversion → drive): a two-arm match firing each arm, the one-arm read-side shape, the `(time, write)` per-key view over a two-row stream, and the in-lambda writer-decision shape ``λ p → Case[p == 0 → `commit(p + 1); true → `abort(unit)]`` reading back `` `commit(1) `` at position 0 and `` `abort(unit) `` at position 1.

`TagMap` is covered directly: canonical ordering independent of insertion order, an absent tag reading as `None` rather than an error, a position being relative to its own key set, numeric ordering of `Index` keys, `zip_matching` pairing by tag, and duplicate tags rejected.

`Extent::join` is covered directly (disjoint tags merge; a shared tag joins its payloads; a conflicting shared payload fails the whole join rather than silently picking a side; unrelated scalars have no join; a variant and a bare payload have none in either direction; and every case is symmetric), along with what `UnionOperator::new` declares in each of those cases — including its rejection of a differing non-`Scalar` arm — and both of `ExtractLast`'s emission paths, the extracted value and the default, asserting the emitted column carries every tag with only the occurring one inhabited.

`UnionArm`'s row→payload correspondence is pinned where it matters: `slot_of` resolving a row to *that row's* payload slot rather than to its position in the whole column, and absent rows resolving to nothing rather than being misresolved.

`lambda_elim`'s two scrutinee-`Case` paths are pinned by their emitted point-free shape **and typechecked** — the typecheck being the load-bearing half, since it confirms `variant_project`'s stamped `scrut_ty ⇒ Pᵢ` really does compose between the eliminated scrutinee and the eliminated arm body rather than merely looking like it should. `run` only checks that under the opt-in `deep-typecheck` feature, so the tests ask directly and stay honest in every configuration. The outer-binder case is the sharper one: simplification fuses its left component down to `.time`, a morphism out of the *outer* binder that nothing derived from the scrutinee could have produced.

`Extent::includes` is covered for the case a *positional* implementation gets wrong and the existing tests cannot see: ``{`b{Int}} <: {`a{String} | `b{Int}}``, where the shared arm sits at position 0 in the subtype and 1 in the supertype. The prefix-shaped cases already covered agree under either implementation.

The documented behaviours that had no test now have one: an absent arm projecting empty rather than erroring — the width-subtyping case, and how the `payload_extent` bug above surfaced, `ColumnValue::repeat` broadcasting a singleton union while leaving the arms that did not occur present-and-empty, a union column renumbering its rows densely under `select_indices`/`retain` while each payload follows its own row, and `flat_merge` both reassembling disjoint arms and rejecting overlapping ones.

Rendering is pinned on both sides — `Type::Variant`, `Extent::Union`, and `Value::Union` — including the nullary collapse, the brace reuse for record and nested-sum payloads, a payload that is *not* brace-delimited, and the anonymous positional sum keeping its flat join.

A variant-valued register is verified only at the operator boundary here — both `ExtractLast` emission paths *are* its seed and its writes — because a register's value has to be written down to be driven end to end, and there is no surface syntax for a variant yet. The end-to-end spellings land with that syntax.


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14
                    - #15 ◀
                        - #54
                        - #67
                            - #17
                                - #68

src/ccl/lambda_elim.rs
```diff
@@ -959,6 +959,210 @@ fn elim_lambda_impl(
...
+            Ok(typed_compose(vec![payload_pf, vw]).with_ty(result_ty))
+        }
+
+        // A **scrutinee-`Case`** over a variant inside a lambda body
+        // (`λ param → match scrut { cᵢ(wᵢ) → eᵢ }`): the point-free **union of
+        // tag-restricts**. The branch tags *partition* the scrutinee's domain,
+        // so elimination is the same fan-out the value-`Case` above emits —
+        // keyed on **tag** rather than a boolean first-match gate:
+        //
+        //   ⧺ᵢ ( scrut ▷ variant_project(cᵢ) ▷ (λ wᵢ → eᵢ) )
```
sortalongo: This doesn't typecheck, AFAICT. IIUC, `variant_project` returns a function, and so the second ▷ should be composition. But it's hard to really understanding because the type of `variant_project` isn't anywhere I can see. Can you make sure the types of all built-ins are explicitly documented somewhere, and double check that this lambda-elim types as expected?

src/ccl/lambda_elim.rs
```diff
@@ -959,6 +959,210 @@ fn elim_lambda_impl(
...
+        // shape, not the C-form.
+        //
+        // **Outer-binder arms.** When `eᵢ` closes over the outer binder as well
+        // as its payload (`eᵢ(param, wᵢ)` — e.g. a per-key view
+        // `λ __c → match __c.decision { Commit(w) → (time: __c.time, write: w.i) }`
+        // reading both the record's sibling field and the Commit payload), the
+        // arm zips the outer element alongside the projected payload:
+        //
+        //   ⧺ᵢ ( ⟨id, scrut ▷ variant_project(cᵢ)⟩ ▷ zip ≫ (λ (param, wᵢ) → eᵢ) )
+        //
```
sortalongo, line 1031: Cool!

But I think the `⟨ id, scrut ▷ ...⟩ ` isn't correct. IIUC, it should be something like `scrut ≫ ⟨id, variant_project ...⟩`, since the two take the same scrutinee as input.

sortalongo (commented):
There seem to be no new test cases. Is that intentional? I would've expected a lot of new cases given that this unblocks a whole new class of programs.

src/ccl/lambda_elim.rs
dpmills: Fixed and clarified

src/ccl/lambda_elim.rs
dpmills: Yeah, fixed

dpmills:
None of this code is testable from CHL yet, but there are new unit tests.  I did another pass and added even more, fixing a few minor bugs along the way

sortalongo (approved):

PR #18, opened 2026-07-23
Lowering: enforce def/lambda parameter type annotations (were silently dropped)

dpmills:
`uncurry_params` attached a `user_annotation` only for `Mut[…]` pass-by-reference parameters; every other parameter annotation (`int`, `List[T]`, …) fell through to a bare `Type::Hole`, so the parameter was inferred purely from its body and the declared type was never checked against arguments. A `def g(a: int)` with an identity body accepted an argument of any type — the annotation was documentation with no force, unlike a variable ascription (`x: T = e`), which does enforce its annotation.

Attach the lowered annotation as the parameter's `user_annotation` in both the single-parameter and the multi-argument (tupled) paths, so `emit_lambda`'s checking mode binds the parameter at its declared type and an ill-typed argument is rejected at the call site.

## Interaction with `Mut` pass-by-reference

`Mut[…]` handling is deliberately unchanged. The `Mut` single-param path seeds the *real* binder type (`Mut[V, _]`), not just the annotation, because the body must be typed with the parameter already `Mut` (its `MutRead`/`MutWrite` nodes check against a `Mut`-typed binder, and the domain generalizes per call site) — leaving it a `Hole` and reconciling the annotation afterward fails. A plain `int`/`List[T]` annotation needs no such seeding; it is purely a boundary constraint, so this change only ever writes `user_annotation` and never touches the `Mut` exception.

The one observable interaction: a mutable variable passed to a *non*-`Mut` parameter is read **by value** (its current value is copied; the callee cannot write it back), and the parameter's declared value type is now enforced against it. Before this fix that mismatch compiled — the annotation was silently dropped and the parameter inferred its type from the argument.

## Tests

Full lower→infer path for single- and multi-argument annotated parameters (scalar type, `List[T]` element type, unannotated inference), plus the `Mut`-interaction cases above. Each enforcement test fails on `main` without the lowering change.

Extracted as a standalone fix from the `dmills/map-set-constructors` collections stack, where the gap was discovered.

### DIFF
--- a/src/ccl/lower/functions.rs
+++ b/src/ccl/lower/functions.rs
@@ -113,10 +113,22 @@ pub(super) fn uncurry_params(
             _ => Type::Hole,
         };
         let mut lam = Expr::lambda(params[0].name.as_str(), param_ty.clone(), body_expr);
-        if matches!(param_ty, Type::History { .. })
-            && let TypedExprNode::Lambda { param, .. } = &mut lam.node
-        {
-            param.user_annotation = Some(param_ty);
+        if let TypedExprNode::Lambda { param, .. } = &mut lam.node {
+            if matches!(param_ty, Type::History { .. }) {
+                // `Mut[…]` pass-by-reference: the annotation rides `user_annotation`
+                // for the mutability discipline check (rule 3).
+                param.user_annotation = Some(param_ty);
+            } else if let Some(ann) = &params[0].annotation
+                && mut_param_history_type(&params[0]).is_none()
+            {
+                // Any *other* annotation (`int`, `List[T]`, …) is a
+                // **checking-mode** declaration: attach it so `emit_lambda` binds
+                // the param at its declared type and an ill-typed argument is
+                // rejected at the call site. Without this the annotation was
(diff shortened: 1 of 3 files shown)


dpmills:

- #18 ◀

src/ccl/lower/functions.rs
```diff
@@ -113,10 +113,22 @@ pub(super) fn uncurry_params(
             _ => Type::Hole,
         };
         let mut lam = Expr::lambda(params[0].name.as_str(), param_ty.clone(), body_expr);
-        if matches!(param_ty, Type::History { .. })
-            && let TypedExprNode::Lambda { param, .. } = &mut lam.node
-        {
-            param.user_annotation = Some(param_ty);
+        if let TypedExprNode::Lambda { param, .. } = &mut lam.node {
```
groundlar, line 116: Per slack, the ordering / exclusivity looks weird here, let's add some more tests around `Mut` with annotations

dpmills: Everything works correctly; tests added

groundlar (approved):

PR #19, opened 2026-07-23
ccl: stable NodeId substrate + lineage core + always-on lowering projection

groundlar:
Stable node identity, the lineage data model, and the always-on lowering projection — reviewable on its own, with no pass recording rewrites yet. Nine commits: the substrate, then four from review round 1, three from round 2, and one from round 3 — each one concern.

**Start here:** `src/ccl/lineage.rs` is the data core; `src/ccl/provenance.rs` and `src/ccl/design/provenance.md` frame how it's consumed.

## The substrate

Every IR node now carries a stable `NodeId`. There are exactly two ways to build a node and the recorder sees precisely the difference between them: `Expr::new` mints and records a birth, `Expr::preserve` carries an id an earlier node already had and records nothing. A pass that rebuilds a node — reparenting it, renaming a slot — uses `preserve`, so its span and lineage survive as a self-edge. Lowering records a `LoweringLog` that `collapse_lowering` folds once, at the lowering boundary, into an always-on `lowering_projection` (a `SourceProjection`: node → directly-lowered `SourceAttribution`).

`src/ccl/lineage.rs` holds the full data core:

- `RewriteStep` — `Transform`/`Copy` ops, with blame recorded separately from consumption.
- the collapse fold — produces a dense `LineageMap` + per-pane `SourceProjection` with a two-sided leak taxonomy; `LoweringStep` + `collapse_lowering` share its `RootTracker` core.
- the thread-local step recorder — RAII `StepGuard` + `RecorderSession` (precedent: `infer_var::ACTIVE_ARENA`).
- the `Expr::new` / `freshen_node_id` construction hooks, and the sentinel `Default` id that keeps `mem::take` throwaways out of open steps.

The fold, survivor-carry, copy-of-copy composition, N:M products, self-edges, and every `Leak` class are pinned by the module unit-test matrix.

Id uniqueness within a tree is what makes a `NodeId` an *identity*, so it is checked rather than assumed: `assert_unique_node_ids` (with `duplicate_node_ids`) is gated at the post-lowering boundary, immediately before the fold that keys attributions by id. Lowering's own copy sites — uncurry's template discharge, the chained-comparison operand freshens — are what make that a live risk there. Monomorphization freshens every cloned node id (`specialize_use`) so specializations never collide with the original definition; `uniquify` asserts id preservation across its rename.

## Release diagnostics

`infer` returns `Result<Type, Vec<LocatedInferError>>`: each error carries the blame node whose rule raised it, and `compile_program` resolves that node one-hop against the `lowering_projection` into a span-carrying `CompileError::Infer` rendered as an ariadne report.

The node is **mandatory, not optional**. Errors are raised through the inference context (`Typing::raise`), which stamps the node whose rule is currently running, so an unlocated inference error is unrepresentable rather than discouraged — there is no side channel, and nothing positionally aligned to keep in step. The same rule applies to the coalesce walk, which already accumulated several errors and previously rendered all of them without source context.

`CompiledProgram` retains the pre-inference / post-inference / post-desugar IR snapshots, the surface AST, the source text, and the `lowering_projection` the inspector consumes. No pass records steps yet: the recorder surface carries `dead_code` allows removed once the passes adopt the recorder.

## The review round

- **Design doc + reference hygiene.** `design-provenance.md` → `design/provenance.md`, retitled, linked from the design hub and `ccl/CLAUDE.md`; the wire section rewritten to describe what the wire carries instead of asserting a version number. The `TypedExpr: Hash` claim corrected in the doc *and* in `provenance.rs` — there is no `Hash` impl; the hand-written `PartialEq` is what excludes `node_id`. The `§`-style cross-doc refs this PR had added are gone (`ccl/CLAUDE.md` bans them): in-repo targets use the form `./ci.sh doc_refs` validates, vault targets have their substance inlined. Two of those refs were also pointing at the wrong section, and one cited a heading that does not exist.
- **Alloc-free child walkers.** `walk_children`/`walk_children_mut` own the two child-structure matches and allocate nothing; `child_exprs` is derived from `walk_children`; `child_exprs_mut` is deleted (no callers outside `expr.rs`). This reverses a regression the substrate commit introduced — `walk_children` had become an iteration over a freshly allocated `Vec`, on a function called from every recursive tree walk. The fix is a widened lifetime (`walk_children<'a>(&'a self, f: impl FnMut(&'a TypedExpr))`), which lets a child borrow escape the closure and so removes the reason the allocating form existed.
- **Blame at the raising frame**, then **raise through the context.** `current_node_id` is no longer left dirty for a later reader; it is restored on every path and read only at raise time. Coalesce and the scope-validity walk get a blame node per error.
- **Test-only location stripping.** `LocatedInferError::bare` is test-gated, so no production affordance exists for discarding the error↔node pairing.

## Review round 2

- **`preserve` as a construction mode.** Every id-preserving rebuild used to go `Expr::new(…)` — which mints and fires `on_mint` — and then overwrite the id, so the log recorded a birth for an id that ended up on no node. No leak class catches that, because they all enumerate from the tree. `Expr::preserve(id, node)` enters the id at construction and `with_node_id` is deleted, making it unrepresentable rather than detectable. The sequencing constructors move to the module that owns the node kinds, in mint/preserve pairs (`Expr::expr_stmt` now carries `body.ty`, since an `ExprStmt`'s type *is* its body's), which lets `mut_elim`'s two local copies be deleted rather than moved. Twelve struct literals that minted a raw `NodeId::fresh()` — bypassing `Expr::new`, so the node got an identity with no recorded birth — now go through the constructors, or through `Expr::throwaway` where they were only building a node to render in a panic message.
- **`StepKind` collapses; `copy_frame` names the copy-only shape.** `StepKind::Copy` — the declared-origin copy — was never constructed outside the recorder's own tests, and it was superseded rather than merely unused: `freshen_node_id` is the single funnel every duplication goes through and it fires `on_copy`, so a frame *discovers* copies with exact per-node origins where a declared origin asserts one origin for all births. With it gone the enum had one variant, so `step` takes `consumed` directly. `copy_frame(label)` names the lowering frame that consumes nothing and mints nothing; its blame/nature were inert there, and the two call sites had been passing *different* natures with a comment each explaining the argument did not matter.
- **One structural rule for `Nature::Source`.** Two rules were in play and disagreed: `lower_expr` re-tags every lowered expression root as `Source` (positional), while each interior mint site picked by judgment. The rule is now stated once and is positional — **`Source` ⟺ the node is the root of a lowered `Spanned<ChlExpr>`** — so no lowering arm makes the call; 48 sites move to `tag_image`. The cost is explicit: an interior image (a call's callee, each comparison of `a < b < c`, every statement-level image) is no longer `Source`, and a comprehension's `Cast` root now is. Nothing is lost from the record — all of them keep the `"lower.image"` label, which is the datum consumers read.
- **Dead code, measured.** Nine of the fifteen `dead_code` allows in `lineage.rs` were unnecessary — the compiler confirms in both CI profiles — and the five that remain now say what makes each live. `map_node` is deleted (zero call sites). Removing an allow also un-roots its exclusive callees, which is why the count is nine rather than seven.
- **The serde feature was compiled by nobody.** `serde` is default-OFF and no gate pass turned it on, so the hand-written `Serialize` impls were never type-checked. The gate encodes a real cross-crate fact, so the fix is a pass: `./ci.sh clippy_serde`. The workflow also gains the `doc` step `ci.sh` already ran locally — that is what makes a `[`identifier`]` doc reference a *checkable* citation, the same argument as citing a heading by text rather than by section number.

There is deliberately **no produced-side leak class** ("every id a step claims to produce is carried by some node"). It is not decidable against the domain the fold works over: lowering tags the nodes inside a refinement predicate, but the fold's node set is the `walk_children` domain, which excludes predicate interiors, so every predicate id would read as a violation. `design/provenance.md` records that, and the two cases such a check would report once the domains agree.

## Review round 3

- **The `label` axis is bounded rather than re-decided.** Making `Nature::Source` structural *moved* the judgment call onto `label` instead of removing it, and the disagreement moved with it — an `ExprStmt` at a statement's span is `"lower.image"` at five sites and `"lower.stmt_seq"` at nine. Rather than invent a label rule (which would mean relabelling one of those groups — a lowering-semantics change this PR should not carry), the docs now **state the weak guarantee**: `"lower.image"` means only *the rule that minted this node considered it an image*, and no consumer may read it as a cross-site classification. The whole nature/label taxonomy is marked **provisional** — nothing branches on it today, which is exactly why it drifted into two rules. One genuine double-attribution is fixed: `stmts.rs`'s bare-write arm tagged the `MutWrite` *and* its enclosing `ExprStmt` `"lower.image"` at the same span.
- **Doc drift the `Nature` commit introduced.** `lower_expr` still stated the old *semantic* rule, contradicting `tag_source`'s "no arm calls this"; the chained-comparison comment claimed `Nature::Source` directly above a `tag_image` call; and `lowering_projection`'s field doc presented a two-way split when there are three tag shapes, `Machinery` + `"lower.image"` being the common one. These mattered more than usual because that commit's stated goal was that the rule be applicable without reading the fold.
- **The preserve surface: the visibility half, not the field half.** `preserve` / `throwaway` / `expr_stmt_preserving` drop to `pub(crate)` — free (all twelve callers are inside `crate::ccl`) and it closes a real hole, since `pub mod ccl` + `pub use expr::*` let an external crate construct a node at another's id. Making the field private is **not** taken: it requires eliminating every field-wise rebuild, which is churn that would make the code worse. The field doc instead states the real invariant — a literal *mints nothing*, so what is forbidden is `NodeId::fresh()` inside a literal, not a write to the field.
- **Two counts that were generous to me.** The `TODO(preserve)` set was one site short, not twenty: of 41 id-preserving literals only **5** reach into another node for its id (the shape `preserve` is for; four were marked, `transact_phase`'s was the miss), while **34** are field-wise rebuilds that deliberately stay literals. Folding those 34 in would be a regression — the literal's exhaustive field check is load-bearing, where the chained form silently defaults `ty` to `Hole` — and the `Option`-taking setter it needs would have no caller of its own, which is the same family-symmetry argument that deleted `map_node` and `child_exprs_mut` last round.
- **A fourth clippy pass.** The `test-helpers` fix reopened the same hole for a different configuration: the self dev-dependency unifies that feature into the lib under `--all-targets`, which *every* clippy pass used, so the library was only ever compiled with `test-helpers` ON. `./ci.sh clippy_lib` closes it. `CLAUDE.md` described "BOTH clippy passes" in five places; it is now a table of what only each of the four covers.
- **The gate was red, and the pipe hid it.** `./ci.sh` failed at `doc_refs` on the pre-round-3 tip — the rebase commit cited *"The id domain"* three times, which was a bold bullet in `design/provenance.md`, not a heading (now `### The id domain`). The first run here was `./ci.sh | tail`, which reported exit 0 on the red tree. A citation this stack added (`solver/scheme.rs` → "One known exception, scoped and unfixed") also did not resolve; that lead is promoted to a real heading. Verified against #47's checker on a merge probe: no conflicts, and its stricter doc-prose citation check passes on the merged tree, so this can land in either order.
- **`Expr::expr_stmt`'s type carry changed behaviour at an unremarked site** — `channelize`'s lifted-prefix spine went from `Hole` to the body's type. A latent fix (`Hole` is `has_type_residue`), now pinned by a unit test. `mut_elim::hoist_feeds` also implemented that carry twice; both copies are gone.

`./ci.sh` is green on every commit — all four clippy passes (debug, release, `serde`, lib-only) plus the doc build and the doc-reference check.

Followed by #44, which extends the same rule to the post-inference check family so `Mut`-discipline and ambiguous-program diagnostics render with source context too.


src/ccl/design/provenance.md
groundlar, line 1: Move to `design/`, cleanup some opaque LLM prose like "schema 3"

src/ccl/infer/api.rs
```diff
@@ -146,6 +146,17 @@ pub struct TypeInferenceContext {
...
+    /// Node ids for the inference errors returned by the last [`infer`] run,
+    /// aligned **positionally** with the returned `Vec<InferError>` (entry `i`
+    /// is the blame node for error `i`, `None` when unknown). Inference stays
+    /// ignorant of the source projection; it only surfaces these ids, and
+    /// `compile_program` drains them via
+    /// [`take_infer_error_nodes`](Self::take_infer_error_nodes) to resolve each
+    /// to a source span (against the `lowering_projection`) while it is in scope. This
+    /// is the side-channel that lets the public [`infer`] keep its
+    /// location-free `Result<Type, Vec<InferError>>` signature unchanged.
+    infer_error_nodes: Vec<Option<crate::ccl::provenance::NodeId>>,
```
dpmills: Aligned vectors across multiple separated parts of code feels brittle; any better options? Flowing a side channel out of the context makes the the infer api more complicated; why is keeping the infer api the same important?

src/ccl/lower/exprs.rs
```diff
@@ -438,6 +517,50 @@ mod tests {
         assert_eq!(symbolic(&ccl), expected);
     }
 
+    /// Regression: a chained comparison shares each middle operand between two
+    /// adjacent pairs. A bare clone used to put the same `NodeId`s in the tree
+    /// twice, tripping `assert_unique_node_ids` at the post-inline boundary. The
```
dpmills: assert_unique_node_ids isn't defined yet.  Should that be moved down into this PR?

src/ccl/expr.rs
```diff
@@ -921,80 +1133,87 @@ impl TypedExpr {
...
+        }
+        out
+    }
+
+    /// Invoke `f` on each direct child Expr by shared reference, in child order.
+    ///
+    /// Delegates to [`child_exprs`](Self::child_exprs) — the single immutable
+    /// child-structure definition — so the two cannot drift out of sync. Leaf
+    /// nodes yield an empty `Vec` (no allocation); only internal nodes pay a
+    /// small transient `Vec`, which is immaterial for compile-time tree walks.
```
dpmills: Is "immaterial" really true here? This is an extra allocation for every single expr in the tree, in a function that is called a huge number of times through the codebase.  Is the refactor really necessary? Worth pushing for alternatives.  If there aren't any, we could probably use some sort of small-vector optimization to keep exprs with a small number of children on the stack (SmallVec<2> perhaps).  Nearly all exprs have <=2 children

src/ccl/design-provenance.md
```diff
@@ -0,0 +1,165 @@
...
+> The design of record — the full decision log, the collapse algorithm, the
+> recorder mechanism, and the adoption sequencing — lives in the lineage-redesign
+> design doc in the internal vault. This file summarizes
+> the shipped shape; where they disagree, that doc wins.
+
+## The two identity primitives (`src/ccl/provenance.rs`)
+
+- **`NodeId`** — a `Copy` newtype giving each IR expression node a stable,
+  never-reused identity (its own atomic counter, distinct from `Uid`). It rides
+  inline on `TypedExpr` and is **excluded from `PartialEq`/`Hash`** — provenance
```
dpmills: I don't think TypedExpr has Hash

src/ccl/design/provenance.md
```diff
@@ -0,0 +1,165 @@
+# provenance & lineage — design and implementation
+
```
dpmills, line 2: Capitalization.  Also, mention this briefly from at least one higher level doc so that it is findable

src/ccl/lineage.rs
```diff
@@ -0,0 +1,2072 @@
...
+    attr: &SourceProjection,
+) -> SourceAttribution {
+    // No *pass* step may carry `Nature::Source`: it is
+    // lowering-only. The guard sits on the attributing ARM, not on projection
+    // entries — an *inherited* Source tag on a preserved id in a later pane is
+    // legal and reaches the projection by clone, never through this fn.
+    debug_assert!(
+        !nature.is_source(),
+        "a pass step carries Nature::Source (label {label:?}, via {via:?}) — \
+         Source is lowering-only (§2.5)"
```
dpmills: We should either cite heading text which can be automatically maintained or remove refs like this entirely (there are several places like this in the PR)

src/ccl/infer/context.rs
```diff
@@ -80,6 +80,13 @@ pub(super) struct InferCtx {
     /// `in_let_rhs`) so RHS-local variables are minted deeper than the
     /// defining scope and become generalizable at the binding site.
     pub(super) level: Level,
+    /// The `NodeId` of the node currently being emitted, maintained by the
+    /// [`emit_node`](super::emit::emit_node) wrapper: set on entry, *restored
+    /// only on success* so that on the fail-fast error path it is left pointing
+    /// at the innermost node that failed. This is provenance metadata for
+    /// diagnostics, surfaced to the `compile_program` boundary via the public
+    /// `infer`'s side-channel — it never influences inference itself.
+    pub(super) current_node_id: Option<crate::ccl::provenance::NodeId>,
```
dpmills: Is this safe? Seems like it assumes single errors, but we're trying to have as much of our code be able to return multiple errors as possible.  We should design APIs with that in mind

src/ccl/provenance.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+// Wire shape (inspector, feature `serde`): a bare JSON number. A `NodeId` is an
+// opaque handle the client round-trips, so it serializes as its underlying
+// `u64` (the one place the numeric value is observed — see [`NodeId::as_u64`]),
+// not as a struct. Hand-written rather than `#[serde(transparent)]` because the
+// inner field is private.
+//
+// TODO(wire-stability): shipping the raw mint-order value makes the golden
```
dpmills: This could end up quite annoying; should we just address it now?

src/ccl/design/provenance.md
groundlar: Both done.

Moved to `src/ccl/design/provenance.md` next to the sibling design docs.

On "schema 3": you were right that it was opaque, and the section was worse than opaque — it named a version whose 1 and 2 appear nowhere in the repo, and it duplicated the "Inspector consumers" section above it. The two are merged into one section that says what the wire *carries*, with no version number, plus a sentence on where versioning actually lives (`meta.schema`, owned by the inspector crate — nothing under `ccl/` reads it).

While in there I also dropped a dangling reference the doc-refs checker could not catch: `provenance.rs` cited *"the design-provenance doc's `Pass` and `SyntheticKind` stay separate section"*, and no such section exists. The rationale was already stated in full at the citation site, so the pointer is gone rather than repaired.

src/ccl/design/provenance.md
groundlar: Done, both halves.

Title is now `# Provenance & lineage — node identity and source attribution`, matching the sibling docs' sentence-case style (`# Lowering — CHL to CCL`, `# Mutability, transactions, and feeds`).

Findability: a row in the `src/ccl/design/README.md` doc table (the hub `/docs/design.md` links to), plus a sentence there marking it as the one cross-cutting concern in that table — every pass in the pipeline both preserves node identity and records what it rewrote. And a new "Node identity and lineage" section in `src/ccl/CLAUDE.md` pointing at it, since that is what an agent reads before changing how a pass constructs or clones nodes.

src/ccl/design-provenance.md
groundlar: Correct, thanks — `TypedExpr` has no `Hash` impl at all. `#[derive(Debug, Clone)]` plus a hand-written `PartialEq`, and it is that `PartialEq` which deliberately excludes `node_id`.

Fixed in the doc *and* in `provenance.rs`, whose "Identity-contamination rule" section made the same wrong claim: it now says the hand-written `PartialEq` exists precisely to exclude the id, notes that nodes are never hashed by value, and distinguishes that from `NodeId`'s own `Hash` (which is there so it can be a map key).

src/ccl/expr.rs
groundlar: You were right on both counts: it is a real regression, and "immaterial" was hand-waving. Before this PR `walk_children` was a direct `match` with no allocation; the PR inverted it to iterate `child_exprs()`, so every internal node heap-allocated on recursive paths like `count_free_with_visited`, `collect_expr_fv`, and channelize's walks.

I went with neither the status quo nor `SmallVec` — the allocation turns out to be avoidable outright, so there is nothing to optimize. Fixed by **inverting the delegation**: `walk_children` / `walk_children_mut` own the two matches and allocate nothing, and `child_exprs` is derived from `walk_children`.

The reason the allocating form seemed necessary was a signature detail. The old

```rust
pub fn walk_children(&self, f: impl FnMut(&TypedExpr))
```

desugars to a higher-ranked bound, `for<'x> FnMut(&'x TypedExpr)`: the closure must work for *any* lifetime, chosen per call by the callee, so a child borrow provably cannot escape into an enclosing binding — which is why find-node-by-id needed the `Vec`. Naming the lifetime and tying it to `self`:

```rust
pub fn walk_children<'a>(&'a self, f: impl FnMut(&'a TypedExpr))
```

makes the borrow outlive the call, so a value-returning recursion can go through the closure form. I verified that (including the `&'a mut` fan-out, which stays sound because the borrows are of disjoint fields) in a standalone repro before touching `expr.rs`.

Two consequences worth flagging:

- **`child_exprs_mut` is deleted.** It was `pub` with zero callers outside `expr.rs` — it existed only to *be* the primitive — so inverting it would have left an allocating public method nobody calls. Four lines to restore if you want the pair symmetric.
- `child_exprs` survives for the callers that want a slice: one unit test and five sites in `src/inspector_model/query.rs`, i.e. the inspector's cold snapshot path, where a `Vec` genuinely is immaterial — and the docs now say that rather than claiming it globally. Only `find_node` there was structurally forced, and with the widened lifetime even it no longer is.

The drift concern the refactor was for is fully preserved: the match arms still live in exactly one place per mutability.

src/ccl/infer/api.rs
groundlar: Agreed on the brittleness, and the answer to your question is: it was not important. That was inertia, not a requirement — so the side channel is gone.

`infer` now returns `Result<Type, Vec<LocatedInferError>>`. `infer_error_nodes`, `take_infer_error_nodes`, the positional-alignment contract, and the `.chain(repeat(None))` length-skew guard in `compile_program` are all deleted. That guard is the tell you were picking up on: it existed only because the alignment could not be checked.

The cost was small — exactly one non-test caller inspects those errors — and it is not just tidier, it is more capable: coalesce (pass 2) *already* returned several errors, and under the old scheme every one of them got `node_id: None`, so no coalesce error ever rendered with source context. Now `1 + "a"` underlines the expression.

Two follow-through details, since they answer the "any better options" part:

- The location is **mandatory**, not `Option`. Errors are raised through the inference context (`Typing::raise`), which stamps the node whose rule is running, so an unlocated inference error is unrepresentable rather than discouraged — no partially-constructed intermediate, no `debug_assert` standing in for an invariant.
- It stays a wrapper beside `InferError` rather than a field inside it, because `InferError: PartialEq` is load-bearing (about a dozen whole-value `assert_eq!` comparisons), and hand-built test trees have no lowering projection for a span to resolve against. The wrapper is flattened into `CompileError::Infer { error, span }` at the boundary once the span is known — node-id-keyed inside the compiler, inline span at the diagnostic boundary.

The check family got the same treatment in the follow-up (#44), so this is now one rule with no exceptions rather than two conventions.

src/ccl/infer/context.rs
groundlar: Fair, and the smell was real: that design only worked because *emission* is fail-fast, and it was already failing the multi-error cases — coalesce and the scope check both return `Vec` and every one of their errors got `node_id: None`.

Restructured so the blame does not depend on fail-fast at all. Errors are attributed **where they are raised** rather than by reading a cursor after the walk unwinds: `Typing::raise` is the only way to build a located error, and it stamps the node whose rule is currently running. `current_node_id` is maintained by each walk's wrapper (set on entry, restored on exit, error path included) and is read only at raise time — so it never holds state for a later reader to interpret, which was the actual problem with the old "restore only on success" trick.

Consequences:

- **Coalesce gets genuine per-error blame.** `CoalesceCtx.errors` holds located errors, each stamped by the frame that raised it, with no parallel vec and nothing to keep aligned. Same for the scope-validity walk.
- **Check mode is symmetric.** It implements the same affordance, so its accumulated errors carry nodes too — and, worth noting for the "we want to accumulate as much as possible" goal, `CheckCtx` is the existence proof that the *shared* typing rules support accumulation: it runs the same `emit_*` family and records every failure instead of returning at the first.
- **Emission stays fail-fast for now**, with `TODO(multi-error)` on `emit_node` naming the three decisions it needs and does not get for free: a poison value to continue with after a failed constraint, cascade suppression so one bad subtree does not report at every ancestor, and what to do with the run-wide `ConstrainCache` after a failed constraint has written into it. The blame mechanism is already accumulation-ready — N errors get N nodes with no further change — so that work is independent of this thread.

Also pinned by test: an emit error is blamed on the node that raised it, not an ancestor that propagated it (the `Var`, not the enclosing `BinOp`).

src/ccl/lineage.rs
groundlar: Agreed — and `ccl/CLAUDE.md` already bans these ("Section numbers across docs (`§4.6`) — cite the heading instead"), so this PR was adding refs the conventions forbid. All five that it introduced are gone:

- The two `§2.5` guards here, and three `§2.14` refs in `lower/functions.rs`.
- Where the target is an in-repo doc, they now use the checkable form `./ci.sh doc_refs` validates: the doc path plus the exact heading text in quotes, e.g. `` see `design/provenance.md`, "The lineage model (`src/ccl/lineage.rs`)" ``.
- Where the target is the vault (nothing can check that), the substance is inlined instead of pointed at — so the `Nature::Source` guard now states *why* Source is lowering-only, and root-carry is defined where it happens rather than cited.

Two things this turned up. The `§2.14` refs on the root-carry comments were simply **wrong** — §2.14 is the `SourceAttribution` decision; root-carry is §12.4(7) — which is the argument for your point better than I could make it. And I left the ~7 pre-existing `§` refs in `channelize.rs` / `lower/loops.rs` / `lower/stmts.rs` alone, since CLAUDE.md calls that a separate migration; three more live in the inspector commits upstack and will get the same treatment when those PRs come round.

src/ccl/lower/exprs.rs
groundlar: Good catch, and yes — moved down into this PR.

It belongs here: the function is a standalone tree walk with no dependency on the recorder adoption, and the invariant it checks (id uniqueness within a tree) is *this* PR's contribution — the `chained_compare_freshens_shared_operands` test above is a regression test for exactly that, and it was hand-rolling the same walk the checker exists to perform. It now calls the real checker (`assert_unique_node_ids` plus `collect_tree_ids`, both `pub(crate)`), so the test asserts through the mechanism instead of duplicating it.

It also has a live call site here rather than sitting unused: `assert_unique_node_ids(&expr, "post-lowering")` immediately before `collapse_lowering`. Uniqueness is the precondition for keying anything by `NodeId`, so gating it right before the fold that does exactly that is the honest place — and lowering's own copy sites (uncurry's template discharge, the chained-comparison operand freshens) are what make it a live risk at that boundary.

The adoption PR keeps its four downstream boundary calls, and its description no longer claims to introduce the gate.

src/ccl/provenance.rs
groundlar: Agreed it will be annoying, and it is now settled rather than a "likely fix" — but I would not land it at the bottom of the stack, for three reasons:

1. **Nothing in `ccl/` is affected.** The renumbering is a serialization-time transform over a built snapshot; `NodeId` stays the internal identity. No pass, fold, or projection changes.
2. **The churn does not exist yet at this depth.** The corpus and both byte-exact gates arrive with the inspector-crate commit. There are no fixture bytes to protect here, and the code would reference snapshot structures that do not exist until the inspector-model commit.
3. **It gets cheaper by waiting.** The fixture-hardening work slims `paneLinks` out of the goldens and bumps the schema in that same region, which *reduces* the surface the walk has to cover.

So: it lands where the snapshot is serialized, with the re-bless against the corpus. The cost of waiting is bounded and already being paid (whole-corpus id diffs on each re-bless, with the classify-every-diff discipline as the mitigation). One thing that would change the answer: if this and the adoption commit merge to `main` while the inspector commits are still in review, then every change here re-blesses the whole corpus in the downstack PRs — in that sequencing I would pull it in before the inspector-crate commit merges.

I did tighten this comment: it now states the mechanism in one paragraph (a deterministic snapshot walk assigning dense first-encounter indices, `NodeId` staying internal) with no forward pointer to a specific commit, since that is the uncheckable-reference class from the other thread.

The two constraints that were buried here — the numbering must be snapshot-**global**, not per-pane (or `paneLinks` edges stop connecting, and the frontend's untagged `Set<number>` collections assume cross-pane uniqueness), and it must apply to every id-bearing wire position including edge endpoints — moved to the plan's canonicalization-walk ticket in the vault, with the justification that neither is a property of `NodeId`: both are properties of what reads the wire, so they belong with the walk that has to satisfy them.

dpmills:
Second pass. I re-verified each of the round-1 threads rather than taking the replies at face value; all eight are genuinely addressed, several better than what I asked for. Full `./ci.sh` is green on the head commit.

One new top-level question first, then the smaller findings.

## `Nature::Source` vs `Nature::Machinery` — what is the rule?

Reading the lowering code I can't reconstruct why a given site picked one over the other, and I think that's because there are really **two rules in play that disagree**:

- **Positional.** `lower_expr` unconditionally re-tags the *root* of every lowered `Spanned<ChlExpr>` as `Source` (`lower/mod.rs`, `lower_expr` → `tag_source(lowered, expr.span)`), "overwriting any interim tag an arm put on its own root". Under this rule `Source` means *"sits at the root of a lowered source expression"* — a positional fact, mechanically decidable, and the fold's `reimage_ok`/last-tag-wins path exists specifically to let it win.
- **Semantic.** At each interior mint site the author picks by judgment whether the node "is a direct image of something the user wrote" (`tag_machinery`'s doc: *"a one-to-one translation of something the user wrote"*).

These collide exactly where it's most visible. `lower/comprehension.rs` has **zero** `tag_source` calls and ten `tag_machinery` — every node is manufactured under the semantic rule — and then the positional rule flips whichever of the outer `Lambda`/`Cast` ends up at the root back to `Source`. The comment at `comprehension.rs:195-197` says as much. So "is a comprehension's root a direct image?" has opposite answers depending on which rule you read, and the node that ends up tagged `Source` is a `Cast` wrapper the user never wrote.

To be fair to the current state: source *linkage* isn't lost in a comprehension — every machinery node carries a real span (`gspan` / `element_span`), and the user-written element and body expressions go through `lower_expr` separately and keep their own precise spans. What's lost is **granularity**: the whole scaffold collapses onto the generator's iterable span or the element span, so anything raised inside it underlines the entire generator rather than the sub-expression. That's the part I'd want to understand the intended behaviour of.

So, concretely: **is there a concise, precise rule for choosing between the two?** Some possibilities, in rough order of how much I'd like them:

1. `Source` ⟺ this node is the root of a lowered `Spanned<ChlExpr>`. Purely positional, so `tag_source` stops being a judgment call — arms only ever call `tag_machinery`, and `lower_expr` owns the one `Source` emission. The `Cast`-is-a-Source oddity remains but becomes *defined* rather than accidental.
2. `Source` ⟺ the node's span is exactly the construct's span and its kind corresponds 1:1 to a source construct. Semantic, matches the docs, but then `lower_expr`'s blanket re-tag has to go and comprehensions legitimately have no `Source` node at all.
3. The distinction isn't carrying its weight yet — `tag_machinery`'s own doc says no consumer behaves differently on the bit, `Source` null-compresses on the wire, and `Nature::Expansion` is unconstructed outside `lineage.rs`'s tests. If the per-site `label` is the primary datum, collapse `Nature` to what's actually decidable and recover a finer taxonomy from a label-keyed remap when a consumer earns it.

Whichever it is, I think it wants to be stated once, in `design/provenance.md` and on `tag_source`, in a form a person adding a lowering rule can apply without reading the fold. Right now the guidance is split across `tag_source`, `tag_machinery`, `lower_expr`, and `RootTracker::transform`'s `reimage_ok` parameter, and no single one of them is sufficient.

## Round-1 threads — verified

| Concern | Verified |
|---|---|
| Aligned vectors in `infer` | `infer` → `Result<Type, Vec<LocatedInferError>>`; `infer_error_nodes` / `take_infer_error_nodes` / the `.chain(repeat(None))` skew guard are all gone |
| `assert_unique_node_ids` not defined yet | Moved down, live call at `context.rs:756`, and the chained-compare test calls the real checker instead of hand-rolling it |
| `walk_children` allocation | `walk_children` / `walk_children_mut` own both matches and allocate nothing; `child_exprs` derived; `child_exprs_mut` deleted. The widened `'a` does what the reply claims |
| `TypedExpr: Hash` | Corrected in the doc *and* `provenance.rs` |
| Doc title / findability | Retitled; linked from `design/README.md` and `ccl/CLAUDE.md` |
| `§` cross-doc refs | Zero `§` refs added anywhere in the diff |
| Single-error assumption | `Typing::raise` stamps at the raising frame; `emit_node`, `check_node`, and `coalesce_node` all restore the mark on every path including errors; coalesce / check / scope-walk accumulate with per-error blame, pinned by end-to-end span tests for emit, coalesce, unbound-var, and ariadne rendering |
| Wire-stability TODO | Deferred with a real argument and a named trigger condition — fine by me |

The multi-error rework and the walker inversion in particular came back stronger than what I asked for.

## New findings

**1. Orphaned doc comment in `uniquify.rs`.** `collect_node_ids` was inserted between `assert_all_binders_minted`'s doc block and the function itself (`uniquify.rs:304-321`). The Barendregt-invariant paragraphs now render as the opening of `collect_node_ids`'s docs, and `assert_all_binders_minted` has none.

**2. Seven of the fifteen `#[allow(dead_code)]` in `lineage.rs` are stale.** I removed them and built: `Op`, `LineageLog`, `StepKind`, `OpenStep`, `group_copies`, `step`, and `StepGuard` are all live. Same for `TypedExpr::with_node_id`, whose allow says *"Used by the pass-adoption layer; unused within the substrate alone"* — it has six call sites in `mut_elim.rs` in this PR. These allows are the stated marker for "delete when the passes adopt the recorder", so a wrong one is worse than none: it hides which surface is actually unreachable. Genuinely dead today: `RewriteStep`, `attribute`, `collapse`, `flush_into`, `RecorderSession::new` / `into_log`, `Nature::Expansion`, `StepKind::Copy`, `map_node`, `wire_str`, and `child_exprs` (one test caller).

**3. `map_node` is speculative retention.** Its doc says *"Currently unused in-tree… Retained as the owned-transform member of the traversal family."* Zero call sites. Family symmetry isn't the `CLAUDE.md` bar for an abstraction. The PR description also claims ids are "preserved by `with_node_id`/`map_node` across every pass rebuild", which is only half true.

**4. `with_node_id` mints a phantom id — worth settling before the adoption PR.** Every call site is `Expr::new(…)` (fires `on_mint`) `.with_ty(…).with_node_id(preserved)`, so a real id is minted and thrown away. Same shape in `subst.rs:565-570`: a `Mapping::Rename` builds a `Var` via `TypedExpr::var` and then overwrites its id with `occurrence_id`. Today nothing captures it, but once `mut_elim::run` and friends are wrapped in `step()` frames, each of those seven sites records a `produced` id that exists nowhere in the tree. No `Leak` class catches it — a phantom is live, is not an input, and never reaches the output, so `Dropped` / `Unexplained` / `ProducedLive` all stay silent.

This reads as a missing primitive rather than a bug to patch: *preserve* is a construction mode, and there's currently no way to express it. `channelize.rs` and `mut_elim.rs` already hand-roll it ~20 times as `TypedExpr { node, ty, user_annotation, node_id: preserved }` struct literals — which is most of why those diffs read as churn. An `Expr::preserving(id, node)` that constructs *without* calling `NodeId::fresh()` would collapse that duplication, make `with_node_id` unnecessary, and make the phantom unrepresentable rather than merely absent.

**5. `design/provenance.md` describes a system that doesn't exist yet.** It calls itself "the as-built reference", but `CompiledProgram::pass_lineage`, `materialize_panes`, `SpanIndex`, `build_inspect_tree`, `src/inspector_model/`, `stage::dense_edges`, and "the census ratchet" are all absent from the tree — I grepped for each. `doc_refs` validates doc→doc and code→doc, not doc→code, so nothing catches it. Since we squash-merge, the version that lands on `main` is roughly half aspirational. No need to move anything up the stack — just mark those sections as planned rather than as-built, so a reader on `main` can tell which is which.

**6. Dangling identifier references — the same uncheckable class as the `§` refs, all added by this PR:**

- `context.rs:630` cites `collect_ids`; it's `collect_tree_ids`
- `context.rs:1198` cites `CoalesceCtx::blame_frame`; the field is `current_node`
- `context.rs:1164` cites `provenance.origins` — no such thing
- `lower/exprs.rs:522` — "tripping `assert_unique_node_ids` at the post-inline boundary"; there is no post-inline call site in this PR
- `Cargo.toml:18` — "added in I3", a transient ticket identifier
- `lineage.rs:263` — "the native schema-3 attribution". This is the "schema 3" prose the very first review comment asked to remove; fixed in the doc, missed here

**7. The `serde` feature is never built by CI — but first, does it need to be a feature?** `ci.sh` has no `--features` / `--all-features` pass, so the hand-written `Serialize for SourceAttribution`, `Serialize for NodeId`, and `Nature::wire_str` are compiled by nobody (`wire_str`'s dead-code warning in a default build is the symptom). Before adding a CI pass, I'd rather we decided whether the gate earns its keep: it buys skipping a `serde` + `serde_derive` compile for a crate that already pulls in `logos`/`chumsky` proc-macro chains, and it costs a `cfg`-divergent build configuration that nothing checks. If the answer is "just depend on serde unconditionally", these three items stop being conditionally-dead and the CI question disappears. If we do keep the feature, it needs a build in the gate — I confirmed both `cargo build --features serde` and `cargo clippy --features serde --all-targets -- -D warnings` pass today, so it's cheap either way.

**8. `mut_elim.rs:1233` re-rolls the uniqueness walk.** The test-local `duplicate_ids` duplicates `context.rs::duplicate_node_ids` — the exact duplication removed from `lower/exprs.rs` this round. `assert_unique_node_ids` is `pub(crate)` and reachable from there.

### Minor

- `on_copy` doesn't guard `NodeId::PLACEHOLDER` the way `on_mint` does. Probably unreachable, but the asymmetry is free to fix and would otherwise surface as a confusing `CopyOfUnknown`.
- `step()`'s `nature` argument is inert for copy-only lowering frames — the two call sites pass *different* values (`Source` at `exprs.rs:387`, `Machinery` at `functions.rs:247`) and each needs a comment explaining the argument doesn't matter. `flush_into_lowering` already asserts such a frame has no births, so its nature is provably unread by `collapse_lowering`'s `Copy` arm. A `copy_frame(label)` constructor would remove both comments and the possibility of a misleading `Nature::Source`. (This also feeds the top-level question above — a `Nature` you can pass arbitrarily with no effect is a sign the axis isn't pinned down yet.)
- The 10-line `bare` helper is duplicated verbatim in `tests/inference_variants.rs` and `tests/type_check.rs`; the crate already has a `test-helpers` feature for exactly this.

Nothing here blocks on correctness — the fold, the leak taxonomy, and the blame plumbing all hold up under reading, and the test matrix is thorough. The `Nature` question and items 4 and 5 are the ones I'd want settled before this merges, since the adoption PR builds directly on all three.

groundlar:
Thanks — the re-verification is appreciated, and the `Nature` question was the right thing to lead with. Three commits on top: the round-2 fixes, the `Nature` rule, and the CI/test items. Round-1 threads resolved.

## `Nature::Source` — option 1, and it was cheaper than either of us priced it

Gone with **the structural rule**: `Source` ⟺ the node is the root of a lowered `Spanned<ChlExpr>`. Decided in one place (`lower_expr`'s re-tag, last tag wins through the fold's re-image path), so no arm makes a judgment call — `tag_source` is no longer called by any arm at all.

The migration turned out to be free of the naming work I expected, because `tag_source`'s body was already `lowering_leaf(id, span, Nature::Source, "lower.image")` — **every one of the 48 sites was already emitting the same label**. So they move to a new `tag_image` (Machinery nature, same `"lower.image"` label) and *only the nature changes*. No new label vocabulary, no coverage change, and the fold's `reimage_ok` path stays load-bearing for the one remaining `Source` emitter.

Your framing was better than mine on the cost, so it is written down rather than glossed. On `tag_source` and in `design/provenance.md`: a node can be a one-to-one image of what the user wrote and still not be `Source` because it is not a *root* — a call's callee `Var`, each comparison of `a < b < c`, the projection of `x[i]`, and every statement-level image, since statements are not `Spanned<ChlExpr>`s and therefore have no `Source` node at all. And the converse you spotted: a comprehension's `Cast` root now *is* `Source`, by definition rather than by accident. What leaves `nature` stays in the `label` — which is what `tag_machinery` already claimed was the primary datum, so this makes that claim true instead of aspirational.

One more instance of the same disagreement, which I think argues your point better than the comprehension does: `Expr::expr_stmt` at a statement's span was `Source` at three sites in `stmts.rs` and `Machinery`/`lower.stmt_seq` at eight across `stmts.rs`, `loops.rs`, and `transactions.rs`. Same node kind, same span role, opposite answers — so the codebase was already internally inconsistent on the most common statement node, independent of which rule you pick.

## 4 — the phantom, and why there is no leak class for it

Fixed as you framed it: a missing primitive, not a bug to patch. `TypedExpr::preserve(id, node)` enters the id at construction, `with_node_id` is deleted, and I audited that no mint-then-overwrite path remains.

The sequencing constructors moved to `expr.rs` rather than getting preserve-siblings in `mut_elim`, because the fork between the two was not real: `Expr::expr_stmt` left `ty` as `Hole` while `mut_elim::expr_stmt_typed` carried `body.ty`, but "an `ExprStmt`'s type is its body's" holds in both phases — during lowering `body.ty` *is* `Hole`, so the carry is the identity. Unifying it let `mut_elim`'s two helpers be deleted rather than moved, and the mint/preserve pair now lives in the module that owns the node kind.

**I did not add a produced-side leak class, and I want to be explicit about why**, because I tried it first and it found two real things before I concluded it could not be checked in closed form yet.

It is not decidable against the domain the fold works over. Lowering tags the nodes inside a refinement predicate — they lower through `lower_expr` like anything else — but the fold's node set is `collect_tree_ids`, the `walk_children` domain, which excludes predicate interiors by design (`context.rs:567`). So every predicate id is produced, live, uncopied, and legitimately absent from the output set. A filtered comprehension reports exactly three violations: `Var`, `Lit`, `BinOp` for `x > 1`. Every existing leak class enumerates from the *tree*, which is why predicate exclusion is free for them and fatal for this one — `duplicate_node_ids` excludes predicates too, for the same reason (`context.rs:595`).

The two things it found on the way, both recorded in `design/provenance.md`:

1. **Uncurry's projection template root is an undeclared discard.** Its two interior ids reach the tree as the freshened copies the frame captures, but root-carry replaces the *root* with each occurrence's own id, so that one is tagged and then carried by nothing. The comment at `functions.rs` claimed all three "compose away as born-copied-discarded" — true for two of three. Corrected in place.
2. **Twelve struct literals minted a raw `NodeId::fresh()`**, bypassing `Expr::new` and therefore `on_mint` — the exact dual of the phantom: a node with no recorded birth. Five were building a node only to render it in a panic message (now `Expr::throwaway`, carrying `PLACEHOLDER`, so a diagnostic cannot perturb the log it may be reporting on), two were test fixtures, the rest go through the constructors. I checked whether these were tripping `Leak::Unexplained` at the adoption commit: they were not — the passes holding them open no frames, and the one reachable case (`subst`'s transport rebuild, via `infer.mono_wrappers` → `apply_type`) lands in a predicate interior, which both sides of the ledger exclude. So a latent inconsistency, not a live bug.

The remaining hand-rolled preserves — the `TypedExpr { … node_id: src.node_id }` literals you flagged as churn — are correct as they stand and carry a `TODO(preserve)` each. They never mint, so they were never part of this class. They are, though, exactly why the twelve raw mints were easy to miss: a preserve and a mint differ by one token in otherwise identical five-line literals, in the same two files.

## 2 — the count is nine, and "the allow is unnecessary" ≠ "the item is dead"

Your argument is right and I want to sharpen the measurement, because the two halves come apart.

Measured by stripping all fifteen and running both CI profiles: **nine** are unnecessary, not seven. The extra two are `RewriteStep` and `flush_into`, which are live from this module's tests — and since the gate lints `--all-targets`, an allow is unnecessary if tests reach the item. Worth noting the mechanism: removing an allow *un-roots* its exclusive callees, because rustc treats an allowed-dead item as a live root. Dropping `collapse`'s allow exposes `seeded`/`roots_of`; dropping `RecorderSession::new`'s exposes `ActiveLog::Pass`. That is why the sets do not decompose the way a per-item reading suggests.

The other half is where I did not follow you, and I think deleting would have broken the stack:

| Item | Status here | Upstack |
|---|---|---|
| `child_exprs` | one test caller | **3 production calls** in `inspector_model/query.rs` from the inspector-model commit |
| `collapse` (+ `attribute`, `seeded`, `roots_of`) | test-reachable only | **production** at the two pane boundaries in `context.rs`, from the adoption commit |
| `RecorderSession::new` / `into_log` | test-only | **5 production sites each** from the adoption commit |
| `Nature::Expansion` | test-only construction | **~16 production sites** from the adoption commit |
| `Nature::wire_str` | serde-gated, so compiled by nobody | unconditional call in `query.rs` from the inspector-model commit |

So those five allows are doing exactly the job their comments claim, and the comments are what was wrong — they now say *what* makes each live rather than asserting it, without naming a commit (`ccl/CLAUDE.md` bans stack-position references, which is the same class as the `§` refs from round 1).

`StepKind::Copy` was the one genuinely dead target your argument had, and it is deleted — but "unused" undersells it. It was **superseded**: `freshen_node_id` became the single funnel every duplication path goes through and it fires `on_copy(old, new)` per node, so a `Transform` frame *discovers* copies with exact per-node origins, where a declared origin asserts one origin for all births. The declared form is strictly weaker. With it gone the enum had one variant, so `step` now takes `consumed` directly.

`map_node`: deleted. Zero call sites at any commit in the stack, and family symmetry is not the bar.

## The minors

- **`copy_frame(label)`** replaces the inert `nature` at both frames, so a misleading `Nature::Source` on a copy frame is unrepresentable. Reading the adoption commit's frames while planning the rebase turned up a limit worth stating: the two folds' `Op::Copy` arms differ — `collapse_lowering` mirrors the origin's attribution verbatim (nature genuinely unread), while the *pass* fold builds `RewriteTag { nature: step.nature, … }` from it. So `copy_frame` is documented as lowering-specific; a pass copy-frame must name its nature. `infer.mono_specialize` is precisely that case, and converting it mechanically would have silently retagged every monomorphization copy from `Expansion` to `Machinery`.
- **`on_copy`** guards `PLACEHOLDER` on both sides now, as `on_mint` does.
- **`bare`** — there was already a `#[cfg(test)]` `LocatedInferError::bare`, which integration tests cannot see because they are separate crates, so each had copied it. Widening the gate to `#[cfg(any(test, feature = "test-helpers"))]` plus a self dev-dependency is what the feature was for; both local copies are gone.
- **`uniquify`'s orphaned doc block** reattached; **`mut_elim`'s** test asserts through `assert_unique_node_ids` instead of re-rolling that walk.

## 5, 6, 7 — docs, references, CI

**5.** `design/provenance.md` now carries a status legend, and every section describing unbuilt code is marked **Planned** — the pass adoption, the pane folds, materialization, and the whole inspector-consumers section. You were right that roughly half of it was aspirational while calling itself the as-built reference.

**6.** All six corrected, and one becomes mechanically checkable rather than merely fixed: written as an intra-doc link, `[`collect_tree_ids`]` is validated by rustdoc's `broken_intra_doc_links`, which `ci.sh doc` denies. It caught one of my own dangling references on the first run, which is the best evidence I can offer for the form. The other five cannot be gated that way — two are doc comments on `#[cfg(test)]` items, one is a `//` comment, one is TOML — so they got the round-1 treatment instead. I would rather say that than imply the class is closed.

**7.** Keeping the feature: it encodes a real cross-crate fact, that a separate binary crate opts into the wire types, rather than a build-config preference. So the fix is the pass — `./ci.sh clippy_serde`, wired into the workflow. Worth flagging that the hole was slightly wider than you found: the workflow had **no `doc` step** either (`ci.sh` ran it locally, GitHub CI did not), so the intra-doc-link gate above was not actually gating. Both steps are in now.

**8.** `mut_elim`'s `duplicate_ids` gone; the test calls the real checker.

---

Two things still open, neither in this PR:

- **This PR conflicts with `main`** and did before these commits — #36 landed and touches `infer/api.rs`, `infer/check.rs`, and both inference test suites. Rebasing the stack onto it is its own pass.
- The upstack commits need the API port these commits imply — `StepKind` → `consumed`, `with_node_id` → `preserve`, and the `mut_elim` helpers that are now in `expr.rs`. The adoption commit's diff should get *smaller*, since it currently deletes allows this commit already dropped.

dpmills:
## The `test-helpers` self dev-dependency reopens the same hole, for a different configuration

With `cambra = { path = ".", features = ["test-helpers"] }` in `[dev-dependencies]`, `--all-targets` unifies that feature into the lib. Every clippy pass in the gate uses `--all-targets`, so the library is now **only ever compiled with `test-helpers` ON**, and the feature-off configuration — what a consumer actually builds — is compiled by nothing in `./ci.sh`.

Demonstrated, not inferred. Adding ungated library code that calls the `test-helpers`-gated `LocatedInferError::bare`:

```
cargo clippy --all-targets -- -D warnings   → exit 0
./ci.sh (full gate)                         → exit 0
cargo build --lib                           → error[E0599]: associated function `bare` not found
```

`cargo build --lib -v` on the clean tree confirms the mechanism: no `feature="test-helpers"` reaches the lib compilation, but `--all-targets` shows it. Edition 2024 / resolver 3 keeps the feature out of plain builds, which is exactly why this is silent rather than loud.

One line closes it — `ci_clippy_lib() { cargo clippy --lib -- -D warnings; }`. I confirmed it fails (101) with the probe in place and passes on the clean tree. This is the same argument the commit makes for `clippy_serde`: a configuration nothing compiles is a configuration that rots.

## The `Nature` judgment call moved to `label` rather than going away

The strongest argument in your reply was that `Expr::expr_stmt` at a statement's span was `Source` at three sites and `Machinery`/`lower.stmt_seq` at eight — "same node kind, same span role, opposite answers." That disagreement is still in the tree. It is now on the `label` axis, which this commit promotes to "the datum consumers read":

- `tag_image` (`"lower.image"`) on an `ExprStmt` at a statement span — 6 sites: `stmts.rs:549`, `600`, `606`; `transactions.rs:310`; `loops.rs:701`, `729`
- `tag_machinery(…, "lower.stmt_seq")` on the same shape — 8 sites: `stmts.rs:708`, `712`; `transactions.rs:260`, `271`; `loops.rs:776`, `789`, `799`, `811`

And `stmts.rs:548-549` double-attributes: the `MutWrite` *and* its enclosing `ExprStmt` both get `"lower.image"` at the same `stmt.span`, so two nodes each claim to be the image of one statement. Under any semantic reading the sequencing wrapper is manufactured — the user wrote no sequencing operator.

`design/provenance.md` states the label split in the same semantic language the nature split used to use (`tag_image` = "an image of source text" vs `tag_machinery` = "manufactured plumbing") with no decision rule attached. So someone adding a lowering rule still cannot answer "which of these do I call?" without guessing — which was the original complaint.

Either resolution works:

1. State a label rule that is mechanically applicable, the way `Source` now is.
2. Say explicitly that `label` is per-rule judgment and that `"lower.image"` carries no cross-site guarantee — so no consumer builds on it.

It might also be worth adding a call out in the design that the Nature/label stuff isn't mature yet and that we anticipate changes as the consumption of this data evolves

## Doc drift the `Nature` commit introduced

These matter more than usual, because the commit's stated goal is that the rule be applicable without reading the fold.

1. **`lower/mod.rs:607-608`** — `lower_expr`'s own doc still states the *old* semantic rule: "Interior nodes an arm builds are each tagged at their mint site — `LoweringContext::tag_source` for a direct image of something the user wrote, `tag_machinery` for manufactured plumbing." That contradicts `tag_source`'s new "No arm calls this," and `tag_image` doesn't appear at all. This is the function that *is* the rule's decision point, so it's where a reader lands first.
2. **`lower/exprs.rs:380-381`** — "Each pair comparison is a `Nature::Source` direct image of its `<op>` in the chain" sits directly above a `ctx.tag_image(...)` call. The chained comparison is one of the three examples the new rule doc uses as a *cost* case.
3. **`context.rs:473-475`** — `lowering_projection`'s field doc still presents a two-way split ("a `Nature::Source` direct-image tag … a `Lower/Machinery/label` tag for lowering-manufactured plumbing"). The third case, `Machinery` + `"lower.image"`, is now the common one — and this is the consumer-facing field.

## Smaller

- **`context.rs:542` still says "in I3."** Round 2 flagged the transient-ticket class; `Cargo.toml` got fixed, this one didn't. `git log -S` puts it in `35e4dcd4`, so it's this stack's. Same uncheckable class as the `§` refs.
- **`transact_phase.rs:1021` is an unmarked hand-rolled preserve.** `Expr::preserve`'s doc tells the reader to `grep TODO(preserve)` to find them; the sweep covered `channelize` and `mut_elim` and missed this one, so the greppable set is incomplete on day one.
- **`mut_elim::hoist_feeds` now implements the type carry twice.** `mut_elim.rs:582-585` and `896-898` still do `let body_ty = body.ty.clone(); body = Expr::expr_stmt(feed, body); body.ty = body_ty;`, which the constructor now does itself. Deleting the save/restore is the point of moving the invariant into `expr.rs`; leaving it gives it a second place to drift.
- **`Expr::expr_stmt`'s type carry changes behavior at a site the commit message doesn't account for.** The message covers lowering (identity — `body.ty` is `Hole`) and `mut_elim` (already carried). But `channelize.rs:374` previously produced `ty: Hole` on the lifted-prefix spine and now produces the body's type. That reads as a latent fix — `Hole` is `has_type_residue`, so an escaping one would trip `assert_no_type_residue` — but it's an unremarked retype in a post-inference pass. Worth a sentence, and a test if the path is reachable in typed mode.
- **`preserve`/`throwaway` are `pub`, where `with_node_id` was `pub(crate)`.** An external crate can now read `node_id()` off one node and construct another at the same id. The `node_id` field doc (`expr.rs:635-638`) still says "external crates still go through the accessor methods," and justifies `pub(super)` by "the pass submodules that build `TypedExpr` struct literals directly … can set it" — which is the pattern this commit set out to eliminate. Relatedly, "unrepresentable rather than detectable" is a little strong: 17 struct literals plus `subst.rs:95`'s `e.node_id = node_id` still bypass both constructors. The mechanical close (private field, plus the `Option`-taking annotation setter the doc names) is real follow-up rather than a blocker — but the field doc shouldn't be arguing for the loophole.
- **`CLAUDE.md` still says "BOTH clippy passes"** (lines 14, 19, 148, 175). There are three now. Same reasoning as everything else here: an undocumented gate step is a step people skip.

dpmills (approved):

groundlar:
**Round 3 — all items landed in `c9425ab`.** Full `./ci.sh` green, all four clippy passes.

- **`test-helpers` reopens the hole** → `./ci.sh clippy_lib`, in the gate and the workflow, exactly as you specified. `CLAUDE.md`'s "BOTH clippy passes" was wrong in five places; it is now a table of what only each of the four compiles.
- **The `Nature` judgment moved to `label`** → your option 2. `"lower.image"` means only *the minting rule considered this an image*; no consumer may read it as a cross-site classification. A label rule would have to relabel one of the two `ExprStmt` groups, which is lowering semantics this PR should not carry. The maturity call-out is in, on `tag_image` and in `design/provenance.md`. `stmts.rs:548` is fixed — the sequencing wrapper is manufactured under any reading, so it is `"lower.stmt_seq"` now.
- **Doc drift** → all three (`lower_expr`, the chained-comparison comment, and `lowering_projection`'s field doc, which now enumerates three tag shapes rather than two), plus "in I3".
- **The unmarked hand-rolled preserve** → marked. I did *not* fold the id-preserving literals into `Expr::preserve` generally, though: most of them are field-wise rebuilds — destructure, swap one child — where the literal's exhaustive field check is load-bearing. Omit a field and it fails to compile, where `preserve(id, node).with_ty(…)` silently defaults `ty` to `Hole`, i.e. type residue surfacing (if at all) as a distant runtime assertion. The `Option`-taking annotation setter the conversion needs would also have no caller of its own, existing only to make the conversion expressible — the same argument that deleted `map_node` last round. So `Expr::preserve`'s doc now distinguishes the two shapes instead of implying one deferred class.
- **`preserve`/`throwaway` visibility** → `pub(crate)`; free, since all twelve callers are inside `crate::ccl`. The private field I did not take: it is coupled to those field-wise rebuilds, so it is a trade rather than a mechanical close. The field doc now states the real invariant instead of arguing for the loophole — a literal *mints nothing*, so the hazard is `NodeId::fresh()` **inside** a literal, not a write to the field. That also scopes "unrepresentable rather than detectable" instead of withdrawing it: mint-then-overwrite genuinely is unrepresentable now, and the one residual mutation (subst's clone re-root) is named `re_root`.
- **`expr_stmt`'s retype at `channelize.rs:374`** → pinned by a unit test rather than left unremarked; `hoist_feeds`' two copies of the same carry are deleted.

Two things worth flagging beyond the list.

**The gate was red before this, and a pipe hid it.** `./ci.sh` failed at `doc_refs`: the rebase commit cited *"The id domain"* three times, and that was a bold bullet in `design/provenance.md`, not a heading — now `### The id domain`. My first run was `./ci.sh | tail`, which reported exit **0** on the red tree, since the pipeline's status is `tail`'s.

**The third of those citations was invisible to the checker**, because its path and heading text fell on different lines and the code→doc scan was line-scoped — i.e. #47. I had started the same fix and dropped it once yours appeared, since yours is a strict superset. Verified on a merge probe: no file conflicts with this PR, and #47's checker (including the new doc-prose citation check) passes on the merged tree, so the two land in either order. One citation this stack added, `solver/scheme.rs` → "One known exception, scoped and unfixed", would have failed your stricter check; that lead is promoted to a real heading here, and `provenance.md`'s truncated "Sharing is an invariant" now names the full one.

PR #20, opened 2026-07-23
ccl: the passes adopt the lineage recorder; both pane boundaries assert leak-free

groundlar:
The five IR passes between the retained inspector panes now record their rewrites, and both pane boundaries assert leak-free.

**Start here:** `compile_program` installs the `RecorderSession`s and retains the per-pass logs on `CompiledProgram::pass_lineage` [`Mono`, `Inline`, `Transact`, `Letrec`, `Desugar`]; `materialize_panes` folds them; `src/ccl/design/provenance.md` documents the model.

Each pass records its rewrites as `RewriteStep`s:

- `mono` (`infer/solve.rs`): a consume-nothing frame around the clone freshen captures per-origin `Copy` steps; `coalesce_generalized_let` is two steps — a produced-empty discard of the def subtree plus a narrow transform (original `let` → wrapper `Let`s), keeping the per-node `Copy` edges out of the bipartite product.
- `inline`: beta/alias `Transform` discards (`record_drops` diffs the consumed subtree against the reduced result) + fan-out `Copy` capture.
- `transact_phase`: one `transact.commit` frame; consumed = the `begin()` markers, writing-block spines, and register bindings, collected by precise spine walks that mirror what `strip`/`build_writer` actually drop.
- `mut_elim` (the letrec/mutation-elimination phase): `letrec.loop` / `letrec.bare_write` frames with spine-walk consumed sets; `flatten_spine` and `hoist_writer_body` carry ids on their 1:1 reparents.
- `channelize`: cluster / lift_defer / feed_union / fan-out / drop steps, with the cluster and lift origins threaded (`ClusterDefer`, the outer-`Let` id into `try_lift_defer`).

Duplication now freshens ids at its source, unconditionally: `Subst` compound replacements, inline fan-out clones, and channelize source clones. Root-carry is the general rule in the substitution engine (`freshen_interior_node_ids` re-mints only the interior and restores the occurrence's id on the replacement root), so a transact/letrec `subst_env` read keeps its read-site `Var` id and stays a self-edge. That makes id-uniqueness a tree invariant; the tripwire that checks it (`assert_unique_node_ids`) is introduced downstack with its first call site at the lowering boundary, and this commit adds the remaining four calls at the post-inline/transact/letrec/desugar boundaries.

`materialize_panes` folds the logs at the two pane boundaries (pre→post-inference; post-inference→post-desugar) into per-pane `SourceProjection`s + dense `LineageMap`s, and `assert_leaks_clean` asserts zero leaks of every class at both.

`census_ratchet` pins per-category attribution counts per stage and asserts structurally that no pane contains an unresolved row — the lowering fold covers every node.

Every `RewriteStep` here is recorded as a byproduct of performing the rewrite — never reconstructed by diffing trees afterwards — so a pass that mints where it should preserve shows up as a leak at the boundary rather than as silently wrong attribution. That is what the zero-leak assertions buy: the recorder surface the substrate landed with is exercised for the first time, and 13 of its 15 `dead_code` allows come off here (the two that remain are `Nature::wire_str`, whose consumer is the inspector wire, and the `Copy` op's `origin` accessor).

**Conditional induction.** `transform_chain`'s statement-`Case` arm (the `if p: acc += e` shape) fans the post-`Case` remainder and the entering read-your-writes env across every branch, so several subtrees reach the tree more than once. Each fan-out site is keep-first — the first path carries the original ids, later paths take an attributed `Copy` — covering the arm predicates' negated guard echoes, the remainder splice, and the per-branch env clones. `collect_block_spine_ids` mirrors the new arm: it consumes the `Case`, every branch guard, and each branch body's spine. Guards are consumed uniformly rather than predicted, because which arm predicates survive is data-dependent (the trailing carry arm's is always discarded, as is any arm whose write set matches the carry).

A write's value is inlined into the env as a *template*, not left on the spine, so the template's own ids never reach the output — reads instantiate it with a freshened interior and the read site's id on the root. Its ids are therefore consumed with the write, and the terminal takes a `fresh_copy` into the write set. This moves `mutation_loop`'s census row: total 48 → 46 (main's shared `decision_record`/`disjoin_guards` construction mints two fewer nodes) and 3 nodes `Source`(1)/`Synthetic(Lower)`(2) → `Derived(Letrec)` (the final write value is now a recorded `Copy`, so it keeps its span but takes the step's nature). Both movements are named on the row.

**Flush order.** A frame's captured `Copy`s now straddle its own `Transform`, partitioned by origin: copies of ids the frame *consumes* fold before it (while the origin is still live), copies of ids it *births* fold after (once they are). Both orders are required, so neither can be global, and the split is per `(origin, fresh)` pair — one deep freshen of a subtree mixing consumed and newly born nodes lands on both sides. The `Transform` stays whole because it carries root inheritance; splitting it into birth and consume halves would hand every born node an empty root set and sever it from its ancestry without tripping the audit. `copies_straddle_the_frame_transform_by_origin` pins both directions.

`conditional_induction_panes_are_leak_free` materializes panes for six conditional-induction shapes. The leak and id-uniqueness assertions at the post-desugar boundary only run under `materialize_panes`, so the compilation-pipeline suite compiles these programs without ever checking them — the coverage has to be explicit.

### DIFF
--- a/src/ccl/transact_phase.rs
+++ b/src/ccl/transact_phase.rs
@@ -70,6 +70,7 @@ use crate::ccl::{
     ccl_utils::is_free_in_value,
     letrec::check_letrec_causal,
     mut_elim::{fold_induction_loop, hoist_feeds, register_value_tys},
+    provenance::NodeId,
 };
 
 /// Recognize a **fed-out register read** and rewrite it to an as-of join, *before*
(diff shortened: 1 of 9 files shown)


src/ccl/transact_phase.rs
```diff
@@ -548,6 +549,108 @@ pub fn collect_txn_registers(expr: &Expr) -> HashSet<Name> {
...
+/// declaration nodes themselves are dropped by `rebind_letrec`). Everything
+/// preserved — register value expressions (`subst_env` containers + id-carried read
+/// leaves), the iteration source, the register inits, the read-only feed spine, and
+/// the enclosing `For` (retired by `mut_elim`, not here) — is a self-edge and
+/// is NOT collected.
+///
+/// Mirrors what [`strip`]/[`build_writer`]/[`build_letrec`] actually drop; a
+/// drift surfaces at the leak gate as `Dropped` (a vanishing node left
+/// unconsumed) or `Unexplained` (a preserved node wrongly consumed).
+fn collect_txn_consumed(expr: &Expr, txn_registers: &HashSet<Name>, out: &mut Vec<NodeId>) {
```
dpmills, line 462: Even through these are checked, having them be complex, out of line logic isn't clean.  inline.rs has a general-purpose-looking helper called record_drops; can we use that or some other general thing? There's similar stuff in mut_elim that can probably be improved in the same way

dpmills:
Aside from the comment about how complex transformations work, just a couple of nits:
#### 1. Duplicated recording helpers should be hoisted

Two small helpers are duplicated across passes:

- **`fresh_copy`** (a one-line `freshen_node_ids_deep` wrapper) is byte-identical in
  `src/ccl/transact_phase.rs:1443` and `src/ccl/mut_elim.rs:196`.
- **"collect every node id in a subtree"** exists four times: `collect_all_ids`
  (`mut_elim.rs`), `collect_all_ids_tx` (`transact_phase.rs`), `collect_tree_ids`
  (`context.rs`), and `collect_ids` (`inline.rs`) — the same walk.

Suggest hoisting both to `expr.rs`/`lineage.rs`. Small on its own, but this is already
at three/four copies and will keep growing as passes adopt the recorder.

#### 2. Stale doc comment in `materialize_panes`

The doc comment says boundary 2 folds "the concatenated **Inline + Desugar** logs," and
the `desugar_map` field comment repeats "Inline + Desugar" — both omit Transact + Letrec.
The code folds `pass_lineage[1..]` (all four passes) and the `assert_leaks_clean` message
correctly lists `Inline+Transact+Letrec+Desugar`. Please update the two comments to match
the code.

#### 3. Session-install inconsistency

There are two idioms for installing the recorder:

- mono/transact/letrec install `RecorderSession::new()` + `into_log()` inline in
  `compile_program`;
- inline and channelize encapsulate it in `*_with_lineage` wrappers.

mono genuinely can't be wrapped (it runs inside `infer`), but transact/letrec could use
the wrapper form for symmetry with inline. Worth picking one idiom where possible.

#### 4. `assert_unique_node_ids` isn't run at the post-inference (post-mono) boundary

The tripwire runs at post-inline / post-transact / post-letrec / post-desugar, but not
right after `infer` (where mono freshens clone ids). A mono-freshen regression is caught
transitively at post-inline (inline runs next), so it's covered — but an explicit check
immediately after inference would localize the failure to the pass that caused it. One line.

groundlar:
Closing in favor of the discussed monoid / unique-node-id machinery changes.

PR #24, opened 2026-07-23
Drop static finiteness assumptions from the transaction as-of path

dpmills:
## Summary

A fed-out `Txn` register read folds to `AsOf` because its history domain is the commit-sequencing domain `Type::Txn` — not because of any derived finite/infinite classification of the reading trigger. Two checks in the transaction path baked such a classification in, and they contradicted each other. This removes both, so the compiler no longer makes a static guess about whether a source terminates.

## Changes

- **`live_register_read` / `is_live_final_or_default`** used `!has_enumerable_extent(domain)` to mean *"is this a transactional register?"* — an imprecise proxy: an induction accumulator over an opaque extent also answers non-enumerable and would have been mis-rewritten to an as-of read. Replaced with the exact `matches!(domain, Type::Txn)`, which is both more precise and free of any finiteness framing.

- **`check_live_reads_resolved` / `contains_live_trigger`** rejected an unrecognized fed-out read sitting beside a `DataSource` trigger, using the source *type* as a proxy for "runtime-infinite". This directly contradicted `has_enumerable_extent`, which treated `DataSource` as finite/enumerable. `Type::DataSource` conflates a finite batch source (`StdinDataSource`) with a live stream (`http_serve`), so any static answer is a guess — and the guard could reject a valid batch program. Removed it: a register read beside a genuinely non-terminating trigger now simply runs continuously, as intended. The guard was already documented in tests/comments as a retired "band-aid", and no test asserted it firing.

- With both callers gone, **`Type::has_enumerable_extent` is dead and removed**, along with its whole-type enumerability classification.

## Verification

`./ci.sh` green — all tests, both clippy passes (debug + release), and doctests. No dangling references to any removed symbol.


dpmills:

- #24 ◀

groundlar (approved):

PR #28, opened 2026-07-23
CHL type annotations: parenthesised application, Caps primitives, record & tuple types

dpmills:
Converge CHL type-annotation syntax on the decided target (`docs/chl-spec.md`, "term/type syntax split"): type application is parenthesised at both the term and type level, and primitive type names are capitalized. The transitional spellings — bracket application (`Mut[V, Txn]`, `List[T]`) and lowercase primitives (`int`/`str`/`bool`) — are retired; they no longer parse as annotations. No dual dialect.

## Changes

- **`lower_type_annotation` / `mut_annotation_parts`**: accept `Int`/`Bool`/`String` and parenthesised type application (`Mut(V, Txn)`, `List(T)`); the bracket forms are dropped.
- **New annotation forms**: record types `{name: T, …}` (via `Expr::Record`) and tuple types `{T, U}` via a new colon-free `Expr::BraceGroup`. Braces are type syntax — a brace group is rejected in value position with a hint toward `[…]` / `(…)`.
- **Display unification**: the CCL `Type` Display for `Mut` now renders `Mut(value, domain)`, matching the surface spelling so the rendered form no longer drifts from what a programmer writes.
- **Corpus + docs migration**: the CHL test/demo programs and the CCL symbolic renderings in comments and design docs move to the paren/Caps spelling. The now-vacuous §8 "Spelling note" is deleted (it described a difference that no longer exists), and the §6 surface-type table marks which types are writable (record/tuple now are; `Dict` is not).

The `fanout` north-star now pins the more precise `unknown type application: Feed(…)` (its blocker is that `Feed` isn't a resolved type constructor yet).

## Verification

`./ci.sh` green — all tests, both clippy passes (debug + release), doctests, and doc-refs. New tests cover the capitalized primitive, parenthesised application, record- and tuple-type annotations, the brace-group parse/classification, and the value-position rejection.


dpmills:

- #28 ◀

sortalongo (approved):

PR #29, opened 2026-07-23
Corpus: redesign the storefront north-star to match the announcement example

sortalongo:
The storefront north-star corpus and `docs/chl-spec.md` should describe the same target syntax. This PR gets the corpus there — the transaction-handle form, a redesign whose domain rules live in the types, a port onto PR69's decided refinement/variant syntax — then closes the gap the redesign opened: the spec now records the surface features it relies on, so it stays the source of truth for what's decided versus open.

## Transaction-handle form

`storefront` (v0/v1), `txn_kv`, and `ledger_balance` switch from `with begin() as t:` to the **[Decided]** `with t = begin():`.

## Storefront redesign

Domain rules move into the types: `Dollars`/`Qty`/`ItemPricing`/`SKU` aliases, a `FullMap` inventory keyed by the value-dependent `SKU`, `static assert` lifted to `quote`'s codomain, HTTP-lib request validation, restriction spelled `filter`. `v1` keeps the diff-is-the-upgrade property — only `is_promo_spent` and `quote` differ from `v0`.

## Adopting PR69's decided syntax

Ports both files onto `{T where p(_)}` refinements and backtick-tagged variants ([§6.4](docs/chl-spec.md#64-refinement-syntax), [§6.5](docs/chl-spec.md#65-variants)); the pinned lex blocker moves from the `\` lambda to the `` ` `` variant tag.

## Spec: recording the new surface

New [§6.7](docs/chl-spec.md#67-type-alias-statements) documents type-alias statements (**[Tentative]**, matching §6.1's note); `FullMap` and `Real` join the type list; §6.4 now allows a top-level refinement to name a top-level binding, as `SKU` does; the function-contracts and HTTP Direction notes gain the codomain-lift and derived-validation cases this redesign exercises. `restrict` vs `filter` is flagged **[Open]**, unresolved.

## Verification

`./ci.sh` is green: shellcheck, doc-refs, `cargo fmt`, all four clippy configurations, the doc build, and the full test suite, including the `programs` target (21/21, both files pinned on `invalid token`).


dpmills (approved):

docs/chl-spec.md
```diff
@@ -664,6 +666,17 @@ point of use. Mutual recursion between top-level functions is
...
+> **Direction [Tentative] — `Real` and `/`.** The target language has a
+> fractional type, `Real`, and a division operator `/` on it. Neither
+> exists today: `Real` is no type the checker knows, `/` is not lexed
+> (§1.8), and there is no fractional literal (§1.7), so a quantity that
+> wants a fraction has nothing to write it with. Everything else is
+> **[Open]** — what `Real` denotes (exact rationals, a decimal, or the
+> `f64` §12 rules out), whether `/` is total or partial at zero
+> (*Partiality*, §3), whether `//` survives beside it, and how an `Int`
+> literal acquires the type in a `Real`-typed position. Wanted by the
+> north-star `storefront`, which prices in `Real`.
```
dpmills: No objection to adding a Real type, but just a note that barely any systems use reals/floats for currency.  Using integers for the number of cents for dollar-only systems is common, and multi-currency stuff is more complicated.
Probably should delete the last sentence here, and use Int in the example

docs/chl-spec.md
```diff
@@ -2333,8 +2512,18 @@ with parser-level support that lowering rejects:
 - **Classes / `try`** — not in the language. `with` **is** a keyword
   (§1.6), but only for transaction blocks `with begin():` (§8.2); it does
   not carry Python's general context-manager meaning.
-- **Float arithmetic** — no `f64` type at the surface.
+- **Float arithmetic** — no `f64` type at the surface, and no `/`
```
dpmills: This section is for stuff that's not described earlier; no need for duplicate entries

docs/chl-spec.md
```diff
@@ -1662,6 +1716,18 @@ details to change:
...
+  key is present. Lookup is `m[k]: V` — there is no missing case, so
+  nothing to match (§3.9). A total map is typically *earned* by refining
+  the key type down to keys known to exist, rather than by promising
+  totality over an open type: `FullMap({String where _ in ks}, Int)` is
+  total because its domain says which strings it holds (§6.4). That
+  shifts "the lookup hits" from a runtime invariant to an obligation on
+  whatever constructs the map. **[Open]**: what that obligation is at a
+  construction site (a literal must cover the domain; a domain that grows
+  must extend the map), and whether totality survives operations that
+  narrow or widen `K`.
```
dpmills: Types can't be narrowed or widened; they have a defined meaning.  This can be phrased  more generally as "[Open]: Precise semantics and limitations"

dpmills (approved):

docs/chl-spec.md
sortalongo: Fair point — dropped the storefront justification from the Direction note (it stays as a general, ungrounded future direction), and switched the corpus to `Dollars = Int`. Also changed `sale = list_price / 2` to `list_price // 2` in v1's `quote` — same result for the pinned example (10 // 2 == 5), just using an operator that actually exists.

docs/chl-spec.md
sortalongo: Removed the two new entries (Type-alias statements, Total maps) and reverted Float arithmetic to its original one-liner — all three are now fully covered in §6.7/§6.3/§3.3, so restating them here was pure duplication.

docs/chl-spec.md
sortalongo: Reworded to that framing — now reads "[Open]: precise semantics and limitations," with the narrow/widen claim dropped and just the construction-site point kept.

PR #31, opened 2026-07-23
CHL records: (name=value) values; braces are type-only; retire dict

dpmills:
Record *values* are now written with parentheses and `=`: `(x=1, y=2)`. The
parentheses are the product constructor, shared with tuples (`(1, 2)`) — a
comma-list of bare expressions is a tuple, a comma-list of `name=value`
fields a record (docs/chl-spec.md, §2.4/§3.11).

Braces are now **type syntax only** at the term level: `{x: T}` is a record
type (`Expr::BraceRecord`), `{T, U}` a tuple type (`Expr::BraceGroup`); both
lower as types in annotation position and are rejected in value position. The
old brace record *value* `{x: 1}` no longer denotes a value.

Dict is retired end to end. `Dict(K, V)` is renamed to `Map(K, V)` in the
spec and design docs, with the map *literal* settled as `[k -> v, …]`
(collection syntax, [Decided]/unimplemented). The dead `Expr::Dict` node and
its brace-dict parse/lowering path are deleted — expression-key braces
(`{"k": v}`) are now a plain parse error pointing at `[k -> v]`.

Also strips rustpython/migration history from the parser design doc and a
couple of "old X" codebase-history notes from the CCL design docs.

Migrates the record-value corpus (`{k: v}` → `(k=v)`), keeping record-*type*
annotations (`{name: T}`) in braces. reachability's record-term syntax now
parses, so its pin advances to the `rec` blocker.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -99,7 +99,7 @@ plus the requisite `DEDENT`s are emitted, so every `INDENT` is paired.
 ### 1.4 Implicit line continuation
 
 Inside `(...)`, `[...]`, or `{...}` (any depth, any combination), newlines
-and indentation are **ignored**. Multi-line list, tuple, dict, record,
+and indentation are **ignored**. Multi-line list, tuple, record,
 and call expressions are written naturally:
 
 ```python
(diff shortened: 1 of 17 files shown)


dpmills:

- #28
    - #31 ◀

sortalongo (approved):

PR #32, opened 2026-07-24
Preserve refinement-predicate Rc sharing across inference and planning

dpmills:
Preserve refinement-predicate `Rc` sharing across inference and planning

Refinement predicates are immutable `Rc<TypedExpr>`s so that one predicate riding many type slots — a comprehension's filtered domain appears on its source, map, cast target, and consumer-contract types — is a single shared allocation. But every pass that "updates" a predicate *rebuilds* it, and none of the rebuilding passes preserved that sharing. By planning time one structural predicate had scattered into roughly one `Rc` per occurrence, defeating planning's `Rc`-keyed compile memo, which then recompiled the same predicate once per occurrence — superlinearly on nested comprehensions.

On the motivating doubly-nested comprehension, inference left **7** distinct predicate `Rc`s where 3 suffice. It now leaves **3**, with no two of them structurally equal.

## One memo, threaded through every rebuilding pass

`ccl_utils::PredMemo` is the pass-scoped memo that maps an origin predicate `Rc` to a single rebuild, so occurrences that enter a pass sharing one `Rc` leave sharing one `Rc`. Its protocol makes misuse inexpressible:

- `begin(&mut Refinement) -> Option<(Origin, Expr)>` — on a hit, re-points the slot and returns `None`; on a miss, hands back an owned copy plus an `Origin` token.
- `finish(refinement, origin, rebuilt)` — installs the rebuild and records it.
- `finish_unchanged(refinement, origin)` — the transform had nothing to do: keep the origin `Rc`, record `origin ↦ origin`.

`Origin` is minted only by `begin` and consumed only by the two `finish`es, so the keepalive that makes address-keying sound cannot be forgotten or confused with a rebuild. That keepalive is load-bearing, not bookkeeping: overwriting a predicate slot can drop the last reference to the origin and free an address a later `Rc::new` in the same walk reclaims, at which point an unrelated predicate collides with the entry and inherits its rebuild.

Every rebuilding pass now threads it — `uniquify`, constraint emission, `coalesce`, `retype`, `subst`'s both modes, `inline`, `simplify`, and planning's predicate compilation. Two of those were *first-pass* splitters, so no later pass could have recovered the sharing:

- **Constraint emission.** `emit_bare_predicate` rebuilds every predicate it types and threaded no memo, so the very first walk over the tree split sharing. A nested comprehension's inner filter is reached twice — once on the term-level `Cast`'s `target`, once on the copy of that `Cast` inside the enclosing comprehension's own filter predicate. It now threads a memo via a `Typing::pred_memo` accessor, so Emit and Check both inherit it. Sharing the memoized copy also makes those occurrences share inference variables, which is correct rather than incidental: a bare predicate is closed but for `REFINEMENT_BINDER`, so two occurrences of one refinement `Rc` denote the same refinement over the same base and their slots have a single solution.
- **`retype`.** `retype_pred_expr` was a hand-rolled near-duplicate of `retype_predicate_slots` that omitted the binder slots and had no `Cast` arm at all, so a `Cast` nested inside a predicate was reached only by the copy that could not see it. A predicate *is* a term, so the two walks are now one function: `retype_in_type` hands each predicate back to `retype_predicate_slots`. Closing that also picked up the `For` and `LetRec` binder scopes the old walk dropped into its catch-all arm.

`walk_refined_predicates_mut` is the type-walk combinator built on the memo. Its callback returns whether it *changed* the predicate; returning `false` keeps the origin `Rc`, so a pass preserves sharing for predicates it merely walks past rather than reallocating them per node visited. `coalesce` drives the same protocol inline instead of through the combinator — its transform needs `&mut CoalesceCtx`, which holds the memo — with the scope-independence precondition the `Rc`-only key relies on written down at the call site.

## `Expr::walk_type_slots`: one enumeration of a node's type slots

A predicate can ride a node's `ty`, its `user_annotation`, a `Cast`'s `target`, or any binder's declared type — each an independent `Rc`. That set was enumerated by hand in six places, two of them incompletely, and both omissions were live defects:

- `distinct_predicate_rcs` — the sharing metric itself — recursed with `Expr::walk_children`, which reaches neither the binder slots nor `Cast.target`. Since `target` is where a comprehension filter's predicate is written (and where `lambda_elim` and operator conversion read it from), the metric could not observe a split in the one slot that mattered most.
- `inline_and_beta_reduce` missed the binder slots, so a UDF use inside a binder-slot predicate could survive as a dangling `Var` once the enclosing `Let` was dropped.

`Expr::walk_type_slots{,_mut}` is now the single source of truth for that set, built on the existing `walk_binders`. The metric, `retype_predicate_slots`, `inline_and_beta_reduce`, and `planning::compile_refinement_predicates` all share it, so a new type-slot-bearing variant updates one match instead of every pass and a pass cannot acquire a blind spot by omission.

## Guards

`tests/predicate_sharing.rs` asserts, end to end and post-inference, that **no two `Rc`-distinct refinements reachable from an inferred tree are structurally equal** — compared with `Refinement`'s own type-blind `PartialEq`. That is exactly the shape a split leaves: one origin term rebuilt into several `Rc`-distinct but value-equal copies. It needs no threshold, so nothing drifts, and the failure message prints the duplicated predicates in symbolic form. The property is asserted at every comprehension nesting depth 2–6, and `ccl_utils::reachable_refinements` exposes the traversal the count derives from so the test can ask that question directly.

Three unit tests pin the metric's own coverage of `Cast.target`, `Lambda.param.ty`, and `Let.binding.ty` on hand-built nodes. Hand-built rather than program-derived on purpose: once sharing is preserved, a cast's `target` and its `ty` hold the same `Rc`, so a fixture-based coverage check goes vacuous the moment the bug is gone.

Per phase, a debug-only tripwire checks that `retype` — a rewrite-only pass — never *increases* the distinct-predicate count. That count is the necessary check; the equal-but-distinct assertion is the sufficient one, and both are documented as such.

## Steering

`Refinement` has exactly two constructors, so a bare struct literal (which states neither intent) no longer appears anywhere:

- `born(Rc)` — a genuinely new predicate term, with no prior identity to preserve.
- `sharing(&Rc)` — a second occurrence of an *existing* predicate, deliberately aliased. This is what lowering does when one filtered domain rides four slots, and it is the sharing every rebuilding pass then has to preserve.

A warning doc-comment on `Refinement.predicate` points rewrites at the memo.

## Docs

`src/ccl/design/type-inference.md` gains a "Sharing is an invariant, not an optimization detail" subsection: the full list of rebuilding passes, why the keepalive is load-bearing, why the cost lands in planning, and what each of the two guards does.

## Noted, not fixed

`retype_predicate_slots`, `subst::rewrite_expr_go`, and `infer::api`'s scope walks each spell out "walk children, each under the binders that scope it" by hand. Unlike the type-slot set, the scoping is genuinely non-uniform — `Let` binds only its body, `LetRec` binds everything, a `Case` branch binds its guard and body — so a `walk_children_scoped` combinator would have to encode that structure. Flagged at the call site rather than folded in here.

## Also carried: the `deep-typecheck` gate

This branch additionally contains **#33** — gating the per-operation `debug_typecheck` behind an opt-in `deep-typecheck` cargo feature (`Cargo.toml`, `ci.sh`, `.github/workflows/ci.yml`, `src/ccl/infer/api.rs`). That PR was merged into *this* branch rather than into `main`, so it rides along here and will land on `main` under this squash. Its own description has the rationale; the short version is that the per-operation typecheck is a bisection tool, too expensive to run unconditionally, and CI sets `DEEP_TYPECHECK=1` so automated runs keep the coverage.

Worth knowing when reading this diff: `ci_test` is `cargo test -q ${DEEP_TYPECHECK:+--features deep-typecheck}`, and the predicate-sharing work below was verified both with the feature off and on.

### DIFF
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -184,9 +184,7 @@ pub fn refine_codomain(morphism: Expr, bare_predicate: &Expr) -> Expr {
     // structurally equal to the cast demand.
     let refined = Type::Refinement(
         Box::new(codomain),
-        Refinement {
-            predicate: Rc::new(bare_predicate.clone()),
-        },
+        Refinement::born(Rc::new(bare_predicate.clone())),
     );
     set_codomain(morphism, refined)
 }
@@ -300,12 +298,7 @@ pub fn cast_target_refinement(target: &Type) -> Option<Refinement> {
 /// inference fills them in by unifying against the value being cast.
 pub fn refined_fn_type(base_domain: Type, predicate: Expr, codomain: Type) -> Type {
     Type::fun(
-        Type::Refinement(
-            Box::new(base_domain),
-            Refinement {
-                predicate: Rc::new(predicate),
-            },
-        ),
+        Type::Refinement(Box::new(base_domain), Refinement::born(Rc::new(predicate))),
         codomain,
     )
 }
@@ -397,12 +390,7 @@ fn refine_with(base: Type, predicate: &Expr) -> Type {
         return base;
     }
     let bare = bare_predicate_of_fn(&base, predicate.clone());
-    Type::Refinement(
-        Box::new(base),
-        Refinement {
-            predicate: Rc::new(bare),
-        },
-    )
+    Type::Refinement(Box::new(base), Refinement::born(Rc::new(bare)))
 }
 
 /// Count free occurrences of `name` in `expr`, including occurrences in
@@ -757,46 +745,128 @@ where
     ty.walk_children(|child| walk_refined_predicates(child, visited, f));
 }
 
+/// Pass-scoped memo that **preserves refinement-predicate `Rc` sharing** across
+/// a single rebuild pass.
+///
+/// Predicates are immutable `Rc<TypedExpr>`s (see [`Refinement::predicate`]), so
+/// any pass that must "update" a predicate — resolve embedded inference vars
+/// (`coalesce`), restamp binder types (`retype`), eliminate lambdas
+/// (`lambda_elim`), simplify, or substitute — *rebuilds* it into a fresh `Rc`
+/// rather than mutating in place. When one predicate term rides several type
+/// slots as a shared `Rc` (a comprehension's filtered domain appears on its
+/// source, map, cast, and consumer-contract types), rebuilding each occurrence
+/// independently **splits** that sharing. The split is invisible to correctness
+/// — the rebuilds are value-equal (refinement equality is structural) — but it
+/// defeats every downstream `Rc`-identity dedup: `Rc::ptr_eq` fast paths and, in
+/// particular, planning's per-predicate compile memo, which then recompiles one
+/// predicate once per occurrence (superlinearly, on nested comprehensions).
+///
+/// This memo re-points every occurrence that entered a pass sharing one origin
+/// `Rc` at a *single* rebuilt `Rc`, so sharing survives the pass. It is the one
+/// mechanism every predicate-rebuilding pass threads; the alternative —
+/// reunifying afterward by structural hash/eq — would reintroduce exactly the
+/// per-predicate hashing cost the `Rc` boxing exists to avoid.
+///
+/// **The memo must be pass-scoped.** One instance threaded across the *whole*
+/// tree walk of a pass. A fresh memo per node/subtree re-shares only within that
+/// node and still splits across the tree — the exact bug this type prevents.
+/// The sharing invariant is guarded end-to-end by `tests/predicate_sharing.rs`
+/// and, per phase, by asserting [`distinct_predicate_rcs`] does not grow across
+/// a transform-only pass.
+///
+/// Keyed on [`PredicateId`] (the origin `Rc`'s address). Sound only while that
+/// address cannot be reused during the pass, which [`Self::record`] guarantees
+/// by retaining a `keepalive` clone of every origin `Rc` for the memo's
+/// lifetime: without it, overwriting a `refinement.predicate` could drop the
+/// origin `Rc`, freeing an address a later `Rc::new` in the same pass could
+/// reclaim, so an unrelated predicate landing there would collide and inherit
+/// this entry's rebuild.
+#[derive(Default)]
+pub struct PredMemo(HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>);
+
+impl PredMemo {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// The rebuilt predicate for `refinement`'s origin `Rc`, if an occurrence
+    /// sharing it was already rebuilt in this pass. Clones the shared `Rc` out
+    /// so the caller holds **no** borrow of the memo across its own transform —
+    /// which is what lets a pass whose transform needs `&mut Ctx` (with the memo
+    /// living in that `Ctx`) look up, transform, then [`record`] without a
+    /// borrow conflict.
+    ///
+    /// [`record`]: Self::record
+    pub fn rebuilt_for(&self, refinement: &Refinement) -> Option<Rc<Expr>> {
+        self.0
+            .get(&refinement.predicate_id())
+            .map(|(_, rebuilt)| Rc::clone(rebuilt))
+    }
+
+    /// Record that every occurrence sharing `original` rebuilds to `rebuilt`.
+    /// Retains `original` as the keepalive pinning its address for the memo's
+    /// lifetime (see the type-level note on address reuse) — pass the origin
+    /// `Rc` you cloned *before* overwriting the slot, not the rebuilt one.
+    pub fn record(&mut self, original: Rc<Expr>, rebuilt: Rc<Expr>) {
+        let key = Rc::as_ptr(&original);
+        self.0.insert(key, (original, rebuilt));
+    }
+}
+
 /// Rebuilding analog of [`walk_refined_predicates`]: invoke `f` on a *mutable
 /// copy* of each predicate and reinstall the (possibly rewritten) result as a
-/// fresh `Rc`. `memo` maps each original predicate's identity to a
-/// `(keepalive, rebuilt)` pair, so every occurrence that shared one predicate
-/// term in `ty` is re-pointed at the *same* rebuilt term. The callback
-/// receives `memo` so it can recurse when a predicate's own subexpressions
-/// carry further refinements.
+/// fresh `Rc`, **preserving sharing** via a pass-scoped [`PredMemo`] (which the
+/// callback also receives, so it can recurse when a predicate's own
+/// subexpressions carry further refinements). Every occurrence that shared one
+/// predicate term is re-pointed at the same rebuilt term.
 ///
-/// The dedup itself is a **performance / structural-sharing optimization, not
-/// a correctness requirement**: `f` is a deterministic rewrite, so rebuilding
-/// each occurrence independently would yield *value-equal* predicates (refinement
-/// equality is structural) — the memo only makes them the *same* `Rc` rather
-/// than *equal* `Rc`s, saving the recompute and keeping `ptr_eq` fast paths and
-/// any downstream `Rc`-keyed dedup effective. It keys on [`PredicateId`]
-/// (`Rc::as_ptr`), the one residual pointer-identity dependency, sound only as
-/// long as that address cannot be reused for the rest of the walk — which is
-/// exactly what the `keepalive` clone guarantees. Without it, overwriting
-/// `refinement.predicate` below would drop the original `Rc` (if this was its
-/// only strong reference), freeing an address that a later, unrelated
-/// `Rc::new` in the same walk could reclaim; a subsequent predicate landing on
-/// that address would then collide with this entry and wrongly inherit its
-/// `rebuilt` value. (Planning's predicate-compilation memo has the identical
-/// shape — see [`crate::ccl::planning`]'s `PredMemo`.)
-pub fn walk_refined_predicates_mut<F>(
-    ty: &mut Type,
-    memo: &mut HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>,
-    f: &mut F,
-) where
-    F: FnMut(&mut Expr, &mut HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>),
+/// The caller **must** pass one memo for the whole pass, not a fresh one per
+/// call — see [`PredMemo`] for why.
+/// Count the **distinct** refinement-predicate `Rc`s reachable from `expr`'s
+/// type slots (every node's `.ty`, `user_annotation`, and binder-slot types).
+/// Pure address-set arithmetic — no structural hashing — so it is cheap enough
+/// for a `debug_assert!` guard.
+///
+/// This is the per-phase sharing check: a *sharing-preserving* transform maps N
+/// distinct origin predicate `Rc`s to ≤ N distinct rebuilt `Rc`s, so this count
+/// is **non-increasing** across a transform-only pass (one that rewrites but
+/// does not introduce predicates). A pass that splits sharing strictly
+/// increases it, tripping the guard at the exact phase that regressed.
+pub fn distinct_predicate_rcs(expr: &Expr) -> usize {
+    fn in_type(ty: &Type, seen: &mut HashSet<PredicateId>) {
+        if let Type::Refinement(_, r) = ty
+            && seen.insert(r.predicate_id())
+        {
+            // A predicate's own subexpressions carry further refinements.
+            in_expr(&r.predicate, seen);
+        }
+        ty.walk_children(|c| in_type(c, seen));
+    }
+    fn in_expr(e: &Expr, seen: &mut HashSet<PredicateId>) {
+        in_type(&e.ty, seen);
+        if let Some(a) = &e.user_annotation {
+            in_type(a, seen);
+        }
+        e.walk_children(|c| in_expr(c, seen));
+    }
+    let mut seen = HashSet::new();
+    in_expr(expr, &mut seen);
+    seen.len()
+}
+
+pub fn walk_refined_predicates_mut<F>(ty: &mut Type, memo: &mut PredMemo, f: &mut F)
+where
+    F: FnMut(&mut Expr, &mut PredMemo),
 {
     if let Type::Refinement(_, refinement) = ty {
-        let original_rc = Rc::clone(&refinement.predicate);
-        let original = Rc::as_ptr(&original_rc);
-        if let Some((_, rebuilt)) = memo.get(&original) {
-            refinement.predicate = Rc::clone(rebuilt);
+        if let Some(rebuilt) = memo.rebuilt_for(refinement) {
+            refinement.predicate = rebuilt;
         } else {
+            let original = Rc::clone(&refinement.predicate);
             let mut pred = (*refinement.predicate).clone();
             f(&mut pred, memo);
             let rebuilt = Rc::new(pred);
-            memo.insert(original, (original_rc, Rc::clone(&rebuilt)));
+            memo.record(original, Rc::clone(&rebuilt));
             refinement.predicate = rebuilt;
         }
     }
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -429,21 +429,22 @@ pub fn count_free(name: &Name, expr: &Expr) -> usize {
 /// its free-var count is collected on first encounter and short-circuited on
 /// subsequent encounters.
 fn count_free_with_visited(name: &Name, expr: &Expr, visited: &mut HashSet<PredicateId>) -> usize {
-    // Every type slot of the node counts: `ty`, the user annotation, and a
-    // `Cast`'s target (where its refinement is written syntactically —
-    // pre-inference no `ty` slot carries the same refinement, so it must be
-    // walked explicitly or a predicate-only occurrence goes unseen).
-    let in_type = count_free_in_type_with_visited(name, &expr.ty, visited)
-        + expr
-            .user_annotation
-            .as_ref()
-            .map_or(0, |t| count_free_in_type_with_visited(name, t, visited))
-        + match &expr.node {
-            TypedExprNode::Cast { target, .. } => {
-                count_free_in_type_with_visited(name, target, visited)
-            }
-            _ => 0,
-        };
+    // Every type slot the node carries counts, via the same
+    // [`Expr::walk_type_slots`] the rewriting passes use. Getting this set wrong
+    // is not a cosmetic under-count: several passes *skip work* when `is_free`
+    // says no (`inline_in_type_predicates`' early return, `subst`'s inert-subtree
+    // and vacuous-predicate checks), so a slot this walk cannot see is a slot
+    // those passes silently decline to rewrite. Enumerating it here by hand is how
+    // it fell out of step with `walk_type_slots` in the first place.
+    //
+    // A binder's declared type is in the *enclosing* scope — a binder does not
+    // bind in its own type — so occurrences there are unshadowed and count, which
+    // is why this runs before the shadowing logic below. Occurrences are
+    // deduplicated by `visited`, so a predicate riding both a binder slot and the
+    // matching position of `expr.ty` (the usual case, and increasingly so as
+    // sharing improves) is still counted once.
+    let mut in_type = 0;
+    expr.walk_type_slots(|ty| in_type += count_free_in_type_with_visited(name, ty, visited));
     let in_node = match &expr.node {
         TypedExprNode::Var(n) => (n == name) as usize,
 
@@ -782,16 +783,54 @@ where
 /// an unrelated predicate landing there would collide and inherit this entry's
 /// rebuild. This is why a pass uses `PredMemo` rather than a bare map — the
 /// keepalive is not optional bookkeeping.
+///
+/// # Key-determined vs context-determined transforms
+///
+/// The key is the predicate `Rc`'s address **and nothing else**, so *reusing* an
+/// earlier occurrence's result is only sound for a transform that is a function
+/// of the predicate term. Passes divide on this, and the division decides which
+/// protocol below they use:
+///
+/// - **Key-determined** — [`begin`](Self::begin) + [`finish`](Self::finish). The
+///   result depends on the term plus context that is invariant across the
+///   occurrences sharing it: `uniquify`, `coalesce` (whose predicates resolve out
+///   of one live constraint graph, so two occurrences of one term resolve
+///   identically), `retype`, `simplify`, `inline` and `subst` *within one active
+///   substitution*. A hit skips the transform.
+/// - **Context-determined** — [`begin_always`](Self::begin_always) +
+///   [`finish_shared`](Self::finish_shared). The result depends on something the
+///   key does not name, so the transform **must run at every occurrence**; only
+///   the resulting *term* is unified. Constraint emission is the case: it binds
+///   `REFINEMENT_BINDER` to a domain supplied per occurrence, and skipping it
+///   would leave that occurrence's domain without the constraints the predicate
+///   imposes on it.
+///
+/// A third shape is neither: when the context is *part of the transform's
+/// identity* rather than a parameter of it — `subst`'s active substitution, which
+/// by design acts differently in different scopes — the fix is to scope the memo
+/// so the context is constant within it, rather than to widen the key. See
+/// `subst::rewrite_type_go`.
+///
+/// Mixing them up is a silent wrong answer, not a compile error, which is why the
+/// two protocols are named rather than left as one.
 #[derive(Default)]
-pub struct PredMemo(HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>);
+pub struct PredMemo {
+    entries: HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>,
+    /// Bumped whenever a predicate slot is pointed at a *different* `Rc`. Lets
+    /// [`walk_refined_predicates_mut`] observe re-pointings performed inside a
+    /// callback's own recursion — which the callback's `changed` bit cannot
+    /// report, because a nested memo *hit* mutates the callback's copy without
+    /// the callback doing anything. See [`Self::revision`].
+    revision: u64,
+}
 
 /// A predicate's **origin** `Rc`, captured before its slot was overwritten —
 /// the memo key, and the keepalive pinning that key's address.
 ///
-/// Only [`PredMemo::begin`] mints one and only [`PredMemo::finish`] /
-/// [`PredMemo::finish_unchanged`] consume one, so it cannot be confused with a
-/// rebuild: the "did you pass the origin or the rebuild?" mistake the raw
-/// `record(Rc, Rc)` shape invited is not expressible.
+/// Only [`PredMemo::begin`] / [`PredMemo::begin_always`] mint one, and only the
+/// `finish*` methods consume one, so it cannot be confused with a rebuild: the
+/// "did you pass the origin or the rebuild?" mistake the raw `record(Rc, Rc)`
+/// shape invited is not expressible.
 pub struct Origin(Rc<Expr>);
 
 impl PredMemo {
@@ -812,14 +851,22 @@ impl PredMemo {
     /// its own transform — which is what lets a pass whose transform needs
     /// `&mut Ctx` (with the memo living in that `Ctx`) drive this protocol
     /// without a borrow conflict.
-    pub fn begin(&self, refinement: &mut Refinement) -> Option<(Origin, Expr)> {
-        if let Some((_, rebuilt)) = self.0.get(&refinement.predicate_id()) {
-            refinement.predicate = Rc::clone(rebuilt);
+    pub fn begin(&mut self, refinement: &mut Refinement) -> Option<(Origin, Expr)> {
+        if let Some((_, rebuilt)) = self.entries.get(&refinement.predicate_id()) {
+            let rebuilt = Rc::clone(rebuilt);
+            self.point_at(refinement, rebuilt);
             return None;
         }
-        let origin = Origin(Rc::clone(&refinement.predicate));
-        let copy = (*refinement.predicate).clone();
-        Some((origin, copy))
+        Some(self.open(refinement))
+    }
+
+    /// Open a rebuild for a **context-determined** transform: one whose result
+    /// depends on something the memo key does not name, so it must run at *every*
+    /// occurrence. Always returns the [`Origin`] token and a copy — there is no
+    /// skip — to be closed with [`finish_shared`](Self::finish_shared), which
+    /// unifies the term while leaving the per-occurrence work already done.
+    pub fn begin_always(&self, refinement: &Refinement) -> (Origin, Expr) {
+        self.open(refinement)
     }
 
     /// Close a rebuild: install `rebuilt` into `refinement` and record
@@ -827,8 +874,27 @@ impl PredMemo {
     /// re-pointed at this same term.
     pub fn finish(&mut self, refinement: &mut Refinement, origin: Origin, rebuilt: Expr) {
         let rebuilt = Rc::new(rebuilt);
-        refinement.predicate = Rc::clone(&rebuilt);
+        self.point_at(refinement, Rc::clone(&rebuilt));
         self.insert(origin, rebuilt);
+    }
+
+    /// Close a [`begin_always`](Self::begin_always) rebuild: point the slot at the
+    /// term an earlier occurrence produced if there is one, otherwise record this
+    /// occurrence's `rebuilt` as that shared term.
+    ///
+    /// The caller's transform has already run — that is the whole point — so this
+    /// only decides *which* term the occurrences share. Discarding `rebuilt` on a
+    /// hit is safe precisely because refinement identity is type-blind: the
+    /// occurrences denote one refinement, and the winning term's embedded type
+    /// slots are as good as the discarded one's.
+    pub fn finish_shared(&mut self, refinement: &mut Refinement, origin: Origin, rebuilt: Expr) {
+        match self.entries.get(&Rc::as_ptr(&origin.0)) {
+            Some((_, shared)) => {
+                let shared = Rc::clone(shared);
+                self.point_at(refinement, shared);
+            }
+            None => self.finish(refinement, origin, rebuilt),
+        }
     }
 
     /// Close a rebuild the transform left **untouched**: keep the origin `Rc` in
@@ -842,8 +908,37 @@ impl PredMemo {
     /// from once per *occurrence* into once per distinct predicate.
     pub fn finish_unchanged(&mut self, refinement: &mut Refinement, origin: Origin) {
         let kept = Rc::clone(&origin.0);
-        refinement.predicate = Rc::clone(&kept);
+        self.point_at(refinement, Rc::clone(&kept));
         self.insert(origin, kept);
+    }
+
+    /// A counter that advances every time this memo points some predicate slot at
+    /// a *different* `Rc`.
+    ///
+    /// [`walk_refined_predicates_mut`] brackets its callback with this to learn
+    /// whether the copy it handed over was mutated by the memo itself — a nested
+    /// memo *hit* re-points a slot inside that copy without the callback doing
+    /// anything, so the callback's own `changed` bit cannot report it. Comparing
+    /// revisions is exact for that question and costs one integer.
+    pub fn revision(&self) -> u64 {
+        self.revision
+    }
+
+    /// Point `refinement` at `to`, counting it as a re-pointing when the `Rc`
+    /// actually changes (see [`revision`](Self::revision)).
+    fn point_at(&mut self, refinement: &mut Refinement, to: Rc<Expr>) {
+        if !Rc::ptr_eq(&refinement.predicate, &to) {
+            self.revision += 1;
+        }
+        refinement.predicate = to;
+    }
+
+    /// Mint the [`Origin`] token and the working copy for a miss.
+    fn open(&self, refinement: &Refinement) -> (Origin, Expr) {
+        (
+            Origin(Rc::clone(&refinement.predicate)),
+            (*refinement.predicate).clone(),
+        )
     }
 
     /// Retain `origin` as the keepalive pinning its address for the memo's
@@ -851,7 +946,7 @@ impl PredMemo {
     /// `rebuilt`.
     fn insert(&mut self, origin: Origin, rebuilt: Rc<Expr>) {
         let key = Rc::as_ptr(&origin.0);
-        self.0.insert(key, (origin.0, rebuilt));
+        self.entries.insert(key, (origin.0, rebuilt));
     }
 }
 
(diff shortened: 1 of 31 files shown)


dpmills:

- #32 ◀
    - #33

src/ccl/subst.rs
groundlar, line 1: - subst.rs:37–42 module doc is now stale: "still rebuilds each predicate as a fresh Rc" — no longer true for the vacuous case.

src/ccl/ccl_utils.rs
```diff
@@ -728,46 +716,128 @@ where
...
+/// address cannot be reused during the pass, which [`Self::record`] guarantees
+/// by retaining a `keepalive` clone of every origin `Rc` for the memo's
+/// lifetime: without it, overwriting a `refinement.predicate` could drop the
+/// origin `Rc`, freeing an address a later `Rc::new` in the same pass could
+/// reclaim, so an unrelated predicate landing there would collide and inherit
+/// this entry's rebuild.
+#[derive(Default)]
+pub struct PredMemo(HashMap<PredicateId, (Rc<Expr>, Rc<Expr>)>);
+
+impl PredMemo {
```
groundlar: This is kinda weird in that callers have to do the assignment with a bare `Rc` returned from `rebuilt_for()`, and `record(Rc, Rc)` is pretty easy to swap args (both flagged in the comments).

Claude suggested a nice shape:

```
    /// The origin `Rc`, captured before its slot was overwritten. Only
    /// `PredMemo::begin` produces one, so it cannot be confused with a rebuild.
    pub struct Origin(Rc<Expr>);

    impl PredMemo {
        /// On a hit, re-points `refinement` and returns `None`. On a miss, returns
        /// the origin token and a mutable copy for the caller to transform.
        pub fn begin(&self, refinement: &mut Refinement) -> Option<(Origin, Expr)> {
            if let Some(rebuilt) = self.rebuilt_for(refinement) {
                refinement.predicate = rebuilt;
                return None;
            }
            let origin = Origin(Rc::clone(&refinement.predicate));
            let copy = (*refinement.predicate).clone();
            Some((origin, copy))
        }

          /// Installs the rebuild into `refinement` and records the mapping.
          pub fn finish(&mut self, refinement: &mut Refinement, origin: Origin, rebuilt: Expr) {
              let rebuilt = Rc::new(rebuilt);
              refinement.predicate = Rc::clone(&rebuilt);
              self.0.insert(Rc::as_ptr(&origin.0), (origin.0, rebuilt));
          }
    }
```
Then call sites are pretty nice:
```
      // walk_refined_predicates_mut
      if let Some((origin, mut pred)) = memo.begin(refinement) {
          f(&mut pred, memo);
          memo.finish(refinement, origin, pred);
    }
```

dpmills:
Thanks — the repro was decisive, and it turned out to be pointing at more than either of us thought. All addressed in 1824e010.

## The blocking items

**1. The metric's blind spot — confirmed, and the fix generalized.** You were right that the doc described a traversal the code didn't perform. Rather than bolt on three arms, I named the shape they belong to: **`Expr::walk_type_slots{,_mut}`** is now the single source of truth for "every type slot a node carries" (`ty`, `user_annotation`, `Cast.target`, and every binder's declared type via the existing `walk_binders`). That shape was already duplicated **six** ways — the metric, `planning::compile_refinement_predicates`, `retype_predicate_slots`, `inline_and_beta_reduce`, `count_free_with_visited`, and a `uniquify` test helper. The first four now share it, so a new type-slot-bearing variant updates one match and a pass can't acquire a blind spot by omission.

That sweep found a second instance of the same defect, unrelated to the metric: **`inline_and_beta_reduce` also missed the binder slots.** It walked `ty`, the annotation, and `Cast.target`, but not `Lambda.param.ty` / `Let.binding.ty` / `Case` pattern binders / `For` / `LetRec` — so a UDF use inside a binder-slot predicate could survive as a dangling `Var` once the enclosing `Let` was dropped, exactly the failure mode its own comment describes for `Cast.target`. Fixed by the same walk.

**Your suspected cause was the right shape but the wrong pass.** With the metric corrected I measured at each phase boundary:

| phase | distinct `Rc`s | structurally-equal duplicates |
|---|---|---|
| after lowering + `uniquify` | 3 | 0 |
| after constraint emission | 5 | 2 |
| after `coalesce` | 5 | 2 |
| after `retype` | **7** | 4 |

Two separate defects, and `coalesce` is not one of them:

- **`retype` (5 → 7)** — your diagnosis, exactly. `retype_pred_expr` was a hand-rolled near-duplicate of `retype_predicate_slots` missing the binder slots and with no `Cast` arm at all. Since a predicate *is* a term, the two walks are now **one** function: `retype_in_type` hands each predicate back to `retype_predicate_slots`, whose slot coverage comes from the combinator. The duplication was the bug, so deduplicating is the fix — and it also picked up the `For`/`LetRec` binder scopes the old walk dropped into its catch-all.

- **Constraint emission (3 → 5)** — the deeper one, and the reason your `Cast.target` arm tripped the tripwire on an input where 2 of the 4 duplicates predate `coalesce` entirely. `emit_bare_predicate` rebuilds every predicate it types and threaded **no memo**, so the *first* pass over the tree split sharing before any later pass could preserve it. A nested comprehension's inner filter is reached twice — once on the term-level `Cast.target`, once on the copy of that `Cast` inside the enclosing comprehension's own filter predicate. Emit now threads a `PredMemo` through a `Typing::pred_memo` accessor, so Emit and Check both inherit it. Sharing the memoized copy also makes those occurrences share inference variables, which I think is right rather than incidental: a bare predicate is closed but for `REFINEMENT_BINDER`, so two occurrences of one refinement `Rc` denote the same refinement over the same base and their slots have a single solution. Full suite green either way, including with `--features deep-typecheck`.

Net: **3 distinct `Rc`s, 0 duplicates** — one per structural predicate the program contains, held from lowering all the way through inference. Your #38 reported 7 where 3 suffice; that's now 3.

**2. Vacuous tests.** Agreed on both, and I took your suggestion for the assertion. The tests now assert **no two `Rc`-distinct refinements reachable from an inferred tree are structurally equal** — compared with `Refinement`'s own type-blind `PartialEq`, which is the semantically exact notion here. No threshold, nothing to drift, and the failure prints the duplicated predicates in symbolic form. `ccl_utils::reachable_refinements` exposes the traversal the count derives from so the test can ask that sharper question. The nesting test asserts the same property at every depth 2–6 instead of a growth bound.

Your input-shape note is carried over as a comment on the fixture. One thing worth flagging: I first wrote your `metric_sees_every_predicate_it_claims_to` as a differential check against a `ty`-only walk, and it **self-defeats once the split is fixed** — with sharing preserved, a cast's `target` and its `ty` hold the same `Rc`, so a `ty`-only walk reaches all 3 and the test goes vacuous the moment the bug is gone. So the metric's coverage is pinned by your three hand-built unit tests instead (`Cast.target`, `Lambda.param.ty`, `Let.binding.ty`) — those can't go vacuous. Thresholds are gone rather than re-derived.

The `retype` count tripwire stays, with its doc now saying what a count can and cannot catch: necessary check, while equal-but-distinct is the sufficient one.

## Non-blocking

**3. `subst`'s keepalive.** Right, and it was worse than a fourth site — there were **six**. `planning::predicates::PredMemo` (a `HashMap` type alias) and `uniquify`'s `memo` field each carried their own re-derived keepalive paragraph. All three are now the shared `PredMemo`, so "centralized" is true rather than aspirational. Agreed on not writing a test — address reuse isn't deterministically reproducible, which is the argument for the type-level fix.

**4 & 5.** `shadow()` sunk into the branches that use it; the vacuous path now records `origin ↦ origin`, so the `is_free` scan runs once per distinct predicate instead of once per occurrence — and supplies the keepalive from #3, as you noted.

**6. `inline`'s comment.** Both claims were false; thanks. Rather than delete the sentences I made them true: the memo is hoisted to the scope of the whole `[name ↦ lambda]` sweep. Scoped to the *rewrite* rather than the `inline` pass deliberately — a different binder's rewrite maps the same origin `Rc` to a different result.

I took the generalization too. `walk_refined_predicates_mut`'s callback now returns `changed`, and `PredMemo::finish_unchanged` keeps the origin `Rc` when it's `false` — so *every* caller preserves sharing for predicates it merely walks past, not just `subst`. `simplify` reports its own change bit; `inline_in_type_predicates` reports `is_free(name, pred)`. I went with `-> bool` over "compare the rebuild": `Refinement`'s equality is type-blind, so it would report `retype` as a no-op when retyping is *entirely* type slots.

**7. `coalesce`'s memo key.** Your reasoning is what I concluded too, so it's now written down at the call site rather than left load-bearing-but-unstated: sharing means literally the same term with the same inference variables, resolved out of one live constraint graph, so both occurrences resolve identically and the first result is the only result. The converse — two refinements that should resolve differently sharing an `Rc` — can't arise, because a shared `Rc` is only ever created by copying one occurrence of one refinement. I left out the `debug_assert`: comparing "the scope this hit would have resolved under" needs the resolution to have happened, which is the thing being skipped, so the assertion would have to re-run `coalesce_node` to check it. The precondition as prose plus the end-to-end guard seemed the honest trade; happy to revisit if you'd rather pay for it.

**Nit.** `Refinement { predicate }` literals are gone. `born` couldn't be the single spelling because it can't express the other intent — a second occurrence of an *existing* predicate. `Refinement::sharing(&Rc)` names that (what lowering does when one filtered domain rides four slots), so the two constructors cover every case and a bare literal states neither. The remaining step to make it enforced rather than conventional is privatizing the field, which needs an accessor at every `r.predicate` read; say the word and I'll do it separately.

## Noted, not fixed

`retype_predicate_slots`, `subst::rewrite_expr_go`, and `infer::api`'s scope walks each spell out "walk children, each under the binders that scope it" by hand. Unlike the type-slot set, the scoping is genuinely non-uniform (`Let` binds only its body, `LetRec` binds everything, a `Case` branch binds its guard and body), so a `walk_children_scoped` combinator would have to encode that structure — bigger than this PR should carry. Flagged at the call site so it isn't lost.

`src/ccl/design/type-inference.md` gains a "Sharing is an invariant, not an optimization detail" subsection: the full list of rebuilding passes, why the keepalive is load-bearing, why the cost lands in planning, and what each of the two guards does.

#38 is superseded — its five tests all pass now (the three metric-coverage ones are carried over near-verbatim, and the two program-level ones as the structural-equality assertion). Want me to close it, or would you rather confirm against it first?

src/ccl/subst.rs
dpmills: Fixed. The module doc now says the in-place mode rebuilds a predicate it *touches* and keeps the `Rc` for one it merely walks past — same as transport mode — and pulls the memo out into a shared paragraph, since both modes now thread the same `ccl_utils::PredMemo` rather than the mode-specific `BTreeMap` this described.

src/ccl/ccl_utils.rs
dpmills: Agreed on both counts, and I took the shape almost verbatim — thanks, it's much better. `record(Rc, Rc)` with two same-typed args and a bare `Rc` for the caller to assign was exactly as swappable as you say.

`PredMemo` now exposes:

- `begin(&mut Refinement) -> Option<(Origin, Expr)>` — on a hit, re-points the slot and returns `None`; on a miss, hands back the `Origin` token plus an owned copy.
- `finish(refinement, origin, rebuilt)` — installs and records.
- `finish_unchanged(refinement, origin)` — the transform had nothing to do: keep the origin `Rc`, record `origin ↦ origin`.

`insert` is private, so `Origin` is minted only by `begin` and consumed only by the two `finish`es — the keepalive can't be forgotten and can't be confused with a rebuild. Call sites read as you sketched:

```rust
// walk_refined_predicates_mut
if let Some((origin, mut pred)) = memo.begin(refinement) {
    if f(&mut pred, memo) {
        memo.finish(refinement, origin, pred);
    } else {
        memo.finish_unchanged(refinement, origin);
    }
}
```

Two deltas from your sketch:

- **`finish_unchanged` as a third operation.** This is what generalizes the vacuous fast path `subst` had grown privately (your #5 and #6): the callback returns `changed`, and on `false` the origin `Rc` stays in the slot. So a pass preserves sharing for predicates it walks past instead of reallocating them per node — which is what makes `inline`'s memo claim structurally true rather than asserted.
- **`begin` takes `&self`, not `&mut self`.** Only `finish`/`finish_unchanged` need `&mut`, which is what keeps `coalesce_type_predicates` working: its transform is `coalesce_node(&mut ctx)` and the memo lives *in* that ctx, so it needs the lookup's borrow released before the transform runs.

The protocol also absorbed three more hand-rolled copies of this same dance — `subst`'s `BTreeMap`, `planning::predicates::PredMemo`, and `uniquify`'s `memo` — so it's now one implementation rather than five.

dpmills:
## Review round two (#39): the memo protocol, split by what determines its result

Pushed as e388a81e. #39's five tests are carried here with authorship preserved (`tests/predicate_sharing_review.rs` plus one lib test); the control still passes and none of the five needed editing.

All five findings trace to one root cause: `PredMemo` is keyed on the predicate `Rc`'s address **and nothing else**, but not every pass threading it computes a function of the predicate term. Where the transform depends on something the key does not name, reusing an earlier occurrence's result is a silent wrong answer. That distinction is now part of the API rather than a property each call site is trusted to have:

- **Key-determined** — `begin`/`finish`; a hit skips the transform. `uniquify`, `coalesce`, `retype`, `simplify`, `inline`, and `subst` within one active substitution.
- **Context-determined** — `begin_always`/`finish_shared`; the transform runs at every occurrence and only the resulting term is unified.
- A third shape wants the memo **scoped** rather than the key widened: when the context is part of the transform's *identity* instead of a parameter of it.

| # | Finding | Fix |
|---|---|---|
| 1 | `subst`'s memo crossed shadowing boundaries | `Subst::recurse_shrunk` — fresh memo whenever the substitution actually shrinks |
| 2 | Emission's memo hit dropped the occurrence's own domain | `begin_always`/`finish_shared`: emit per occurrence, unify only the term |
| 3 | `rewrite_expr_go` never rewrote binder-slot types | on `Expr::walk_type_slots_mut` |
| 4 | `count_free`/`is_free` missed binder-slot predicates | on `Expr::walk_type_slots` |
| 5 | `changed` couldn't see a nested memo hit's re-pointing | `PredMemo::revision()`; the combinator returns its own bit |

Finding 1 is the sharpest, and it is the third shape: substitution acting differently in different scopes is the *point*, so "same term ⇒ same result" is exactly what shadowing denies. A memo carried across a shadowing binder crossing was wrong in both directions — an outer rebuild served to an occurrence where the variable is rebound, and a `finish_unchanged` vacuity decision made under a restricted substitution served to a later occurrence under the full one. All five shrink sites now route through one helper, so the invariant holds by construction rather than by argument.

Finding 2 was a defect in this PR's own emit change. Emission binds `REFINEMENT_BINDER` to a domain supplied per occurrence and `emit_cast` mints a fresh one per cast node, so skipping it left the second occurrence's domain under-determined — the worst failure mode. **Sharing did not regress**: the motivating comprehension still leaves inference with 3 distinct predicate `Rc`s and no structurally-equal duplicates, because the split that memoization was preventing was in the *term*, not in the emission. I re-derived that rather than assuming it.

Findings 3 and 4 were the last two hand-rolled type-slot enumerations. Finding 4 is the one with teeth: three fast paths now skip work when `is_free` says no, so a slot the free-variable walk couldn't see was a slot those passes declined to rewrite.

**Planning's compilation**, which #39 flagged as the same shape, keeps the skipping protocol — recompiling per occurrence is the superlinearity this work exists to remove — because `base` reaches only the compiled term's *type* slots and refinement identity is type-blind. Reuse is sound provided the bases agree, which nothing enforced. `predicates::CompileMemo` now records each entry's base in debug builds and checks it on the hit path; release builds are exactly a `PredMemo`. It does not fire anywhere in the suite.

`./ci.sh` clean (both clippy passes, doc, tests) and green with `--features deep-typecheck`. The design doc's sharing section carries the taxonomy.

dpmills:
## The memo now keys on the predicate *and the conditions it was rebuilt under*

Pushed as 2a066b27. Every defect the review found — and the two the follow-up audit found — was the same mistake, so this replaces the argument each pass was trusted to have made with a mechanism.

### `PredMemo<C>`

`C` is what the rebuild depends on besides the term. An entry is reused only when it was recorded under an equal `C`, so supplying the wrong one costs a sharing opportunity rather than producing a wrong answer. Each pass now states its dependency:

| pass | `C` |
|---|---|
| `simplify` | `()` — a function of the term |
| `uniquify`, `coalesce`, `retype`, `inline` | `()`, justified at the call site |
| `subst` | `Subst` — the active substitution |
| planning | `Type` — the refinement base its compilation reads |

Two of the `()` claims are load-bearing and now say so: `inline`'s rests on its binder arms *skipping* a shadowed subtree rather than descending into it, and `retype`'s on freshening giving monomorphization clones distinct predicate `Rc`s.

**Two mechanisms delete.** `Subst::recurse_shrunk` was a weaker fix for the same problem — it threw entries away at every substitution shrink, so sibling shadowed scopes couldn't share with each other; carrying the `Subst` as `C` keeps every scope's entries live at once. `predicates::CompileMemo` (a wrapper struct plus a debug-only side map plus a hit-path assertion) becomes `PredMemo<Type>`, and a base mismatch now *recompiles* instead of panicking in debug and quietly borrowing another occurrence's element type in release.

`TermMemo` is the separate type for sharing *allocations* without sharing *results* — constraint emission, which binds `REFINEMENT_BINDER` to a domain `emit_cast` mints fresh per cast node. Which of the two a pass is entitled to is visible in its signature now.

### The protocol is a closure, so a rebuild cannot be half-done

`rebuild`/`rebuild_always` take the transform as a closure. The token shape they replace could be left open — dropping it discarded the rebuild, left the occurrence on its origin `Rc`, and memoized nothing, so a pass returning early between the halves rewrote every occurrence *but one*. Both leaks that existed are unrepresentable now rather than caught at runtime, and `Origin` leaves the public API.

The closure is also what lets the memo be a cheap clonable handle. That matters because three passes own their memo inside the very context their transform needs mutably — `uniquify`'s `self.expr` → `self.ty`, `coalesce_node` → `coalesce_type_predicates`, `subexpr` → `emit_node` → `emit_cast`. Reaching a handle needs only `&ctx` and no borrow is held across the callback, so those transforms re-enter the memo freely; threading the memo as a parameter instead would have meant changing `Typing::subexpr` and every emit rule.

### Verification

All five review guards and the metric-coverage tests pass unchanged (only their call syntax moved). Sharing is unaffected — still 3 distinct predicate `Rc`s with no structurally-equal duplicates on the motivating comprehension.

And, measured against `main` for the first time, the payoff this branch exists for — compiling a nested filtered comprehension over `[1, 2, 3, 4]`, release:

| depth | `main` | this branch |
|---|---|---|
| 3 | 11.7ms | 4.8ms |
| 5 | 248ms | 28.2ms |
| 6 | 1.2s | 62.3ms |

The refactor itself is performance-neutral against the pre-refactor tip, so carrying a `Subst` per entry and scanning a per-key entry list cost nothing measurable. Net −86 lines.

dpmills:
## Rebased onto `main` with literal singletons (#36), and they turned out to need something

Pushed as 02bd741f. The rebase itself was clean — every conflict was an import union — but the suite was not: `nested_comprehension_shares_predicate_rcs` and `filtered_join_nesting_stays_shared` both failed. A two-generator join over `[1, 2, 3]` left inference with **13** distinct predicate `Rc`s, 9 of them redundant: `4x __elem == 1`, `4x __elem == 2`, `4x __elem == 3`.

Measuring at the phase boundaries said these were **not a split**: post-lowering the tree has one predicate and no singletons at all, since `emit_node` mints them. Each literal *occurrence* was getting its own `{Int | __elem == 1}`. So the guards were reporting something real but not what they were written to catch — their premise is "structurally equal ⇒ was shared ⇒ got split", and per-occurrence minting makes structurally-equal-but-never-shared predicates routine.

**Rather than weaken the guard, I made the premise true again.** `InferCtx` now keeps a pass-scoped table from a literal *value* to its singleton predicate, so every occurrence of `1` shares one `Rc`. That join goes 13 distinct / 9 redundant → **4 / 0**.

Sharing singletons is unusually safe, for a reason worth naming: the predicate is **ground** — built fully typed, closed but for `REFINEMENT_BINDER`, a pure function of the `Lit`. None of the context-dependence that makes sharing delicate elsewhere applies, because there is nothing left to resolve. `singleton_predicate` is split out of `lit_singleton` so that is a documented property rather than an observation.

### What the combination costs and buys

Compiling a nested filtered comprehension, release:

| depth | `main` before #36 | `main` with #36 | this branch |
|---|---|---|---|
| 3 | 11.7ms | 50.2ms | 8.5ms |
| 5 | 248ms | 1.1s | 69.3ms |
| 6 | 1.2s | 5.4s | 151.6ms |

#36 costs ~4–5× on that shape because every literal adds a refinement; this branch more than absorbs it.

On a literal-heavy program (comprehension over a list with repeating values), the interning is what pays:

| literals | without interning | with |
|---|---|---|
| 40 | 5.4ms | 3.6ms |
| 80 | 7.6ms | 5.2ms |
| 160 | 15.4ms | 8.6ms |

The gap widens with literal count, which is the expected shape — planning compiles once per distinct predicate `Rc`.

Also: `lit_singleton`'s `Refinement { predicate }` literal became `Refinement::born`, restoring the two-constructor rule, and `Lit` gained `Eq + Hash` to key the table (total — no floats).

### Noted, not fixed

`mut_elim::erase_mut` is another hand-rolled type-slot walk and it omits `Cast.target` — the same shape as the ones this branch converted, but about `Type::History` erasure rather than predicates. Whether a `History` can reach a cast target is a separate question in a separate subsystem, so I left it alone rather than widen this branch further.

`./ci.sh` clean, green with `--features deep-typecheck`.

dpmills:
Correction to the "Noted, not fixed" item above — `mut_elim` **is** fixed, in 2aefe97f. Investigating it changed the picture.

It is not one hand-rolled walk but **three**, sharing the same `Cast.target` blind spot: `erase_mut` (the erasure), `contains_mut_type` (the *release-mode* post-condition asserting the erasure was complete), and `infer::collect_expr_errors` (the strict wall downstream). An eraser and its own checker sharing a blind spot is the failure mode where the guard cannot observe what it guards — a history on a cast target would survive erasure, pass the post-condition, pass the wall, and reach `channelize`/`lambda_elim`/planning silently.

**Latent, not live.** Instrumenting the pass input across the whole suite found no `Cast.target` carrying an `Overwrite` history, and there is a structural reason: `emit_cast` gives a cast the function type `{D | r} ⇒ V`, while a register's type is `Mut(V, dom)` — never that shape. So the fix is behaviour-neutral, which the suite confirms.

`erase_mut` and `contains_mut_type` now walk the same `Expr::walk_type_slots{,_mut}` set, so they cannot disagree about which slots exist: a slot is either erased *and* checked, or neither. Net −21 lines.

**The strict wall stays as it is**, and after looking at #44 that is now a documented decision rather than an open question. #44 makes every inference error carry a blame node and dedups on `(node, error)` — but `error` includes the `at` display label its own docs call "not the location", and routing the wall through `walk_type_slots` produces reports that differ *precisely* in that label (a lambda's param hole reachable both as `expr.ty`'s domain and as `param.ty`). #44 settles the wall's design as per-node blame with per-binder labels, which depends on those labels differing; widening the coverage would work against it. Details and two findings for that author are on #44 — including that the wall today silently passes an unresolved `Hole` in a `Cast.target` or a `user_annotation` (both return `Ok(())`; verified).

Coverage and dedup are coupled: widening needs a label-independent key, and a label-independent key is what makes widening safe. Neither belongs here alone.

src/ccl/expr.rs
```diff
@@ -1201,6 +1201,53 @@ impl TypedExpr {
...
+    /// [`crate::ccl::ccl_utils::distinct_predicate_rcs`].
+    pub fn walk_type_slots(&self, mut f: impl FnMut(&Type)) {
+        f(&self.ty);
+        if let Some(annotation) = &self.user_annotation {
+            f(annotation);
+        }
+        if let TypedExprNode::Cast { target, .. } = &self.node {
+            f(target);
+        }
+        self.walk_binders(|b| f(&b.ty));
```
groundlar: **Blocking — this walk omits `TypedBinding.user_annotation`, and converting `mut_elim` onto it is a live regression.**

A binder carries *two* independent type slots, `b.ty` and `b.user_annotation`, and the annotation is load-bearing rather than decorative: lowering writes a mutable variable's `Mut(V, D)` history there (`lower/stmts.rs` for `x := e` via `let_bind_annotated`; `lower/functions.rs:120`/`:153` for a `Mut[…]` param), and `infer::api::binder_is_mut` treats it as *authoritative over* `b.ty` precisely because an explicit annotation wins.

Before this branch `erase_mut_in_binding` erased both slots and `contains_mut_type`'s `binder_has_mut` checked both. Now both go through `walk_type_slots{,_mut}` (`mut_elim.rs:127`, `:162`), so the erasure stopped erasing the annotation *and* the post-condition stopped being able to see that. That is exactly the failure mode the commit message says it closed:

> An eraser and its own checker sharing a blind spot is the failure mode where the guard cannot observe what it guards.

**It is live, not latent.** Adding an independent binder-annotation check to `contains_mut_type` fails 3 of 6 `cli_driver_convergence` tests on the ordinary `x := 0` accumulator path. Nothing downstream catches it because the strict wall doesn't check binder annotations either (`infer/api.rs:569-600`) — which is why the suite is green.

A guard has to enumerate the slot itself, or it inherits the same blind spot. Sketch, in `mut_elim`'s test module:

```rust
/// Deliberately independent of `contains_mut_type`: the post-condition and the
/// erasure share one walk, so if that walk misses a slot the post-condition
/// cannot report it.
fn binder_annotation_has_mut(expr: &Expr) -> bool {
    fn ty_has_mut(ty: &Type) -> bool {
        matches!(ty, Type::History { kind: HistoryKind::Overwrite, .. })
            || ty.fold_children(false, |acc, t| acc || ty_has_mut(t))
    }
    let mut found = false;
    expr.walk_binders(|b| {
        if let Some(ann) = &b.user_annotation {
            found |= ty_has_mut(ann);
        }
    });
    found || expr.any_child(binder_annotation_has_mut)
}

#[test]
fn phase_erases_mut_from_a_binder_annotation() {
    let (mut tree, _, _) = direct_mirror_sum();
    // As lowering actually emits `x := 0`: the history rides the annotation.
    let TypedExprNode::Let { binding, .. } = &mut tree.node else { panic!() };
    binding.user_annotation = Some(Type::History {
        value: Box::new(Type::Base(BaseType::Int)),
        domain: Box::new(Type::Hole),
        kind: HistoryKind::Overwrite,
    });
    assert!(binder_annotation_has_mut(&tree), "sanity");

    let out = run(tree);

    assert!(!binder_annotation_has_mut(&out), "{}", symbolic(&out));
}
```

Currently fails with the `Mut` still on the `let x` binding after the phase. A matching combinator-level guard stamping a distinct `Type::DataSource` marker into each of `node_ty` / `node_annotation` / `param_ty` / `param_annotation` and asserting `walk_type_slots` reaches all four reports `["node_ty", "node_annotation", "param_ty"]`.

**Fix** — extend both walks:

```rust
self.walk_binders(|b| {
    f(&b.ty);
    if let Some(annotation) = &b.user_annotation {
        f(annotation);
    }
});
```

I ran this: it turns both guards green and the full suite stays green. It's also the right direction for the other four converted sites — `subst::rewrite_expr_go`, `inline_and_beta_reduce`, `count_free_with_visited` and `retype_predicate_slots` never covered binder annotations either, so widening the combinator closes four omissions at once rather than restoring one.

Until it's fixed, three claims are false and actively steering passes: this doc-comment's "single source of truth for which type slots a node carries", `walk_type_slots_mut`'s "any new type-slot-bearing variant must appear in both", and `design/type-inference.md:311-314`.

src/ccl/expr.rs
```diff
@@ -1201,6 +1201,53 @@ impl TypedExpr {
...
+    }
+
+    /// Mutable analog of [`walk_type_slots`](Self::walk_type_slots), for passes
+    /// that rewrite type slots in place. Kept in lockstep with it — any new
+    /// type-slot-bearing variant must appear in both.
```
groundlar, line 1260: *(Anchored here because `TypedExprNode::Transact` at `src/ccl/expr.rs:348` is outside this PR's diff — this is the claim it falsifies.)*

**`Transact`'s sequencing domain is a second type slot `walk_type_slots` misses.**

It's the only directly-carried `Type` outside that walk's match (`ty`, `user_annotation`, `Cast.target`, binder types), so the walk's exhaustiveness claim — "the single source of truth for which type slots a node carries", "a new type-slot-bearing variant then updates one place instead of every such pass" — is already false for it.

Latent today, and for a real reason: `Transact` is born by `plan_loops` *after* every pass currently on the walk. But nothing links the walk to the variant list, so nothing makes the next slot-bearing variant land in it either — which is the same "acquire a blind spot by omission" this abstraction exists to prevent, just one level up.

Failing guard, hand-built since no pipeline stage produces a `Transact` early enough:

```rust
#[test]
fn walk_type_slots_reaches_a_transact_domain() {
    let mut txn = TypedExpr::new(TypedExprNode::Transact {
        keys: vec![TransactKey { name: Name::raw("k"), init: leaf("seed") }],
        writers: vec![WriterSite {
            read_keys: vec![Name::raw("k")],
            write_keys: vec![Name::raw("k")],
            source: leaf("src"),
            body: leaf("decide"),
        }],
        domain: Type::DataSource("transact_domain".into()),
    });
    txn.ty = Type::DataSource("node_ty".into());

    let mut seen = Vec::new();
    txn.walk_type_slots(|ty| if let Type::DataSource(n) = ty { seen.push(n.to_string()) });
    assert!(seen.contains(&"transact_domain".to_string()), "reached {seen:?}");
}
```

Reports `reached ["node_ty"]`.

Either cover the slot, or scope the docs to say which nodes the claim holds for and why `Transact` is excluded — I'd take either, but not silence. Happy for this to be a follow-up rather than in this PR; the thing I'd not want is the third instance arriving before the second is written down.

groundlar (approved):

groundlar:
*(The body of [my review above](https://github.com/cambra-dev/cambra/pull/32#pullrequestreview-4813057009) was lost on submit — the two inline comments on `src/ccl/expr.rs` are intact. Reposting the prose here.)*

The core of this is right, and it got materially better across the rounds. `PredMemo<C>` — keying on the predicate *and the conditions the rebuild depends on* — is the correct abstraction: it turns "each call site is trusted to have made an argument" into a mechanism where a wrong `C` costs a sharing opportunity instead of a wrong answer. The closure protocol beats the token pair (a half-done rebuild is unrepresentable rather than caught), splitting `TermMemo` out so "reuse the answer" vs. "reuse only the allocation" is visible in the type is the right call, and depth 6 at 1.2s → 62ms against `main` while absorbing #36's 4-5x justifies the surface area.

I confirmed `./ci.sh` green on the tip, all five #39 guards and the three metric-coverage tests present and passing, and #42's `Case`-arm regression resolved in `main` — though by a *different* fix than #42 proposed (join on a fresh var, `emit.rs:1120-1135`, rather than strip-both-sides), which is probably worth a note on #42 before it stays closed.

Two blocking-ish findings inline on `walk_type_slots` and `Transact.domain`. The rest:

### Doc rot: four live references to the retired token protocol

Comments have to make sense against the current code, and these describe an API that no longer exists:

- `src/ccl/design/type-inference.md:292-295` — "whose `begin`/`finish` protocol maps an origin `Rc` to a single rebuild; `finish_unchanged` is the stronger case". Line 358 of the same doc explains that shape was retired.
- `src/ccl/infer/emit.rs:304-305` — "uses `PredMemo`'s `begin_always` / `finish_shared` protocol". Those methods don't exist, and it isn't a `PredMemo` method at all — it's `TermMemo::rebuild_always`.
- `src/ccl/infer/solve.rs:1053` — "`begin` hands back an owned copy and drops its borrow".
- `tests/predicate_sharing_review.rs:182` — "so `finish_unchanged` discards them".

### The open thread on `solve.rs` is still open

My earlier nit on the `coalesce_type_predicates` doc block is unaddressed, and it's the same block as above with a second problem: `:1049-1057` has two adjacent comments both explaining the handle clone, one of them via the dead `begin`. Collapse to one and drop the API restatement — the `C = ()` justification is the part that belongs at this call site, and it's good.

### Nits

- **`simplify_once` (`simplify.rs:164`)** walks only `expr.ty`, not `walk_type_slots_mut` — the last rewriting pass not on the combinator. Not a regression, and post-`lambda_elim` the missed slots are likely empty, but it's the odd one out in a change whose whole point is one enumeration. Same for `channelize::erase_chan_domains`, which additionally *does* cover binder annotations — so it's currently more complete than the shared walk.
- **`channelize.rs:722`** — "shared coverage with `mut_elim::erase_mut` via `walk_binders_mut`" is stale; `erase_mut` is on `walk_type_slots_mut` now, and per the inline finding the coverage is no longer shared in the direction that matters.
- **`PredMemo<Type>` / `PredMemo<Subst>` lookup cost** — each hit does a structural `Type`/`Subst` compare, and planning clones the base per refinement node. The measurements say neutral and the entry-list-of-1 comment explains why, but the `Type` case compares refinement predicates deeply; a sentence in `predicates.rs` on why that stays cheap would preempt the question.
- **`Refinement.predicate` is still `pub`**, so `born`/`sharing` are conventional rather than enforced. No bare literals remain anywhere, so keeping the privatization as the stated follow-up is fine by me.

src/ccl/expr.rs
dpmills: You're right, and thank you for the specificity — this was a live regression I introduced, not a latent gap. Fixed in 8ef7b211.

I verified both halves before changing anything: `x := e` does lower via `let_bind_annotated` (`lower/stmts.rs:545`), putting the `Mut(V, D)` on the binder's **annotation**, and `binder_is_mut` reads the annotation as authoritative over `b.ty`. Then your guard, added verbatim in intent, failed exactly as you said:

```
a history survived on a binder annotation: let x : Int = 0
in letrec __hist : ([0, 2] ⇒ {commit: Bool, writes: (Int)}) = …
```

So `mut_elim` had been erasing that slot and checking it, and my conversion silently dropped both — the failure mode the commit message claimed to close, introduced by the commit that claimed it. Worth stating plainly.

I took your fix over restoring `mut_elim`'s local helper, for the reason you gave: widening the combinator closes the same omission in `subst::rewrite_expr_go`, `inline_and_beta_reduce`, `count_free_with_visited`, and `retype_predicate_slots` at once. Full suite green, and I kept your point about the guard needing to enumerate the slot *independently* — `phase_erases_mut_from_a_binder_annotation` does not go through `contains_mut_type`, since a post-condition sharing the erasure's walk cannot report what that walk misses.

The three false claims are now true rather than deleted: the walk's doc, `walk_type_slots_mut`'s lockstep note, and `design/type-inference.md` all name the binder annotation, with the reason it is load-bearing rather than decorative.

src/ccl/expr.rs
dpmills: Confirmed, and taken — with the scoping option rather than coverage, for a reason your comment anticipated.

I enumerated every directly-carried `Type` in the AST rather than fixing the one you named. The full inventory is six: `TypedExpr.ty`, `TypedExpr.user_annotation`, `TypedBinding.ty`, `TypedBinding.user_annotation`, `Cast.target`, and `Transact.domain`. Your two comments are exactly the two the walk was missing — the binder annotation is now covered (it was a live regression, see the other thread), and `Transact.domain` is now a *named* exclusion.

Why scoped rather than covered: `plan_loops` births `Transact` **before** `planning::run` calls `compile_refinement_predicates` (`context.rs:673` then `:677`). So covering the slot is not the no-op the "born after every pass on the walk" framing suggests — it would newly expose the sequencing extent to planning's predicate compilation. `build_transact_store` only distinguishes `Txn` from an extent and never inspects refinements there, so I could not convince myself either way about what compiling a loop extent's refinement does downstream. That is a behavioural change in the recurrence engine, and I'd rather it not ride along in a predicate-sharing branch.

Rather than leave it as prose, the exclusion is now pinned:

```rust
#[test]
fn walk_type_slots_covers_every_carried_type_slot() { … }
```

It stamps a distinct `DataSource` marker into every carried slot and asserts exactly which the walk reaches — including that a `Transact` reports `["node_ty"]` and not its domain. So if someone covers the slot, the test fails and the docs have to be updated with it; the exclusion cannot go silent, and a third instance arriving before the second is written down is the case it guards against. Both the walk's doc and `design/type-inference.md` state the exception and the reason.

Happy to do the coverage as a follow-up if you'd like it — my hesitation is only about evaluating the planning-side effect from this branch, not about the direction.

dpmills:
Thanks — and sorry, I worked the two inline comments first and missed the prose until it was pointed out. All of it is addressed now, in 363ee175 (the two inline findings were 8ef7b211).

## Doc rot — all four, plus the open thread

Every one was real; they described an API that no longer exists.

- `design/type-inference.md` — now states what `rebuild` does rather than `begin`/`finish` with `finish_unchanged` as "the stronger case".
- `infer/emit.rs` — you're right that it was doubly wrong: not just retired methods but the wrong *type*. It is `TermMemo::rebuild_always`, and saying so makes the point the paragraph exists for — emission shares terms, not results.
- `infer/solve.rs` — this was also your open thread, and the two problems were one: the second of the two adjacent comments explained the handle clone *through* the dead `begin`. Collapsed to one, keeping the `C = ()` justification.
- `tests/predicate_sharing_review.rs` — section header fixed.

The design doc keeps exactly one mention of the token shape, deliberately: the paragraph explaining why it was retired. That reads as an alternative-that-doesn't-work note rather than history, so it stays.

## `simplify_once` — measured, and kept narrow

I tried your suggestion before deciding, and the measurement argued against it, so I've documented the reason instead of making the change:

- Across the whole suite, **no** node reaching that walk carries a refinement in a slot `expr.ty` does not reach. Your "likely empty" is exactly right, and it is empty for two structural reasons: this runs after lambda elimination has removed the `Lambda` nodes whose `param.ty` would carry a domain refinement, and `lambda_elim::run` follows it with `sync_cast_targets`, which re-points every `Cast` target at the already-simplified `ty` rather than needing it simplified separately.
- Widening it costs ~30% of compile time at depth 3 on the nested-comprehension benchmark (11.1ms → 14.6ms; neutral by depth 6) for zero additional coverage.

Both reasons are now at the call site, so the asymmetry reads as a decision rather than an oversight. Happy to reverse if you'd rather have the uniformity at that price.

`channelize::erase_chan_domains` I left on its own walk for a reason the stale comment was hiding: it visits binder slots, then children, then the node's own `ty`/annotation, in three ordered phases around a `Let`-discharge that mutates the domain map — one combined call would collapse that ordering. The comment now says what it covers and why, instead of claiming shared coverage with `erase_mut`.

## Nits

`predicates.rs` gains the sentence you asked for. The short version: a refinement's base sits *below* it, so comparing `{Int | p}`'s base compares `Int` and never `p`, and the per-key entry list is a singleton in every measured shape — so a hit is one compare of a small type, and the per-node clone is that same shallow value.

`Refinement.predicate` stays `pub` with privatization as the follow-up, per your read.

## #42

Good catch, and noted there. `main` fixes it by minting a fresh result variable and relating every arm to it, so the pairwise `prev` comparison the asymmetry lived in is gone entirely — not the symmetric deep strip #42 proposed. Worth recording because they differ on the second defect that PR raised: a symmetric strip discards both arms' witnesses, whereas the join intersects them, which is what keeps the "every arm establishes it" case refined.

dpmills:
Correcting my previous comment on `simplify_once`: **you were right, I was wrong, and the evidence I cited against you was invalid.**

I claimed "across the whole suite, no node reaching that walk carries a refinement in a slot `expr.ty` does not reach." That probe used `eprintln!` under a plain `cargo test`, which captures output for *passing* tests — so I was grepping an empty capture buffer, not an empty result. A false negative I then reported as a measurement.

Checking the property with a `debug_assert!` instead of a print fails immediately, on ordinary programs:

```
simplify's predicate walk reads `expr.ty` only, but a refinement rides a slot it
cannot reach: let x : 7 = 7 in letrec __hist : (∅ ⇒ …)
```

A `Let` binding's declared type routinely carries a **literal singleton** — `x := 0` lowers to `let x : 0 = 0` — so that slot has been non-empty since #36 landed, which this branch rebased onto. Three `cli_driver_convergence` cases hit it on the plain accumulator path.

So it is a real split rather than a stylistic inconsistency, and worse placed than most: an unvisited occurrence keeps its origin `Rc` while a visited sibling moves to the rebuild, in the one pass whose output the post-pass `typecheck` compares by *structural* refinement equality. Latent only because nothing currently rewrites a singleton predicate — which is exactly the kind of "latent" this branch has twice found to be one change away from live.

Fixed in 4cb49abb: `simplify_once` is on `Expr::walk_type_slots_mut`, your original suggestion. Re-measured honestly (best of three, release, nested filtered comprehension):

| depth | narrow | widened |
|---|---|---|
| 3 | 8.8ms | 9.9ms |
| 5 | 75.1ms | 82.5ms |
| 6 | 167.2ms | 192.7ms |

~10-15%, not the ~30% I quoted from a single noisy run. Correctness settles it regardless, but the number I used to argue against you was also inflated.

I removed the `debug_assert!` that caught this rather than keeping it: widening the walk retires the assumption it was checking. `channelize::erase_chan_domains` still keeps its own walk, and that reason does hold — it visits binder slots, then children, then the node's own slots in three ordered phases around a `Let`-discharge that mutates the domain map, so one combined call would collapse the ordering. Happy to be checked on that one too.

Two lessons I'm taking from this beyond the fix: a probe that reports absence needs to prove it can report presence, and "measured" is not a synonym for "true" when I'm the one who wrote the measurement.

PR #33, opened 2026-07-24
Gate per-operation debug_typecheck behind opt-in `deep-typecheck` feature

dpmills:
Gate per-operation debug_typecheck behind opt-in `deep-typecheck` feature

`debug_typecheck` runs a full `check_pre_desugar` over the whole (sub)tree, and
pass-internal helpers (lambda elimination, substitution, simplify) call it after
*every* rewrite to localize which operation first produced an ill-typed tree.
That is O(subtree) per rewrite, so on nested comprehensions it is superlinear
(O(rewrites) × O(subtree)) and dominated *debug* compile time — release builds
compiled it out, so it only ever taxed `cargo test` / local iteration. The
lambda-elimination algorithm itself is near-linear (release compiles the deepest
cases measured here in milliseconds).

`compile_program` already runs `check_pre_desugar` at every pass boundary; those
walls are the always-on correctness net. The per-operation version only adds
localization, so gate it behind an opt-in `deep-typecheck` feature (off by
default, even in debug). Enable it — `cargo test --features deep-typecheck` — to
bisect a miscompile to its introducing operation.

Automated GitHub runs keep the check on: `ci_test` enables the feature when
`DEEP_TYPECHECK` is set, and the workflow's test step sets it. So CI still
guards against a pass emitting an ill-typed tree, while a bare local `./ci.sh`
and normal `cargo test` stay fast. Verified the whole compilation-pipeline suite
still passes with the feature on, so nothing relied on the per-op check to catch
a live error.

Effect on the compilation-pipeline test binary (debug, parallel, feature off):
~3.2s -> ~0.27s (~11x). Combined with the predicate-Rc-sharing fix this stacks
on, the binary is down from ~6.8s to ~0.27s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,11 @@ jobs:
         # Read more about the "hidden ifs" and general inanity here
         # https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions
         if: steps.filter.outputs.code == 'true' && (success() || failure())
-        run: ./ci.sh test
+        # Enable the opt-in per-operation typecheck (the `deep-typecheck`
+        # feature) in automated runs so it keeps guarding against a pass that
+        # produces an ill-typed tree — it is off by default locally because it is
+        # superlinear on nested comprehensions (see `ci_test` / `debug_typecheck`).
+        run: DEEP_TYPECHECK=1 ./ci.sh test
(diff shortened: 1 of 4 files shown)


dpmills:

- #32
    - #33 ◀

groundlar (approved):

PR #34, opened 2026-07-24
ci.sh: add `fast` inner-loop gate for local iteration

dpmills:
ci.sh: add `fast` inner-loop gate for local iteration

The full `./ci.sh` runs six phases; on a one-file edit the two heaviest are the
release clippy pass and the doc build, and clippy's `--all-targets` roughly
doubles the debug clippy time to check test targets that `cargo test` then
recompiles anyway (clippy-driver fingerprints separately from rustc, so nothing
is shared — the old `ci_clippy` comment claiming it "warms the cache for cargo
test" was wrong, now corrected).

`./ci.sh fast` runs just what an edit→check loop needs: fmt + debug clippy
(lib/bins, no `--all-targets`) + tests. It skips the release clippy pass, the
doc build, shellcheck, and the doc-ref check — all of which the full `./ci.sh`
(and CI) still gate on before push. Roughly a third of the full gate's time
after a one-file edit. CLAUDE.md now points the iteration loop at it while
keeping the full gate as the pre-push requirement.


dpmills:

- #34 ◀

sortalongo (approved):

PR #35, opened 2026-07-24
Fix false "value-dependent dependent function" panic on nested filters

dpmills:
Fix false "value-dependent dependent function" panic on nested filters

Filtering a filtered comprehension at nesting depth ≥3 —
`[a for a in [b for b in [c for c in L if p] if q] if r]` — panicked in lambda
elimination with "value-dependent dependent function unsupported: `__elem`
occurs in the cast value". It was a false positive: nothing is actually
value-dependent.

The refinement element binder `REFINEMENT_BINDER` (`__elem`) occurs *only* inside
refinement predicates, and each occurrence is bound by its enclosing refinement.
But `count_free_in_type_with_visited` counted those bound occurrences as free, so
`is_free(__elem, value)` returned true whenever `value` merely carried a type
slot with a nested refinement over `__elem` — which a nested filter's predicate
does. That tripped lambda-elim's guard against genuine value-dependent dependent
functions.

Fix: `count_free` of the element binder in a type is always 0 — it is never free
there, since a type binds nothing else and every `__elem` sits under a
refinement that binds it. Other names in a predicate remain free references to
the enclosing scope and are still counted.

Adds pipeline regression cases (depth-2 and the previously-panicking depth-3)
that now compile and evaluate correctly.

Known follow-up (separate, deeper): the *type* of a nested filtered
comprehension still grows exponentially — inference stamps the refined element
type onto ~4-6 slots of each enclosing comprehension, and `Type` is a `Box`-tree
(subtrees are copied, not shared), so it compounds ~4-6x per level (compile:
~10ms at depth 3, ~5s at depth 7). Realistic nesting (≤5) is sub-200ms. The real
fix is representational (share `Type` subtrees) and is out of scope here.

### DIFF
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -691,6 +691,17 @@ fn count_free_in_type_with_visited(
     ty: &Type,
     visited: &mut HashSet<PredicateId>,
 ) -> usize {
+    // The only variable a type can bind is the refinement element binder
+    // ([`crate::ccl::REFINEMENT_BINDER`]): it occurs *only* inside refinement
+    // predicates, and each such occurrence is bound by its enclosing refinement.
+    // So it is never free in a type — counting its (bound) occurrences would
+    // falsely report the binder as free, e.g. tripping lambda-elim's
+    // "value-dependent dependent function" guard when a predicate merely carries
+    // a nested refinement over `__elem`. Every *other* name in a predicate is a
+    // free reference to the enclosing lexical scope and is counted.
+    if name.is_elem() {
+        return 0;
+    }
     let mut count = 0;
     walk_refined_predicates(ty, visited, &mut |pred, vis| {
         count += count_free_with_visited(name, pred, vis);
--- a/src/ccl/ccl_utils.rs
+++ b/src/ccl/ccl_utils.rs
@@ -415,6 +415,12 @@ fn refine_with(base: Type, predicate: &Expr) -> Type {
 /// nodes treat their `name` field as a use of the defer-handle variable,
 /// so writes to that defer count as occurrences too.
 ///
+/// A [`Type::Refinement`] is a binding form too — it binds
+/// [`crate::ccl::REFINEMENT_BINDER`] (`__elem`) in its predicate — so occurrences
+/// of that one name inside a type are **bound, never free** (see
+/// [`is_free_in_type`]). Every other name in a predicate is a free reference to
+/// the enclosing lexical scope and is counted.
+///
 /// Used by:
 /// - [`is_free`] — the bool wrapper for "does `name` appear at all?"
 /// - [`crate::ccl::channelize`] — to detect when a defer's value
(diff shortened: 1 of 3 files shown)


dpmills:

- #35 ◀

groundlar (approved):

src/ccl/ccl_utils.rs
groundlar, line 686: Update for element binder

groundlar (approved):
opt: Consider modifying `lambda_elim.rs::is_free_detects_var_in_partial_tuple_refinement()` to check this property:

```
// Refinement(Int, pred = Var(__elem)) attached to a Lit: the element binder
// is bound by the refinement, so it is not free in the expression...
assert!(!is_free(&Name::elem(), &expr));
// ...while any other name in the same predicate still is.
assert!(is_free(&Name::raw("k"), &expr_with_pred_mentioning_k));
```

src/ccl/ccl_utils.rs
dpmills: Done in 4f198097. `is_free_in_type`'s doc now states the carve-out outright — it's the contract callers reason about (lambda-elim's value-dependence guard among them), so "always `false` for `REFINEMENT_BINDER`" belongs there rather than only in the implementation comment:

> **Always `false` for `crate::ccl::REFINEMENT_BINDER`** — a type's only binding form is the refinement, and every `__elem` occurrence sits under the refinement that binds it, so the element binder is never a free reference to anything a substitution could supply.

Also added the general statement one level up, on `count_free`, since that's where the shadowing rules are documented and `is_free`'s doc defers to it: a `Type::Refinement` *is* a binding form, it binds `__elem` in its predicate, so occurrences of that one name inside a type are bound rather than free — while every other name in a predicate stays a free reference to the enclosing scope.

dpmills:
Good call — taken, in 4f198097. The fix asserted something no test stated, which is exactly the gap.

`is_free_detects_var_in_partial_tuple_refinement` is now `is_free_in_type_sees_free_names_but_not_the_bound_element_binder`, with both halves on **one** fixture rather than two. The predicate is `__elem > x`, so a single refinement mentions the bound element binder *and* a free reference to the enclosing scope:

```rust
// pred = `__elem > x`: `__elem` is bound by the enclosing refinement,
// `x` is a free reference to whatever scope the type sits in.
let pred = Rc::new(Expr::binop(
    Expr::var(Name::elem()),
    BinOpKind::Compare(CompareKind::Greater),
    Expr::var("x"),
));
// Tuple([Int, {Int | __elem > x}]) on a Lit(42) — the node itself has no free
// vars, so both answers come entirely from `is_free_in_type`.
assert!(is_free(&Name::raw("x"), &expr));
assert!(!is_free(&Name::elem(), &expr));
```

One fixture rather than two on purpose: it makes the test discriminating in both directions. An implementation that answered uniformly — everything free (the pre-fix behaviour) or nothing free in a type — fails one half. I verified that by reverting the `is_elem()` early return: the new assertion is the one that fires, with the pre-existing `x` assertion still passing.

The `any`-vs-`all` tuple regression the test was originally written for is preserved — the refinement still rides only the second component — and the doc comment now names both properties it's guarding.

PR #36, opened 2026-07-26
Type a literal by which literal it is: the singleton refinement

dpmills:
Types a literal by **which** literal it is: `5 : {Int | __elem == 5}`, printed `5`.

This both unlocks reasoning about literals, and also exercises the type system more which keeps bugs from sneaking in

## Why

A literal knows more about itself than its base does, and that is exactly what a proof
obligation needs: `a[0]` can only discharge against `Array(3, T)`'s index range if `0`'s
type says it *is* `0`. Typing `5` as plain `Int` throws that away at the one place it is
free to keep. This is the prerequisite for total `[]` lookup.

## What it makes reachable

Refinements were rare enough before this that several rules assumed their absence. Every one
of the following was already wrong for a user-written refinement; literals are just the first
thing that makes them reachable.

- **A binary operator no longer propagates its operands' refinements.** Arithmetic's
  `∀α. α → α → α` shares *one* variable across both operands and the result, so a refinement
  reaching α claimed the operator preserved it. Invisible while operands merely join —
  distinct refinements intersect to none — and wrong when they do not, since intersecting a
  set with itself is that set: `x = 2; x + x` typed as `{Int | __elem == 2}`. The sum is 4.
- **A mutable register takes no refinement** from its initializer or any single write. A
  register is not one value but the sequence its writes produce, so its value type is the
  join over all of them. `x := 0` pinned it to `{Int | __elem == 0}`, rejecting later writes
  and failing the invariant `History` edge against a `Mut(Int)` parameter.
- **The annotation reconcile is one-way.** `bind_annotation`'s own comment called the reverse
  edge a "KNOWN OVER-RESTRICTION … currently unreachable", to be removed "when variant/range
  annotations become source-reachable". `x: Int = 1` makes it reachable. Conflicts now
  surface at coalesce rather than eagerly.
- **A propagated error in `check` is no longer discarded.** Most rules accumulate via
  `require_sub`, but `emit_case`'s `EmptyCase` and `emit_node`'s `UnboundVariable` can only
  return — so an empty `Case` reaching post-inference typecheck was **silently accepted**.
  Pinned by a test that fails against the old driver.
- **`__elem` is bound by the refinement it rides**, so it is never free *in a type*. The
  free-variable walk reported it free, making `lambda_elim` reject ordinary terms because
  some literal inside carried a predicate.
- **Beta reduction discharges a refined parameter** when the argument's type entails it,
  rather than refusing outright — substituting the argument is what establishes the
  precondition. The assert still fires on one the argument does not establish, and is now on
  a reachable path: a generalized list-returning UDF applied to a literal gets its domain
  specialized to that argument's singleton.
- **Every merge point joins** rather than adopting one input's type — the largest part of
  this change, below.
- Also: `+` finds `Concat` through a refined `String`; feed operands compare modulo refinements.

## Every merge point joins

A merge point — a list's elements, a `Case`'s arms, a register's seed and writes, a channel's
contributions — is not one value but whichever input the runtime supplies. A refinement is a
fact about *a value*, so a witness survives the merge only if **every** input establishes it.
Four places instead adopted one input's type, which reads as correct until the inputs carry
different refinements:

- **`emit_case`** derived the node's type from the first arm *stripped of refinements* and
  related the others to it. Stripping is not the join, and it is not variance-stable: a
  collection carries its extent as a refinement on its `Fun` **domain**, so relating a refined
  arm to a stripped sibling inverts at the contravariant edge into `[0, N] <: {[0, N] | p}`,
  rejecting two arms that are *literally the same expression*. (`D <: {D | p}` is never a real
  obligation in this language — acquiring a witness is an explicit `cast` — so an occurrence of
  one means an erasure manufactured it.) The arms now flow one-way into one shared variable,
  refinements included, exactly as a list's elements do.
- **`emit_transact`** let a register key's seed be the *only* lower bound on its value type, so
  `flag := False` resolved the register — and every read of it — to `{Bool | __elem == False}`,
  typing the writers' `True` as `False`. Now stripped: the register law the `MutWrite` rule
  already states.
- **Letrec recognition** (`planning/loops.rs`) rebuilt that same value type from the seed
  expression, and `transact_phase` read it back off the seed binding. Same law, same fix.
- **`channelize`'s channel union** took operand 0's codomain as the shared element type — its
  own comment noted this was sound only "under that invariant". With `c << 1` and `c << 2`
  contributing `{Int | __elem == 1}` and `{Int | __elem == 2}`, the channel typed as the `1`.
  It now keeps the witnesses every contribution establishes.

What the join answers, pinned by tests: arms depositing different singletons intersect to none
(`1 if c else 2` is an `Int`; `(1, 2) if c else (3, 4)` is `(Int, Int)`, and the same at depth);
a restriction *every* arm establishes is kept (identical filtered comprehensions stay filtered;
`5 if c else 5` is still the `5`); arms whose extents differ accumulate both witnesses on the
domain, which is where a function type's join puts them. Arms of incompatible base types are
rejected as an `IncompatibleBounds` collision at coalesce — where a heterogeneous list literal
or `CollectionUnion` reports it — rather than eagerly at the arm relation.

Collection-valued merges are where this is the difference between compiling and an internal
panic, since `compile_program` panics on a consistency-wall failure. Both of these type and
clear the wall:

```python
xs = [1, 2, 3]
c = 1 > 0
if c:
    [x for x in xs if x > 1]
else:
    [x for x in xs if x > 1]
```

```python
xs = [1, 2, 3]
[[x for x in xs if x > 1]]
```

### The consistency wall stays strict

`check`'s reconcile (a node's rule-reconstructed type vs. the type inference recorded on it)
compares refinements too. A rule that rebuilds a node's type from its children rebuilds its
refinements with it, so a recorded refinement the reconstruction lacks is a real disagreement
about the node — in practice a merge point that took one input's refinement — and this is the
check best placed to say so. Relaxing it to compare modulo refinements, by stripping both
operands or by a refinement-blind relation, hides exactly that class.

### The second-class `Mut` rule reads the argument, not the selection

A mutable read derefs into the arms' join exactly as it derefs into a tuple element, so a
`Case` over two registers types as their *value* — a read of whichever register, like `(x, y)`
is a pair of reads — and rule 1 has no `Mut` on the selection to reject. It does not need one:
what the rule protects is a selected register reaching a position that writes through it, which
is its *argument* clause, and `bump(x if c else y)` is rejected there. Keying the rejection on
the selection's type would also make it arm-order-dependent (`x if c else y` rejected,
`1 if c else x` not), which is not a property the rule is about.

### The ordering invariant counts witnesses for every read but one

`types_agree_modulo_unread` — the end-of-pass check that every type the coalesce walk read
still resolves to what was stamped, i.e. that *specialization only adds bounds to variables the
walk has not yet read* — compares refinement layers along with the base skeleton. A witness is
lattice content like a record field, so a bound determines it as much as it determines the base,
and one arriving on an already-read variable is exactly the staleness the check exists to catch.

`ReadPurpose` carves out the single read that cannot keep that property: the one in
`specialize_use` that produces a specialization's memo key, whose resolution is *not* what the
use node ends up carrying (that is the clone's own coalesced type). The pin that immediately
follows the read is what moves witnesses there, and the key's only consumer is
`Specialization::use_ty`'s equality, where a late-arriving witness can only cause a **miss** —
a fresh private clone, at worst a wasted one, never a mis-typed use. Its base skeleton is still
held fixed, since a stale skeleton could make the lookup *hit* the wrong clone. A unit test pins
both modes.

## Notes

- **Specialization is per distinct use type**, so `f(1)` and `f(2)` mint separate clones.
  Program size is not a concern today. Keying modulo refinements is the better rule — a
  refinement changes no code — but the clone is *pinned* to its use type, so a shared clone
  would carry one use's refinement and reject the others; it needs the clone built at the
  stripped type throughout. Recorded on the `specs` field, along with why the memo's
  candidate-vs-entry asymmetry is conservative in the safe direction.
- **Tests assert the real refined answers**, via `int_lit`/`str_lit`/`bool_lit` helpers
  wrapping the production `lit_singleton`, so they construct exactly what inference does.

Docs: `docs/chl-spec.md` §3.1 for the user-visible rule, `type-inference.md` for the
mechanism and the lists above.

## Compile cost

A refinement on every literal is not free, and it lands on a curve `main` already knows
about: nested *filtered* comprehension types grow exponentially because inference stamps the
refined element type onto several slots per level and `Type` is a `Box`-tree, so subtrees are
copied rather than shared. Measured on
`[a₀ for a₀ in [a₁ for a₁ in … [1, 2, 3, 4] … if a₁ < 3] if a₀ < 3]`, release build:

| depth | before | after | ratio |
|---|---|---|---|
| 3 | 11.7ms | 50.2ms | 4.3× |
| 4 | 51.4ms | 282ms | 5.5× |
| 5 | 248ms | 1.1s | 4.4× |
| 6 | 1.2s | 5.4s | 4.5× |
| 7 | 6.3s | 25.3s | 4.0× |

The **exponent is unchanged** — ~4.8× per level before, ~4.3× after — so this is a constant
~4–5× factor on an already-exponential curve, not a worse asymptotic. The remedy is the same
representational one that curve already needs (share `Type` subtrees instead of copying them),
which is out of scope here. Realistic nesting (≤5) stays comfortably sub-second.

Specialization keyed per distinct use type costs separately: `f(0)`, `f(1)`, … previously
shared one clone at `Int` and now mint one each. On a 32-call chain that is 20.8ms → 48.7ms
(~2.3×), the same shape of cost. Keying modulo refinements removes it but needs the clone
built at the stripped type throughout — see the `specs` field doc.

The merge-point rules above change constant work, not the shape of either curve. Re-measured
on the final tip, depth 5 is ~0.95s and depth 6 ~4.7s against the 1.1s / 5.4s recorded here,
so the table is if anything conservative.


dpmills:

- #36 ◀

src/ccl/infer/check.rs
```diff
@@ -270,7 +271,26 @@ impl Typing for CheckCtx {
...
+        // Reconstructing the pointful predicate here and requiring it to match would
+        // contradict this mode's own contract (see `type_annotation_predicates`):
+        // a rebuilt predicate can never equal a compiled one, though the two denote
+        // the same restriction. Every other node here trusts its recorded type or
+        // derives it from children; a literal is the one that could rebuild it, and
+        // must not.
+        TypedExprNode::Lit(lit) => {
+            // Recorded through `require_sub`, which *accumulates* into `ctx.errors`;
+            // `check` discards a propagated `Err` (see its `let _ =`), so returning
+            // one here would drop the diagnostic on the floor.
```
groundlar: nit: The `let _ =` was removed in this commit (check.rs:426), and `CheckCtx::require_sub (check.rs:96)` never returns `Err`, it pushes and returns `Ok`.

src/ccl/simplify.rs
```diff
@@ -217,12 +217,18 @@ fn take(expr: &mut Expr) -> Expr {
...
 /// rewriting, and a BinOp inside a lambda body that has been desugared to
 /// `Apply(Tuple, Builtin(BinOp(Add)))` is rewritten via its `Builtin` head.
+/// Whether `ty` is `String`, **ignoring refinements**. A refined string
+/// (`{String | __elem == "a"}`, the singleton a literal carries) is still a string,
+/// and which runtime operator `+` dispatches to is a question about the base.
+fn is_string(ty: &Type) -> bool {
+    crate::ccl::ccl_utils::strip_refinements(ty) == Type::Base(BaseType::String)
+}
+
 fn try_string_add_to_concat(expr: &mut Expr) -> bool {
```
groundlar: Looks to have b0rked the fn comment

src/ccl/inline.rs
```diff
@@ -565,6 +571,34 @@ fn is_name_in_function_position(expr: &Expr, name: &Name) -> bool {
...
+///
+/// Compared with the type-blind predicate relation
+/// ([`eq_refinement_predicate`](crate::ccl::eq_refinement_predicate), via `Refinement`'s
+/// `PartialEq`), because the two copies legitimately differ in inference metadata.
+/// This is deliberately a *syntactic* entailment, not a solver call: post-inference
+/// there is no constraint graph left, and the case that must succeed — an argument
+/// whose type carries the very refinement the parameter acquired *from* it — is an
+/// equality. Anything subtler is exactly what should trip the assert and get a real
+/// `restrict` lift.
+fn refinement_discharged_by(arg_ty: &Type, param_ty: &Type) -> bool {
```
groundlar, line 583: move above tests banner or below cfg?

groundlar (approved):
There's a regression introduced here, find tests and suggested fix in 

https://github.com/cambra-dev/cambra/pull/42

overall these are minor concerns, so address and merge at will

dpmills:
## Review response

Thanks — the regression was real, and chasing it turned up three more instances of the same mistake. The description now reflects the final state; here is what changed and why.

### The `Case`-arm regression, and the fix that isn't a strip

Tests from #42 are carried in as their own commit, with the deliberately-failing one now passing and its `BUG` comment replaced by the rule it described. The suggested fix — strip **both** sides of the arm relation — clears the panic but is not the relation the rule wants:

- It discards a restriction *no branch widens*. Two identical filtered comprehensions genuinely satisfy their filter, and the strip types the `Case` as the bare extent `[0, 2]`. For a collection the domain refinement *is* the extent, so widening it claims positions no branch produces.
- Stripping is not variance-stable in general, which is why the asymmetric version failed in the first place. `D <: {D | p}` is never a real obligation here — acquiring a witness is an explicit `cast` — so an occurrence of one means an erasure manufactured it.

So the arms **join** instead: each flows one-way into one shared variable, refinements included, exactly as a list's elements already do. The join is what decides which witnesses survive, so nothing has to strip. Arms whose extents differ accumulate both witnesses on the domain (a function type's join meets its domain) rather than one arm's filter being picked — the second defect noted in #42.

### Three older instances of the same mistake

Once `Case` joined, the consistency wall started reporting merge points elsewhere that adopt one input's refinement. None involve a `Case`; all reach the wall through `check_pre_desugar`:

| site | symptom |
|---|---|
| `emit_transact` register key | the seed is the value type's only lower bound, so `flag := False` types the writers' `True` as `False` |
| letrec recognition + `transact_phase` | rebuild that value type off the seed expression |
| `channelize`'s channel union | takes operand 0's codomain — `c << 1` / `c << 2` types the channel as the `1` |

A list literal *containing* a filtered comprehension (`[[x for x in xs if x > 1]]`) panicked on the wall too, with no `Case` involved — through the join variable the list-element commit introduced.

### Why the wall stayed strict

The reconcile could have been relaxed instead — stripping both operands, or a refinement-blind relation threaded through the solver (both were tried). Either one is green, and either one hides all four bugs above. Keeping it strict is what forced each merge point to join, so that is where it landed.

### The second-class `Mut` check on conditionals

With the join, a `Case` over two registers derefs to their value (as a tuple of reads already does), so rule 1 no longer has a `Mut` on the selection to reject and `x if True else y` lost its discipline diagnostic. Two things made this a non-issue rather than a hole:

- A logical `Case` in tail position is rejected by lambda elimination today regardless of `Mut`, on `main` too — so no working program changes behaviour.
- The hazard the rule protects is a selected register reaching a position that writes through it, which is rule 1's *argument* clause. `bump(x if c else y)` is still rejected, and the test moved there — a form that survives the `Case` support coming next.

An arm-level version of the check was written first and then dropped: it would have been scaffolding for a diagnostic on programs that cannot compile, and actively wrong once `Case` support lands, since a conditional over two registers is a *read*.

### The ordering invariant

`types_agree_modulo_unread` had stopped counting refinement layers, justified as "a join legitimately removes one". Instrumenting the check across the suite (log every drift instead of asserting) says that was too broad: **28,792 reads, 3 drifts, all three at one site** — `specialize_use`'s memo-key read — and in *both* directions, including a witness *appearing* (`pick = \lo, hi -> …` at `pick(8, 0)`), which the stated justification does not cover. Zero drifts at any read whose resolution is stamped on a node.

So `ReadPurpose` now splits the two: witnesses are compared for every `Stamp` read (restoring the pre-branch strength) and excluded only for the memo key, where a late witness can at worst cause a miss and mint a redundant clone. Two doc comments about the memo were wrong and are corrected — an entry's key is the clone's *coalesced* type, not the pre-pin snapshot the lookup compares against.

### Nits

All three were real:

- `is_string` had been inserted between `try_string_add_to_concat`'s doc comment and the function, so the doc attached to the wrong item and the rewrite rule lost its own.
- `refinement_discharged_by` — production code — sat below the `Tests` banner.
- The `Lit` arm's comment pointed at a `let _ =` this branch removed, and claimed `require_sub` can return `Err`.

### Open, not blocking

Why the witness becomes visible only across the pin, rather than through the emit-time `arg <: domain` edge, is not understood — i.e. whether resolution before the pin drops a witness already reachable in the graph. Nothing depends on it (the key tolerates a miss), but it is the one thing here that is bounded rather than explained.

src/ccl/infer/check.rs
dpmills: Right on both counts — fixed. The comment now says `require_sub` accumulates into `ctx.errors` and returns `Ok` (which is how a rule reports without short-circuiting the walk), and notes that `check` collecting a propagated `Err` is for the rules that have nothing to accumulate into, rather than pointing at the `let _ =` this branch removed.

src/ccl/simplify.rs
dpmills: It had — `is_string` went in between `try_string_add_to_concat`s doc comment and the function, so the doc attached to the helper and the rewrite rule lost its own. `is_string` now sits above the doc comment with its own.

src/ccl/inline.rs
dpmills: Above the banner — it is production code, not a test helper. Moved, with the blank line after the banner restored.

dpmills:
## The open question, closed: a negative position cannot see its lower bounds

Measured what the graph holds around the specialization pin, since the one carve-out in the ordering invariant was resting on a bound rather than an explanation.

**Nothing arrives late.** The argument's witness reaches the instantiation as a **lower** bound on the domain variable — `(8, 0) <: ?dom`, the emit-time `arg <: domain` edge — and a domain is a *negative* position, where coalescing intersects **upper** bounds. The witness is already in the graph when `specialize_use` reads there; it is just not on the side that read consults. The pin then adds the clone's parameter variable as an upper bound of `?dom`, having driven the same information into it, and *that* is the path that makes it visible:

| | `?dom` bounds | use type resolves to |
|---|---|---|
| pre-pin | `lower=[(8, 0)] upper=[?52]` | `((Int) ⇒ Int)` |
| post-pin | `lower=[(8, 0)] upper=[?52, (?89)]` | `((8) ⇒ Int)` |

So the drift is a property of *when* the read is taken relative to the pin, not of any bound going stale — which is exactly why every `Stamp` read is stable and only this one moves. The carve-out now says that instead of asserting a bound.

## What it turned up: the memo misses per call site, not per type

The `specs` doc said specialization is per distinct use type. It is weaker than that. Because the candidate key is resolved pre-pin (witness invisible) and the stored entry is the clone's coalesced type (witness present), a definition whose clone type acquires a witness only across the pin gets a fresh clone at **every call site** — two calls at the *same* literal miss each other:

```
pick(8, 0) + pick(8, 0)
  SPEC candidate_key=((Int) ⇒ Int) entries=[]                → miss, clone 1
  SPEC candidate_key=((Int) ⇒ Int) entries=[((8) ⇒ Int)]     → miss, clone 2
```

Whether it happens depends on how the body uses the parameter, not on arity:

| definition | call | candidate key | entry | memo |
|---|---|---|---|---|
| `\a, b -> a + b` | `h(1, 2)` twice | `((1, Int) ⇒ Int)` | same | **hit**, one clone |
| `\x -> x + 1` | `f(1)` twice | `(Int ⇒ Int)` | same | **hit**, one clone |
| `\x -> sum([v for v in xs if v >= x])` | `g(8)` twice | `(Int ⇒ Int)` | same | **hit**, one clone |
| `\lo, hi -> sum([v for v in xs if v >= lo])` | `pick(8, 0)` twice | `((Int) ⇒ Int)` | `((8) ⇒ Int)` | **miss**, clone per site |

Sound either way — each clone is private and correctly typed, and both programs evaluate correctly (`pick(8, 0) + pick(2, 0)` is 62) — but it is program size per call site, and it makes the ~2.3× specialization cost in the description a floor rather than a ceiling for such shapes.

Both candidate fixes were already noted on the `specs` field and neither is in scope here: key modulo refinements (needs the clone built at the stripped type throughout, or a shared clone carries one use's refinement and rejects the others), or resolve the key *after* the pin. The finding is recorded on `SpecializeFrame::specs` and `ReadPurpose::SpecializationKey` with the measurements, so whoever takes either fix starts from the mechanism rather than the symptom.

PR #37, opened 2026-07-27
BaseType: single-source keyword spelling via `keyword()`/`from_keyword()`

groundlar:
The surface spelling of each primitive type (`Int`/`UInt`/`String`/`Bool`/`Unit`) was hand-written at four independent sites — three writers (`Display for Type` in `src/ccl/ty.rs`, `Display for BaseType` in `src/interpreter/types/extent.rs`, `fmt_extent` in `src/pretty_graph.rs`) and one reader (`name_type` in `src/ccl/lower/stmts.rs`). A renamed or added `BaseType` variant could silently diverge across them.

This collapses the spelling into a single source of truth on `BaseType`:

- `BaseType::keyword(&self) -> &'static str` — the surface spelling (a `const fn`).
- `BaseType::from_keyword(&str) -> Option<Self>` — its total inverse (`from_keyword(b.keyword()) == Some(b)` for every variant), covered by a round-trip test.

All three writer sites now format through `keyword()`, and `name_type` recognises primitives through `from_keyword()`. Adding a `BaseType` variant is now a compile error at every site until it is handled, rather than a silent divergence.

Because the round-trip is total, type annotations now also accept `UInt` and `Unit` — previously `name_type` accepted only `Int`/`String`/`Bool`. The inference wildcard `_` stays handled separately in `name_type`; it is not a `BaseType`.

Not addressed here: `Type::Txn` and `Type::History` `Display` still hand-spell `Txn`/`Mut`/`feed` outside any shared helper. Consolidating the full type-surface vocabulary is a separate change, and a parser-level "reserved type keyword" concept was investigated and rejected as incoherent with CHL's capitalization-defined, open type namespace.

### DIFF
--- a/src/ccl/lower/stmts.rs
+++ b/src/ccl/lower/stmts.rs
@@ -881,8 +881,8 @@ fn is_mut_annotation(annotation: &Spanned<ChlExpr>) -> bool {
 ///
 /// Recognised forms:
 /// - Capitalized primitive names (`Caps` means type — `docs/chl-spec.md`):
-///   `Int` → [`Type::Base`]([`BaseType::Int`]), `String` → `String`,
-///   `Bool` → `Bool`.
+///   any [`BaseType`] spelling ([`BaseType::keyword`]) → [`Type::Base`], i.e.
+///   `Int`, `UInt`, `String`, `Bool`, `Unit`.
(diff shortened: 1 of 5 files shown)


groundlar:

- #37 ◀

dpmills (approved):

PR #39, opened 2026-07-27
tests: failing guards for five predicate-sharing review findings

groundlar:
Review response to the predicate-`Rc`-sharing work (`dmills/predicate-rc-sharing`, which this branch is based on). Five red tests, one per finding, plus one passing control.

**CI is expected red.** That is the deliverable: each test asserts a property the branch currently violates. `cargo test --test predicate_sharing_review` and `cargo test --lib review_tests`.

## The findings

Detail is in each test's doc comment; this is the index.

1. **`subst`'s memo crosses shadowing boundaries.** One `PredMemo` spans all four places the active substitution shrinks, so a rebuild from an outer scope is served to an occurrence whose scope rebinds the substituted variable — and, via `finish_unchanged`, a vacuous decision made under a restricted substitution skips a later occurrence under the full one. *(`rewrite_does_not_leak_an_outer_rebuild_into_a_shadowing_scope`, with `rewrite_respects_shadowing_when_the_occurrences_are_not_shared` as the passing control.)*
2. **Constraint emission's memo hit drops the occurrence's own domain.** `emit_bare_predicate` is parameterized by the domain the element binder is bound to; `emit_cast` mints a fresh one per cast node, so the second occurrence emits nothing and its domain is left unbounded. *(`each_occurrence_constrains_its_own_domain`.)*
3. **`subst::rewrite_expr_go` never rewrites binder-slot types.** A predicate on `param.ty` is left stale against the same predicate on `ty` — the defect `retype_pred_expr` had. *(`rewrite_reaches_the_lambda_param_type_slot`.)*
4. **`count_free`/`is_free` does not reach binder-slot predicates.** Three fast paths now skip work on `!is_free` — `inline_in_type_predicates`' early return, `subst`'s inert-subtree check, and `subst`'s vacuous-predicate check — so the coverage gap against `walk_type_slots` decides whether a rewrite happens. *(`is_free_sees_a_predicate_on_a_let_binding_slot`.)*
5. **The combinator's `changed` bit cannot see a nested memo hit's re-pointing.** `finish_unchanged` then discards the copy and memoizes the staleness. `simplify`'s fixpoint loop repairs the result on a later iteration, so this is latent — the fix is for the combinator to return the bit, since it is what knows. *(`walk_preserves_a_repointing_made_by_a_nested_memo_hit`.)*

## The pattern: one memo protocol, several different transforms

`PredMemo` is keyed on the predicate `Rc`'s address alone. That key determines the result only for a transform that is a function of the predicate term. The passes now threading it split in two:

- **Key-determined** — `uniquify`, `coalesce`, `retype`. The rebuild depends on the term plus context the branch argues is invariant across occurrences.
- **Context-determined** — the transform depends on something the key does not name: `subst` (the *active substitution*), constraint emission (the *domain*), planning's compilation (the refinement's *base*, which `fn_of_bare_predicate`/`bare_predicate_of_fn` both consume).

`coalesce_type_predicates` is the only one with a written precondition: resolution reads inference variables out of one live constraint graph, so two occurrences of one term resolve identically. That does not transfer to the second group, and for `subst` it inverts — substitution's job is to act differently in different scopes, so "same term ⇒ same result" is what shadowing denies. Findings 1 and 2 are where I think this is live rather than merely undocumented; planning's base-keyed compilation is the same shape and wants at least the same paragraph, or a `debug_assert!` that the bases agree on the hit path.

Formerly unreachable code paths are exposed by `Rc` sharing; the `rewrite_respects_shadowing_when_the_occurrences_are_not_shared` control illustrates this with shared/unshared versions of the same tree.

For `subst`: not letting the memo cross a shrink (fresh memo when `restricted != *self`) keeps the win and closes both directions.

`Expr::walk_type_slots{,_mut}` unified the type-slot walks; the two remaining hand-rolled copies are exactly findings 3 and 4. Converting `count_free_with_visited` is the highest-value follow-up.

### DIFF
--- a/src/ccl/infer/emit.rs
+++ b/src/ccl/infer/emit.rs
@@ -1282,3 +1282,52 @@ pub(super) fn emit_transact<C: Typing>(
     }
     Ok(Type::Record(fields))
 }
+
+#[cfg(test)]
+mod review_tests {
+    use super::*;
+    use crate::ccl::{BinOpKind, CompareKind, Lit, TypedExpr};
+    use std::rc::Rc;
+
+    /// Finding 2: [`emit_bare_predicate`]'s transform is parameterized by
+    /// `domain`, which is not part of the memo key. `emit_cast` passes a *fresh*
+    /// inference variable per cast node, so two occurrences of one shared
+    /// predicate `Rc` are typed against two different domains — and the memo
+    /// makes the second one emit nothing at all, leaving its domain with none of
+    /// the constraints the predicate imposes on `REFINEMENT_BINDER`.
+    #[test]
+    fn each_occurrence_constrains_its_own_domain() {
+        // `__elem > 0` — using the element binder is what makes the predicate
+        // constrain the domain it is typed against.
+        let shared = Rc::new(TypedExpr::binop(
+            TypedExpr::var(Name::elem()),
+            BinOpKind::Compare(CompareKind::Greater),
+            TypedExpr::lit(Lit::Int(0)),
+        ));
+        let mut r1 = Refinement::sharing(&shared);
+        let mut r2 = Refinement::sharing(&shared);
+
+        let mut ctx = InferCtx::new(std::collections::HashMap::new());
+        let d1 = ctx.fresh();
+        let d2 = ctx.fresh();
+        emit_bare_predicate(&mut r1, &d1, &mut ctx).expect("first occurrence types");
+        emit_bare_predicate(&mut r2, &d2, &mut ctx).expect("second occurrence types");
+
+        let bound_count = |t: &Type| {
+            let Type::Infer(v) = t else {
+                panic!("fresh() yields a variable");
+            };
+            let b = v.bounds.borrow();
+            b.lower.len() + b.upper.len()
+        };
+        assert!(
+            bound_count(&d1) > 0,
+            "sanity: typing `__elem > 0` at `d1` constrains `d1`",
+        );
+        assert!(
+            bound_count(&d2) > 0,
+            "the second occurrence's domain must be constrained by the predicate \
+             too — the memo hit skipped emission and left `d2` unbounded",
+        );
+    }
+}
--- a/tests/predicate_sharing_review.rs
+++ b/tests/predicate_sharing_review.rs
@@ -1,0 +1,223 @@
+//! Failing guards for the review findings on the predicate-sharing work.
+//!
+//! Each test states one property the current code violates. They are written
+//! against the smallest construction that exhibits the defect, not against a
+//! source-level fixture, so a fix can be verified without reasoning about which
+//! program shape happens to produce the sharing.
+
+use cambra::ccl::ccl_utils::{
+    PredMemo, distinct_predicate_rcs, is_free, walk_refined_predicates_mut,
+};
+use cambra::ccl::subst::Subst;
+use cambra::ccl::{
+    BinOpKind, CompareKind, Expr, Lit, Name, Refinement, Type, TypedExpr, TypedExprNode,
+};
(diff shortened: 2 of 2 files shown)


dpmills:
All five fixed in e388a81e, with your tests cherry-picked onto #32 (authorship preserved) so they land as the guards. The control still passes and none of the five needed editing.

**Your diagnosis of the root cause is the fix.** "The key determines the result only for a transform that is a function of the predicate term" is the thing the branch was missing, and splitting the passes into key-determined and context-determined is now part of `PredMemo`'s API rather than a property each call site is trusted to have:

- **Key-determined** — `begin`/`finish`, a hit skips the transform.
- **Context-determined** — `begin_always`/`finish_shared`, the transform runs at every occurrence and only the resulting *term* is unified.

Your third category turned out to want a third answer, which I think is the most interesting part. For `subst` the context isn't a *parameter* of the transform, it's part of the transform's *identity* — as you put it, "same term ⇒ same result" is exactly what shadowing denies. So neither protocol applies: the fix is to scope the memo rather than widen the key. `Subst::recurse_shrunk` hands the inner scope a fresh memo whenever `shrunk != *self`, at all five places the substitution shrinks (including the `Name::elem()` shadow, which never actually shrinks — but going through the same helper means the invariant holds by construction rather than by that argument). Each memo stays key-determined internally, and with no shadowing `shrunk == *self` so one memo still spans the whole walk. Your suggested shape, essentially verbatim.

**Finding 2 was mine and you're right about it.** Emission binds `REFINEMENT_BINDER` to a domain supplied per occurrence and `emit_cast` mints a fresh one per cast node, so the second occurrence was left under-determined rather than visibly wrong — the worst failure mode. It now emits at every occurrence and unifies only the term. Worth noting for the record: **sharing did not regress** — the motivating comprehension still leaves inference with 3 distinct `Rc`s and no structurally-equal duplicates — because the split memoization was preventing was in the *term*, not in the emission. Discarding the loser's term is sound on exactly the ground you identified: refinement identity is type-blind, and each occurrence has already discharged its own obligation.

**Findings 3 and 4** were the last two hand-rolled type-slot walks, as you predicted. Both now go through `walk_type_slots{,_mut}`. Agreed that 4 was the higher-value one — three fast paths gate on `is_free`, so the coverage gap decided whether a rewrite happened at all. `visited` keeps the dedup, so a predicate riding both a binder slot and the matching position of `expr.ty` is still counted once.

**Finding 5**: took your framing — the combinator is what knows. Concretely `PredMemo::revision()` advances whenever a slot is pointed at a different `Rc`, and `walk_refined_predicates_mut` keeps the copy when the callback reports a change *or* the revision moved during the call. It also returns its own `changed` bit now, which `simplify` folds into its fixpoint rather than using only the callback's — so the latency you noted is gone too, not just the staleness.

**Planning's compilation.** I went with the `debug_assert!` rather than the paragraph, and it kept the skipping protocol, because recompiling per occurrence is the superlinearity this branch exists to remove. The reason it can: `base` reaches only the compiled term's *type* slots, and refinement identity is type-blind, so two occurrences compile to terms that compare equal — provided the bases agree, which held by argument and nothing else. `predicates::CompileMemo` now records each entry's base in debug builds and checks it on the hit path; release builds are exactly a `PredMemo`. It does not fire anywhere in the suite, which is the evidence the precondition holds rather than an assumption that it does.

One thing I did *not* do: your closing suggestion that converting `count_free_with_visited` is the highest-value follow-up — I did it here rather than deferring it, since finding 4 made it load-bearing rather than tidy.

`./ci.sh` clean (both clippy passes, doc, tests) and green with `--features deep-typecheck`.

Shall I close this as superseded, or would you rather verify against it first?

PR #40, opened 2026-07-28
Binder scoping: one exhaustive scoped-children walk, four folds over it

dpmills:
Five independent walkers each encoded "which children does this node's binder scope over": `count_free_with_visited` and `count_free_in_value` (`ccl_utils.rs`), `collect_expr_fv` and `Subst::rewrite_expr_go` (`subst.rs`), and `Uniquifier::expr` (`uniquify.rs`). They agreed, but every one of them ended in `_ => walk_children(…)`, so adding a binding form to `TypedExprNode` compiled clean in all five and silently got the wrong scope in each. `LetRec`, `For`, and `MutWrite` are all recent additions, so this is a path the codebase demonstrably walks.

New `src/ccl/scope.rs` states the rules once. `for_each_scoped_item` visits each direct child paired with the binders in scope for it, plus the name *occurrences* the node makes itself, and its `match` is **exhaustive with no wildcard arm** — a new variant must declare its scope before the crate compiles.

Four of the five are now folds over it:

- `count_free_with_visited` / `count_free_in_value` — shadowed if the scope shadows the name, else recurse.
- `collect_expr_fv` — the same fold, accumulating instead of counting.
- `Subst::rewrite_expr_go` — folds over `for_each_scoped_item_mut`, an adapter *derived* from the same walk (a `&mut` traversal cannot hold borrows into the node it mutates, so binder names are cloned and children are enumerated by `walk_children_mut`; a node that declares no binder skips the collection pass entirely, so only binder-introducing nodes pay for it). It has no arms left of its own: with the handle-target retarget folded in as well, the descent decides nothing about binding structure.

`TypedExpr::walk_binders` — the enumeration of binding *slots*, which `uniquify`'s post-pass mint assertion runs on — loses its `_ => {}` arm for the same reason the scoped walk never had one. A wildcard there would let a newly-added binding form be invisible to the assertion meant to catch a forgotten mint: the identical failure mode, one layer down. Adding a dummy binding variant now fails to compile in five places — `walk_children`, `walk_children_mut`, `walk_binders`/`walk_binders_mut`, `for_each_scoped_item`, and the `&mut` walk's name half — which together are the whole guarantee.

### A scope is a type

`Binders` is the scope a node opens over a child: `Ambient`, `One(&binding)`, or `Group(&bindings)`. It borrows the node's binder slots **in place**, so the walk allocates nothing — which matters because it sits under `is_free`, which capture-avoiding substitution consults once per mapped binder per node, so collecting a `LetRec` group into a `Vec` would be an allocation per group *per query*. `tests/scope_walk_allocations.rs` pins that with a counting allocator.

It also owns `Binders::shadows` — "does this scope shadow this name", the question every consumer actually asks, and the one that would otherwise be re-spelled as `binders.iter().any(|b| &b.name == name)` once per fold.

The `&mut` walk yields **scopes**, not per-child binder lists: `ScopedItemMut::Scope` comes once ahead of the run of children a scope covers, which is well-defined exactly because of the consecutiveness invariant this module documents and tests. So a consumer builds per-scope state once per scope, and the binder-name clones the `&mut` side cannot avoid are one per scope rather than one per child.

Substitution needs both halves of that. `Subst::shadow_all` returns a `Cow` and so borrows outright when the scope names nothing the substitution acts on — the common case, since a substitution's domain is usually a single binder; folding `shadow` over a group cloned the map once per binder even on a miss. And `assert_no_capture` is a release-active `assert!` whose `range_mentions` walks every `Discharge` replacement term, so a `LetRec` group of `n` binders owes `n` checks — one per binder of the one scope it opens — where a per-child callback performs `n * (n + 1)`. Release, 40-deep letrec spine, 32-node discharge term, per rewrite:

| group width | per child | per scope |
| --- | --- | --- |
| 1 | 355µs | 302µs |
| 8 | 4.62ms | 2.18ms |
| 32 | 44.5ms | 11.8ms |
| 64 | 152ms | 26.6ms |

`capture_assert_probe` counts the calls, so "once per scope crossed" is a test and not a comment.

`rewrite_expr_go` no longer pre-tests inertness before descending under a binder. The recursive call opens with the same test against the same substitution, so the pre-test walked every child subtree twice; and returning early on it skipped `assert_no_capture`, which is the tripwire for a broken uid-preservation invariant and belongs at every binder crossing, not only the ones where the substitution happens to still be live. That deliberately widens where the assertion fires, and it stays an `assert!` rather than a `debug_assert!`: what it catches is a uid minted outside lowering or a copy that broke uid preservation, whose consequence is silently-wrong output. Per-scope checking is what makes keeping it release-active cheap.

### Name occurrences

Two name-occurrence kinds are distinguished rather than left implicit. `ScopedItem::VarRef` is a variable use — a `Var`, or a `Feed`/`Define`/`MutWrite` write target. `ScopedItem::KeyRef` is a `Transact` register key: a label of the record the node denotes, not a variable, so free-*variable* analyses skip it. That was already the de-facto split; naming it is what lets one walk serve consumers that disagree about it. It is not hypothetical — `planning::run` calls `simplify`, which substitutes, *after* `plan_loops` has produced `Transact`, so post-planning trees do reach `Subst` and `is_free`.

The `&mut` walk surfaces them too, from its own wildcard-free match, and that is what retires substitution's last node-local `_ => {}`: the set of names a node mentions of its own was declared in `scope.rs` and then re-derived behind a wildcard in `subst.rs`, which is the shape this change exists to remove, one field over. `mut_walk_surfaces_the_same_names` checks the two walks agree over the whole corpus, and `Subst::retarget` splits "unchanged" out of `handle_target` so visiting every name occurrence in the tree costs no `Name` clone.

### What stays outside

`uniquify` does not fold over the walk, and its module docs say why: it does not observe binders, it *mints* them — it needs `&mut` access to the binding slot, must mint a whole `LetRec` group before descending into any definition, and unwinds its own environment stack. What it does now share is the binder enumeration: `assert_all_binders_minted` checks its output through the exhaustive `TypedExpr::walk_binders`, whose agreement with the scoping rules is a test in `scope.rs`, so a binding form the pass forgets to mint fails the post-pass assertion instead of passing silently.

Substitution's transport mode, `Subst::apply_expr_inner`, rebuilds nodes rather than walking them: folding it in would mean cloning each subtree at every binding node, turning substitution down a `let` spine quadratic in its depth.

Both walks are generic in their callback rather than taking `&mut dyn FnMut`. `for_each_scoped_item` sits under `is_free`, so the fold body wants to inline; neither walk is itself recursive — the recursion lives in the consumers, one closure type each — so monomorphization stays flat.

### Tests

`scope.rs` carries a corpus with one instance of every `TypedExprNode` variant, and its completeness is *checked*, not prompted: a macro declares `variant_name`'s arms and derives `VARIANT_COUNT` from that same list, so a new variant makes the match non-exhaustive, adding the arm bumps the count, and the bumped count fails `corpus_covers_every_variant` until the corpus gains an instance. Two instances of one variant fail it too — that is what would otherwise mask a missing one.

Over that corpus: the scoped walk's child sequence is pointer-identical to `walk_children`'s, the binders it declares match `walk_binders`, a binder-introducing scope's children are consecutive and the scope is never re-entered once left (the ambient scope recurs freely — a `Case` returns to it for every payload-free branch), the `&mut` walk pairs the same binder names and opens one `Scope` per run, both walks surface the same name occurrences, and the seven rules come out as written.

`ir.md` gains "Binding structure lives in one place" — the rules, the three matches that are exhaustive and why, the three invariants a consumer may rely on, and the two passes that stay outside the fold.

### DIFF
--- a/src/ccl/design/ir.md
+++ b/src/ccl/design/ir.md
@@ -51,7 +51,7 @@ Renaming happens at the name level only — the tree shape is untouched, so this
 
 ### Binding structure lives in one place
 
-Which children a node's binders scope over is stated once, in `ccl/scope.rs`'s `for_each_scoped_item`: it visits each direct child paired with the binders in scope for it, plus the name *occurrences* the node makes itself. Every scope-aware walk in the crate is a fold over it — free-occurrence counting (`ccl_utils`'s `count_free`, `count_free_in_value`), free-variable collection (`subst`'s `collect_expr_fv`), and capture-avoiding substitution (`Subst::rewrite_expr`, via the `for_each_scoped_child_mut` adapter).
+Which children a node's binders scope over is stated once, in `ccl/scope.rs`'s `for_each_scoped_item`: it visits each direct child paired with the binders in scope for it, plus the name *occurrences* the node makes itself. Every scope-aware walk in the crate is a fold over it — free-occurrence counting (`ccl_utils`'s `count_free`, `count_free_in_value`), free-variable collection (`subst`'s `collect_expr_fv`), and capture-avoiding substitution (`Subst::rewrite_expr`, via the `for_each_scoped_item_mut` adapter).
 
 The rules themselves:
 
@@ -63,11 +63,17 @@ The rules themselves:
 - `Feed` / `Define` / `MutWrite` — the `name` field is a *use* of the binder it names, not a binder.
 - `Transact` — introduces no binder; its keys are labels of the register record the node denotes, so they are surfaced as a distinct occurrence kind that free-*variable* analyses skip while an identity-sensitive consumer still folds them in.
 
-The walk's `match` is exhaustive with **no wildcard arm**, deliberately: before this existed, the walkers ended in `_ => walk_children(..)`, so a new binding form compiled clean in all of them and silently got the wrong scope in every one. Now it is a compile error until the new form declares its scope. `TypedExpr::walk_binders` — the enumeration of binding *slots*, which the scoped walk is checked against and which `uniquify`'s post-pass mint assertion runs on — is exhaustive for the same reason: a wildcard there would let a new binding form be invisible to the assertion that is supposed to catch a forgotten mint, which is the same failure mode one layer down.
+Every `match` that decides one of these rules is exhaustive with **no wildcard arm**, deliberately: before this existed, the walkers ended in `_ => walk_children(..)`, so a new binding form compiled clean in all of them and silently got the wrong scope in every one. Now it is a compile error until the new form declares its scope. That covers three matches, each closing the same failure mode at a different layer:
+
+- `for_each_scoped_item` — which children each binder scopes over.
+- `for_each_scoped_item_mut`'s name half — which names a node mentions of its own, so that a rename retargets a new node's name reference instead of walking past it. Its agreement with the immutable walk is a corpus test.
+- `TypedExpr::walk_binders` — the enumeration of binding *slots*, which the scoped walk is checked against and which `uniquify`'s post-pass mint assertion runs on. A wildcard here would let a new binding form be invisible to the assertion that is supposed to catch a forgotten mint.
 
 Two passes stay outside the fold and say why in their own docs: `ccl/uniquify.rs` *mints* binders rather than observing them, and substitution's transport mode (`Subst::apply_expr_inner`) rebuilds nodes rather than walking them.
 
-Consumers may rely on three properties of the walk, each asserted in `scope.rs`'s tests: the `Child` items come in `walk_children` order (so the `&mut` adapter can pair binder lists positionally), the binders it declares are exactly `walk_binders`', and a scope's children are consecutive with a binder-introducing scope entered only once. The last is what lets a consumer build per-scope state once per scope rather than once per child — `Subst::shadow_all` restricting a substitution across a `LetRec` group, for instance, which borrows instead of cloning whenever the group names nothing the substitution acts on.
+A scope is a type, `Binders`, not a collected list of binder references: it borrows the node's binder slots in place, so the walk allocates nothing even for a `LetRec` group — which matters because it sits under `is_free`, and substitution consults that once per mapped binder per node. `Binders` also owns `shadows`, the question every consumer actually asks, so "does this scope shadow this name" is answered once rather than at each fold.
+
+Consumers may rely on three properties of the walk, each asserted in `scope.rs`'s tests: the `Child` items come in `walk_children` order (so the `&mut` walk can pair scopes with children positionally), the binders it declares are exactly `walk_binders`', and a scope's children are consecutive with a binder-introducing scope entered only once. The last is what lets a consumer build per-scope state once per scope rather than once per child, and the `&mut` walk hands it the change point directly, as a `Scope` item ahead of the run of children it covers. Substitution needs both halves of that: restricting the substitution across the scope (`Subst::shadow_all`, which borrows instead of cloning whenever the scope names nothing the substitution acts on) and the Barendregt no-capture check, which is release-active and walks every replacement term — so a `LetRec` group of *n* binders owes *n* checks, and paying per child instead would cost *n*(*n*+1).
 
 ### Application shape
 
--- a/src/ccl/design/ir.md
+++ b/src/ccl/design/ir.md
@@ -49,6 +49,26 @@ After uniquification, shadowing ceases to exist for every downstream pass: plain
 
 Renaming happens at the name level only — the tree shape is untouched, so this does not over-normalize the way strict ANF or SSA conversion would.
 
+### Binding structure lives in one place
+
+Which children a node's binders scope over is stated once, in `ccl/scope.rs`'s `for_each_scoped_item`: it visits each direct child paired with the binders in scope for it, plus the name *occurrences* the node makes itself. Every scope-aware walk in the crate is a fold over it — free-occurrence counting (`ccl_utils`'s `count_free`, `count_free_in_value`), free-variable collection (`subst`'s `collect_expr_fv`), and capture-avoiding substitution (`Subst::rewrite_expr`, via the `for_each_scoped_child_mut` adapter).
+
+The rules themselves:
+
+- `Lambda` — `param` scopes over `body`.
+- `Let` — `binding` scopes over `body` only; `bound_expr` is outside it (CCL's `let` is non-recursive).
+- `LetRec` — every group binder scopes over every binding's definition and over `body`.
+- `For` — `target` scopes over `body`; `iter` is outside.
+- `Case` — each branch's `pattern.binding` scopes over that branch's `guard` and `body`, and nothing else.
+- `Feed` / `Define` / `MutWrite` — the `name` field is a *use* of the binder it names, not a binder.
+- `Transact` — introduces no binder; its keys are labels of the register record the node denotes, so they are surfaced as a distinct occurrence kind that free-*variable* analyses skip while an identity-sensitive consumer still folds them in.
+
+The walk's `match` is exhaustive with **no wildcard arm**, deliberately: before this existed, the walkers ended in `_ => walk_children(..)`, so a new binding form compiled clean in all of them and silently got the wrong scope in every one. Now it is a compile error until the new form declares its scope. `TypedExpr::walk_binders` — the enumeration of binding *slots*, which the scoped walk is checked against and which `uniquify`'s post-pass mint assertion runs on — is exhaustive for the same reason: a wildcard there would let a new binding form be invisible to the assertion that is supposed to catch a forgotten mint, which is the same failure mode one layer down.
+
+Two passes stay outside the fold and say why in their own docs: `ccl/uniquify.rs` *mints* binders rather than observing them, and substitution's transport mode (`Subst::apply_expr_inner`) rebuilds nodes rather than walking them.
+
+Consumers may rely on three properties of the walk, each asserted in `scope.rs`'s tests: the `Child` items come in `walk_children` order (so the `&mut` adapter can pair binder lists positionally), the binders it declares are exactly `walk_binders`', and a scope's children are consecutive with a binder-introducing scope entered only once. The last is what lets a consumer build per-scope state once per scope rather than once per child — `Subst::shadow_all` restricting a substitution across a `LetRec` group, for instance, which borrows instead of cloning whenever the group names nothing the substitution acts on.
+
 ### Application shape
 
 Function application is a single `Apply(Box<Expr>, Box<Expr>)` node. Single-argument CHL calls `f(a)` lower to `Apply(a, Var(f))` directly. Multi-argument CHL calls `f(a, b, ...)` lower to `Apply(Tuple([a, b, ...]), Var(f))` — the arguments are tupled so the call shape matches how multi-arg CHL lambdas are uncurried at lowering time (see [lowering.md](lowering.md)).
--- a/src/ccl/scope.rs
+++ b/src/ccl/scope.rs
@@ -2,11 +2,12 @@
 //!
 //! Every scope-aware pass — free-variable counting, capture-avoiding
 //! substitution, α-uniquification — needs the same answer to one question:
-//! *which binders scope over which of a node's children?* [`for_each_scoped_item`] is the single place that answers it. Its
-//! `match` is **exhaustive with no wildcard arm**, deliberately: a new
-//! [`TypedExprNode`] variant must declare its scope here before the crate
-//! compiles, instead of falling into a `_ => walk_children(..)` catch-all in
-//! five separate passes and silently getting the wrong one.
+//! *which binders scope over which of a node's children?* [`for_each_scoped_item`]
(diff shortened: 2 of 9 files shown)


dpmills:

- #40 ◀
    - #41

src/ccl/design/ir.md
```diff
@@ -49,6 +49,26 @@ After uniquification, shadowing ceases to exist for every downstream pass: plain
...
+- `For` — `target` scopes over `body`; `iter` is outside.
+- `Case` — each branch's `pattern.binding` scopes over that branch's `guard` and `body`, and nothing else.
+- `Feed` / `Define` / `MutWrite` — the `name` field is a *use* of the binder it names, not a binder.
+- `Transact` — introduces no binder; its keys are labels of the register record the node denotes, so they are surfaced as a distinct occurrence kind that free-*variable* analyses skip while an identity-sensitive consumer still folds them in.
+
+The walk's `match` is exhaustive with **no wildcard arm**, deliberately: before this existed, the walkers ended in `_ => walk_children(..)`, so a new binding form compiled clean in all of them and silently got the wrong scope in every one. Now it is a compile error until the new form declares its scope. `TypedExpr::walk_binders` — the enumeration of binding *slots*, which the scoped walk is checked against and which `uniquify`'s post-pass mint assertion runs on — is exhaustive for the same reason: a wildcard there would let a new binding form be invisible to the assertion that is supposed to catch a forgotten mint, which is the same failure mode one layer down.
```
groundlar: > exhaustive with **no wildcard arm**

🚀

groundlar:
The architecture is right and I'd like to see this land: replacing five `_ => walk_children(..)` fallthroughs with one wildcard-free walk turns a demonstrated hazard into a compile error, and the corpus test suite is unusually well aimed — pointer-identity against `walk_children`, agreement with `walk_binders`, and the consecutiveness property that the `&mut` adapter's positional pairing actually depends on. The `VarRef`/`KeyRef` split earns its keep, and the justification for it (`planning::run` calls `simplify` after `plan_loops`, so post-`Transact` trees really do reach `Subst`) is a live code path rather than a hypothetical. `uniquify` sharing only the binder *enumeration* is the right amount of sharing and the module docs say why.

Three things to fix first. I've pushed them as tests in #61 (draft, based on this branch) — three fail on purpose, two pass and are worth keeping. Specifics and measurements are there; summary below.

### 1. The Barendregt check moved from once-per-scope to once-per-child

`assert_no_capture` is a release-active `assert!`, and its `range_mentions` does a full `is_free` walk of every `Discharge` replacement term. The new descent invokes it inside the `for_each_scoped_child_mut` callback, so a node whose binders scope over several children pays the check once per child rather than once per scope. A `LetRec` group of `n` binders opens one scope covering `n + 1` children — it owes `n` checks and performs `n * (n + 1)`:

| group size | owed | performed |
| --- | --- | --- |
| 1 | 1 | 2 |
| 4 | 4 | 20 |
| 8 | 8 | 72 |

`Case` is the same shape at 2x, guard and body being two children of one scope. Both are regressions against the arms this replaces, which asserted once per group binder and once per branch — so the change trades a per-child map clone in `shadow_all` for per-child work that costs considerably more than the clone did.

The consecutiveness invariant this PR documents and tests is exactly what makes per-scope expressible: track the previous binder list, assert only when it changes. Worth doing on top: `range_mentions` depends only on the name, not on the subtree, so the range's free-variable set can be computed once per `rewrite_expr` (`collect_expr_fv` already exists) and each crossing becomes a set lookup — `O(crossings * |range|)` becomes `O(|range| + crossings)`.

The "release timing unchanged within noise" measurement is believable for a suite without wide `LetRec` groups or large discharge ranges; it just isn't robust to one. Worth re-measuring on a wide group.

Separately and deliberately: dropping `under_binder_mut`'s pre-inertness return means a crossing is now checked even when the substitution is dead in the body. I agree with the reasoning, and #61 pins the behavior green. But note the panic surface widened, and this is `assert!` rather than `debug_assert!` — a uid-preservation bug that previously produced quietly-wrong output now aborts release builds in strictly more places. If that's intended, the commit message is the place to say so; if it's a tripwire, `debug_assert!` is the cheaper form.

### 2. The walk allocates once per `LetRec` visit, on the `is_free` hot path

`for_each_scoped_item`'s `LetRec` arm collects the group into a `Vec<&TypedBinding>` so each child can be handed the same binder slice. But the walk sits under `is_free`, which substitution consults once per mapped binder per node, so that allocation is paid per `LetRec` node *per query* — where the per-pass arms it replaces tested shadowing with `bindings.iter().any(..)` and allocated nothing. Measured in #61 with a counting allocator: a 10-deep `letrec` spine, one `is_free_in_value` query, 10 allocations. `for_each_scoped_child_mut` compounds it with `n + 1` cloned `Vec<Name>`s per group.

Both dissolve if `binders` becomes a type rather than a slice:

```rust
pub enum Binders<'a> {
    None,
    One(&'a TypedBinding),
    Group(&'a [(TypedBinding, TypedExpr)]),  // borrowed in place
}
```

with `shadows(&Name) -> bool` and `names()`. Allocation-free for every variant, and it collapses the duplication in (3) — which is the "recurring shape already being duplicated" justification, not an aesthetic one.

### 3. `binders.iter().any(|b| &b.name == name)` is already duplicated four ways

Character-identical in `count_free_with_visited` and `count_free_in_value`; `binders.iter().map(|b| b.name.clone())` in `collect_expr_fv` and in `for_each_scoped_child_mut`. "Does this scope shadow this name" is *the* question this module exists to answer, and it is currently answered at the call sites. Either collapse it into the `Binders` type above or note it in the PR body so the third duplication isn't the one that gets noticed.

### Smaller

- **The commit message is stale post-rebase.** It says the remaining arms in `rewrite_expr_go` are "a `Cast`'s target type, a handle node's write target" — the `Cast` arm is gone, since the base's `walk_type_slots_mut` reaches it generically. This body squash-merges into `main`, so worth the one-word fix.
- **The remaining `_ => {}` in `rewrite_expr_go`'s node-local match** is narrower than it was (type slots are generic now) but still hides the handle-target arm: the walk *names* every self-made variable occurrence as a `ScopedItem::VarRef`, then the `&mut` adapter drops them, so substitution re-derives the set behind a wildcard. #61's `a_rename_retargets_every_varref_the_scoped_walk_surfaces` pins the current contract and passes; surfacing `&mut Name` from the adapter is the structural fix.
- **`shadow_all`'s `contains_key` guard is load-bearing and unremarked** — it is what keeps the `Cow` borrowed, and without it `to_mut()` clones unconditionally and the `Cow` buys nothing. The doc explains why `Cow` but not why the guard; one clause protects it from a future simplification.
- **`scopes.resize(children - 1, ..)`** is safe only because scoped children arrive in increasing index order — the adjacent comment explains sparseness, not monotonicity. A `debug_assert!(scopes.len() < children)` names the invariant that's actually holding it up.
- **Corpus completeness is prompted, not enforced**, as the doc comment honestly says. Asserting that `variant_name` over the corpus has no duplicates and matches a `VARIANT_COUNT` const would close it cheaply.
- **`binds_any()` doesn't short-circuit** — `walk_binders` with a bool, so a `LetRec` sets it `n` times. Trivial, but `any_child`-style short-circuiting already exists in that file.

groundlar (approved):

src/ccl/scope.rs
```diff
@@ -0,0 +1,660 @@
...
+/// binder *names* are cloned up front and the children are then enumerated by
+/// [`TypedExpr::walk_children_mut`]; only the binder-introducing nodes allocate.
+/// The two enumerations agreeing is the invariant this pairing rests on — it is
+/// documented on [`for_each_scoped_item`], checked pointer-for-pointer by this
+/// module's corpus test, and its arity half is asserted here.
+///
+/// Name *occurrences* (`ScopedItem::VarRef` / `KeyRef`) are not surfaced: a
+/// caller that rewrites them (only [`crate::ccl::subst`] does, to retarget a
+/// renamed defer handle) reaches them through its own node match.
+pub fn for_each_scoped_child_mut<F>(e: &mut TypedExpr, f: &mut F)
```
groundlar: (blocking comment, see top-level comment)

dpmills:
All three blocking findings fixed in `77bd2a8e`, and they turned out to be one finding: the walk handed each child a `&[&TypedBinding]` slice, which left the *scope* implicit, so every consumer re-derived it per child. Your `Binders` sketch is what I took, with `Ambient` in place of `None`.

**1. Per-scope, not per-child.** `for_each_scoped_item_mut` now yields `ScopedItemMut::Scope` once ahead of the run of children a scope covers, so the change point is handed to the consumer rather than reconstructed by it — the per-child callback made per-scope work inexpressible, which is why it read as a cost bug rather than a mistake. The descent restricts and checks there. Your two probe tests are in as permanent guards (the `#[cfg(test)]` counter with them): "once per scope crossed" is a cost invariant of a release-active `assert!`, so a comment wouldn't hold it.

Re-measured on a wide group, as you asked. Release, 40-deep letrec spine, 32-node discharge term, per rewrite:

| group width | per child | per scope |
| --- | --- | --- |
| 1 | 355µs | 302µs |
| 8 | 4.62ms | 2.18ms |
| 32 | 44.5ms | 11.8ms |
| 64 | 152ms | 26.6ms |

You were right that the old "unchanged within noise" claim wasn't robust to one; it's replaced by this table in the PR body.

I did **not** take the `range_mentions` set-lookup on top, and the reason is that the set can't be hoisted to `rewrite_expr` without making the common case worse: computing it costs one `collect_expr_fv` walk per range entry *unconditionally*, while the per-scope check costs zero on a tree that crosses no binder — which most `discharge_in_place` targets are. It would need to be lazy, and a `OnceCell` threaded through both `rewrite_expr_go` and `rewrite_type_go` alongside `memo` is more machinery than the remaining cost justifies now that it's linear. Happy to revisit if a profile disagrees.

On `assert!` vs `debug_assert!`: intended, and the PR body now says so. What it catches is a uid minted outside lowering or a copy that broke uid preservation, and the consequence is silently-wrong output rather than a slow build — so the widened surface is the feature. Making it per-scope is what makes keeping it release-active cheap.

**2. No allocation.** `Binders::Group(&bindings)` borrows the group in place. Your `tests/scope_walk_allocations.rs` is in and green; I kept `is_free_in_value` as the probe for exactly the reason you gave.

**3. `shadows` collapsed.** It's a method on `Binders` now, so the four call sites ask instead of answer. The `&mut` side still clones binder names — a `&mut` walk can't hold a borrow into the node it mutates — but once per scope rather than once per child, which falls out of the same change.

### Smaller

- **Stale commit message** — fixed, and more than the one word: the whole node-local match is gone (see below), so the sentence describing its remaining arms went with it.
- **The remaining `_ => {}`** — taken structurally rather than left pinned. `for_each_scoped_item_mut` surfaces `VarRef`/`KeyRef` from its own match, exhaustive with no wildcard, and `mut_walk_surfaces_the_same_names` checks the two walks agree over the whole corpus rather than over four hand-written nodes. That does state the name rule twice — but the previous arrangement stated it twice too, with the second copy wildcarded and in another module, so this is the same count with exhaustiveness on both and a test between them. `rewrite_expr_go` now has no node-local match at all. Your `a_rename_retargets_every_varref_the_scoped_walk_surfaces` is kept as the end-to-end half. One thing that fell out: visiting every name occurrence in the tree would have cost a `Name` clone per node, so `handle_target` is split into `retarget` returning `Option<Name>` — "unchanged" is now distinguishable from "changed to the same thing".
- **`shadow_all`'s guard** — documented as load-bearing, with the `to_mut()`-clones-unconditionally reason.
- **`resize` monotonicity** — the `resize` is gone; scope runs are recorded with explicit first-child indices, and `debug_assert_eq!(next_run, runs.len(), …)` names the increasing-index invariant that lets one forward pass do the pairing.
- **Corpus completeness** — closed rather than tightened. A macro declares `variant_name`'s arms and derives `VARIANT_COUNT` from that same list, so a new variant makes the match non-exhaustive → adding the arm bumps the count → the bumped count fails `corpus_covers_every_variant` until the corpus gains an instance. A count declared separately wouldn't have closed it: nothing would have made anyone bump it. The no-duplicates half is in too.
- **`binds_any` short-circuit** — left alone, and I think this one's a non-finding: `walk_binders` has no early exit to take, and the callback is `|_| binds = true`, so an `any_child`-style guard would only avoid re-storing a `bool` that's already `true`. There's no `f` whose call is being saved. Say the word if you'd rather have it for symmetry.

`./ci.sh` green. `#41` is now behind by this commit and wants a restack.

src/ccl/scope.rs
dpmills: Taken — the per-child callback is gone. `for_each_scoped_item_mut` yields `ScopedItemMut::Scope` once ahead of the run of children a scope covers, so substitution restricts and checks per scope; details and the re-measurement are in the top-level reply.

src/ccl/design/ir.md
dpmills: It now covers three matches, not one — the scoped walk, the `&mut` walk's name half (new here, and what retires substitution's last `_ => {}`), and `walk_binders`. `ir.md` lists them as three instances of the same failure mode at different layers.

PR #44, opened 2026-07-28
ccl(infer): every inference error carries its node — locate the check family

groundlar:
Follows the review rounds on the substrate (#19): inference errors now carry a **mandatory** blame node, and the post-inference check family carries one too, so the errors users actually hit render with source context instead of as bare `error: …` lines.

**Start here:** `LocatedInferError` in `src/ccl/infer/api.rs` (the type and its docs), then `record_error` beside it, then `resolve_infer_errors` in `src/ccl/context.rs` — the one conversion from an inference error to a compile error.

### One rule, no modes

`check_fully_typed`, `typecheck`, `check_pre_desugar`, `check_mut_discipline`, `check_mut_write_targets`, and `ccl::infer::check` all return `Vec<LocatedInferError>`. There is deliberately no "located for the paths that render, bare for the paths that panic" split: *does this error end up in a rendering pipeline?* is a question that would have to be re-answered at every push site and every time a wall changes from panicking to reporting.

The 32 `typecheck(…).expect(…)` walls are untouched — `.expect` needs only `Debug` — and a hand-written `Debug for LocatedInferError` prints the error's own message plus the node, so panic payloads stay readable.

`impl IntoCompileErrors for Vec<InferError>` is **deleted**. Its four callers all had the lowering projection in scope, so the "fallback for callers without the projection" described a caller that cannot exist; no path can now silently emit a span-less inference diagnostic.

### Blaming a binder slot

A `TypedBinding` is not an `Expr` and has no `NodeId`, so a binder's type errors are blamed on the node that *declares* it, whose span covers the binder. One node can therefore carry two distinct errors — its own type's and its binder's — which is why the dedup key below is a pair.

### Dedup: `(node, error)` by value

The check walks recurse through *types*, and one type can mention the same unsolved variable several times: `Fun{?0, Tuple[?0, ?0]}` stated one fact about one node three times, each report carrying the same label because the label renders the enclosing node. Those are equal values at the same node, so exact equality is the whole key — no per-variant logic. `record_error` collapses them.

Not keyed on the node alone (that would drop distinct errors sharing a node), and reports of one cause at *different* nodes are kept: each node's type genuinely is unresolved. Measured on `λ x → (x, x)` with everything unsolved: 8 reports → 5, over 4 nodes. Presenting those as one finding is the renderer's job, and is left alone here.

### Tests

- `mut_discipline_error_renders_with_source_context` — the payoff, end to end: a rule-3 violation now renders an ariadne report underlining `b = a` at line 2.
- `binder_slot_errors_are_blamed_on_the_declaring_node`.
- `one_variable_reports_once_per_node` — pins both halves: the within-node collapse *and* that the four per-node reports survive it.

### Deliberately not here

Emission stays fail-fast; `TODO(multi-error)` on `emit_node` names the three decisions accumulation needs (a poison value to continue with, cascade suppression, and the run-wide `ConstrainCache` after a failed constraint). Check mode already accumulates over the same shared rules, so the rules demonstrably support it.

`IncompatibleBounds`'s `origin`/`context` labels are the other contributing sites of a collision; `TODO(structured-context)` records what making them locatable needs — they move onto the wrapper as label-with-node pairs, mirroring `ParseErrorInfo::context: Vec<(String, Span)>`, since a span inside `InferError` would break its location-free contract.

### DIFF
--- a/tests/compilation_pipeline/mutability.rs
+++ b/tests/compilation_pipeline/mutability.rs
@@ -468,6 +468,38 @@ fn cross_channel_cycle_is_rejected() {
     expect_compile_error(code, "mutually recursive cycle");
 }
 
+/// A `Mut`-discipline violation renders with source context, not as a bare
+/// `error: …` line.
+///
+/// These are user errors, and every one of them used to reach the terminal
+/// span-less: the check walks reported an error naming the offending expression
+/// only as a pre-rendered label, so `CompileError::Infer` carried `span: None`
+/// and the ariadne path was skipped. They now carry the node their rule raised
+/// them at, which `compile_program` resolves against the lowering projection —
+/// so this asserts the *report*, not just the message: the underlined source line
+/// is what tells a user where to look.
+#[test]
+fn mut_discipline_error_renders_with_source_context() {
+    let code = "a := 0\nb = a\nb";
+    let mut ctx = GlobalContext::default();
+    let consumer: Box<dyn Consumer> = Box::new(|| {});
+    let errs = match compile_program(&mut ctx, code, consumer) {
+        Ok(_) => panic!("expected a Mut-discipline error, but the program compiled"),
+        Err(e) => e,
+    };
+    let rendered = render_errors(&errs, "mut.chl", code);
+    assert!(
+        rendered.contains("not annotated `Mut`"),
+        "expected the rule-3 message, got:\n{rendered}"
+    );
+    // The ariadne report quotes the offending line and points at it; a span-less
+    // diagnostic would contain neither.
+    assert!(
+        rendered.contains("b = a") && rendered.contains("mut.chl:2"),
+        "expected an underlined report at line 2, got:\n{rendered}"
+    );
+}
+
 /// Rule 3: an unannotated copy of a mutable variable (`b = a`) aliases it — rejected, to
 /// force the author to disambiguate a value copy (`b: Int = a`) from seeding a
 /// new mutable variable (`b: Mut(Int) := a`).
--- a/src/ccl/context.rs
+++ b/src/ccl/context.rs
@@ -13,8 +13,8 @@ use crate::{
     ccl::{
         Expr, Type, channelize,
         infer::{
-            InferError, TypeInferenceContext, check_mut_discipline, check_mut_write_targets,
-            check_pre_desugar, infer, typecheck,
+            InferError, LocatedInferError, TypeInferenceContext, check_mut_discipline,
+            check_mut_write_targets, check_pre_desugar, infer, typecheck,
         },
         inline, lambda_elim,
         lineage::{Leak, RecorderSession, SourceProjection, collapse_lowering},
@@ -193,6 +193,37 @@ impl CompileError {
     }
 }
 
+/// Resolve each inference error's blame node to a source span and wrap it as a
+/// [`CompileError`].
+///
+/// The single conversion from an inference error to a compile error: every
+/// inference error carries the node its rule raised it at, and this is where that
+/// node becomes a span — one hop through the lowering projection, no fold, on the
+/// always-on release path. There is deliberately no span-less alternative
+/// conversion: a caller that has inference errors is inside `compile_program`,
+/// where the projection is in scope, so "forgot to resolve" cannot happen by
+/// picking the wrong helper.
+///
+/// `span: None` survives only for a node the projection does not cover — one
+/// minted after lowering (monomorphization, or a pass that mints while checking a
+/// rewritten tree) — which renders as a plain `error: …` line.
+fn resolve_infer_errors(
+    errs: Vec<LocatedInferError>,
+    projection: &SourceProjection,
+) -> Vec<CompileError> {
+    errs.into_iter()
+        .map(|located| {
+            let span = projection
+                .get(&located.node_id)
+                .and_then(|attr| attr.spans.first().copied());
+            CompileError::Infer {
+                error: located.error,
+                span,
+            }
+        })
+        .collect()
+}
+
 /// Build an ariadne report for an inference error pinned to a source span.
 ///
 /// Mirrors the parse/lower report builders: the `Debug` impl of [`InferError`]
@@ -268,18 +299,6 @@ impl From<ConversionError> for CompileError {
 /// per inference error.
 pub trait IntoCompileErrors {
     fn into_compile_errors(self) -> Vec<CompileError>;
-}
-
-impl IntoCompileErrors for Vec<InferError> {
-    fn into_compile_errors(self) -> Vec<CompileError> {
-        // Fallback for callers without the `lowering_projection` (i.e. not
-        // `compile_program`): no span resolution, so `span: None`. The
-        // `compile_program` path resolves spans explicitly and constructs the
-        // `Infer` variant itself rather than going through `.errs()`.
-        self.into_iter()
-            .map(|error| CompileError::Infer { error, span: None })
-            .collect()
-    }
 }
 
 impl IntoCompileErrors for lambda_elim::LambdaElimError {
@@ -818,18 +837,7 @@ pub fn compile_program(
     // an id the projection doesn't cover (a node minted after lowering, e.g. by
     // monomorphization) degrades to a span-less diagnostic.
     if let Err(errors) = infer_outcome {
-        return Err(errors
-            .into_iter()
-            .map(|located| {
-                let span = lowering_projection
-                    .get(&located.node_id)
-                    .and_then(|attr| attr.spans.first().copied());
-                CompileError::Infer {
-                    error: located.error,
-                    span,
-                }
-            })
-            .collect());
+        return Err(resolve_infer_errors(errors, &lowering_projection));
     }
     debug!("Inferred:\n{}", symbolic(&expr));
     debug!("Inferred (typed):\n{}", symbolic_typed(&expr));
@@ -844,9 +852,9 @@ pub fn compile_program(
     check_pre_desugar(&expr).map_err(|errs| {
         if errs
             .iter()
-            .all(|e| matches!(e, InferError::UnresolvedInfer { .. }))
+            .all(|e| matches!(e.error, InferError::UnresolvedInfer { .. }))
         {
-            errs.into_compile_errors()
+            resolve_infer_errors(errs, &lowering_projection)
         } else {
             panic!("Inference created invalid expr: {errs:?}")
         }
@@ -860,7 +868,7 @@ pub fn compile_program(
     // `user_annotation`s. Unlike the surrounding `check_pre_desugar` walls
     // (compiler-bug backstops), these are user errors: aliasing or nesting a
     // mutable reference.
-    check_mut_discipline(&expr).map_err(|errs| errs.into_compile_errors())?;
+    check_mut_discipline(&expr).map_err(|errs| resolve_infer_errors(errs, &lowering_projection))?;
 
     // Enforce that every `:=` / `+=` write targets a mutable variable (a write is
     // never a shadowing rebind of an immutable). Post-inference so binder types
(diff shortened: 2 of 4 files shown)


dpmills:
Drive-by from the predicate-sharing branch (#32), where I hit the same duplicate-report phenomenon from the other side and backed off. Two observations — both offered as information, not asks.

## 1. The dedup key includes a label the type docs call "not the location"

`record_error` keys on `(node, error)`, and `error` carries `at: String` — documented on the variant as "Display label for the message … not the location". So the key collapses two reports only when they happen to share a label.

In `one_variable_reports_once_per_node`'s own fixture, the lambda's `Fun` domain and its `param.ty` hold **the same `var`**. The two surviving reports at the lambda node are:

```
UnresolvedInfer { id: var, at: "λ x → (x, x)" }   // from expr.ty's Fun domain
UnresolvedInfer { id: var, at: "x" }              // from param.ty
```

One variable, one node, one span — two diagnostics, kept apart by a cosmetic string. The test's comment reads that as "distinct errors (the labels differ)", and for a lambda whose domain *is* its binder's type I'd argue it's one fact stated twice, which is the thing the dedup exists to collapse. Worth a second look; entirely possible you considered it and want both.

## 2. The wall silently passes holes in `user_annotation` and `Cast.target`

The complementary failure — not duplicates but silence. Both of these return `Ok(())` today:

```rust
// a hole riding only the cast's target slot
let e = Expr::cast(Expr::lit(Lit::Int(1)).with_ty(Type::Base(BaseType::Int)), Type::Hole)
    .with_ty(Type::Base(BaseType::Int));
assert!(check_fully_typed(&e).is_ok());   // passes — the hole is never visited

// a hole riding only the annotation
let e = Expr::lit(Lit::Int(1))
    .with_ty(Type::Base(BaseType::Int))
    .with_user_annotation(Type::Hole);
assert!(check_fully_typed(&e).is_ok());   // passes
```

`collect_expr_errors` visits `expr.ty` plus hand-written binder arms; neither annotations nor cast targets are in that set. Since this PR is what makes the check family authoritative, that seemed worth surfacing.

## Why they're coupled

I tried exactly this fix on #32 — routing the wall through `Expr::walk_type_slots` (the single-source-of-truth slot walk that branch adds: `ty`, `user_annotation`, `Cast.target`, and every binder's declared type). It closes the silence, and it **double-reports**: a lambda's unresolved param type becomes reachable both as `expr.ty`'s domain and as `param.ty`, so two existing tests fail on duplicated errors. I reverted it and left the wall alone.

Post-this-PR those two reports would land on the *same node id* — so the user sees two underlines of one span — and `record_error` won't collapse them, because the labels differ. So:

- widening coverage needs a dedup key that doesn't depend on the label;
- a label-independent key is what makes widening safe.

Neither seems right alone, which is why I stopped rather than pushing a half of it into #32. If you'd rather own both here, `walk_type_slots` will be on `main` with #32 and is a drop-in for the enumeration. If you'd rather keep this PR to locations, I'm happy to pick up the coverage half afterwards — just say which.

Nothing here blocks the PR as scoped; the located errors and the `IntoCompileErrors` deletion both read well to me.

groundlar:
Rebased onto #19's round-2 commits — content unchanged, base moved. `./ci.sh` green on this commit, including the new serde-feature pass.

One interaction worth naming, since it touches this PR's subject: #19 widened `LocatedInferError::bare`'s gate from `#[cfg(test)]` to `#[cfg(any(test, feature = "test-helpers"))]`. That method existed already but integration tests could not see it — separate crates — so `tests/inference_variants.rs` and `tests/type_check.rs` had each copied it verbatim. Both copies are gone and they call the library's one implementation.

That does not weaken the rule this PR is about. The helper is still unavailable in any production configuration (`test-helpers` is default-off and dev-dependency-only), so there is still no affordance for discarding the error↔node pairing on a path that renders — it just has one implementation instead of three.

dpmills:
Findings below, roughly in order of how much they matter.

## 1. `at` is overloaded, and *that* is the missing distinction — not the dedup key

This confirms my first comment, but I think it points somewhere slightly different.

The double report reproduces on a hand-built minimal tree — `λ x → 0` where the lambda's `Fun` domain and its `param.ty` hold the same var, everything else concrete:

```rust
let var = InferVar::fresh(0);
let e = Expr::new(TypedExprNode::Lambda {
    param: TypedBinding { name: "x".into(), ty: Type::Infer(var.clone()), user_annotation: None },
    body: Box::new(Expr::lit(Lit::Int(0)).with_ty(Type::Base(BaseType::Int))),
})
.with_ty(Type::Fun {
    name: None,
    domain: Box::new(Type::Infer(var.clone())),
    codomain: Box::new(Type::Base(BaseType::Int)),
});
check_fully_typed(&e).unwrap_err()
```

```
[Unresolved inference variable 0 in expression: λ x → 0 [NodeId(6)],
 Unresolved inference variable 0 in expression: x        [NodeId(6)]]
```

One variable, one node, one span, two errors — distinguished only by "in expression: λ x → 0" versus "in expression: x". The second is actively misleading: `x` is a *binder name*, and the message template says "in expression".

**How visible this is today.** I could not get there from CHL source: an unused UDF, a UDF applied at one type, an unused parameter, an empty list — all compile clean, so I never reached `check_pre_desugar`'s residual-`Infer` branch, which is the one path that would hand this pair to a user. The other user-facing consumers of the check family (`check_mut_discipline`, `check_mut_write_targets`) raise one error per node and don't produce a label-differing pair. So today the duplicate lands in `.expect` panic payloads — developer-facing — and it becomes user-facing exactly when the walls widen to cover the missing slots. Which is the coupling described above, from the other direction: the duplicate isn't urgent on its own, but it is a precondition for finding 2 rather than a consequence of it.

The fix isn't simply "drop `at` from the key", because `at` is doing two different jobs:

- In `collect_type_errors` for `expr.ty`, `at` is `symbolic(expr)` — a rendering of the blamed node, i.e. **redundant with `node_id`**.
- In the binder arms, `at` is `binding.name.base()` — the only thing identifying **which slot of the node** the error is about, because a `TypedBinding` has no id.

So a label-independent key is safe for the `UnresolvedHole` / `UnresolvedInfer` family and *unsafe* for the binder-named family: two same-named binders in one `LetRec` / `Case` group (`base()` drops the uid) with equal types would collapse into one report.

That is the same missing distinction the PR docs already name from the other side ("a `TypedBinding` is not an `Expr` and has no `NodeId`"). The structurally complete version is to make the blame a **node plus slot** rather than a node — `{ node: NodeId, slot: Slot }` with `Slot` as `NodeType | Annotation | Binder(usize) | CastTarget`. Then:

- dedup keys on `(node, slot, cause)` and `at` becomes purely cosmetic, as its own docs already claim it is;
- the lambda double-report collapses because you *decided* the domain slot and `Binder(0)` are the same finding, once and centrally, instead of by string coincidence;
- `LetRec` / `Case` binder errors stop being N reports underlining the whole group with no way to say which binder;
- the renderer can say "in the parameter `x`" without parsing a label.

That reads as a justified abstraction by the CLAUDE.md test — it makes explicit a concept the code currently encodes in a display string — and it is also the prerequisite for finding 2, so it isn't speculative.

## 2. The coverage question, answered: the gap is confirmed, and the tool is already on `main`

All three of these return `Ok(())` on this branch, probed directly:

| slot | result |
| --- | --- |
| `Cast.target` = `Type::Hole` | `Ok(())` |
| node `user_annotation` = `Type::Hole` | `Ok(())` |
| **binder** `user_annotation` = `Type::Hole` | `Ok(())` |

Two things sharpen the earlier version of this. First, [`Expr::walk_type_slots`](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/expr.rs#L1569) is **already on `main`**, not pending, and its own doc says in bold that *"a walk that visits only `b.ty` will silently skip every mutable binder in the program"*, plus *"a pass cannot acquire a blind spot by omission"*. [`collect_expr_errors`](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/infer/api.rs#L682) is a hand-enumeration of exactly the set that doc warns against, and this PR is what promotes it to authoritative. Second, the binder-annotation miss is the one that matters most: that annotation is where lowering writes `Mut(V, D)`, and where `binder_is_mut` reads it as authoritative over `b.ty`.

`src/ccl/design/type-inference.md` also describes this wall as asserting "every `ty` and every `TypedBinding::ty`" is concrete — accurate to the code, but a reader will take it as "every type slot".

None of this is a defect this PR introduces, and keeping the PR to locations is defensible. But the honest answer to "say which" is that the coverage half wants the slot abstraction anyway, so doing both in one change is cheaper than doing them in either order.

## 3. `resolve_infer_errors`' doc overstates when `span: None` can happen

[The doc](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/context.rs#L207) says `span: None` survives *only* for a node minted after lowering. There is a second, systematic source: per `src/ccl/design/provenance.md`, the lowering projection covers "every `walk_children` node — refinement-predicate interiors stay outside", and `collect_type_errors`' `Refinement` arm deliberately recurses into `refinement.predicate` with the predicate's own node ids. So a predicate-borne `UnresolvedHole` / `UnresolvedInfer` is **guaranteed** span-less, and not because of minting — and since `check_pre_desugar`'s all-`UnresolvedInfer` branch returns to the user, that is a user-reachable bare `error: …` line.

Either blame the enclosing refined-type node for predicate-interior errors (it is in the projection and does have a span), or amend the doc to name predicate interiors as the second case. Cheap either way, but wrong as written.

## 4. The two panic walls throw away the thing this PR just built

[Both](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/context.rs#L859) [walls](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/context.rs#L909) do `panic!("… created invalid expr: {errs:?}")`, which now prints `… [NodeId(6)]` — an id a human cannot resolve. But `lowering_projection` *and* `code` are in scope at both sites, so these could be `panic!("{}", render_errors(&resolve_infer_errors(errs, &lowering_projection), "<program>", code))` and print an underlined report instead.

These are compiler-bug backstops, and the developer staring at one is precisely who benefits most from source context. The PR's own framing — *"does this error end up in a rendering pipeline?" is a question that would have to be re-answered* — argues for not leaving these as the exception. The hand-written `Debug` impl is still worth keeping for the 32 `.expect` walls that have no projection.

## 5. `record_error`'s stated bound is wrong, and the scan is quadratic

[The comment](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/infer/api.rs#L462) says the list is "single-digit on the failure path". It isn't — it's roughly one error per unresolved node, and this PR's own test produces 5 from a 4-node expression. `errors.contains` scans *all* prior errors, not just the current node's, so it adds a quadratic term. Measured in release on a wide-tuple probe with all-distinct vars:

| errors | with dedup | without dedup |
| --- | --- | --- |
| 1000 | 0.93 ms | 0.42 ms |
| 4000 | 12.9 ms | 5.0 ms |
| 8000 | 60.7 ms | 19.9 ms |

Clean 4×-per-doubling both ways — the error path was already quadratic before this PR, since each error's `at` is an O(subtree) render allocated per error. So the dedup is a ~3× constant on an existing shape, not a new asymptote. I also tried reordering the struct fields so derived `PartialEq` compares the cheap `node_id` before the deep `error`: only ~20%, so that isn't the fix.

Not worth optimizing now, but the comment should change, since "single-digit" is exactly what would justify leaving the scan alone. If the slot abstraction from finding 1 lands, a `HashSet<(NodeId, Slot, Cause)>` falls out for free.

## Minor

- **This PR makes [the `IntoCompileErrors` trait doc](https://github.com/cambra-dev/cambra/blob/04006bd7/src/ccl/context.rs#L298) stale**: it still says "the inference stage flattens its `Vec<InferError>` into one `CompileError` per inference error", describing the impl this PR deletes. Inference no longer uses the trait at all.
- `one_variable_reports_once_per_node`'s comment ("Distinct errors (the labels differ) at one node") pins finding 1's behaviour as intended. If it is one fact stated twice, that assertion and its comment are the thing to change — worth not enshrining accidentally.
- `Case`-pattern and `LetRec` binder errors are blamed on the group node, so the underline is the whole `case` / `letrec`. Truthful ("the span covers the binder") but wide; subsumed by finding 1.

tests/compilation_pipeline/mutability.rs
```diff
@@ -468,6 +468,38 @@ fn cross_channel_cycle_is_rejected() {
...
+/// A `Mut`-discipline violation renders with source context, not as a bare
+/// `error: …` line.
+///
+/// These are user errors, and every one of them used to reach the terminal
+/// span-less: the check walks reported an error naming the offending expression
+/// only as a pre-rendered label, so `CompileError::Infer` carried `span: None`
+/// and the ariadne path was skipped. They now carry the node their rule raised
+/// them at, which `compile_program` resolves against the lowering projection —
+/// so this asserts the *report*, not just the message: the underlined source line
+/// is what tells a user where to look.
```
groundlar, line 480: rm context about "used to X", just mention what we're enforcing

groundlar:
this was forgotten: we want something similar eventually but not a priority atm

PR #45, opened 2026-07-29
Type conditional collections as dependent sums and compile them to a gated union

dpmills:
Two collections over different domains have no join: meeting `[0, 1] ⤇ Int` with `[0, 2] ⤇ Int` drops a row and leaves nothing in the type recording it, so `sum([1, 2] if c else [1, 2, 3])` was rejected outright. This adds the dependent sum — a data function carrying its Σ binders, `Σ (𝐷 : 𝐾). 𝐷 ⤇ 𝑉` — with a `box` builtin as the only way in, a four-constructor `TypeKind` lattice classifying what a binder ranges over, and a planning pass that rewrites a conditional collection into a gated union so it executes. `sum(box([1, 2]) if c else box([1, 2, 3]))` now compiles and answers, CHL's five collection types get one representation and a design of record ([`collections.md`](src/ccl/design/collections.md)), and the `zip`-of-generators gallery program that the incomparable-kinds rejection blocked now answers `20`.

## The type

`FunKind::Data` gains a slot — `Data(Option<Rc<Vec<Witness>>>)` — so a sum is a data function binding its witnesses over its domain, the way `Fun::name` already carries the Pi binder over the codomain. Occurrences are `Type::WitnessRef` leaves rather than `Infer`s, so nothing can unify a conditional collection down to one arm. `TypeKind` is a type of types with four constructors (`Enumerated`, `UIntRanges`, `SubtypesOf`, `Type`) differing only in how each states which types it classifies, which is what keeps Σ subtyping to a single rule with no case per kind. Spec: [type-inference.md, 4.7 Dependent sums](src/ccl/design/type-inference.md#47-dependent-sums) and [Type kinds](src/ccl/design/type-inference.md#type-kinds).

## Subtyping and consumption

One rule relates two sums, pairing binders positionally under an α-rename and drawing a kind premise, a domain premise and a codomain premise ([The Σ rule](src/ccl/design/type-inference.md#the-σ-rule)). Nothing puts a plain value into a sum and nothing forgets a witness, so a boxed arm meeting an unboxed one is a conflict rather than a silent dissolve. A consumer carries a `FunKind::Var` that the collection reaching it pins, which is how one aggregate accepts both a plain collection and a sum ([Consuming a sum](src/ccl/design/type-inference.md#consuming-a-sum-pinning-the-consumers-kind)).

## In the solver

The candidate list comes from the lattice join at coalesce, not from the `Case`: a conditional whose arms are both parameters has nothing to enumerate at the definition and still yields a sum at the call site ([Where the candidates come from](src/ccl/design/type-inference.md#where-the-candidates-come-from)). Compaction merges kinds and bodies over a fresh binder both contributions α-convert onto, with each merge law derived from a subtyping rule rather than declared; `subst.rs` carries the change of scope an edge between two binder scopes crosses; and `debug_assert_no_free_witness` asks the escape question at three phase boundaries — after inference, after lambda elimination, after planning.

## Compiling it

`planning/conditionals.rs` is new and takes the value-`Case` fan-out over from `lambda_elim`: each arm restricted by its first-match path condition, the restricted arms unioned, so exactly one leg is non-empty and the union's extent is the domain the witness selected. A new `TypedExprNode::Realize` re-views that union at the type the `Case` had. It is deliberately not a `Cast` — a cast's whole rule is the subtype obligation, and only the gates reconcile a sum with a tagged union ([Planning asserts the type it replaces](src/ccl/design/type-inference.md#planning-asserts-the-type-it-replaces)). A `let`-bound conditional is copied to each consumer, and a site's witnesses are compiled together as arm tuples; both are forced rather than chosen, argued in [collections.md, Compiling a conditional collection](src/ccl/design/collections.md#compiling-a-conditional-collection).

## Surface and spec

`box` is specified in [chl-spec, 7.5 `box`](docs/chl-spec.md#75-box), including that a boxed value cannot be boxed again — `box` takes a plain collection, so `box` of a sum is the plain-versus-sum kind conflict. The collection-type direction moves to `[Decided]`, with the five types and their lookups in [6.3](docs/chl-spec.md#63-direction-collection-types-decided), and [Joining the types of several values](docs/chl-spec.md#joining-the-types-of-several-values) is a new section collecting every construct typed by a join. Telling `Set` from `Map` stays [Open] and the operation layer stays [Planned], both scoped in `collections.md`.

## Formal model

Four new Lean files: `TypeKind.lean` (containment transports membership, transitivity, and that a refusal never lands on a member), `TypeKindIsALattice.lean` (both bound directions, row by row), `TypeKindMerge.lean` (the merge commutative, idempotent, associative), and `SigmaIsTheElementwiseReading.lean` (why the kind premise is unswapped, with the swapped premise exhibited unsound). Two differential oracles join the three existing ones, for `TypeKind::refuses` and for the kind merge, because the model's `CompactTy` has no binder slot, so the polar-merge oracle filters every kind-bearing case out before the wire. No sum reaches an oracle for that reason, which is why the binder correspondence `𝜌` is asserted at its derivation site instead ([What checks each premise](src/ccl/design/type-inference.md#what-checks-each-premise)).

Coverage: `tests/compilation_pipeline/sums.rs` is new (`box` end to end, determined witnesses, no `Case` anywhere), `tests/compilation_pipeline/conditionals.rs` grows the undetermined half including arms that share a domain base and differ only by a refinement, and `tests/type_check.rs` carries the type-level rules.


dpmills:

- #10
    - #11
        - #45 ◀
            - #156
                - #157
                    - #158
                        - #159

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
+# Collections
+
+Cambra's surface language (CHL) has lists, arrays, sets and maps. The runtime has one
+collection primitive: a **data function** `𝐷 ⤇ 𝑉`, a function whose domain is the data
+([type-inference.md, 4.6 Data vs compute functions](type-inference.md#46-data-vs-compute-functions)).
+This document is how the five collection types those four spellings and the abstract
+`Collection(𝑇)` name are that one primitive over different domains, and how they compile.
```
sortalongo: Doesn't parse.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
+# Collections
+
+Cambra's surface language (CHL) has lists, arrays, sets and maps. The runtime has one
+collection primitive: a **data function** `𝐷 ⤇ 𝑉`, a function whose domain is the data
```
sortalongo:
```suggestion
Cambra's surface language (CHL) has lists, arrays, sets and maps. The IR has one
collection primitive: a **data function** `𝐷 ⤇ 𝑉`, a function whose domain is the data
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+[type-inference.md, Subtyping for sums](type-inference.md#subtyping-for-sums). A collection
+type is a Σ over a witness **kind**, and this document says which kind each surface type
+picks.
+
+The **kind axis** is the part still open, and only for one pair. The witness kind already
+tells four of the five apart: `Array(𝑛, 𝑇)` is a bare data function over one range,
+`List(𝑇)` a Σ over `UIntRanges`, `Collection(𝑇)` a Σ over `Any`. `Set(𝐾)` and `Map(𝐾, 𝑉)`
+are the pair it cannot, sharing the `Keyed(𝐾)` kind and its key parameter and differing
+only in a codomain that is `unit` for one — so the operation layer has nothing to dispatch
+on between them.
```
sortalongo:
```suggestion
`List(𝑇)` a Σ over `UIntRanges`, `Collection(𝑇)` a Σ over `Any`. `Set(𝐾)` and `Map(𝐾, 𝑉)`
share the `Keyed(𝐾)` kind and its key parameter, differing
only in a codomain that is `unit` for one.
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+> **What is built.** A mechanism's section arrives in the branch that implements it, so a
+> section being here means its mechanism is built unless it is tagged `[Planned]` — which
+> the operation layer is, throughout. Where a `[Planned]` feature has an interim behavior in
+> today's code, that is tagged `[Interim]`: an unfinished state on the path to the design,
+> not a shim to remove.
```
sortalongo:
```suggestion
> **What is built.** Everything in this document is implemented unless it is tagged `[Planned]` (e.g.
> the operation layer). Where a `[Planned]` feature has an interim behavior in
> today's code, that is tagged `[Interim]`.
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+## The kind is declared, not read off the shape
+
+Which side of `𝐷 ⤇ 𝑉` holds the payload is not fixed by the shape. `Set(𝐾)` iterates its
+**keys** (the domain); `List(𝑇)` iterates its **values** (the codomain) — opposite sides
+of the same arrow — and `Map(UInt, 𝑉)` and a filtered `List` can share a shape while one
+must iterate entries and the other values. So which side is the payload is a fact about
+the collection's *kind*, and operations (`Iterable`, `Index`, `Membership`, `Ordering` —
```
sortalongo: "Kind" is already overloaded here, and in a super confusing way. Above, it's the kind of the sigma witness. Here, it's just another word for the collection type. I suggest doing a pass for this conflation, and calling this usage the collection's "type" (`List(T)`) or "type constructor" (`List`).

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+[Operations](#operations-how-the-trait-layer-is-realized-planned)) dispatch on the
+declared kind rather than reading it back from the arrow. For the same reason
+**"keyed-ness" is not a primitive**: there is no structural keyed property, only a
+per-kind choice of what `for`-in surfaces.
```
sortalongo:
```suggestion
per-kind choice of what `for ... in` surfaces.
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+Subtyping is the other axis and reads the shape as usual — `box(arr) <: List(𝑇) <:
+Collection(𝑇)`, each edge the ordinary Σ-width rule at that kind
+([type-inference.md, The width rule](type-inference.md#the-width-rule)) — so the two axes do
+not coincide: a `Set(𝐾)` is structurally a collection of `unit` and iterates `𝐾`.
+
+> **Tentative — the ambiguous pair gets a nominal type.** `Set` and `Map`, and only
+> those two, become **nominal type heads** — the `type` strength of
+> [chl-spec, Direction: term/type syntax split](../../../docs/chl-spec.md#61-direction-termtype-syntax-split-decided),
+> distinct from a structural `=` alias — landing when nominal types do. The other three
```
sortalongo:
```suggestion
> those two, become **nominal types**. The other three
```

"type head" is LLM jargon. Just call them "type constructors rather than coining a new term.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+> kinds need nothing: the witness kind already discriminates a bare range `Fun`
+> (`Array`), `UIntRanges` (`List`), `Keyed(𝐾)` (`Set`/`Map`) and `Any` (`Collection`),
+> so `Set` versus `Map` is the one pair it cannot tell apart — they share a kind and its
+> key parameter and differ only in a codomain that is `unit` for one. A nominal head
+> makes that difference readable from the type; a `Keyed(𝐾)` domain paired with a `unit`
+> codomain cannot say it.
```
sortalongo:
```suggestion
> key parameter and differ only in a codomain that is `unit` for one. 
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+> Nothing in [`Type`] carries a collection kind today, so this states an intent rather
+> than an implemented property.
+>
+> Three consequences:
+>
+> - **The head wraps the Σ rather than abbreviating it.** A `Map` that elaborates to
+>   `Σ (𝐷: Keyed(𝐾)). 𝐷 ⤇ 𝑉` and vanishes is structural again, and `Set` is `Map` again.
+>   Trait dispatch matches the head, so `Set` and `Map` are known nominal names rather
+>   than ordinary library declarations.
```
sortalongo: This is just repetition.

```suggestion
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+> - **Only the lateral relation becomes a decision.** `Array <: List <: Collection` is
+>   untouched. Width-to-top survives for the nominal pair without a declared edge:
+>   widening to `Collection(𝑉)` forgets which head it had, and the head's content
+>   `Σ (𝐷: Keyed(𝐾)). 𝐷 ⤇ 𝑉` has a kind contained in `Any`, so the edge falls out of the
+>   existing rule. What has to be stated is one rule — a nominal collection head is a
+>   subtype of its content — plus the head's parameter variance, once per constructor.
```
sortalongo: I can't understand anything in this paragraph. What the "lateral relation"? What's "width-to-top"? What's "the head's parameter variance"? What is "once" counting?

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+> - **The head's compact representation cannot be an [`AtomKey`].** `AtomKey` is a
+>   discrete set: reflexive arms only, atoms merge by union, and two shapes at one
+>   position is a collision. Because a head widens by being forgotten, two different
+>   heads meeting at one position join through their content rather than colliding, so a
+>   tag alongside the existing `fun` slot that the join drops is likely sufficient. If a
+>   full slot is needed, the shape to copy is [`Type::History`] — a fixed-arity
+>   constructor with type children and its own slot merged componentwise, differing in
+>   that `history_slot` recurses invariantly. Deciding between the two belongs to
+>   designing nominal types.
```
sortalongo: This seems like an implementation detail. If there is important content in here, it needs to be drawn out.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+>
+> **Sequencing.** This is a prerequisite for key and entry iteration, which is the
+> `Set`-versus-`Map` distinction, so until it lands `for`-in binds the codomain for every
+> kind.
```
sortalongo: What does sequencing have to do with the difference between `Set` and `Map`? Those are unordered collection types.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+> If nominal types land after the operation layer is needed, the interim to reach for is
+> **uniform entry iteration**: a `Set`'s entry is `(𝐾, unit)` and the projection to `𝐾` is
+> lossless, so `Map` gets correct entry iteration without the distinction existing. That
+> is surface-visible — `for k in s` would bind a pair — so it is a spec decision, not a
+> silent one.
+
+## The five collection types
+
+With `𝐷` a witness domain and `𝑛` a length:
```
sortalongo, line 35: Why is this way down here? We're 100 lines into the design doc, and just now defining the key terms that were discussed that whole time. Move it up to the top.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+- **`Array(𝑛, 𝑇)`** = `[0, 𝑛) ⤇ 𝑇` — domain `UIntRange(n)`, length *static*.
+  Ordered. Lookup `arr[𝑖] : 𝑇` — total, because the bound `{𝑖 | 𝑖 < 𝑛}` is
+  statically dischargeable. This is the shape the compiler builds for a list
+  literal today.
+- **`List(𝑇)`** = `Σ (𝐷: UIntRanges). 𝐷 ⤇ 𝑇` — *some* index range, which one not
+  known statically. Ordered. Lookup `lst[𝑖] : Option(𝑇)` — the bound is not
```
sortalongo:
```suggestion
- **`List(𝑇)`** = `Σ (𝐷: UIntRanges). 𝐷 ⤇ 𝑇` — *some* index range, which one not
  necessarily known statically. Ordered. Lookup `lst[𝑖] : Option(𝑇)` — the bound is not
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+- **`Set(𝐾)`** = `Σ (𝐷: Keyed(𝐾)). 𝐷 ⤇ unit` — a key domain with trivial codomain;
+  the domain *is* the payload. Unordered. Membership `𝑒 in 𝑠` discharges `𝑒`'s
+  presence in the domain.
+- **`Map(𝐾, 𝑉)`** = `Σ (𝐷: Keyed(𝐾)). 𝐷 ⤇ 𝑉` — a key domain. Unordered. Lookup
+  `𝑚[𝑘] : Option(𝑉)` in general, `: 𝑉` when presence discharges. Membership `𝑘 in 𝑚`.
+- **`Collection(𝑇)`** = `Σ (𝐷: Any). 𝐷 ⤇ 𝑇` — the witness ranges over *every*
+  domain; the domain rides along **in the value** (retained, not sealed — a
+  domain-generic consumer holds it abstract). Unordered. The ⊤ of the kind order, and
+  nothing more: keyed-ness lives in `Keyed(𝐾)`, so `Collection(𝐾)` is not
+  load-bearing for `Map`/`Set`.
```
sortalongo: I've said this before, but I really don't like calling this `Any`. `Any` has a _ton_ of precedent to mean "the supertype of all small types, and this is not that. This should be called `Type`, the type of small types. That's the standard name in dependent type theory.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+  marker.
+- **Order.** Sequentiality is deduced from loop-carried dependencies and ordering
+  is supplied as an `Ord` given, never fabricated. A positional domain (`UIntRange`, `Txn`,
+  an induction domain) is totally ordered by construction, so `Array` and `List` are
+  ordered; a keyed or opaque domain carries no order and an order-dependent operation over
+  one needs an `Ord[𝐾]` instance rather than a fabricated one.
+- **Views (`keys` / `values` / `items`).** `Map`'s projection operations, each a
+  **lazy view** — no copy: `keys(m) : Collection(𝐾)` (the key set), `values(m) :
+  Collection(𝑉)` (the map's own arrow), `items(m) : Collection((𝐾, 𝑉))`
+  (`𝑘 ↦ (𝑘, 𝑚(𝑘))`). Turning a `Map` into a `Collection(𝑉)` is a **re-pairing**
```
sortalongo:
```suggestion
  Collection(𝑉)` (the map's own arrow), `items(m) : Collection({𝐾, 𝑉})`
  (`𝑘 ↦ (𝑘, 𝑚(𝑘))`). Turning a `Map` into a `Collection(𝑉)` is a **re-pairing**
```

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+  (project the key set, re-introduce over domain `{𝑘 | 𝑘 ∈ keys}`) — runtime-free,
+  since `m` already carries its keys as its domain. `values(m)` is nonetheless the form to
+  write, because `for x in m` binds entries while `for x in (m : Collection(𝑉))`
+  binds values, and the explicit projection is what makes which one is meant
+  visible — matching [chl-spec §6.3](../../../docs/chl-spec.md#63-direction-collection-types-decided).
+  What *enforces* that is the **iteration element**, not a withheld subtyping edge:
+  the implicit `Map <: Collection(𝑉)` edge holds (⊤ absorbs every kind), and
```
sortalongo: > `for x in (m: Collection(V))`

> the implicit `Map <: Collection(V)` edge

`Map(K, V)` is not a subtype of `Collection(V)`, is it? I think that's undesirable—we want the explicit projections being presented here. And it's also not supposed to be if `Map` is a nominal type. The whole point of nominal types is to opt out of structural subtyping edges like this.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+  `sum(m)` is rejected once `sum` lowers through iteration, because a `Map` yields
+  `(𝐾, 𝑉)` entries and entries cannot be summed. Until then `sum(m)` means
+  `sum(values(m))`, and the per-kind iteration element is what changes that.
+
+## Realizing a conditional collection
```
sortalongo: "realization" is massively over-indexed in this PR. See my AI review comment on getting rid of the term.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+**Realization** replaces a conditional collection's Σ with a term the runtime can drive. A
+`Case` typed `Σ 𝐷 ∈ {𝐷ᵢ}. 𝐷 ⤇ 𝑉` becomes the gated union `⧺ᵢ (armᵢ | π̂ᵢ)` over the same
+domains, each leg gated by the path condition its `if`/`elif` compiled to. The gates are
+exclusive and exhaustive, so exactly one leg is non-empty and the union's extent is the
+domain the witness selected — which is what makes the union and the Σ the same collection.
+It runs in planning (`planning/conditionals.rs`) and asserts its pre-realization type
+rather than relating the two by a typing rule ([type-inference.md, Realization asserts
+rather than rewrites](type-inference.md#realization-asserts-rather-than-rewrites)).
```
sortalongo: This explanation is backwards. It starts with the case expression's type, rather than its definition. It would be way clearer with a simple example, which would show the code, then demonstrate the resulting type and corresponding dataflow.

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+
+**This is the only Σ the compiler can remove**, and therefore the only one that compiles.
+Nothing reads a witness off a value at runtime, so a Σ that reaches op-conversion with no
```
sortalongo: What? The compiler shouldn't be "removing" any types. Should this say "the only Σ-typed term the runtime can currently evaluate"?

src/ccl/design/collections.md
```diff
@@ -0,0 +1,248 @@
...
+twice, by its path condition `π̂ᵢ` and by `𝑝` rewritten to read that arm.
+
+The rewrite is what makes the filter expressible. A predicate may name a plain arm, but it
+may not name the gated union: reading the union needs the `iterate`/`restrict` markers that
+a predicate is forbidden to contain. Rewriting `𝑝` against `armᵢ` puts it somewhere a
+predicate is allowed to be.
+
+**So a restricted conditional is copied to each consumer that restricts it.** The legs carry
+one consumer's demand, so a `let`-bound conditional filtered by two consumers needs two sets
+of legs. Planning copies the conditional into each restricting consumer and drops the
```
sortalongo: This seems like a very inefficient thing to do—duplicating both upstream legs to do filter pushdown of filters from distinct consumers. Sometimes it's the right thing to do, but usually not. I suggest not implementing it for now, since deciding correctly requires having data, and it's a very unintuitive optimization.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+to a consumer that **discharges membership**: a filtered list keeps the source's index
+numbering, so `{[0, 2] | 𝑝} ⤇ Int` coarsened to `[0, 2] ⤇ Int` licenses proving
+`1 ∈ [0, 2]` and typing `xs[1] : Int` while the `Filter` may have removed index 1.
+That makes the rule and the lookup design one choice, not two: a domain's *predicate*
+must never be a proof source, membership riding the key against the domain's identity.
+
+## 4.7 Dependent sums
+
+A dependent sum `Σ 𝑤 ∈ 𝐾. 𝐵[𝑤]` is a pair: a **witness** `𝑤` — a type, drawn from a
+[witness kind](#witness-kinds) `𝐾` — and a value of `𝐵[𝑤]` at that witness. The sum is how
```
sortalongo: Let's switch away from the `∈` syntax. That really only applies to the `Enumerated` kind. The more general-purpose notation for our usage is `Σ W : K. B[W]`.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+`Type::Sigma` stores a [`SigmaType`]: a witness (a binder id plus its kind, [`Witness`])
+and a body. Occurrences of the witness in the body are `Type::WitnessRef` leaves naming the
+binder. The body takes one of exactly three shapes, and mentions its binder in all of them:
+
+- **Factored** — the body is a data function over the witness: `Σ 𝐷 ∈ 𝐾. 𝐷 ⤇ 𝑉`. The
+  witness is a *domain*, and the sum reads "a collection over some domain in `𝐾`, with
+  element type `𝑉`". A conditional collection has this shape, and so do `List` and
+  `Collection`.
+- **Unfactored** — the body is the witness itself: `Σ 𝑇 ∈ {𝑇ᵢ}. 𝑇`, "a value of one of
+  these types". What `box` builds ([Only a term builds a sum](#only-a-term-builds-a-sum)).
```
sortalongo: This concept has a name: "pointed set". Let's use it.

> what `box` builds

Is that right? I thought `box(array)` only put the domain in the witness position, not the whole type.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+contained in the universe, and the universe in nothing narrower, which is what rejects
+consuming a `Collection` where a concrete domain is demanded.
+
+#### An unresolved candidate becomes a kinding constraint
+
+A membership test needs a shape, and a computed collection's domain is still a variable
+when the entry term is emitted. So the requirement is recorded as a **kinding constraint**
+on that variable — `InferBounds::kinds`, merged by conjunction, non-polar because which
+kind a position must satisfy is one fact at either polarity — and discharged when the
+position resolves.
```
sortalongo: This smells wrong. A "kind" is just another kind of type. It can't be nonpolar. I think it's just that we don't have functions of kinds, so kinds never show up in the negative polarity. Correct here and have the implementation blow up if a kind is found in a negative position.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+- **The candidate position is invariant**, so `𝑎` is pinned to the argument's type exactly
+  and `box` never widens first. `box(5)` is `Σ 𝑇 ∈ {5}. 𝑇`, which is the point: retaining
+  the alternatives instead of joining past them is the whole service.
+- **A singleton does not collapse.** `Σ 𝑇 ∈ {𝑈}. 𝑇 <: 𝑈` holds by witness-forgetting
+  ([Subtyping for sums](#subtyping-for-sums)) and not conversely, so a one-candidate box
+  sits strictly below its content and is usable wherever the content is. That is what
+  makes boxing free where the arms agree.
+- **Nesting flattens**, so `box` is idempotent. It applies at materialization rather than
+  at construction, because `box`'s scheme writes `Σ 𝑇 ∈ {𝑎}. 𝑇` with `𝑎` a variable that
+  has no shape until the solver resolves it.
```
sortalongo: This contradicts the type signature above. `box(box(5)): Σ T ∈ (Σ U ∈ Int@5 . U) . T`

Making it idempotent seems hard to do rigorously, since traversing the isomorphism requires proving that the witness is an enumerated singleton. Why do it in the first place? I'd prefer to just let it be what it is.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+there is one term per witness kind.
+
+Read it as a statement about *entering*, not about the syntactic origin of every
+`Type::Sigma` value: joining two sums forms a third, and must, since a conditional over two
+boxed collections has to have a type. That join builds nothing new — its candidates are the
+joined sums' — so nothing reaches a sum except through a term that says so. A demand never
+forms one.
+
+**`box` — the enumerated sum.** `Σ 𝑇 ∈ {𝑇ᵢ}. 𝑇`, where the witness kind is the listing and
+the body is the bare witness:
```
sortalongo: This notation is bad, and has pervaded this PR. `{Tᵢ}` is _supposed_ to denote a set of types, but it collides with our record type syntax `{T,}`. The notation is also sloppy about cardinality: is `Tᵢ` one thing, or a set of things? The only way to tell is that it's surrounded by braces.

I'd prefer to use the following notation: `T ∈ [T₀, ...]`. This reuses our collection syntax `[ ]` and the `...` makes clear that there are more, without introducing a bound name (`i`) out of nowhere.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+### Subtyping for sums
+
+Two rules generate the relation, and one absence shapes it:
+
+- **Width** — `Σ <: Σ`, the only rule between two sums ([below](#the-width-rule)).
+- **Witness-forgetting** — `Σ 𝑇 ∈ {𝑇ᵢ}. 𝑇 <: 𝑈`, for a `𝑈` not itself a sum, exactly
+  when every `𝑇ᵢ <: 𝑈`. A Σ value is a pair, so a consumer valid at every candidate is
+  valid on it. Distribution instantiates the witness, so the rule applies only where the
+  body is the witness and the kind lists its candidates; a described kind lists nothing to
+  distribute over and is a mismatch.
```
sortalongo: Why make this implicit? I understand that it's more ergonomic, but my intuition is that we should have an explicit dereferencing operator. If something is in a box, you have to take it out of the box to use it. The type of the unboxed value is `lub(∀ i. Tᵢ)`.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+
+A Σ denotes a union — `Σ (𝑤: 𝐾). 𝐵[𝑤]` is `⋃{𝐵[𝑑] | 𝑑 ∈ 𝐾}` — so the rule is
+union-on-the-left: every instantiated body of the left sum lies below some instantiated
+body of the right.
+
+```
+    ∀ 𝑑 ∈ 𝐾₀. ∃ 𝑒 ∈ 𝐾₁. 𝐵₀[𝑑] <: 𝐵₁[𝑒]
+    ────────────────────────────────────
+      Σ (𝑤: 𝐾₀). 𝐵₀  <:  Σ (𝑤: 𝐾₁). 𝐵₁
+```
```
sortalongo: I'm pretty sure this is wrong. A sigma is a _disjoint_ union, not a regular union. So it's not "every left body lies below some right body". It's much more direct than that. It should work exactly like variant subtyping, but with a quantified tag.

BTW, I'm not an expert on typing rule syntax, but I strongly suspect logical quantifiers like ∀ and ∃ are a no-no. Type systems are intentionally much more restrictive than higher order logic—that's where they get their power. So the rule should look something like this:

```
    Γ ⊢ K₀ : Kind    Γ ⊢ K₁ : Kind    Γ ⊢ K₀ <: K₁
    Γ, W₀: K₀ ⊢ B₀    Γ, W₁: K₁ ⊢ B₁    Γ, W: K₀ ⊢ B₀[W₀ ↦ W] <: B₁[W₁ ↦ W]
    ─────────────────────────────────────────────────────────────────
             Σ (W₀ : 𝐾₀). 𝐵₀  <:  Σ (W₁ : 𝐾₁). 𝐵₁
```

I think it's worth double checking that the implementation checks all of these premises. The substitutions into the same context on the last premise seems particularly easy to miss.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+
+What the pairing edge is follows from the body shape, with no per-shape stipulation
```
sortalongo: "pairing edge" is silly claude jargon

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+function type — a collection being indexed or iterated — is decided by an arm that
+destructures the sum and emits edges against a **free** witness reference; the edges
+outlive both types of the constraint that fired the arm, so no rule over those two types
+can state them ([Consuming a sum: naming the
+witness](#consuming-a-sum-naming-the-witness)). Witness-forgetting is that arm's fallback
+for a sum the consumer cannot read as a collection; the `𝑈` there may itself be the
+consumer's function type.
+
+#### The width rule
+
```
sortalongo: Something should describe the intuition for the term "width" here. The analogy is with variant and record subtyping, where "narrower" types are subtypes of "wider" types. In a sigma type, the "width" is an aspect of both the witness kind and the body.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+- **Listing under listing** hands the pairing back as an obligation — the per-pairing
+  test is subtyping, which only the solver runs. The no-graph callers
+  ([`KindObligations::holds_structurally`]: the compact-domain lattice and the
+  post-coalesce check) discharge it by the identity instance, sound rather than complete.
+- **Listing under description**: the description admits the candidate
+  ([`TypeKind::admits`]). An unshaped candidate becomes a [kinding
+  constraint](#an-unresolved-candidate-becomes-a-kinding-constraint) rather than an
+  optimistic acceptance.
+- **Description under description**: the same description; `Any` on the right contains
+  every kind.
```
sortalongo: Use the specific type kind, not synonyms.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+
+**An unfactored sum of collections below a factored one** decomposes through the fibered
+view — each candidate as `(domain, element)` — exactly as factored/factored does, the
+difference being that an unfactored sum carries one element type per candidate rather than
+one shared, so the element edges are emitted per candidate (`data_candidate_fibers`). With
```
sortalongo: What's "the fibered view"? Never defined. Another case where a simple example would be much better than this vague description.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+
+So two collections over distinct domains have **no** join. `[1, 2] if c else [1, 2, 3]` is
+a domain conflict naming both domains, because no rule puts a data function below a sum,
+so there is no upper bound to find ([Only a term builds a
+sum](#only-a-term-builds-a-sum)).
+
+**`box` is what supplies one.** Each arm becomes a one-candidate sum, and the join of two
+sums is the sum over both candidates — `Σ 𝑇 ∈ {𝑇ₓ, 𝑇ᵧ}. 𝑇`, whose witness is the runtime
+branch discriminant. That Σ is the least upper bound *of the boxed arms*: it is a different
+element of the lattice from either, and it loses neither domain.
```
sortalongo: This isn't what `box` does as described above. The earlier description said box was a pointed set. `Σ T ∈ [T₀, T₁] . T ⤇ V` is not a pointed set.

I can think of a few resolutions to this problem:
1. Create a typing rule that make `Σ T ∈ [T₀, T₁] . T ⤇ V` the LUB of `Σ F₀ ∈ [T₀ ⤇ V] . F₀` and `Σ F₁ ∈ [T₁ ⤇ V] . F₁`. This seems doable, but is quite complicated machinery to embed in the type lattice, and that scares me a bit.
2. Make `box` part of a trait, so it can have different types depending on what it's given. This seems promising.
3. Use explicit conversion functions like `array_to_list`, `list_to_collection`, etc. This seems like a recipe for a mess of conversions.

I'd favor 2, but am open to ideas.

sortalongo (commented):
Still WIP, but have tons of comments, so pushing out so we can pipeline.

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+
+| law | join property |
+|---|---|
+| candidates are the arms' | union |
+| nested conditionals flatten | associativity |
+| identical arms dedup | idempotence |
+| the codomain is shared | the codomain join, which may **fail** |
+| unboxed collection arms have no upper bound | a domain conflict, not a silent narrowing |
+| scalar arms produce no Σ | the same join, intersecting refinements |
+
```
sortalongo, line 1790:
```suggestion
| scalar arms produce no Σ | the same join, union refinements |

```
right?

src/ccl/design/type-inference.md
```diff
@@ -1202,31 +1066,491 @@ domains; `a_data_domain_relates_only_to_itself` pins all four directions plus th
...
+`emit_case` constrains every arm into a fresh result variable rather than requiring
+equality, so the arms meet at their join: homogeneous arms join to their common type, and
+data-collection arms with distinct domains form the sum.
+
+#### The codomain join
+
+**A Σ is lossless on the domain and lossy on the codomain.** The shared body's element type
+is the ordinary covariant join `𝑉₀ ⊔ 𝑉₁` of the arms' codomains, forgetting which domain
```
sortalongo, line 1798: A Σ doesn't have a domain or codomain. I think this is assuming that there's a structure `Σ W: Kind . D ⇒ C`, and talking about the join on two sigmas of that structure. But I haven't seen a definition of that join anywhere. Not having a definition is concerning since it's hard to evaluate its soundness. Should it be here, or did I miss it elsewhere?

sortalongo (commented):
more WIP. Pivoting to other PRs until you do the simplifaction we discussed.

src/ccl/design/collections.md
dpmills: Applied verbatim.

src/ccl/design/collections.md
dpmills: It didn't. Rewritten: "This document gives each surface type — those four spellings and the
abstract `Collection(𝑇)` — as that one primitive over a different domain, and says how each
compiles."

src/ccl/design/collections.md
dpmills: Applied, with two corrections the suggestion predates: the kind is `SubtypesOf(𝐾)`, not
`Keyed(𝐾)` — no `Keyed` variant was ever built, `TypeKind` has four — and `Any` is now
`Type` per your other comment.

src/ccl/design/collections.md
dpmills: Applied verbatim.

src/ccl/design/collections.md
dpmills: Gone; they're "declared type constructors" now. Zero occurrences of "type head" in the doc.

src/ccl/design/collections.md
dpmills: Applied verbatim, in the intro.

src/ccl/design/collections.md
dpmills: Deleted. The nominal-types blockquote is gone entirely; the wraps-or-abbreviates point is
stated once, as the first of the three questions.

src/ccl/design/collections.md
dpmills: Deleted. "lateral relation", "width-to-top" and "the head's parameter variance" have zero
occurrences in the branch. What that paragraph was reaching for is now one sentence:
"Subtyping is the other axis and reads the shape as usual — `box(arr) <: List(𝑇) <:
Collection(𝑇)`, each edge the ordinary Σ rule at that kind."

src/ccl/design/collections.md
dpmills: Cut. The `history_slot`-recurses-invariantly and componentwise-merge material is out of the
section; it now carries the three questions and nothing about compaction.

src/ccl/design/collections.md
dpmills: Nothing, and that was the confusion — the paragraph meant sequencing *the work*, not
collection ordering. Rewritten as "**Until this is settled** `for ... in` binds the codomain
for every collection type", with the interim named.

src/ccl/design/collections.md
dpmills: Applied verbatim.

src/ccl/design/collections.md
dpmills: Renamed, in the doc and in the code: `TypeKind::Type`, and `Collection(𝑇)` = `Σ (𝐷 : Type). 𝐷 ⤇ 𝑇`.

src/ccl/design/collections.md
dpmills: The brace correction is applied. One deviation: the line reads "the map's own **function**"
rather than "arrow" — a later vocabulary pass settled on "function" throughout, matching
`Type::Fun` and "data function". Say the word if you'd rather have "arrow" here.

src/ccl/design/collections.md
dpmills: Reordered exactly that way. The section now opens with the program, then gives the type it
produces, then what planning does with it.

src/ccl/design/collections.md
dpmills: Your wording is the wording: "**A conditional collection is the only Σ-typed term the
runtime can currently evaluate.**" (One note: a later branch in the stack moved this
paragraph and reintroduced "the only Σ the compiler can remove". Fixed there too.)

src/ccl/design/type-inference.md
dpmills: Done, and codified so it stays done — CLAUDE.md's notation section now requires the binder
kinded with `:`. Zero occurrences of a `Σ`-with-`∈` in the tree.

src/ccl/design/type-inference.md
dpmills: Brackets adopted: a listing kind is `[𝐷₀, 𝐷₁]`, or `[𝑇₀, …]` where the length is beside the
point. The binder is kinded with `:` rather than `∈`, per your comment above — so
`Σ (𝐷 : [𝐷₀, 𝐷₁]). 𝐷 ⤇ 𝑉`. That resolves the two comments together; the reason for `[ ]` is
the one you gave, that braces are the record type's.

src/ccl/design/type-inference.md
dpmills: The intuition is now the rule's first sentence, in the terms you named: "a narrower sum is a
subtype of a wider one, as a variant with fewer arms is a subtype of one with more." The
term itself is gone — the section is "The Σ rule", and "width" is left to mean only the
record/variant sense it already had.

src/ccl/design/type-inference.md
dpmills: Gone, zero occurrences.

src/ccl/design/type-inference.md
dpmills: Done. "witness kind", "listed kind" and "described kind" are all at zero; kinds are named
`Enumerated`, `UIntRanges`, `SubtypesOf` and `Type`.

src/ccl/design/type-inference.md
dpmills: Deleted, zero occurrences.

src/ccl/design/collections.md
dpmills: Half applied. The collection usage is "type" / "type constructor" as you asked. Rather than
leave "kind" ambiguous, the intro now reserves it: *""Kind" throughout means that
`TypeKind`, never the collection type itself; a witness is a binder, and the kind classifies
the types it ranges over rather than the binder."* The four senses in the codebase
(`fun_kind`, `type_kind`, `history_kind`, and the surface collection type) are settled with
that split recorded in CLAUDE.md.

src/ccl/design/collections.md
dpmills: Taken as "per-**type** choice" — the comment above asks that "kind" not name the collection
type, and this is that usage. Substance unchanged.

src/ccl/design/collections.md
dpmills: Down to four occurrences in this doc, all in the one subsection describing
`TypedExprNode::Realize`. It is the node's name in the code rather than prose coinage now,
so it stays where the node is discussed and nowhere else.

src/ccl/design/collections.md
dpmills: Agreed that it's undesirable, and the doc no longer asserts it. It is now one of the three
open questions under "Telling `Set` and `Map` apart [Open]", stated from both sides:
structurally the edge falls out of the Σ rule today, and withholding it is exactly what a
declared type constructor would be for. Nothing in `Type` carries a collection type
constructor yet, so neither side is implemented.

src/ccl/design/collections.md
dpmills: It isn't an optimization, and the doc now says why: "This is β, not a pushdown. Nothing
weighs running the filter inside the leg against running it after the union: a leg is the
site with the witness instantiated, and an occurrence left uninstantiated names a witness
that no longer exists. A cost model has nothing to decide here, and no ordering of the two
gates is available to choose between." A predicate holds its own read of the source, and
realizing the site substitutes one candidate for the witness — both occurrences, because
it's one substitution. Leaving one uninstantiated is not a slower plan, it's an ill-formed
one. If the objection is to the *duplication* rather than to the ordering, that's the
let-bound-conditional copy, and it is a separate paragraph worth taking on its own.

src/ccl/design/type-inference.md
dpmills: You're right about kinds and the doc now separates the two things. It is the *kinding
constraint* that has no polarity, because it isn't a bound: "`α :: 𝐾` asserts what `α` must
resolve to, which is one fact wherever `α` occurs." The second half would misfire, though —
a kinding constraint **does** reach a negative position, since an annotation is a demand:
`r: List(Int) = box(…)` records `UIntRanges` on the domain variable at negative polarity
(`test_kinding_constraint_survives_instantiation`). So an assertion there would fire on a
legitimate program. What sits at that position is the constraint, not the kind.

src/ccl/design/type-inference.md
dpmills: Your reading was right, and that's what the doc now says: `box` makes each arm a
one-candidate sum over its **domain** — `Σ (σ : [𝐷ₓ]). σ ⤇ 𝑉` — so the witness position
holds the domain, not the whole type. "Pointed set" is not used, because there is no longer
a claim for it to name.

src/ccl/design/type-inference.md
dpmills: Idempotence is not claimed anywhere any more (zero occurrences). What the doc states instead
is flattening, as a structural invariant: "A Σ re-entering compaction lands in the same slot
with its candidates enumerated, so `box(box(xs) if p else box(ys)) if q else box(zs)` forms
one flat three-candidate sum. A candidate domain is therefore always a ground data-function
domain, never itself a Σ." So the nested shape you wrote can't arise — not by an isomorphism
anyone has to traverse, but because a candidate is never a sum.

src/ccl/design/type-inference.md
dpmills: There is no implicit one. The doc states two absences as rules: "**No elimination** — no rule
concludes `Σ … <: 𝑈` by forgetting the witness", and "There is no consumption arm either — no
rule, in any spelling, reads a sum at a plain function." A consumer that takes both is
polymorphic over the kind and the value pins it.

The explicit operator exists too, and it is `list(…)` — the middle row of the table under
"The domain join needs `box`": `list(xs) if c else list(ys) : List(𝑉)`, keeping "the rows,
not which range — so lookup is partial". That is the deref you're describing, and its type
is the lub you wrote, for the codomain. For the *domain* there is no lub — that's what `box`
exists for — so an unbox that returned one would have to pick a domain, which is the
row-destroying meet the section opens by rejecting.

src/ccl/design/type-inference.md
dpmills: The premise this rests on is gone: `box` boxes the **domain**, not the whole type, so there
is no `Σ F₀ ∈ [T₀ ⤇ V]. F₀` for it to disagree with — as you suspected in your comment on
1109. `box(xs)` is `Σ (σ : [𝐷ₓ]). σ ⤇ 𝑉` and the join of two of those is
`Σ (σ : [𝐷ₓ, 𝐷ᵧ]). σ ⤇ 𝑉`, one lattice element, both domains kept. So none of the three
resolutions is needed; if you still read a mismatch after that change, worth reopening.

src/ccl/design/type-inference.md
dpmills: Rewritten to your rule, generalized to arity `𝑛`. The `∀`/`∃` phrasing is gone, and the
premises line up one for one: `𝐾₀ⁱ <: 𝐾₁ⁱ` is your kind premise, `𝜌` is your
`[W₀ ↦ W]`/`[W₁ ↦ W]` substitution to a common binder, and your `B₀ <: B₁` is drawn as the
two halves of the body arrow — `𝐷₀‹𝜌› <: 𝐷₁` invariantly (a data domain is the data) and
`𝑉₀‹𝜌› <: 𝑉₁` covariantly.

Your closing suspicion was right on both counts, and both were real bugs:
- **The substitution into a common context was being missed.** A var-to-sum edge drew no
  rename, so the two sides' witnesses were compared as unrelated names. Fixed by deriving
  the correspondence from the binder *identities* rather than from stated kinds.
- **The kind premise wasn't drawn everywhere.** It ran on the live solve only, so the
  post-inference check accepted `Collection <: Map`, `Collection <: List`, and a
  two-candidate sum below a one-candidate one. It now runs at every derivation, and the doc
  says so under "Every premise is drawn at every derivation".

The rule also has a differential oracle against the Lean model now, so the premises are
checked against an independent implementation rather than only by tests.

src/ccl/design/collections.md
dpmills: Moved. "The five collection types" is now the first section, at line 33, and the intro links
forward to it. The comment is still anchored because GitHub followed the section rather than
the old position.

src/ccl/design/type-inference.md
dpmills: Intersect rather than union, and I think the suggestion is a slip. Refinements are
width-subtyped like records — more refinements ⇒ subtype — so they merge with the same
polarity rule as `rec`: positive ⇒ intersect, negative ⇒ union (§"A refined type"). A
conditional's arms meet at their **join**, which is positive, so the shared refinements
survive and the rest drop.

The counterexample is one line: `1 if c else 2` types as `Int`. Under union it would be
`{Int | __elem == 1, __elem == 2}`, which is uninhabited.

src/ccl/design/type-inference.md
dpmills: The structure is stated, at the top of §4.7 rather than here, which is probably why it read
as assumed: a sum is a "data function carrying its binders" — "Every binder scopes over the
**domain**, whose occurrences of it are `Type::WitnessRef` leaves naming the binder; the
codomain is the witness-independent residue, one element type shared across the candidates.
`Σ (𝐷 : 𝐾). 𝐷 ⤇ 𝑉` reads 'a collection over some domain in `𝐾`, with element type `𝑉`'."
So the body is always the arrow, and that is now a stated invariant rather than a shape the
passes happen to build.

The join has the two halves the structure implies, each with its own section: the domain side
is "The domain join needs `box`" (candidates union; no join without `box`), and the codomain
side is this one (the ordinary covariant join, which may fail). Happy to add a forward
pointer from §4.7 to both if that's the missing thread.

src/ccl/design/collections.md
sortalongo: Applied — `SubtypesOf(𝐾)` is the right name and my suggestion predated it. Accepted.

src/ccl/design/collections.md
sortalongo: Accepted. Reserving "kind" for `TypeKind` is better than leaving it ambiguous, and the four-sense split belongs in CLAUDE.md.

src/ccl/design/collections.md
sortalongo: Right — "per-type" is the usage my other comment asked for. Accepted.

src/ccl/design/collections.md
sortalongo: "function", matching `Type::Fun` and "data function". One name per concept.

src/ccl/design/collections.md
sortalongo: The edge stays for now; nothing can withhold it until a declared constructor exists, and `Type` being ⊤ is what `List(𝑇) <: Collection(𝑇)` rides too.

What I do not want is #156 pinning it. `test_keyed_collection_widens_to_collection_like_any_other` asserts `Map(Int, Int)` into a `Collection(Int)` parameter positively, names it "like any other", and argues in its doc comment that there is nothing for the kind lattice to withhold. Today the edge is an unasserted consequence of ⊤; after that test it is specified behavior with a green test defending it, and answering this question my way later costs a red test and a doc rewrite.

Please restate that test and `test_map_consumption_is_kind_blind_interim` as interim pins naming this open question — the way #156 already marks its `g(k)` regressions — rather than as properties.

Separately, #156 reintroduces vocabulary this PR retired: `Any` as the kind name twice (the `Collection(𝑇)` line and that test's comment) and `width-to-top` once. Same drift you caught yourself on the "only Σ the compiler can remove" paragraph, so it is worth a sweep of the stack rather than a spot fix.

src/ccl/design/collections.md
sortalongo: The paragraph exists and answers me — "So a `let`-bound conditional is copied to each consumer", and the closing point that the copy is the only compiling form until a runtime witness exists, so the duplication is an obligation rather than a cost question. I withdraw the objection.

What is missing is the connection, which you had to supply by hand here. Please point the β-not-a-pushdown paragraph at it, so a reader who arrives at the substitution argument asking about cost lands on the paragraph saying why cost is not yet the question.

src/ccl/design/type-inference.md
sortalongo: Accepted. Boxing the domain is what I was reaching for, and dropping "pointed set" along with the claim is right.

src/ccl/design/type-inference.md
sortalongo: Agreed, and your counterexample holds — an annotation is a demand, so a kinding constraint does reach a negative position. Withdrawn; separating the constraint from the kind is the fix, and an assertion there would have fired on a legitimate program.

src/ccl/design/type-inference.md
sortalongo: The invariant is the right replacement for idempotence, and it is prose only. "A candidate domain is therefore always a ground data-function domain, never itself a Σ" has nothing holding it — no `debug_assert` in `compact.rs`, and the `unreachable!`s in `coalesce.rs` guard one candidate per function slot, which is a different invariant. Please assert it where a candidate is installed, with the invariant named in the message.

The PR description still says `box` is idempotent. This repo squash-merges, so that lands on main as the commit message — please drop it there too.

src/ccl/design/type-inference.md
sortalongo: Accepted, and this settles 1109 and 1285 with it. `box` boxing the domain rather than the whole type is what I was reaching for, and it removes the mismatch those two were about: with the join of two one-candidate sums landing as one lattice element, none of the three resolutions I proposed is needed.

On the unbox itself: no implicit elimination, kind-polymorphic consumption, `list(…)` as the explicit lossy exit, and no lub on the domain side. My `lub(∀ i. Tᵢ)` was the codomain only. Take the model as settled for the rest of the stack.

src/ccl/design/type-inference.md
sortalongo: Two corrections, then a TODO.

The differential oracle does not cover this rule. `tests/differential_oracle.rs` drops witnesses before the wire — `AtomKey::Witness(..) => return None`, "the model has no witness — outside the fragment for the same reason `ChanDom` is" — and `SubtypeChecker.lean` relates only `FunKind`, never a `TypeKind` or a binder. What the oracles do cover is the kind-lattice merge (`CompactTypeKind::merge` against `mergeTypeKind`) and `refuses`. So "the premises are checked against an independent implementation rather than only by tests" holds for the kind lattice and not for the Σ rule.

Which premise it is, is why this matters. The ρ substitution to a common binder is the premise that was missing, and it is the one the oracle cannot see. Tests are the only thing standing on it.

TODO, either form: give the model's `CompactTy` a Σ binder slot so the rule reaches the oracle, or assert at the derivation site that the rename was drawn. Whichever is cheaper — what I want is that a regression in ρ is not silent. Please also state in the rule's section which premises are machine-checked and by what; it currently claims no coverage either way, which is how I read the reply as covering more than it does.

src/ccl/design/type-inference.md
sortalongo: Accepted — the premise change resolves it, and there is nothing to reopen. With `box` over the domain there is no `Σ F₀ ∈ [T₀ ⤇ V]. F₀` to disagree with, so none of my three resolutions is needed.

src/ccl/design/type-inference.md
sortalongo: You're right, it was a slip. A join is positive, so refinements intersect; `1 if c else 2` settles it.

src/ccl/design/type-inference.md
sortalongo: Reads fine. Yes to the forward pointer from §4.7 to both halves.

sortalongo (approved):
Approving. The Σ formation, the subtyping rule and the kind lattice are in the shape I want, and the box model — `box` over the domain, no elimination, `list(…)` as the explicit lossy exit — is settled for the rest of the stack. The rule rewrite is a genuine improvement on what I first read, and the two bugs it turned up (the missing ρ rename, the kind premise drawn only on the live solve) are worth the round trip.

Four threads are left open as non-blocking follow-ups rather than change requests:

- **The Σ rule's oracle coverage.** The differential oracle does not reach this rule — witnesses are dropped before the wire and the Lean checker relates only `FunKind`. The ρ substitution, the premise that was actually missing, rests on tests alone. A model binder slot or an assertion at the derivation site, plus a line saying which premises are machine-checked by what.
- **The flattening invariant is prose only.** Nothing holds "a candidate domain is never itself a Σ". Assert it where a candidate is installed.
- **`Map(𝐾, 𝑉) <: Collection(𝑉)`.** Fine to land unwithheld — `Type` being ⊤ is what `List <: Collection` rides too — but #156 should record it as an interim consequence rather than pin it as a property.
- **A cross-reference** from the β-not-a-pushdown paragraph to the let-bound-conditional copy.

One thing to fix before merge, since the repo squash-merges and the body becomes the commit message: the description still says `box` is idempotent, which the doc no longer claims.

src/ccl/design/collections.md
dpmills: Both restated. Each now says what it pins: the state today, the question that decides it, and what the assertion becomes if it's answered your way. The `List` half stays a property.

On the vocabulary — `Any` and `width-to-top` have zero occurrences on any of the five heads, pushed or local; both went on this PR, which is an ancestor of the rest. The test you named had already been renamed at the pushed head, so that part reads as a stale checkout. The real sites were the `is realized` heading on #156 and one `arrow` in prose on #158. Both fixed, and I swept `arrow` across the stack: 16 sites, all the stack's own, `main`'s 58 left for a separate change.

src/ccl/design/collections.md
dpmills: Pointed. "A `let`-bound conditional is copied to each consumer" is a heading now, so the β paragraph links to it — a bold lead-in wasn't citable.

src/ccl/design/type-inference.md
dpmills: Asserted in `Witness::formed`, the one place a binder is paired with what it ranges over; the four constructors that built the struct literally now route through it. Confirmed it fires on materialized candidates, not just variables.

`box` is worse than the doc claimed: `box(box(xs))` is a type error, since `box` takes a plain collection and a sum isn't below one. So the spec's idempotence claim is gone, and so is the flattening example — it used that exact shape. `(box(xs) if p else box(ys)) if q else box(zs)` is the nesting that flattens, and it compiles.

src/ccl/design/type-inference.md
dpmills: Correct on both counts. §4.7 gains "What checks each premise": no sum reaches an oracle, the kind lattice does, and `𝜌` has tests plus a new assert — I took that fork, so the Fun/Fun arm checks each side states one binder identity per position and that two sums pair every binder. `formal/design.md`'s roadmap and coverage table now agree; its concrete-pairs row had read as covering the Σ arm.

The same defect ran the other way in that section: the compaction paragraph described a `sigma` slot merged by `CompactSigma::merge` with laws in `distribute_sigma`. None of the three exists — the code has one `fun` slot and says so. Four more citations named nothing, and five test comments explained consumption as a `Σ <: Fun` rule.

PR #47, opened 2026-07-30
Doc refs: check the references line-based scanning was silently skipping

dpmills:
`check_doc_refs.py` extracted Rust comments one line at a time, so a citation that wrapped across two `///` lines never matched the citation grammar — the opening quote had no closer on its line — and its section title went unchecked. Nothing reported it: the reference was counted and `./ci.sh doc_refs` passed. 7 of 25 quoted titles in the tree were being skipped that way, and 4 of the 7 named headings that no longer exist.

## The fix is at the extraction layer

The unit a sentence is written in is a *comment run*, not a line. `_rust_comments` now yields a `Prose` value — a block comment, or a run of consecutive line comments with markers stripped and line breaks preserved — and `Prose.line_at` maps a match offset back to an exact source line, so a problem points at the reference's own line rather than the comment's first.

A wrapped title's line break, comment marker, and indentation collapse back to the single space they stand for before slugging (`_slug_base` maps each whitespace character to its own hyphen, so this matters).

Wrapping runs are bounded twice — by a line budget, and by never crossing a blank line — because a delimiter allowed to roam far enough pairs with an unrelated one and invents a match out of ordinary prose. One `_wrapping()` helper builds that pattern for every scanner.

An opening quote where a title belongs, with nothing closing it, is now reported as an *unterminated quoted citation*. Treating an unparseable citation as "no citation" is the failure mode this whole change is about.

## The same hole, one level over

`check_markdown_links` also scanned line by line, so a link whose *text* wrapped across the hard-wrapped prose was invisible to it. It now scans whole prose with fenced and inline code blanked out (newlines preserved, so offsets still map to lines). That picked up 2 more links in `src/ccl/design/mutability.md`.

## Check B now reads doc prose, not just Rust comments

A doc citing another doc's section names it the same way code does, and that form was checked nowhere. Three things the doc side needs that the code side does not:

- **Inline code has to survive.** Headings here are full of backticks; blanking the spans mangles the title being cited.
- **The path rule stays lenient until a title appears.** Requiring a unique path resolution produced 14 false hits (`# CLAUDE.md`, a file's own title; two files named `lowering.md`). In a comment the `.md` mention *is* the citation — there is no link syntax to use — so it must resolve. In a doc it is just prose, as legitimately loose as the source paths check C tolerates. A quoted title is what makes it a citation, and that is exactly when the path must be unambiguous.
- **Citations hang off links.** `[mutability.md](mutability.md), "Title"` is the usual shape, so the citation tail is taken from after the closing paren. The link's own path and anchor remain check A's business.

That surfaced 7 stale citations across 5 docs — including a section cited from three places that no longer exists, and a `§4 ("Retire desugar_defers")` whose subject is gone from the target doc entirely. All are now anchor links, which retires six `§`-refs along the way.

## Two bold lead-ins promoted to headings

Several failures cited a **bold paragraph lead-in** rather than a heading. Those read like landmarks but carry no anchor and cannot be checked. The two that code actually points at are now headings — "No aliasing: `Mut` values are second-class (downward-only)" in `mutability.md` and "Apply is one-way" in `type-inference.md` — and the citations name them exactly. Five citations became checkable at those sites, two of which previously carried no doc path at all.

Only these two were promoted. Most lead-ins are list items, which cannot become headings without destroying the list, and converting the rest would roughly double `type-inference.md`'s heading count for no inbound reference.

## Incidental fixes the work surfaced

- `Repo.resolve_suffix` used `lstrip("./")`, which strips *characters* rather than a prefix: any dotfile path (`.github/scripts/...`) lost its leading dot and resolved to nothing.
- The `doc-refs-ignore` opt-out marker is honored by all three checks now, not only check C. Prose that illustrates the reference syntax needs it as much as a migration note does.

Checker tests go 17 → 41. Guidance updated in `CLAUDE.md` ("Referencing docs and sections") and `.github/scripts/doc-refs/README.md`.


dpmills:

- #47 ◀

src/ccl/design/README.md
```diff
@@ -31,7 +31,8 @@ CHL source
 | [optimization.md](optimization.md) | The optimization/compilation passes: inlining, lambda elimination, join/aggregate planning, algebraic simplification, and conversion to tile operators. |
 
 The feed-channelization step (`ccl/channelize.rs`) has no dedicated design doc: its design of
-record is [mutability.md](mutability.md) §4 ("Retire `desugar_defers`"), and the in-depth
+record is [mutability.md](mutability.md), [Compilation pipeline](mutability.md#compilation-pipeline) — feed
```
groundlar:
```suggestion
record is [Compilation pipeline](mutability.md#compilation-pipeline) in `mutability.md`— feed
```

.github/scripts/doc-refs/test_check_doc_refs.py
groundlar, line 1: **Test structure: two of these test a copy of the checker, not the checker**

`TestDocProseCitations.cited()` re-implements the loop body of `check_doc_citations` (`_MD_REF.finditer` → `_citation_tail` → `_cited_titles`), and `test_ignore_marker_skips_a_link` re-implements the ignore-marker filter from `check_markdown_links`. Both can stay green while the real function is broken, and neither notices if the real loop changes.

Look at what the copies leave out, since that's what's now uncovered: the `.md`-leniency rule (`if rel.suffix == ".md" and not titles and not dangling: continue`) — the subtlest decision in this PR, the one that cost 14 false hits to get right — plus check B's ignore-marker path and the whole resolve/report branch. No test calls `check_markdown_links`, `check_doc_source_paths`, or `check_doc_citations` at all, so the new "unterminated quoted citation" diagnostic is only ever exercised as a boolean out of `_cited_titles`.

That's also why the inputs read as opaque: `self.titles(', "Application\n shape"')` only makes sense if you already know "tail" means *the slice after a `.md` ref, in a comment run with `//` markers stripped*. (The class docstring tries to say so and names `check_code_refs`, which doesn't exist.)

The cause is one line: `REPO_ROOT` is a module global used at 12 sites, so `Repo` and all three checks are untestable by construction, and poking privates is the only surface available. Giving `Repo` a root (`def __init__(self, root: Path = REPO_ROOT)`, with the 12 uses taken from `self.root`) makes fixtures possible — the checks already take `repo`:

```python
def check(self, files):  # {relpath: contents} -> [problem details]
    root = Path(self.tmpdir)
    for p, text in files.items(): ...write...
    problems = []
    c.check_doc_citations(c.Repo(root), problems)
    return [p.detail for p in problems]

def test_wrapped_citation_is_checked(self):
    self.assertEqual(self.check({
        "d.md": "# Application shape\n",
        "a.rs": '/// see d.md, "Application\n/// shape".\n',
    }), [])

def test_wrapped_stale_citation_is_caught(self):   # the bug this PR fixes
    problems = self.check({
        "d.md": "# Something else\n",
        "a.rs": '/// see d.md, "Application\n/// shape".\n',
    })
    self.assertIn('cited section "Application shape" not found', problems[0])
```

That reads as *a repo and its verdict*, covers the real path including the line number in `Problem.source`, and — the part that matters — a fixture version of the first test **fails on the parent commit**. None of the 24 new tests do, at the check level.

Worth keeping as unit tests: the places where the logic really is a standalone grammar — `_slug_base`, `_SRC_PATH`, and the `_cited_titles` glue/run rules (`test_rule_quote_is_not_a_title` earns its keep). I'd drop the two re-implementing classes in favor of fixtures.

Smaller things while you're in here:

- `_cited_titles` returns a bare `(list, bool)`, so every helper indexes `[0]`/`[1]` and `TestCitationTitles` calls it twice per assertion. A `NamedTuple("Citations", [("titles", ...), ("dangling", ...)])` would read better here *and* at `check_doc_refs.py:410`.
- `TestRustCommentScanner.spans()` no longer returns spans.
- `" ".join(t.split())` is open-coded in three helpers — that's `_normalize_title` minus the backtick strip.
- Multi-line inputs written as `'\n '` escapes hide the wrap that's under test; `textwrap.dedent` with a triple-quote shows it.
- `TestRepoResolution` builds a real `Repo()`, which `rglob`s the whole tree (`target/` included — `SKIP_DIRS` filters after the walk rather than pruning it). Harmless today, but with a root parameter it becomes a tmp-tree test instead of an integration test in disguise.

groundlar (approved):

.github/scripts/doc-refs/test_check_doc_refs.py
dpmills: Agreed on all of it — fixed in 16b0239e.

You identified the cause correctly: `REPO_ROOT` as a module global is what made `Repo` and the three checks untestable, and testing copies of the loops was me working around that instead of fixing it. `Repo` now takes a root (`def __init__(self, root: Path = REPO_ROOT)`) and the checks read through it (`repo.read(rel)`, `repo.root`), so fixtures work exactly as you sketched.

The two re-implementing classes are gone. Tests now build a `{path: contents}` tree, run the real `check_markdown_links` / `check_doc_source_paths` / `check_doc_citations`, and assert on `"{source}: {detail}"` so line numbers are covered too. What that newly covers is what you predicted was uncovered:

- the `.md`-leniency rule, from both sides — `test_bare_doc_name_in_prose_is_not_a_citation` puts two files named `lowering.md` in the tree so an ambiguity error *would* fire if prose were treated as a citation, and `test_quoting_a_title_demands_an_unambiguous_path` shows the same mention becoming an error once a title is quoted;
- check B's ignore-marker path (`TestOptOut`, with a companion asserting all three checks fire without the marker);
- the resolve/report branch, including the dotfile-path case, which is now a tmp-tree test rather than the integration test in disguise it was;
- the unterminated-citation diagnostic, at the check level rather than as a boolean.

On "none of the 24 tests fail on the parent commit" — that was the part worth taking seriously, so I checked rather than assumed. Substituting the parent's per-line comment scanner back in, `test_wrapped_stale_citation_is_caught` and `test_wrapped_citation_is_checked_and_passes` both fail (and five scanner unit tests with them). Disabling doc-prose scanning and the link-wrapping regex fails six more, across `TestDocProseCitations`, `TestWrappedLinks`, and `TestOptOut`. So each capability this PR adds now has at least one test that fails without it.

One correction to the plan, from writing it: a title that can't close because a blank comment line intervenes reports as *unterminated*, not as clean. I had written that fixture asserting clean, and the failure was the test being wrong rather than the code — the blank line ends the thought, and an unpairable quote is exactly the case the diagnostic exists for. The test now asserts the diagnostic, which documents the behavior better than my original phrasing of it did.

Kept as unit tests, per your list: `_slug_base`, `heading_anchors`, `_SRC_PATH`, the citation run (`TestCitationRun` — `test_rule_quote_is_not_a_title` and the wrap-bound cases), and the comment scanner. Their docstrings now say what `tail` is (the slice after a `.md` ref, markers already stripped) instead of naming a function that no longer exists.

The smaller items are all in:

- `_cited_titles` returns a `Citations` NamedTuple; the call site reads `cites.titles` / `cites.dangling`.
- `spans()` → `comments()`.
- `_one_line` replaces the open-coded `" ".join(x.split())` — it was in four places, counting the one in `check_doc_citations`; `_normalize_title` is now `_one_line` plus the backtick strip.
- Fixture inputs are triple-quoted and `textwrap.dedent`-ed, so the wrap under test is visible in the source. I left the short `tail` fragments in `TestCitationRun` as escapes: there the leading space on the continuation line is significant (it stands for the stripped `///`), and dedent hides it.
- `Repo` prunes `SKIP_DIRS` during the walk now instead of filtering after — good catch, `target/` was being traversed in full on every run.

src/ccl/design/README.md
dpmills: Applied — mine linked the same file twice, which is worse than naming the section once. Taken with one change: your suggestion has `` `mutability.md`— feed `` with no space before the em dash, so I kept the space.

PR #48, opened 2026-07-30
Positional tuple access `t.0`; `[…]` is lookup, not projection

dpmills:
`[0]` was the only way to reach a tuple positionally, and lowering resolved `t[0]` to a projection by *guessing* from the index's shape — a guess it has no types to make. That guess made `xs[0]`, the commonest subscript anyone writes, mean tuple projection and then fail with a structural mismatch. Positional access now has its own spelling alongside the named one, and `[…]` means collection lookup with no case on the index.

The grammar accepts an integer after `.`, held in the existing `Attribute { attr }` slot. `ProjKey` already models "keyed by name or by position", so lowering is the single place the two spellings resolve, and *leading digit* is the whole discriminator since an identifier cannot begin with one. `.0` lexes as `Dot` then `Int` — there is no float token to swallow it — so `t.0.1` is two projections rather than a decimal. A subscript lowers to the application it *is* (`c[k]` → `c(k)`), so a product, having no domain, is no longer reachable through it: `t[0]` is rejected with a hint naming `.0`.

## Diagnostics

A positional projection's requirement is a **dense** tuple, because `Type::Tuple` cannot express "has position `n`" except as "is `n + 1` wide". Two failures were consequently reported against a shape the program never had:

- `t.99` on a 3-tuple said `missing field Index(3)` — the first *absent* position, which is just the value's own width restated. The subtyping edge now reports the **widest** position the requirement demands, which for a projection is the only one genuinely demanded, and the message states it against the found width: `No position .99 for Apply: (1, 2, 3) is a tuple with 3 positions`.
- `r.0` mismatched `{a: 1}` against `(?31)` — the partial requirement, rendered as if it were a type. A `Record`-vs-`Tuple` mismatch is *always* a keying confusion, so it now carries that as a hint, as does its mirror `t.b`.

`MissingField` becomes its own `InferError` variant because there is no second *type* to report: flattening it into a `TypeMismatch` required inventing one (`type_b: Type::Hole`, rendered `_`). Its dual `ExtraTag` has the same shape and is deliberately left alone — that is variant-tag territory, not this change's — with a comment recording it.

## Coverage

Every tuple subscript in the test suite migrates to `.n`; no corpus program in `tests/programs/` used one. Pinned: both keyings project and compose to any depth (`r.p.1`, `t.0.a`, `t.0.1`); `[…]` never projects, literal index or not; `.n` recovers a shape through an inferred parameter and is rejected at the call site for a non-product; the brace type forms `{T, U}` and `{name: T}` project by their own keying; `.-1` is a parse error naming both spellings; and both diagnostics say what the shape actually has.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -324,7 +324,7 @@ noted:
 | 12 | `+` `-` | additive |
 | 13 | `*` `//` | multiplicative |
 | 14 | unary `-` | prefix |
-| 15 | postfix: `f(args)`, `x[i]`, `x.attr` | left-fold |
+| 15 | postfix: `f(args)`, `x[i]`, `x.attr`, `x.0` | left-fold |
 | 16 | atom | literal, name, `(...)`, `[...]`, `{...}`, comprehension |
 
 ```ebnf
@@ -764,7 +764,8 @@ error.
 
 ```
 target[index]    -- subscript: list element or map lookup
-target.attr      -- attribute: record field access
+target.attr      -- attribute: record field by name
+target.0         -- attribute: tuple field by position
 ```
 
 - **Subscript** on a list with integer index `i` denotes the i-th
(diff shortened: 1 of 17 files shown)


dpmills:

- #48 ◀

sortalongo (approved):

PR #50, opened 2026-07-31
Monomorphization: key specializations on instantiation identity, not a resolved type

dpmills:
`specialize_use`'s memo compared two values produced by two different procedures: a candidate's key was `resolve_var_type(&use_expr.ty)` taken *before* its pin, while an entry's was `clone.ty` — the clone's type *after* the pin and a full re-entrant coalesce. Neither is a faithful identity for what a clone gets specialized to, and the mismatch broke the memo in both directions.

## False hits, i.e. a miscompile

A resolved type is a polarity-correct *rendering*: a domain resolves from upper bounds — what the definition body demands — so a position the body never touches is narrowed away, and an argument's refinement, which arrives as a *lower* bound of that domain, is invisible unless `compact_type`'s opposite-polarity fallback happens to fire at that exact position. The clone's interior reads its parameter at a *positive* position and sees precisely those refinements. So two uses differing only in a key-invisible position shared a clone whose interior was resolved against the first use's argument:

```
h = \a, b -> a + b
h(1, 2) + h(1, 5)
```

Both uses keyed on `((1, Int) ⇒ Int)` — element 0's refinement reached the key, element 1's did not — and the shared clone typed `.1` as `2`. The post-inline consistency wall then panicked with `Type mismatch for Apply: expected 5, found 2`. A difference in the *first* argument keyed apart and compiled fine, so the failure was confined to differences the rendering could not see.

## A write-only table

For any definition whose clone type gains a refinement across the pin, no candidate key could ever equal a stored one, so every call site minted its own clone — including character-identical ones — and the table accumulated several entries under one key. The comment justifying the stored key claimed a later same-typed use would resolve through the first pin's extended chains; it cannot, because every use instantiates its own fresh variables, which the first clone's pin never touches.

## `SpecKey`

`SpecKey` replaces the resolved type. It is the pair of **directed reads** of the use's instantiation, kept apart: the *positive* read is the stamping view (a domain from the definition's demands), the *negative* read is the clone's view (a domain from the argument that flowed in, a codomain from the consumer's demand). The negative read is the load-bearing half — it is the polarity the clone's interior reads its parameter at — and the pair is exhaustive, because use-specific information enters an instantiation through exactly two channels: an `arg <: domain` edge and a `codomain <: demand` edge.

Both reads stay *directed*, which matters: following both bound lists at every variable instead walks out of one use's cone and into every other (two calls' arguments meet at the shared variable of an operator's scheme), so every use keys on the whole program's literals and they all compare equal. Keeping the two views separate matters too — a merged view forgets which direction a contribution came from, and the pin the key stands in for is direction-sensitive. Within a read, merging is always union: a key that narrows can only under-split, and under-splitting is a miscompile while over-splitting is a wasted clone. An under-determined position is one canonical empty view rather than a fresh `Infer` placeholder, so two unexercised uses can now share.

Both sides of the comparison are now taken by one procedure at one point in the pin's lifecycle, and an entry stores the key of the use that minted it, with a `debug_assert` that no two entries share a key.

`poly_used_directly_and_through_wrapper_shares_specializations` drops from five specializations to four: the direct `f(1)` and the `f(y)` reached inside `g`'s `Int` clone instantiate `f` identically, so they now group onto one specialization — which is what the test's name and its "the memo is per frame, not per demanding region" comment always asked for.

Consequences carried along: `ReadPurpose::SpecializationKey` becomes `Instantiation` (that resolution no longer keys anything — it seeds the clone's channel-domain pairings and blames a resolution failure, so its refinement exclusion is now justified by *those* consumers being refinement-insensitive rather than by the false claim that a refinement the read could not see can only cause a miss); `HistoryKind` derives `Ord` so a kind can key a `BTreeMap`; and a `specialize_use` doc block that had detached from its function is reattached.

The remaining imprecision is documented on `SpecializeFrame::specs` and in the new "Keying a specialization" design section: the key summarizes the pin's *input*, so two uses differing only in a position the clone never reads still key apart. Closing that means keying on the pin's *output* — the finished clone, deduped structurally — which cannot be a lookup, only a build-then-dedupe. The two compose, so nothing here has to be undone to get there.

## Retiring "witness"

The second commit is a vocabulary sweep with no behaviour change. `design/type-inference.md` defined a *witness* as "the term for a refinement in its role as a black box to the subtyping lattice", and the word had spread through the solver, the design docs, and four identifiers.

The distinction does not earn a word. We already have `base` and `refinement`; "the refinement as the lattice sees it" is a fact about the lattice, not a second kind of object. Worse, the word names precisely a **limitation** — that the lattice accumulates refinements and matches them by identity instead of reasoning about what they imply — so having a term for it quietly asserts that the limitation is permanent.

So `witness` → `refinement` throughout (185 occurrences, 26 files), including `join_witnesses` → `join_refinements` and `type_carries_witness` → `type_carries_refinement` (`channelize`), `input_witnesses` → `input_refinements` (`coalesce`), the flag on `types_agree_modulo_unread`, and the `#### Refinements as witness sets` section, which becomes `#### Refinements on the lattice` with an opening sentence that states the opacity as a property of the lattice rather than as the definition of a term.

## Testing

Four end-to-end regression cases for the miscompile plus its controls (a difference in a key-visible position, and identical calls that must still share), two unit tests for the memo's two halves, and unit tests for `SpecKey` itself. Full `./ci.sh` passes.

## A function's kind is part of its identity

`KeyView.fun` is keyed by `KindMerge`, the way `KeyView.history` is keyed by `HistoryKind` and for the same reason: `𝐷 ⇒ 𝑉` and `𝐷 ⤇ 𝑉` are one shape that compiles to different code, so a clone pinned at `⤇` — whose body iterates the domain — must not serve a `⇒` use that never supplies one. A union at one position keeps the two apart rather than letting one shadow the other.

The kind is read through `KindMerge::of`, the same resolved-from-bounds view compaction takes, rather than off the `FunKind` itself. That distinction is the whole of it: an inferred kind is a *variable*, freshly minted per instantiation, so keying on its identity would give every use its own key and retire sharing altogether. Its bounds are the answer, and reading them mid-solve is the bargain every other field in the key already makes — the guarantee is agreement with the pin, not omniscience about the graph.


dpmills:

- #50 ◀
    - #51
        - #52
            - #60
                - #63
                    - #64
                        - #62

src/ccl/infer/solver/spec_key.rs
```diff
@@ -0,0 +1,784 @@
...
+    };
+    // One walk-wide `ctx` for both reads: its memo is keyed by polarity, so the
+    // two reads share it without contaminating each other.
+    SpecKey {
+        positive: key_go(ty, true, &Subst::id(), &mut ctx),
+        negative: key_go(ty, false, &Subst::id(), &mut ctx),
+    }
+}
+
+fn key_go(ty: &Type, pol: bool, subst_acc: &Subst, ctx: &mut KeyCtx) -> KeyView {
```
groundlar, line 343: `key_go` is now a second parallel walk over `Type` alongside `compact_go`: the same polarity flip on the `Fun` domain, the same no-flip on `History` children, the same `Subst::then(&b.render_subst(), subst_acc)` composition at bound edges, the same `subst_acc.shadow(b)` for the Pi binder, and the same `(uid, pol)` cycle guard.

Their agreement is the whole soundness argument — the module doc's closing guarantee is that *"a bound the key cannot see is one the clone cannot see either… the clone's own resolution walks the same edges from the same side"*. But nothing enforces it. A new `Type` variant, or a change to how `compact_go` composes an edge substitution or flips polarity, has to be mirrored here by hand, and a divergence is **silent** — it produces exactly the class of miscompile this PR exists to fix.

Today the only cross-link is the one-line note on `AtomKey::from_type`, and it points one way.

Minimum ask for this PR: a reciprocal comment on `compact_go` naming `key_go` as the sibling that must move with it, so the coupling is discoverable from both ends.

Worth considering as a follow-up: factor the polarity + accumulated-substitution + cycle-guard skeleton into one traversal parameterized by the per-position accumulator, so the two views become two accumulators over one walk rather than two walks that have to be kept in step by discipline.

src/ccl/infer/solver/compact.rs
```diff
@@ -355,19 +358,19 @@ impl CompactType {
...
-    /// Merge two witness sets. The set-op tracks
+    /// Merge two refinement sets. The set-op tracks
     /// polarity the same way `rec` does — positive ⇒ *intersect*,
     /// negative ⇒ *union* — because refinement-sets width-subtype like
     /// record fields (more refinements ⇒ subtype). At a positive
-    /// position the value reliably carries only the witnesses *both*
+    /// position the value reliably carries only the refinements *both*
     /// sides guarantee; at a negative position a consumer that may
     /// impose either set imposes their union.
     fn merge_refinements(pol: bool, lhs: Vec<Refinement>, rhs: Vec<Refinement>) -> Vec<Refinement> {
```
groundlar, line 368: Follow-up bug request, **not** a defect in this PR — the parent commit fails identically, so nothing here caused it. Raising it on this line because the mechanism points at `merge_refinements`, and because this PR's own framing makes the shape far more reachable.

```
f = \a -> a
f(1) + f(1)
```

panics with:

```
Inference created invalid expr:
[Type mismatch for type of 1 ▷ __mono + 1 ▷ __mono: expected Int, found 1]
```

Characterization from a probe set run against both this commit and `dacbbc2`:

| program | result |
|---|---|
| `f = \a -> a` ; `f(1) + f(1)` | **panic** |
| `f = \a -> a` ; `f(1) - f(1)` | **panic** |
| `f = \a -> a` ; `f(1) * f(1)` | **panic** |
| `f = \a -> a` ; `f(1) + f(2)` | ok |
| `f = \a -> a` ; `2 * f(1) + f(1)` | ok |
| `x = 1` ; `x + x` | ok |
| `def g(a): a` ; `g(1) + g(1)` | **panic** |

The split lines up with the positive arm here: **equal** singletons survive the intersection and get stamped on the binop node, **differing** ones cancel to `Int`. So the arithmetic node inherits its operands' singleton exactly when both operands are the same generalized-binding use at the same literal — which is why `f(1) + f(2)` and `2 * f(1) + f(1)` are fine and `f(1) - f(1)` is not, and why the non-generalized `x = 1; x + x` is unaffected.

Worth its own issue rather than scope creep here. Flagging it on this PR because the stated rule going forward is one specialization per distinct argument tuple with every literal carrying its own singleton — singleton-typed UDF results become the normal case, so this shape will start showing up in real programs.

src/ccl/mut_elim.rs
```diff
@@ -1004,7 +1004,7 @@ pub(crate) fn register_value_tys<'a>(
 /// hand-built test trees do. The written type then stands in, **stripped** — a
 /// register takes no refinement from any single contribution, and an unstripped
 /// one would be a witness acquired by erasure rather than by `cast`
```
groundlar: Sweep miss: the sentence above this line still says *"a witness acquired by erasure rather than by `cast`"* — the retired sense, in the same breath as the citation that was updated.

There is one other survivor, and it is the interesting one: `src/ccl/design/type-inference.md`, in the Σ-type material — *"whose witness 𝑤 is the runtime branch discriminant"*. That is the standard mathematical sense of the word and is **correctly kept**, but a future sweeper (human or grep) will find it and "fix" it. Worth a one-line note there that the retirement covers the *refinement* sense only, so the exception is recorded rather than rediscovered.

src/ccl/infer/solver/spec_key.rs
```diff
@@ -0,0 +1,784 @@
...
+        Type::Base(_)
+        | Type::UIntRange(_)
+        | Type::DataSource(_)
+        | Type::ChanDom(..)
+        | Type::Txn => KeyView::from_atom(
+            AtomKey::from_type(ty).expect("every atomic type classifies as an AtomKey"),
+        ),
+        // A `Hole` contributes nothing — the same "no information here" the
+        // default key denotes.
+        Type::Hole => KeyView::default(),
```
groundlar, line 360: The arm says what it does but not why it can't collide. A `Hole` and a genuinely under-determined `Infer` now key identically, so two uses can share on the strength of it.

That looks right to me — `normalize_annotation` turns a `Hole` into a fresh `Type::Infer` at emit, so a `Hole` should not survive into a use's instantiation type in the first place, and if one did, "no information here" is the honest reading either way. But that is two inferential steps the reader has to take unaided, right next to a decision about sharing. One clause naming the emit-time erasure would close it.

src/ccl/infer/solve.rs
```diff
@@ -438,48 +445,70 @@ struct SpecializeFrame {
     /// The binding's polymorphism level — the freshen cutoff: variables
     /// deeper than this are the quantified ones.
     cutoff: Level,
-    /// Specializations minted so far. Scanned linearly, comparing a candidate use's
-    /// resolved type against each entry's [`Specialization::use_ty`] by `PartialEq`
-    /// — structural, with refinement witnesses comparing by type-blind predicate
-    /// equality — so same-typed uses share one specialization.
+    /// Specializations minted so far, scanned linearly (a binding has a handful).
```
groundlar: Worth qualifying the parenthetical now that it is no longer quite true. The stated rule is one clone per distinct *argument tuple*, and every literal carries a singleton — so for a collection-producing UDF (the ones `inline` leaves cached) `specs` grows with **call sites**, not with distinct types.

The scan is linear in `specs` and each comparison is a deep structural `KeyView` walk, so that is quadratic in call-site count for exactly the definitions this PR expects to clone most.

Not a blocker at current scale, and I would not change the data structure here. But "a binding has a handful" is the assumption the linear scan rests on, and it is the assumption that stops holding first — better to say what makes it hold (and what would break it) than to leave the reader to rediscover it.

groundlar (commented):
Reviewed against the design docs, and verified the fix differentially rather than by inspection: built `dacbbc2` in an isolated clone and ran the same 12-program probe set against both commits. One real miscompile fixed (`h = \a, b -> a + b` at `h(1,2) + h(1,2) + h(1,5)`), zero regressions in the differential, full `./ci.sh` green locally.

The design is right, and the two-directed-reads-kept-apart argument — plus the "why not saturate" paragraph, which is the part most people would get wrong — is worth more than the code. Comments below are one structural concern, one follow-up bug that is *not* yours, and three small documentation asks.

groundlar:
### Context from comparing this against other monomorphizers

Not a blocker — this is the right fix. Two things I turned up that seem worth recording in the design doc, plus one concrete suggestion.

**1. The resolved-type key was a fossil of the pre-`dce4285` architecture.** That commit removed a separate post-coalesce `monomorphize` pass, whose own doc comment read:

> "This is classic monomorphization, **deferred until *after* types are known so it can key specialization on the resolved type** — which is what lets one definition be shared across same-typed uses."

Post-coalesce, a resolved type is all you *can* see: `expr.ty` has been overwritten and the arena is gone. `dce4285` moved specialization into the walk and made var-laden instantiation types available at the point of keying — but the key stayed a resolved type. This PR retires it. Worth a sentence in [Keying a specialization](https://github.com/cambra-dev/cambra/blob/dmills/mono-spec-key/src/ccl/design/type-inference.md#keying-a-specialization), because it makes `SpecKey` an argument *for* in-walk specialization rather than a cost of it — the key is only expressible because monomorphization runs where the graph is still live.

**2. The key is already a function of the post-emission graph.** I ran `emit_node` alone on `let h = λp → p.0 + p.1 in (h(1,2), h(1,N))` — no coalescing, no pin — and computed `spec_key` on both uses:

| | `h(1,2)` vs `h(1,5)` | `h(1,2)` vs `h(1,2)` |
|---|---|---|
| `resolve_var_type` | `((1, Int) ⇒ Int)` — **equal** | equal |
| `spec_key` | **differ** | equal |

So the discriminating information is in the graph the instant emission finishes, and the resolved type collides exactly as the commit message describes. The pin transports it across polarity; it doesn't discover it.

**That suggests a cheap test**, and it closes the verification gap I raised on `SpecializeFrame::specs`: assert that `spec_key` taken right after emission equals `spec_key` taken at specialization time. That turns "both sides computed by one procedure at one point in the pin's lifecycle" from a discipline into a checked property, and it would fail loudly if a future change made the key sensitive to intervening pins.

**3. Where Cambra sits relative to everyone else** — the framing I'd suggest for that design-doc paragraph. rustc keys on `Instance { def, args }`, frozen by typeck; C++ on the template-argument list; Swift on a substitution map; MLton on the type-argument list in a single post-inference pass. All of them key on a *finished value*, because their generic bodies are already typed and instantiation is substitution.

Cambra has no `Type::ForAll` and never types a generic body at all ("The definition's own subtree is never coalesced in place"), so monomorphization here is *the act of typing the body*, not duplication of already-typed code. That puts us in C++'s category — the one mainstream compiler where instantiation genuinely re-runs semantic analysis rather than substituting — and it is the real reason the key must be read off a live graph. A reader arriving from rustc will otherwise assume that's an implementation choice.

For completeness on what is *not* the reason: poly-calls-poly doesn't block a phase split. Discovering the *set* of instantiations is a reachability fixpoint in every monomorphizer (rustc's collector, Swift's pass-manager worklist), and the pre-`dce4285` pass handled it by recursing into each new specialization. The cost of deferral is only that coalesce destructively overwrites `expr.ty` — which is why that pass also needed `rederive_dependent_types`.

**Prior art worth a look:** [Lutze, Schuster & Brachthäuser, *The Simple Essence of Monomorphization* (OOPSLA 2025)](https://se.informatik.uni-tuebingen.de/publications/lutze25simple.pdf) — monomorphization via a flow analysis explicitly inspired by algebraic subtyping (Dolan & Mycroft), i.e. this type system. Closest prior art I found, including on where monomorphization stops being possible (polymorphic recursion, which they detect as cyclic flow).

groundlar (approved):

src/ccl/infer/solver/spec_key.rs
dpmills: Added the comments highlighting the linkage

src/ccl/infer/solver/compact.rs
dpmills: This is what #51 is for.  Since that's already out for review, I'm not going to file an issue for it

src/ccl/mut_elim.rs
dpmills: Done

src/ccl/infer/solver/spec_key.rs
dpmills: Clarified

src/ccl/infer/solve.rs
dpmills: Done

dpmills:
> ### Context from comparing this against other monomorphizers
> Not a blocker — this is the right fix. Two things I turned up that seem worth recording in the design doc, plus one concrete suggestion.
> 
> **1. The resolved-type key was a fossil of the pre-`dce4285` architecture.** That commit removed a separate post-coalesce `monomorphize` pass, whose own doc comment read:
> 
> > "This is classic monomorphization, **deferred until _after_ types are known so it can key specialization on the resolved type** — which is what lets one definition be shared across same-typed uses."
> 
> Post-coalesce, a resolved type is all you _can_ see: `expr.ty` has been overwritten and the arena is gone. `dce4285` moved specialization into the walk and made var-laden instantiation types available at the point of keying — but the key stayed a resolved type. This PR retires it. Worth a sentence in [Keying a specialization](https://github.com/cambra-dev/cambra/blob/dmills/mono-spec-key/src/ccl/design/type-inference.md#keying-a-specialization), because it makes `SpecKey` an argument _for_ in-walk specialization rather than a cost of it — the key is only expressible because monomorphization runs where the graph is still live.

Added

> **2. The key is already a function of the post-emission graph.** I ran `emit_node` alone on `let h = λp → p.0 + p.1 in (h(1,2), h(1,N))` — no coalescing, no pin — and computed `spec_key` on both uses:
> 
> `h(1,2)` vs `h(1,5)`	`h(1,2)` vs `h(1,2)`
> `resolve_var_type`	`((1, Int) ⇒ Int)` — **equal**	equal
> `spec_key`	**differ**	equal
> So the discriminating information is in the graph the instant emission finishes, and the resolved type collides exactly as the commit message describes. The pin transports it across polarity; it doesn't discover it.
> 
> **That suggests a cheap test**, and it closes the verification gap I raised on `SpecializeFrame::specs`: assert that `spec_key` taken right after emission equals `spec_key` taken at specialization time. That turns "both sides computed by one procedure at one point in the pin's lifecycle" from a discipline into a checked property, and it would fail loudly if a future change made the key sensitive to intervening pins.

Once we have #51, this isn't entirely true.  I added a section discussing

> **3. Where Cambra sits relative to everyone else** — the framing I'd suggest for that design-doc paragraph. rustc keys on `Instance { def, args }`, frozen by typeck; C++ on the template-argument list; Swift on a substitution map; MLton on the type-argument list in a single post-inference pass. All of them key on a _finished value_, because their generic bodies are already typed and instantiation is substitution.
> 
> Cambra has no `Type::ForAll` and never types a generic body at all ("The definition's own subtree is never coalesced in place"), so monomorphization here is _the act of typing the body_, not duplication of already-typed code. That puts us in C++'s category — the one mainstream compiler where instantiation genuinely re-runs semantic analysis rather than substituting — and it is the real reason the key must be read off a live graph. A reader arriving from rustc will otherwise assume that's an implementation choice.
> 
> For completeness on what is _not_ the reason: poly-calls-poly doesn't block a phase split. Discovering the _set_ of instantiations is a reachability fixpoint in every monomorphizer (rustc's collector, Swift's pass-manager worklist), and the pre-`dce4285` pass handled it by recursing into each new specialization. The cost of deferral is only that coalesce destructively overwrites `expr.ty` — which is why that pass also needed `rederive_dependent_types`.
> 
> **Prior art worth a look:** [Lutze, Schuster & Brachthäuser, _The Simple Essence of Monomorphization_ (OOPSLA 2025)](https://se.informatik.uni-tuebingen.de/publications/lutze25simple.pdf) — monomorphization via a flow analysis explicitly inspired by algebraic subtyping (Dolan & Mycroft), i.e. this type system. Closest prior art I found, including on where monomorphization stops being possible (polymorphic recursion, which they detect as cyclic flow).

Added this discussion to the design

PR #51, opened 2026-07-31
Type functions: state a relationship between type variables other than subtyping

dpmills:
The constraint solver relates type variables in exactly one way: subtyping. `α <: β` is the only fact it can record. That is enough whenever a rule's result type *is* one of its input types — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is expressible as sharing a lattice position.

It is not enough for a rule whose result is **computed from** its inputs. Addition is the smallest example: the result of `a + b` is neither operand.

## What went wrong without it

The encoding available was to share a variable: `∀α. α → α → α`. A variable carries every lattice dimension, so this states far more than intended, and it is wrong in both directions at once.

| Program | Inferred | Should be |
|---|---|---|
| `\x -> x + 1` | `{Int \| __elem == 1} ⇒ Int` | polymorphic in `x` |
| `y = 1; z = 1; y + z` | `{Int \| __elem == 1}` | `Int` |
| `1 > "a"` | `Bool`, then panics in the interpreter | rejected |

The first two are the two polarities of one mistake. Operand occurrences are *domains* — negative positions, where refinement sets **union** — so one operand's refinement becomes a requirement on the other. The result occurrence is positive, where refinement sets **intersect**, so a refinement both operands carry survives onto the result. The second row needs two operands carrying the *same* refinement (intersecting a set with itself returns it), which is why it stayed hidden.

## The mechanism

`Type::App { fun: TypeFn, args }` is the type "whatever `fun` computes from `args`" — `1 + x` types as `Add({Int | __elem == 1}, α)`.

An `App` **denotes** a type rather than constructing one: `Add(Int, Int)` *is* `Int`, written in a form that does not yet know it. It adds no inhabitants and no subtyping edges to the lattice, so reduction is *normalization*, and it is transient in the same sense as `Infer` — every `Type` that escapes inference is `App`-free.

Reduction is **demand-driven**: materializing an `App` resolves each argument through the ordinary pipeline and applies the rule. Nothing is deposited, so no phase ordering can make the answer wrong. A rule is a pure function of resolved argument types — no inference context, no access to the bound graph — which is also what leaves room for user-declared type schemas on UDFs later.

Four laws make a rule sound: pure, monotone in the subtype order, **declares a `CycleTolerance`**, and normalizing. They are stated in `src/ccl/infer/solver/reduce.rs` with what each buys inference, and summarized in `src/ccl/design/type-inference.md`, "4.7 Type functions".

### A cycle is a recurrence in the program, and a rule says whether it can answer through one

An argument arrives cyclic when resolving it would re-enter the resolution already computing it, which is what a self-read is: `x += 1` makes `value(x) = join(seed, Add(value(x), 1))`, a fixpoint equation, because an accumulator *is* one. Measured: 1,910 cyclic arguments across the suite, and adding a self-read is exactly what creates them.

So `Option<Type>` is `Arg::Known(_) | Arg::Cyclic` — the distinction a rule must reason about is not "absent" but *cyclic*, and there is no later point at which more is known. `TypeFn::cycle_tolerance()` declares `Any` or `AllKnown` and `reduce` enforces it centrally, so a rule that cannot answer cannot forget to say so. It is deliberately **not** per-position: every rule's condition is about *how many* arguments are cyclic rather than which.

Both of today's rules are `Any`, for reasons worth keeping apart — a comparison is constant on its domain so a cyclic operand costs nothing, while arithmetic loses the agreement check but keeps a usable type, which is what lets an accumulator have one.

### The cut is an unrolling, not an iteration

Traced, one materialization of `x := 0; x := x + 1; x` reduces `Add(⟨cyclic⟩, 1) → Int` and then `Add(Int, 1) → Int`, and stops; `x + x` reaches three levels because each operand re-enters separately. The depth is fixed by the shape of the program, never by an iteration count, and in every shape measured the outermost frame — the one whose answer becomes the node's type — sees every argument known.

The consequence is the opposite of the familiar one. Because nothing iterates, whatever a rule answers at a cut is **what the unrolling carries**; there is no later pass that widens it. A rule must already be sound at the cut, not merely improvable. Both current rules are, for free: arithmetic drops to the shared base — the top of the chain the missing operand could have contributed to — and comparison is constant on its domain. A sharper rule that kept what it could see would return whatever a bounded unrolling happened to reach, sound only by accident of how far the program unrolled.

## The operand requirement lives in the result rule

An operator's operands are two **unrelated** scheme variables. Nothing relates them, and nothing needs to: what checks them is the rule that reduces the result.

That works because of one property, which is silent when violated and so is recorded where schemes are defined: **a scheme that wants its operands checked must mention them in its result.** A rule runs when something materializes the application it belongs to, and an application is materialized only when some node's type reaches it. A scheme's operand variables are nobody's node type — a use records `left <: α` and `right <: β`, so `α` and `β` are reachable *from* the operands, but nothing walks *from* them. The only thing that can reach them is the result.

So `1 + "a"` is rejected because the addition's result is `Add(α, β)`; materializing the node runs the rule and finds the conflict. A comparison declaring a bare `Bool` mentions neither operand, so nothing ever materializes them, the rule never runs, and `1 > "a"` types cleanly and then panics. A comparison's result is therefore `Compare(kind, α, β)`, reducing to `Bool` for operands that share a base and to an error for those that do not.

### What that costs

`\x -> x + 1` no longer infers `x : Int`. Its result still resolves — the rule answers from the literal operand — but the parameter is determined by nothing.

That is the honest answer rather than a regression. The lambda works for any two things `+` accepts, so it is polymorphic in its parameter, and `Type` has no `∀` to say so. As a program value it is an **ambiguous program**, rejected downstream exactly as `\x -> [x, x]` already was. Inferring `Int` read as precision; it was a single-numeric-type lattice showing through, and it would have become wrong the moment `Int + Float → Float` existed.

An earlier revision of this branch bought that inference back with a self-referential bound on the scheme's own variables (`α <: CommonBase(α, β)`), so that resolving one operand pulled the other's base. It is gone: redundant as a guard, wrong as a claim about what makes two values addable, and — as a pull — compensating for two defects elsewhere that are fixed below it. `src/ccl/design/type-inference.md`, "Why not a bound on the operand variables" records the argument so it is not re-added.

## Obligations that are not decidable yet are parked

`constrain_go` cannot compare an unreduced `App` against anything: reduction resolves the arguments off the bound graph, and emission runs while that graph is still being built, so an answer read there could be superseded by the next edge.

The obligation is therefore **parked** rather than dropped. `constrain_go` records it with the in-flight substitutions applied, `require_sub` tags it with the node to blame, and `InferCtx::check_parked_obligations` retries it between emission and coalesce — the point at which every edge the program implies exists. Only fully-determined obligations are checked, which keeps the retry from perturbing the graph it reads: a side still holding a variable is one the program never determined, and coalesce reports that as `UnresolvedInfer`.

This is what makes `(1 + 2) and True` and `sum(["a" + "b"])` ordinary diagnostics. Accepting the obligation silently let inference return `Ok`, and the conflict then reached the `check_pre_desugar` consistency wall, which cannot distinguish a user type error from a compiler bug and panics on both.

`InferCtx::constrain_and_claim` is the single choke point every emission-time `constrain_subtype` goes through, with a `debug_assert` that the solver's park list is empty on the way out — a call that bypassed it would silently drop its obligation, which is the defect parking exists to fix.

## The audit of the other operators

Every scheme in the registry was checked against the rule that decides it: **sharing a lattice position between an input and the result is right exactly when the operator *selects* an existing value or *merges* several** — the result then *is* one of those values, so a fact about it survives — and wrong when the operator computes a new one.

`Sum`, `Neg`, `Not`, `Concat` and `BoolLogic` inherit nothing; `Max` shares its element type and should; `final_or_default` and the `get_prev_*` accessors merge; `List` merges across elements; `Proj` selects the field, refinement included. Tests pin the table, and half their cases assert a refinement is *still present*, because dropping it is the same bug from the other side.

`CollectionUnion` is why type functions **coexist** with variable sharing rather than replacing it: its codomain is a real join while its domain is a real computation, and no single scheme can say both. It and `Proj`'s `FieldOf(ρ, k)` are the natural next clients.

## Two inference defects this exposed

Both were found by asking why the retired bound appeared to be load-bearing, and both are fixed in their own commits.

**`groupby` never related its key domain to its key function.** The lowered shape binds `λ __gb_k` whose only occurrence is an operand of the partition predicate's `==`, so the key's type could only arrive backwards along an operator's operand relation — the wrong place for that fact to come from, and a place this PR deliberately removes. `Type::SharedHole(id)` lets lowering state the relation directly: a `Hole` with an identity, where every occurrence of one id normalizes to the same inference variable.

It is carried by the key application and by the **domain of the group-by's own `data_fun` annotation** — the two places the claim is actually about. "A partition function's domain is the type of its keys" is a fact about the group-by; `__gb_k` is a name this lowering invented, and a different encoding of the same construct would not have it. Annotating the domain also makes the edge directional with no new machinery: `bind_annotation` records `inferred <: ann` and a function type is contravariant in its domain, so it reduces to `key_ty <: ⟨the parameter⟩` — produced keys flow *into* the domain, rather than the two being forced equal.

One caveat worth knowing: `symbolic` does not render annotations, so `test_lower_groupby`'s snapshots do not show this relation. `test_groupby_key_type_comes_from_the_key_function` is what covers it.

**Freshening asked `type_level` a question it does not answer.** `type_level` is predicate-blind by design (it drives extrusion, where refinements are lattice-blind), but `freshen_above` used it to decide whether a type needs copying. A quantified variable reachable only through a *nested* refinement's predicate therefore short-circuited, and the specialization clone kept pointing at the definition's own variables — visible as a duplicate specialization, since the clone reached one generalized `let` through both the freshened variable and the original. Freshening now asks `freshen_level`, which counts predicates.

## The strict wall could not see through an application

Making a result an `App` surfaced a hole this area already had: `typecheck` could not catch a corrupted `1 + 2 : String` either, and no test noticed because the one test there used a comparison, whose result was still concrete.

Check reconciles a rule-derived type against the recorded one. For an operator scheme that type is a fresh variable whose lower bound is the `App`, and `constrain_subtype` treats an unreduced application as opaque, so the bound-closure walk stops there. Check now *resolves* a rule-derived type that is still a bare variable before reconciling. Narrow on purpose: every other rule rebuilds a ground type from its children, where resolving is the identity.

## Errors

A reduction conflict has its own error rather than borrowing `IncompatibleBounds`, whose rendering is a specific claim ("won't infer an untagged sum from a collision") that is not what happened: each operand is well typed and nothing collided on a variable. Compaction returns a bag of contributions rather than a `Result`, so a failed reduction rides the position it poisoned (`CompactType::reduce_error`) and `coalesce_compact` raises it where a `Type` is produced — which is a node, with a span.

## Prior work

The name is deliberately GHC's. Associated type synonyms (Chakravarty, Keller & Peyton Jones, ICFP 2005) and open type functions (Schrijvers, Peyton Jones, Chakravarty & Sulzmann, ICFP 2008) are the closest analogue; the substantive difference is that GHC's families are *open*, so the checker must decide entailment between unreduced applications and the rule set needs confluence and termination side conditions. Cambra's set is closed and every rule is normalizing, so there is never an equality to decide between two unreduced applications — reduce, then compare.

Reducing before comparing is the **conversion** discipline of a dependent type checker (Coquand, *An algorithm for type-checking dependent types*, 1996), restricted to a closed first-order rule set. Parking a constraint whose reduction is blocked is Agda's **constraint postponement** (Norell, Chalmers, 2007) at coarser granularity — retried once after emission rather than woken per-solution, which is sound because emission is a single bounded phase.

The lattice underneath is unchanged: MLsub (Dolan & Mycroft, POPL 2017) as presented by Parreaux (ICFP 2020). Because reduction is normalization, a type function adds nothing to it.

Alternatives considered — including named predicate constraints (`(SameBase α β) ⇒ …`, deferred rather than rejected) — are tabulated in the design doc.


dpmills:

- #50
    - #51 ◀
        - #52
            - #60
                - #63
                    - #64
                        - #62

groundlar:
## Review notes: Critical + Important findings

> **Caveat:** these were identified by Claude (`/staff-rust-review`) and have **not** been manually validated by me — Dan is the primary reviewer here. Each finding below was reproduced against a build of `356826f2`, and the parent-commit comparisons come from building `oxsknwns` in a scratch tree, but treat the analysis as a lead to check rather than a verdict. `./ci.sh` is green on this commit.

The design itself reads well and I'd keep it as is — putting the computation in the type is the right call, and "an operand requirement must be reachable from the result type" is a real solver insight, correctly generalized into the scheme registry's docs rather than patched at one call site. I confirmed the headline win: on the parent, `1 > "a"` **compiles clean**; on this commit it is rejected.

I also stress-tested the memoization independently — disabling the cache lookup outright and re-running `type_check` + `compilation_pipeline` gives 533 passed / 0 failed with identical results, so the memo appears behaviour-neutral on the suite.

---

### Critical — an ordinary ill-typed program now panics instead of producing a diagnostic

The new arm in `constrain.rs`

```rust
(Type::Compute { .. }, _) | (_, Type::Compute { .. }) => Ok(()),
```

drops `Add(α, β) <: Bool`. Before this PR the arithmetic result was a plain variable, so bound closure produced `Int <: Bool` → a real `ConstrainError::Mismatch` during inference. Now inference returns `Ok`, and the conflict resurfaces at `context.rs:840`, which panics on anything that is not `UnresolvedInfer`.

Measured, same program, both trees:

| `(1 + 2) and True` | result |
|---|---|
| parent (`oxsknwns`) | `Err([Infer { error: Type mismatch for BinOp: expected Int, found Bool, span: Some(0..16) }])` |
| this commit | `panicked: Inference created invalid expr: [Type mismatch for BinOp: ...]` |

Same for `sum(["a" + "b"])`. A user type error became a compiler panic that blames the compiler.

The commit message and the `constrain.rs` comment both state that the only traffic reaching the arm is the harmless operand-requirement direction — *"across the whole library suite the only obligations reaching it are `Int <: CommonBase(α, β)` and `String <: CommonBase(α, β)`"* and *"`check_node_rule` resolves such a variable before reconciling rather than relying on this arm; nothing else does, **which is why the shape no longer reaches here**."* Instrumenting the arm with an `eprintln!` and running the pristine suite refutes this — 634 obligations in the unchecked direction:

```
   184  Add(?, ?)     <: CommonBase(?, ?)
   154  Equals(?, ?)  <: Bool
    73  Add(?, ?)     <: Int
    52  Greater(?, ?) <: Bool
    40  Sub(?, ?)     <: Int
    37  Less(?, ?)    <: Bool
    35  Mul(?, ?)     <: CommonBase(?, ?)
    …
```

The real fix is the parked-obligation store the comment already names. Short of that, `context.rs`'s wall should not panic on a `TypeMismatch` at a node whose rule involved an operator. Either way the claim in the commit message and the code comment wants correcting — it is the load-bearing justification for the arm.

Illustrative test (fails on this commit, passes on the parent):

```rust
// An operator result flowing into a concrete demand is an ordinary type error, so
// it must be a diagnostic from `infer` — not a panic at the compiler-bug wall.
#[rstest]
#[case::bool_logic("(1 + 2) and True")]
#[case::aggregate(r#"sum(["a" + "b"])"#)]
fn test_operator_result_against_a_concrete_demand_is_a_diagnostic(#[case] code: &str) {
    assert!(!infer_program_err(code).is_empty());
}
```

---

### Important — `spec_key`'s `same_compute` under-splits, which its own doc says it cannot

`KeyView::compute`'s doc: *"this over-splits (a wasted clone) and never under-splits (a clone shared with a use it is wrong for)."* `same_compute` is a length guard plus a subset test:

```rust
a.len() == b.len() && a.iter().all(|(op, args)| b.iter().any(|(o, s)| o == op && same_args(op, args, s)))
```

That is not multiset equality when `a` holds two entries that are `same_args`-equal but structurally distinct — precisely the swapped-argument shape `union` cannot dedupe, because `union` dedupes with derived `==` while `same_args` compares as a multiset for a symmetric operator. Both entries then match the *same* element of `b`, and a genuinely different second entry in `b` is never noticed. Reproduced as a unit test in `spec_key.rs`:

```rust
#[test]
fn same_compute_under_splits() {
    let cb = |a, b| Type::Compute { op: TypeOp::CommonBase, args: vec![a, b] };
    let a = fresh_var(0);
    push_lower(&a, cb(int(), singleton(1)));
    push_lower(&a, cb(singleton(1), int()));   // same multiset, swapped
    let b = fresh_var(0);
    push_lower(&b, cb(int(), singleton(1)));
    push_lower(&b, cb(int(), singleton(2)));   // genuinely different
    assert_ne!(spec_key(&a), spec_key(&b));    // FAILS: keys compare equal
}
```

Whether the exact shape arises end-to-end today I did not establish — but the documented invariant is violated, and the fix is local: make `same_compute` a true multiset match (the same `taken`-vector trick `same_args` already uses), or dedupe in `union` with `same_args` instead of `contains`.

---

### Important — `ARGS_IN_FLIGHT` is not unwind-safe, while `REACHED` is

`ResolutionScope::drop` explicitly clears `REACHED`, with the comment *"leaving a partial one behind (an unwind, say) would misattribute the next chain's reach sets."* `ARGS_IN_FLIGHT` has the identical property and is cleared only by the straight-line `remove` in `resolve_argument`. A panic anywhere inside `compact_type` / `coalesce_compact` (`coalesce.rs:204/372/407`, a `RefCell` double-borrow, stack exhaustion) leaves a uid marked in-flight for the thread's lifetime — after which every reduction touching it silently coarsens. The failure mode is wrong types, not a crash. A six-line RAII guard fixes it, and it's the same pattern already used one function up.

Note this composes with the Critical finding: that one *adds* a reachable panic path through the compiler.

Related: the `debug_assert_eq!` in `ResolutionScope::drop` fires *during* unwinding if a panic followed a bound write, turning a diagnosable panic into an abort. Worth guarding with `!std::thread::panicking()`.

---

### Important — two comments describe the abandoned first design

`ResolutionScope`'s doc still says the memo is per-compaction — *"One compaction is the largest window that cannot contain one… So the memo is cleared on the way out of the outermost call"* — while `reconcile_cache`, ten lines below, says *"Scoping it to one compaction instead is what the first version did, and it costs a factor."* The two contradict each other, and the code does the second. Likewise `compact_type`'s call-site comment: *"dropped on unwind too, so a compaction that panics cannot leave a stale memo behind"* — the memo is generation-scoped and survives; only `REACHED` is dropped.

---

Nits (argument-guard gap for the named compound-argument next clients, `binop`'s `.expect` on the scheme-map lookup, the unreachable `[0,2] + [0,2]` example in §4.6) are minor and I'll leave them out of here unless useful.

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1046,74 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
 
 ---
 
+## 4.6 Type operators
```
sortalongo: I'd prefer to call these "type functions" (if these are well-defined, computable things) or "type families" (if we want them to be open, extensible, or inspectable). Although it's overloaded, "operator" tends to have a specific meaning (e.g. infix operator, dataflow operator) depending on context. This would be coining a new term for an idea that already exists.

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1046,74 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+*Sharing a variable* to mean "these are related" drags the whole lattice along with the base. Arithmetic's `∀α. α → α → α` demonstrated both halves, one per polarity:
+
+* the two operand occurrences are *domains* — negative positions, where refinement sets **union** — so one operand's refinement became a *requirement* on the other, and `λ 𝑥 → 𝑥 + 1` inferred `𝑥 : {Int | __elem == 1}`;
+* the result occurrence is positive, where refinement sets **intersect**, so a refinement *both* operands carried survived onto the result — `𝑥 + 𝑥` where `𝑥` is `2` claiming the sum is `2`. (Distinct refinements intersect to none, which is why this stayed hidden: it needs two operands carrying the *same* refinement, where intersecting a set with itself returns it.)
+
+*Reconstructing the result afterwards* needs a walk order that respects data flow, and the coalesce walk deliberately does not have one: `coalesce_node`'s `Apply` arm visits function before argument (for [monomorphization's ordering invariant](#keying-a-specialization)), so a projection's codomain is read before an operand nested in the argument has been coalesced.
```
sortalongo: Remove this, it's all "this is how stuff used to be" content that shouldn't be added to our docs.

src/ccl/design/type-inference.md
sortalongo, line 1: I think all of these additions would benefit substantially from a rewrite. They bounce around between ideas, don't define terms, and is super verbose and hard to skim. It reads like an essay written in obscure jargon (which I'm not sure exists anywhere beyond Claude's context), not documentation.

Can you rewrite it to have a nice, predictable structure, with matter-of-fact, well-organized content? State the motivation, sketch the solution, then go into details only where helpful and necessary, deferring to the code otherwise.

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1046,74 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+### The audit of the other operators
+
+Every other scheme in the registry was checked against the same distinction, and they are correct as they stand. The rule that decides it: **sharing a lattice position between an input and the result is right exactly when the operator *selects* an existing value or *merges* several** — the result then *is* one of those values, so a fact about it survives — and wrong when the operator *computes* a new one. `Compare` is the one that shares only across its *operands*, so it had the first symptom and not the second, and takes the same fix.
+
+| Operator | Shares? | Verdict |
+|---|---|---|
+| `Sum` | result is concrete `Int` | computes, inherits nothing ✔ |
```
sortalongo: It's not necessarily an Int. That's just a simplifying assumption we're currently making. Does that cause any problems?

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1046,74 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+| `Neg`, `Not`, `Concat`, `BoolLogic` | monomorphic | nothing to inherit ✔ |
+| `Max` | element type with the result | *selects* an element — `max([1, 1])` really is `1` ✔ |
+| `final_or_default`, `get_prev_seq`, `get_prev_txn` | value type across stream, default, result | *merges* — the result is whichever the runtime supplies, so the refinement set intersects, which is the join rule ✔ |
+| `List` | element type across elements | *merges* — `[1, 1]` is a collection of `1`s, `[1, 2]` of `Int`s ✔ |
+| `CollectionUnion` | codomain across operands | *merges* on the codomain; its **domain** is a computation (`Variant({𝑖: dom_𝑖})`) and is hand-rolled in `emit_node` for exactly that reason ✔ |
+| `Proj` | codomain with the projected field | *selects* the field, refinement included ✔ |
```
sortalongo, line 1138: Question for the future: should field projection out of a refined record try to bring some projection of the outer refinement along with it?

```
def get_one(x: {{i: Int, l: List(Int)} where _.i < len(_.l)}) => Int:
  return x.l[x.i]
```

Does `x.i` get a refinement here so the unchecked lookup can happen? I think it can, but some additional machinery is needed.

tests/type_check.rs
```diff
@@ -794,6 +795,202 @@ fn test_lambda_unapplied(#[case] code: &str, #[case] expected: Type) {
     assert_eq!(infer_program(code), expected);
 }
 
+// Nested generic arithmetic that never resolves must still **finish**.
+//
+// An unapplied generic function leaves its operand types unresolved, which is an
+// ambiguous program and rejected downstream — but rejected, not hung on. This is
+// the shape that made inference not terminate: every operand of every arithmetic
```
sortalongo: > the shape that made inference not terminate

This is a vague reference to something readers have no clue about. Explain or remove.

src/ccl/ty.rs
```diff
@@ -425,10 +426,133 @@ pub enum Type {
...
 
+/// A type-level operator: the computation a [`Type::Compute`] is waiting to run.
+///
+/// The set is closed and lives in the compiler. Keeping the operator as *data*,
+/// and its reduction a pure function of resolved argument types, is what leaves
+/// room for user-declared type schemas on UDFs later — a user-defined operator
+/// becomes another variant whose rule is looked up rather than matched.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub enum TypeOp {
+    /// The single base type all arguments share, with every value-level claim
```
sortalongo: I think "base" means the same thing as "supertype" here? We don't need multiple terms for the same idea. This is just the LUB, right?

src/ccl/infer/solver/reduce.rs
```diff
@@ -0,0 +1,198 @@
...
+        // the result whether or not the operands have resolved, which is what keeps
+        // an unapplied `λ x → x > 1` at `Int ⇒ Bool`.
+        TypeOp::Compare(_) => {
+            common_base(op, &available)?;
+            Ok(Type::Base(crate::ccl::BaseType::Bool))
+        }
+    }
+}
+
+/// The single base every argument shares, with all value-level claims dropped.
```
sortalongo: What does common base refer to formally? Is it really just "drop the refinements of a type"? I'd like to understand what these "operators" are more abstractly. What properties must they preserve? Which guarantees do they preserve for inference? T

sortalongo (commented):
The description of this PR is off target and obscure. The motivation is quite simple: we need a way to specify relationships between type variables other than subtyping. Give some examples.

I also think this needs a more rigorous foundation before I think it's safe to make such a big change in the type checker. The design doc is an even more inscrutable version of the PR description, so I really don't have a good idea of what this change is. Can you develop it into a more organized, rigorous development that cites relevant research (specifically, how to handle function applications in dependent types, and how that interacts with type inference) and compares alternatives?

PR #53, opened 2026-07-31
Variant arms are keyed by tag, not by position

dpmills:
A union column's arms were keyed by **position** — `VariantWrap`/`VariantProject` carried a `tag: usize`, "the resolved 0-based position of the constructed tag in the full union". A position is not a stable identity for a tag: coalesce materializes a variant's tags in name order, so `[.b] <: [.a, .b]` renumbers `b` from 0 to 1. Because an arm set is part of a union column's layout, subsumption then needs a renumbering coercion, and nothing inserts one.

`ColumnValue::Union`, `Extent::Union` and `Predicate::Union` were built for *anonymous positional* sums (`FieldKey::Index`), where the position genuinely is the identity — fan-out domains, `a ++ b`. Named tags arriving on top of that representation is where it stopped being adequate.

So arms are keyed by [`FieldKey`] throughout, via a `TagMap<T>` — a tag-keyed sequence in canonical (name) order, with the key-set operations the union/extent/predicate paths kept hand-rolling: `same_tags`, `zip_matching`, `get_or_insert_with`. Runtime variant width subtyping is now *stated* in one place, `Extent::includes`, matched by tag rather than by position, and a projection of a tag the scrutinee does not carry yields an **empty** projection rather than an error — which is what leaves a width-subtype scrutinee well-formed.

## The row→payload partition

`ColumnValue::Union` stored per-arm payload vectors but discarded which row each payload belonged to, and five call sites rebuilt that correspondence: `index_at` by an O(n) scan (making element access linear and so iteration quadratic), `select_indices`, `retain` and `drain_to_value_iter` by running counters, and `VariantProject` by a per-tag walk.

A union column *is* a partition of the row space, and the row→payload correspondence is the whole content of that partition, so `UnionArm { rows, values }` stores it. All five reconstructions go away and `index_at` becomes a binary search per arm.

## Also

`UnionOperator::new_flat`'s disjointness precondition is now asserted rather than assumed — a flat merge is correct only if the arms' runtime domains are disjoint, and merging one position twice is a documented bug.

`design-operators.md`'s `VariantWrap`/`VariantProject` rows described the positional representation; they now describe the tag-keyed one, and say why a position cannot key an arm.

## A merged stream declares the merged tag set

Keying the arms by tag is also what lets a *merged* variant stream describe itself honestly, and one declaration had been left behind at the positional form.

A conditional whose arms carry *different* tags — `.pos(n) if n > 0 else .neg(n)`, or a `match` whose arms construct different tags — panicked at compile time:

    ExtractLast default tiling must match source codomain tiling
      left: Scalar((.neg(Int)))  right: Scalar((.0((.pos(Int))) | .1((.neg(Int)))))

The tile the union operator *builds* was right all along. Its codomain column is the tag-merged sum, `[.neg(rows: []) | .pos(rows: [0], 5)]` — both tags present, the arm that did not occur empty. Only the **declared tiling** was wrong: `UnionOperator::new` kept the arms' codomains apart as an anonymous positional sum whenever they were not literally equal, describing a value shaped like "arm 0's variant *or* arm 1's variant", which is not what any row holds. A tile-conformance check caught the mismatch, and `ExtractLast` — whose default is one of those arms, so width-narrower than the whole — hit it first.

So the fix is to declare what the operator already produces. A union's **domain** keeps every arm apart, because which arm a row came from is exactly what `final_or_default` dispatches on. Its **codomain** does the opposite: the arms are alternative values at one row, so it is their **join**. `Extent::join` is that operation, and for two variants it merges tag maps — a tag they share joining its payloads, a tag only one carries taken as-is. Where the arms have no join the positional sum stays, which is right for a concatenation: there the arm a row came from *is* part of a row's identity.

`ExtractLast` then needs two things. Its `default` must be *representable* in the output value space rather than identical to it — `Extent::includes` already states variant width subtyping at the runtime boundary, so that is the check. And every emission is built at the **declared** extent via `ColumnValue::from_values`, not from the extracted value alone: a variant value carries only its own tag, so a column built from it is narrower than the tiling whenever the collapsed alternatives carry more tags between them. Building at the declared extent keeps every arm present with the ones that did not occur empty, which is the shape a downstream merge or append requires.

That second half is what also fixes a **variant-valued mutable register**, which failed in every spelling — same-tag writes as `Mismatched ColumnValue variants in append`, a `.none` seed with `.some` writes as the same tiling mismatch, and a conditional write likewise. A register's seed and its writes are alternatives in exactly the same sense as a conditional's arms.

Tests cover the join directly (disjoint tags merge; a shared tag joins its payloads; a conflicting shared payload fails the whole join rather than silently picking a side; unrelated scalars have no join), what `UnionOperator::new` declares for each of those cases, and both of `ExtractLast`'s emission paths — the extracted value and the default — asserting the emitted column carries every tag with only the occurring one inhabited.


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14
                    - #15
                        - #53 ◀
                            - #54
                            - #16
                                - #17

dpmills:
Squashed into #15 — the tag-keyed representation is now how that PR introduces variant arms in the first place, rather than a correction applied on top of a positional one. Nothing is dropped: #15's tree is byte-identical to this branch's, plus the review fixes described in its updated body.

Closing in favour of #15.

PR #54, opened 2026-07-31
Add CHL variant support: constructors, `match`, and variant types

dpmills:
CHL's IR has carried tagged sums since variant elimination landed, with no way for a program to write one. This adds the surface — variant construction, a `match` block statement, and a written-down variant type — and then removes three places, all found by exercising it, where the runtime re-derived a type fact it should have been handed. Programs can now build and dispatch on their own sums, and `Extent` no longer runs a second, weaker copy of the type lattice.

Surface: [chl-spec.md §3.15](docs/chl-spec.md#315-variant-constructors) (terms), [§4.10](docs/chl-spec.md#410-match--tag-dispatch) (`match`), [§6.5](docs/chl-spec.md#65-variants) (types). Mechanism: [lowering.md, "Variants and match"](src/ccl/design/lowering.md#variants-and-match).

A backtick introduces a tag in **every** position — `` `some(1) ``, `` case `some(v) ``, ``{`some{Int} | `none}``. Tags are structural and undeclared, so nothing else could tell `some(1)` from a call to a function named `some`.

`match` lowers to the `Case { scrutinee: Some(_), branches }` that variant elimination already compiles, one `Branch` per arm with a `Lit(true)` guard — which makes tag dispatch and the guarded `if`/`elif` chain the *same* first-match rule over one node. A `match` whose only arm is `case _:` names no tag, so it lowers to that arm's body.

**A tag's bracket is the payload's bracket, elided**, so neither level needs a bracket rule of its own. The type side *requires* the elision (`` `a{{a: Int}} `` is rejected) to keep one spelling per type; the term side cannot, since `( e )` is also grouping.

The one genuinely new concept, and the field to push on. A `match` with a `case _:` needs each arm's binder to receive the scrutinee's payload *while* the scrutinee stays free to carry tags no arm names — on the tag axis the scrutinee is the supertype, on the payload axis its payload is the subtype, and one closed edge cannot point both ways. An open arm set drops the missing-tag rejection, which *is* the exhaustiveness check, and keeps the per-tag recursion.

Openness belongs to a demand, never a value, so `Open` appears only on the right of an edge. It rides through compaction (`CompactVariant`) rather than being flattened there, because a diagnostic naming that demand is resolved through the same round-trip. Details: [type-inference.md, "The unified tagged sum"](src/ccl/design/type-inference.md#the-unified-tagged-sum).

Three sites, one shape: a runtime lattice recomputing what inference concluded, in a form that cannot represent the answer.

- A **union's codomain** is now read off its node's type. `Extent::join` has no record rule, so a conditional at two record widths compiled to an anonymous positional sum where the type layer said `{a: Int}`. It had no other caller and is deleted with its eight tests.
- A **list literal's** element extent comes from the list's declared type. `Extent::for_value` drops a `Value::Union`'s tag and guesses a function table's domain from its first binding; it had no callers and is deleted.
- A **flat merge's** domain is the one extent its arms share, asserted rather than reconciled. `dedup_extents` never removed anything, and its fallback built a domain `flat_merge` would have panicked on.

`UnionProducer`'s row-oriented `ColumnValue::Variants` merge goes with them: the solver refuses to infer an untagged sum from a collision, so differing arms are always two `ColumnValue::Union`s, which concatenate.

`tests/compilation_pipeline/variants.rs` covers construction, every bracket spelling, `match` with and without default arms, payload-misuse rejections, variant-valued mutable registers, and mixed-tag list literals.

Two failures are pinned with their causes: two call sites of a `match`-on-parameter UDF panic at the post-inference wall, from **order-dependent specialization**, a `main` defect this surface only *detects*; and a conditional at two record widths now compiles but cannot merge, `UnionProducer` having no record-width narrowing. `match` inside a `for` body is a stated surface limit. The three north-star programs that pinned "the backtick is not lexed" are repinned at their next blocker.


dpmills:

- #57
    - #86
        - #54 ◀
            - #99
                - #100

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+match x:
+    case `some(v):
+        v + 1
+    case `none:
+        0
+```
+
+A pattern spells its tag exactly as a constructor does (§3.15), so an arm
+reads as the inverse of what it matches. `` case `tag(binder): `` binds the
+payload; `` case `tag: `` matches without binding, which is the natural
```
sortalongo: > `` case `tag: `` matches without binding

I'm skeptical that we should allow this for non-unit-typed tags.

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+ordinary local, scoped to its arm. The default arm below takes no backtick:
+`_` is the absence of a tag, not a tag.
+
+Like `if`, a `match` is value-yielding by position: where a value is
+required (a function body, the program value), every arm's block must end
+in a value-yielding statement.
+
+**Each tag is handled by exactly one arm.** Two arms for one tag is an
+error — the arms *partition* the scrutinee's tags, so first-match never
+arbitrates and arm order is not observable.
```
sortalongo: What about guards? Mark that as a tentative direction.

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+type-checked as one.
+
+**Exhaustiveness needs no separate check when there is no default arm.**
+The arms describe the consumer, so the scrutinee is constrained to a
+*subtype* of the arms' tag set (width subtyping, §3.15). A `match`
+therefore cannot be non-exhaustive with respect to its own arms; it fails
+only when the scrutinee's type is fixed independently — by an annotation,
+or by a compiler-produced type such as the `Option(V)` of a checked
+lookup — and carries a tag no arm handles. That is reported as an
+unhandled tag.
```
sortalongo: This bold sentence is misleading. Exhaustiveness _does_ need a check: for whether all of an inferred variant's tags are handled. That check may happen as part of type checking, but it's still there.

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+
+With a default arm that constraint is *dropped*: the scrutinee may carry
+tags no arm names, which is exactly what the default is for. (Keeping the
+subtype requirement would make the default arm unreachable — every
+scrutinee it could fire on would be rejected first.)
```
```suggestion
```

redundant cruft

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+
+The converse is *not* an error: arms for tags the scrutinee cannot carry
+are simply unreachable, and are *kept* — a projection of a tag the value
+never carries contributes nothing, so a dead arm costs nothing and needs
+no pruning. Writing the full `Option` match over a scrutinee that
+inference has pinned to one tag is ordinary code, not a mistake.
```
sortalongo: > a dead arm costs nothing and needs no pruning

Is that true? It smells false.

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+payload, with no nesting, no literal patterns, and no per-arm guard.
+
+A `match` is not yet accepted **inside a `for`-loop body**, which admits
+only assignments, `<<` feeds, `yield`, `if` guards, `with begin():` blocks
+and bare calls. Dispatching per element is written as a `def` that matches
+on its parameter and is called from the loop or a comprehension, which is
+the same first-match rule reached through a call.
+
+> **Direction [Tentative].** `match` is a statement only. A one-line
+> expression form has no settled syntax: the obvious spellings collide
```
sortalongo: Statements are not values. But above, this doc says matches are "value yielding".

There's a difference between "match is a multi-line expression" and "match is a statement". So which is it? I think it's the former, and that makes this a minor directional note.

docs/chl-spec.md
```diff
@@ -1493,6 +1566,89 @@ statement or an inert mistake is decidable from the callee's type.
...
+> with decided syntax — `{…}` is structural-*type* syntax (§2.4) and
+> `a -> b` is pair sugar (§2.4) — so it is deliberately left out rather
+> than given a form that would have to be retracted. Tuple
```
sortalongo: IMO, the obvious spelling is just removing the line breaks:

```
match scrut: case `foo(x): x   case `bar(y): to_x(y)
```

CLAUDE.md
sortalongo, line 1: separate PR?

tests/compilation_pipeline/variants.rs
sortalongo, line 1: probably worth adding some tests with the `<:` annotation too

sortalongo (approved):

docs/chl-spec.md
dpmills: Agreed, fixed.  (_) is now mandatory for non-unit

docs/chl-spec.md

docs/chl-spec.md
dpmills: Removed the whole block; not spec material

docs/chl-spec.md
dpmills: Removed

docs/chl-spec.md
dpmills: Clarified

docs/chl-spec.md
dpmills: Put that in as the candidate, although I'm not 100% sold.  Inline : feels wrong

CLAUDE.md
dpmills: Split

tests/compilation_pipeline/variants.rs
dpmills: Done

docs/chl-spec.md
dpmills: False and irrelevant; removed

PR #55, opened 2026-08-01
Stop copying the bound graph to read it

dpmills:
Materializing a type walks a variable's bounds, and to do that it took a copy of
them: `compact_go`'s variable arm cloned both bound lists and the `parents` set at
every visit. A `Bound` owns a whole `Type` and a `Subst`, so the list copy is deep,
and an inference-heavy program makes millions of visits. A profile of one puts
`memmove` at 8%, `malloc`/`free` at ~4.5%, and `BTreeSet`/`BTreeMap` clone, drop and
search after that, with a long tail of the same — no hotspot, just the walk paying
for a copy of what it is reading.

Neither copy is needed.

**The bound lists are shared, not copied.** `InferBounds` holds `Rc<Vec<Bound>>` in
private fields, so a reader takes the list away from the `RefCell` with a refcount
bump and every write has to go through `lower_mut` / `upper_mut` / `set_lower` /
`clear` — the copy-on-write discipline is compiler-checked, not conventional. A
writer that is itself holding a snapshot does fork the list, and one does:
`constrain_go`'s transitive closure walks a snapshot while the recursion under it
records edges on the same variable. That stays cheap on the ratio, not on never
happening — a fork costs one copy per *push*, so forks are bounded by edges
recorded, where the read side was copying per *visit*, and visits outnumber edges
by orders of magnitude. A fresh variable still costs no allocation: every empty
list is the same shared `Rc`, since a run mints hundreds of thousands of variables
and most never take a bound at all — and teardown hands them back to it rather than
allocating a copy in order to clear one.

**`parents` is a path, so it is a chain of stack frames.** It holds one entry per
variable on the current bound chain, is reset at every structural boundary, and is
only ever asked whether a uid is on it — a `BTreeSet` was answering that by
allocating a copy of itself per visit, for a path a handful of entries deep.
Borrowing the caller's frame costs nothing, and the type now says what the thing is.

## Measured

Best of three, on chains of generic wrappers applied at a concrete type:

| program | before | after |
|---|---|---|
| 2 wrappers | 286 ms | 178 ms |
| 3 wrappers | 2.31 s | 1.31 s |
| 4 wrappers | 12.1 s | 7.84 s |

Roughly 1.6×. Small programs are unchanged, as expected — a three-millisecond
comprehension never reaches the walk. The win is proportional to how much
bound-graph traversal a program does, which is the same axis every other inference
cost sits on.

A profile after the change confirms the mechanism rather than a coincidence: the
`BTreeSet<InferVarId>` and `BTreeMap<Name, Mapping>` clone/drop/search entries are
gone from it entirely. What surfaces behind them is free-variable counting
(`count_free_with_visited` and the type/predicate walks under it, together ~13%),
which is the next thing to look at and is unrelated to this.

Correctness is covered where it already was: the spurious-cycle pruning that
`parents` exists for is what makes `?a <: ?a` coalesce to a variable instead of
being reported as a recursive type, and the coalesce tests pin that.

### DIFF
--- a/src/ccl/infer/api.rs
+++ b/src/ccl/infer/api.rs
@@ -123,8 +123,8 @@ impl Drop for InferArena {
         // edges, so the (otherwise cyclic) refcounts can all reach zero.
         for var in crate::ccl::arena_exit() {
             let mut bounds = var.bounds.borrow_mut();
-            bounds.lower.clear();
-            bounds.upper.clear();
+            bounds.lower_mut().clear();
+            bounds.upper_mut().clear();
         }
     }
 }
--- a/src/ccl/infer/api.rs
+++ b/src/ccl/infer/api.rs
@@ -122,9 +122,7 @@ impl Drop for InferArena {
         // Take back every variable minted during the run and sever its bound
         // edges, so the (otherwise cyclic) refcounts can all reach zero.
         for var in crate::ccl::arena_exit() {
-            let mut bounds = var.bounds.borrow_mut();
-            bounds.lower_mut().clear();
-            bounds.upper_mut().clear();
+            var.bounds.borrow_mut().clear();
         }
     }
 }
--- a/src/ccl/infer/context.rs
+++ b/src/ccl/infer/context.rs
@@ -458,7 +458,7 @@ impl Typing for InferCtx {
         };
         v.bounds
             .borrow_mut()
-            .lower
+            .lower_mut()
             .push(crate::ccl::Bound::with_subst(
                 result,
                 crate::ccl::subst::Subst::discharge(&x, argument.clone()),
--- a/src/ccl/infer/emit.rs
+++ b/src/ccl/infer/emit.rs
@@ -1558,7 +1558,7 @@ mod review_tests {
                 panic!("fresh() yields a variable");
             };
             let b = v.bounds.borrow();
-            b.lower.len() + b.upper.len()
+            b.lower().len() + b.upper().len()
         };
         assert!(
             bound_count(&d1) > 0,
--- a/src/ccl/infer/solver/coalesce.rs
+++ b/src/ccl/infer/solver/coalesce.rs
@@ -729,7 +729,11 @@ mod tests {
         // the path is defensive.
         let v = fresh_var(0);
         if let Type::Infer(state) = &v {
-            state.bounds.borrow_mut().lower.push(Bound::conc(v.clone()));
+            state
+                .bounds
+                .borrow_mut()
+                .lower_mut()
+                .push(Bound::conc(v.clone()));
         }
         match coalesce_compact(&compact_type(&v)).unwrap() {
             Type::Infer(_) => {}
--- a/src/ccl/infer/solver/compact.rs
+++ b/src/ccl/infer/solver/compact.rs
@@ -7,6 +7,7 @@
 //! sibling [`mod@super::simplify_type`] and [`super::coalesce`] modules.
 
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
+use std::rc::Rc;
 
 use smol_str::SmolStr;
 
@@ -531,12 +532,13 @@ struct ParentPath<'a> {
 impl ParentPath<'_> {
     /// Whether `uid` is on the path. Linear in the path's depth, which is the
     /// length of one variable's bound chain.
-    fn contains(mut path: Option<&ParentPath<'_>>, uid: InferVarId) -> bool {
-        while let Some(frame) = path {
-            if frame.uid == uid {
+    fn contains(&self, uid: InferVarId) -> bool {
+        let mut frame = Some(self);
+        while let Some(f) = frame {
+            if f.uid == uid {
                 return true;
             }
-            path = frame.prev;
+            frame = f.prev;
         }
         false
     }
@@ -702,7 +704,7 @@ fn compact_go(
             let uid = state.uid;
             let key = (uid, pol);
             if st.in_process.contains(&key) {
-                if ParentPath::contains(parents, uid) {
+                if parents.is_some_and(|p| p.contains(uid)) {
                     // Spurious cycle (a <: b and b <: a with no
                     // structural intermediary). Drop the bound.
                     return CompactType::empty();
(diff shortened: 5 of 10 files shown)


dpmills:

- #55 ◀

groundlar:
Read through this against the bound-graph readers and writers, and ran a few probes. Both mechanisms look right to me — `ParentPath` naming the path is the kind of change where the type finally says what the thing is, and the COW boundary is on the rare side of a read/write ratio that is orders of magnitude. `./ci.sh fast` is green.

Two things I found worth flagging, plus a few smaller ones.

### Important — teardown allocates instead of releasing (`src/ccl/infer/api.rs:124`)

```rust
bounds.lower_mut().clear();
bounds.upper_mut().clear();
```

For a variable that never took a bound, `lower` *is* the shared `NO_BOUNDS`, so `Rc::make_mut` sees `strong_count > 1` and clones — allocating a fresh `Rc` box to hold an empty `Vec`, then clearing the copy. Twice per variable, over exactly the population the description calls out as dominant ("most of which never take a bound at all").

Measured with the `tests/scope_walk_allocations.rs` harness pattern — 1000 fresh variables in an arena, allocations counted inside `drop(arena)`:

```
severing 1000 empty bound lists allocated 2000 times
```

An `InferBounds::clear()` that does `*self = InferBounds::default()` takes that to 0 (I ran it; `./ci.sh fast` stays green). It's also strictly stronger than `make_mut` + `clear`, not just cheaper: the assignment drops the old `Rc`, which releases the `Bound`s and the cycle when uniquely held, whereas `make_mut` on a shared list clears a copy and severs nothing.

### Important — the COW justification comment is inaccurate (`src/ccl/infer_var.rs:196`)

> Mutation goes through `lower_mut`, which copies only if a reader is still holding the old list; **the passes that write bounds hold none.**

They do. `extrude` (`constrain.rs:1009`) snapshots both lists and then pushes to one of them:

```rust
let (lows, ups) = { let s = tv.bounds.borrow(); (s.lower.clone(), s.upper.clone()) };
if pol {
    tv.bounds.borrow_mut().upper_mut().push(...);   // `ups` is alive → guaranteed fork
```

That fork is unconditional. `constrain_go_impl` (`:645`, `:677`) is the conditional case — it pushes to one list while an outer frame may hold a snapshot of the same variable's other list, which the cycle-break cache narrows but doesn't eliminate.

Counting forks and `Bound`s copied in `lower_mut`/`upper_mut`, on chains of generic record wrappers:

| wrappers | write calls | forks | `Bound`s deep-copied |
|---|---|---|---|
| 2 | 205 | 95 | 22 |
| 3 | 326 | 149 | 56 |
| 4 | 507 | 219 | 182 |

The win is untouched by this — forks are bounded by *edges recorded* where the read side was copying per *visit*, and visits dominate. But as written the comment invites a later maintainer to add a reader-held write path believing it's free. What's true is narrower: a writer holding a snapshot forks once per push, bounded by edge count.

Incidentally, `extrude` can be made fork-free: each branch reads only the list it does *not* push to (positive pushes `upper` and seeds from `lows`; negative mirrors), so snapshotting inside the branches drops the count to zero.

### Nit — `Rc::clone` at the reader sites (`compact.rs:769`–`773`, `spec_key.rs:463`, `scheme.rs:280`, `constrain.rs:649`)

```rust
let primary_bounds = primary.clone();
```

This is the line whose meaning the PR changes — deep copy to refcount bump — and `.clone()` on an `Rc<Vec<Bound>>` reads identically to the old deep copy at a glance. The explicit form is already in use two hunks away (`Rc::clone(&nvs)`). `Rc::clone(if pol { &s.lower } else { &s.upper })` also collapses the `primary` binding.

### Nit — `ParentPath::contains` as an associated function

It takes the `Option` because the empty path has no frame. A method plus `is_some_and` keeps the receiver where a reader expects it:

```rust
impl ParentPath<'_> {
    fn contains(&self, uid: InferVarId) -> bool { /* walk self.prev */ }
}
// caller:
if parents.is_some_and(|p| p.contains(uid)) {
```

### Nit — the COW discipline isn't type-enforced

`lower` / `upper` stay `pub` and four sites assign them directly (`scheme.rs:305`–`306`, `constrain.rs:1029`, `:1046`). Those are a real distinct operation — "install a freshly built list" — so the honest surface is `lower()` / `lower_mut()` / `set_lower(Vec<Bound>)` with the fields private, which makes "all mutation goes through COW" compiler-checked rather than conventional. `clear()` from the first finding falls out of the same surface.

### Not a finding, but worth claiming

`seed_pairings_go` (`scheme.rs:413`, `:425`) and `extrude_invariant` were cloning both lists per visit for pure reads and are now refcount bumps for free. Those hunks currently read as incidental.

groundlar (approved):

src/ccl/infer/api.rs
```diff
@@ -123,8 +123,8 @@ impl Drop for InferArena {
         // edges, so the (otherwise cyclic) refcounts can all reach zero.
         for var in crate::ccl::arena_exit() {
             let mut bounds = var.bounds.borrow_mut();
-            bounds.lower.clear();
-            bounds.upper.clear();
```
groundlar, line 127: blocking comment here, see top-level for findings

dpmills:
All five taken, in `83b00971`.

**Teardown.** `InferBounds::clear()` is `*self = Self::default()`, and the `Drop` is now one call. You're right that it's the stronger operation and not just the cheaper one — the point of teardown is to *drop* the `Rc`, and `make_mut` on a shared list clears a copy and severs nothing. Pinned by pointer identity rather than by an allocation count, which is deterministic and doesn't need a second global-allocator harness: `clearing_bounds_restores_the_shared_empty_list` asserts `Rc::ptr_eq` against `NO_BOUNDS`, which is exactly what a `make_mut` + `clear` fails (it installs a fresh `Rc`), plus one that the holder's refcount is gone and one that a fresh variable shares the empty list.

**The COW comment.** It was wrong, and `extrude` was the unconditional case. Fixed both ways: each branch now snapshots only the list it does *not* push to (positive seeds from `lower` and writes `upper`, negative mirrors), and the doc says what's actually true — a writer holding a snapshot forks once per push, which is bounded by edges recorded where the read side was copying per visit.

Counting forks and `Bound`s copied in `lower_mut`/`upper_mut` over the whole `compilation_pipeline` suite (451 inference runs, ~72k bound writes):

| | forks | `Bound`s deep-copied |
|---|---|---|
| before | 163 | 220 |
| after | 30 | 61 |

So `extrude` was 133 of the 163. What's left is the `constrain_go` case you identified, which is the one the doc now describes.

**Nits.** All three: reader sites say `Rc::clone` explicitly and the `primary` binding is collapsed into `Rc::clone(if pol { s.lower() } else { s.upper() })`; `ParentPath::contains` is a method called through `is_some_and`; and the fields are private behind `lower()` / `upper()` / `lower_mut()` / `upper_mut()` / `set_lower()` / `set_upper()` / `clear()`. `set_lower` is worth its own name for the reason you gave — "install a freshly built list" is a genuinely different operation from "append to the existing one", and the three sites that were assigning the field are now saying so.

`./ci.sh` and `DEEP_TYPECHECK=1 ./ci.sh test` are both green.

src/ccl/infer/api.rs
dpmills: Addressed in `83b00971` — `Drop` is a single `bounds.clear()` now, and `clear()` assigns the shared empty list rather than clearing through `make_mut`. Details and the fork measurements in the top-level reply.

PR #56, opened 2026-08-01
A renamed occurrence keeps its type, so nothing has to re-stamp predicates later

dpmills:
`Mapping::as_expr` materialized a `Rename` as `TypedExpr::var(to)`, and that
constructor leaves `ty` at `Type::Hole` — so substituting a rename produced an
*untyped* node. α-renaming cannot change a term's type: the type belongs to the
position, not to the name. The replacement now carries the type of the occurrence it
replaces, in both `as_expr` and `as_expr_preserving`.

Where the hole did damage is refinement predicates. Nothing types a predicate's
interior after inference — a predicate is a term riding a *type*, outside the walk that
resolves node types — so a hole introduced by a rename survived to the post-inference
wall as `Unresolved inference variable N`. `retype_predicate_slots` existed to sweep
those up, re-deriving each free `Var`'s type from the lexical scope of the final tree.

That pass is deleted. With the rename fixed there is nothing left for it to repair:

- it was a whole-tree walk (plus `retype_in_type`, two scope maps, and a `PredMemo`
  threaded through both) whose entire justification was four tests, all of the shape
  `[sum(x) for x in groupby(…)]`;
- it recovered types by **name lookup**, which is why it needed to know which binder
  kinds are "authoritative" over their uses — a distinction that existed only because
  the real type had been thrown away;
- and it ran on the *final* tree, so it also carried a tripwire that its rewrite never
  split predicate `Rc` sharing.

The symptom worth recording: one source predicate survived as six forced instances with
*different* typing states — two carrying a live variable, four a bare `Hole` — with
nothing to determine which reached the final tree. The two typed ones came through a
real `Discharge` (which clones a typed term); the four holes came through a `Rename`.

## The invariant is now asserted where it is owned

Every term a substitution fabricates goes through `as_expr`/`as_expr_preserving`, so
that is the one boundary that can state "a substitution never leaves a typed occurrence
holding an untyped replacement" — a hard `assert!`, O(1), release included, because the
alternative is silently producing a type nothing can reconstruct. An *untyped*
occurrence carries no obligation, which is what keeps pre-inference substitution
(lowering's uncurry template discharge, the chained-comparison operand freshens) legal.

The phase-boundary walls could not have caught this: the damage was created during
coalesce and repaired inside the same function, so it never reached a boundary. Removing
the repair is what let the existing wall report it — the check was never missing, its
input was laundered.

`type-inference.md`'s "Discharged-argument slot resolution" paragraph described the
defect as the design (a discharged argument's slots are "expected" to be placeholders,
re-stamped by lexical scope); it now states the invariant instead. Its claim of a
"residual under-determination" for `[sum([s.amount for s in g]) for g in groupby(…)]` is
also removed: that program reaches operator conversion and fails there on `curry`, so
inference and every post-inference check pass — it was never a typing error.

## A second instance, found by the assert itself

`channelize`'s defer-lift did the same thing in a different spelling: `try_lift_defer`
and the inner-defer splice both renamed `x → y` by *discharging a bare
`Expr::var(name)`* — a hand-built node, so `Type::Hole`. The assert caught both the
first time it ran. They are now real renames (`desugar_rename` over `Subst::rename`),
which carries the occurrence's type by construction and also keeps the mapping's
**species** honest: `subst` distinguishes rename from discharge deliberately
(`Subst::invert` and `split_renames` are exact only because of it), and spelling a
rename as a discharge of a variable threw that away too.


dpmills:

- #56 ◀

sortalongo (approved):
Nice! I remember being skeptical of `retype_predicate_slots`, but I must have dropped the investigation into its justification.

PR #57, opened 2026-08-03
An unobservable `match` arm's payload is pinned on the variable, not in the slot

dpmills:
A `match` arm naming a tag the scrutinee cannot carry gets no lower bound from width subtyping, so nothing determines that payload's type. The arm is ordinary code, but its variable survived coalesce as a `Type::Infer` and the post-inference wall rejected the program — the shape the most ordinary `Option` match produces. Inference now pins such a payload, recorded on the **variable** rather than in the binder slot so every occurrence agrees, including the enclosing lambda's parameter type where the residual variable actually surfaced. Mechanism and the two ordering constraints that put the pin inside the coalesce walk: [type-inference.md](src/ccl/design/type-inference.md#an-unobservable-arm-payload-defaults-to-unit).

## Which type it pins to

`Unit` when the body ignores its binder. When the body *reads* it, `Unit` contradicts that read: an operator states what it requires of its operands rather than equating them, so `v + 1` leaves the payload unbound while still demanding `Addable` of it. The requirements choose instead. `v + v` narrows them to no single row, and any will do since the arm is unreachable, so one is taken by table order. `pin_unobservable_arm_payload` carries the argument, including why bounded quantification is the real answer.

## Binder slots resolve their predicates

Independent, and reached by this pin's own tests. Resolving a binder slot settled the type's structure but not the refinement predicates riding it. The memo redirects only the occurrences it visits, so a skipped slot kept a pre-coalesce `Rc` full of unresolved variables. All five slots now share one `resolve_binder_slot` doing both halves. Only a *value* binding exposes it — a `groupby`, or a `match` over a conditionally-built collection — since a generalized binding is re-coalesced per specialization.

### DIFF
--- a/tests/inference_variants.rs
+++ b/tests/inference_variants.rs
@@ -7,11 +7,13 @@
 //! integration — without depending on CHL surface syntax for variants
 //! (which is a separate workstream).
 
+use cambra::ccl::ArithmeticKind;
 use cambra::ccl::{
     Branch, FieldKey, Lit, Pattern, Type, TypedBinding, TypedExpr, TypedExprNode,
     infer::{InferError, LocatedInferError, TypeInferenceContext, infer},
 };
 use cambra::interpreter::BaseType;
+use rstest::rstest;
 
 // ---------------------------------------------------------------------------
 // Helpers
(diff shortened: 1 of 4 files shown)


dpmills:

- #57 ◀
    - #86
        - #54
            - #99
                - #100

tests/inference_variants.rs
```diff
@@ -438,3 +433,104 @@ fn payload_mismatch_reject() {
...
+}
+
+/// ``let f = λ x → match x { `a(v) → v; `b(w) → 0 } in `a(1) ▷ f``.
+///
+/// The `b` arm is unreachable (the argument carries only `a`) *and* ignores its
+/// binder, so nothing in the program constrains its payload. Inference must
+/// still produce a fully concrete tree: the variable is pinned to `Unit` in the
+/// constraint graph, so the binder slot **and** the lambda's parameter type
+/// agree on it. Before the pin, `f`'s domain kept a bare `Infer` and the
+/// post-inference wall rejected this program.
```
sortalongo, line 474: This actually seems like a valid case for the empty type.  The type ``{`a(_) | `b(⊥)}``  means "\`b is an empty case", and is isomorphic to ``{`a(_)}``. I don't think it matters much, though. More of an interesting observation.

sortalongo (approved):
FWIW, I found this PR's title and description to be quite confusing as well. It would've been much clearer like

"Pins an unobservable `match` arm's payload's type to Unit"

and then leading with an example in the description.

One thing about Claude's writing style that's been driving me crazy is that it writes everything in the present tense. "payload is pinned" is ambiguous about whether it's talking about the old or new version of the code. If memory serves, PR descriptions often switch the meaning back and forth, which makes it really hard to disambiguate without diving into the PR first.

Using the imperative mood to indicate changes, past tense to indicate the prior state of the code, and timing clauses ("now does X" "after this change" etc) for the new code would make things much easier to follow.

tests/inference_variants.rs
dpmills: Hmm, I'll keep that in mind, yeah.  Once we start trying to prove stuff about variants and eliminating arms etc it might end up working better

PR #58, opened 2026-08-03
Producers keep the release contract

dpmills:
A `release` says the data in that region is never requested and never returned again, but nothing enforced the producer's half, and nine operators broke it — returning positions they had released, or swallowing a release instead of forwarding it and stranding upstream state. Both fail silently: a consumer that already took delivery merges the same values again, and a `Tile::Scalar`'s implicit positions make `merge` append rather than recognize the repeat, so one value becomes two, surfacing wherever it is later broadcast. `TileProducer::get` now checks the contract centrally, and these operators keep it.

## The check

`get` asserts a post-condition on the tile it returns (debug builds): no live data inside the accumulated `obsolete_guard`. `Tile::contains_guarded` defines that as "removing the guarded region would change the tile", so it agrees with `remove_guarded` by construction — including where that deliberately keeps data. Being *pulled* after a release is not a violation, so the assertion names the producer at fault rather than its puller.

Two things give the check reach. `Memo` must not shield what sits beneath it: skipping the upstream pull once the input is drained is an optimization — a conforming input answers empty — so only release builds take it, and debug keeps pulling. And the pipeline harness now releases universally and pulls once more, which must answer empty. They catch different things: the first re-pulls mid-run at every `Memo` in the graph and is what exposed `MapResult`; the second exercises the end state and is what exposed `ExtractAggregate` across 44 tests.

## Returning released data

- **`VariantWrap`** preserves its domain, so a domain release forwards verbatim — but it forwarded only universal ones, leaving its payload producing released positions and the wrap handing them back. The writer-decision path: 53 tests fail reverted.
- **`Memo`** pulled an input it had released in full; it now serves its cache.
- **`InductionStore`** re-requested a fully-released iteration source. `source_fully_released` now ends the drive, and a drained store serves the changelog through a shared `render_store`.
- **`UnionProducer`** pulled every arm. A fully-released arm now contributes an empty but **decided** region — a plain empty tile reads as "not ready" and would hold the union non-terminal forever.
- **`Aggregate`** ignored release entirely, and its accumulator *is* its whole output. It now goes quiet, and releases its input in full — `get_impl` only releases what was delivered, so an aggregate finishing before its source ran dry stranded the rest.
- **`MapAggregate`** ignored a **per-key** release: released keys stayed in its accumulator map and were re-emitted, and the input kept them. A domain predicate now drops exactly those accumulators and forwards verbatim, the input's `domain1` being the same key set.
- **`ExtractAggregate`** read `terminal.index_at(0)` unconditionally, so an empty input aggregation — `⊥`, or an upstream that released what it produced — panicked instead of answering empty.
- **`MapResult`** carried its input's domain through while mapping only live rows, so an input holding logically-deleted rows — what release marks, and what `Restrict` produces — gave a domain and codomain of different lengths. It compacts first.

## Swallowing a release

`ScalarFanInProducer::release_impl` was `// Nothing to do`, and `MapResultProducer` never released its `function`. Both re-read the operand whole on every pull, so both forward the universal release and nothing narrower. `MapResult` also built the input's guard from `self.function.tiling()`: a guard is shaped by the tiling it is handed to, and those two are unrelated.

## Reachability and coverage

The corruption was reachable from source — a conditional whose guard and arms share a computed binding pulls the shared `Memo` three times, and panicked in the broadcast:

    n: Int = 0 - 5
    n if n > 0 else n      # repeat requires single-element ColumnValue, left: 3

Each test is checked in the failing direction. `test_helpers` carries two doubles: `ReleaseSpy` records the guards handed to it, and `QuietSpy` also honors them, for tests that pull again after releasing — a double that re-answers released rows is itself the violation, and trips the assertion before the behavior under test is reached. Reverting the `Memo`, `InductionStore`, or `UnionProducer` fix fails 3, 3, and 1 tests. Three test writers that re-emitted committed positions now share a `ProposalWindow` — the offset window `TransactWriterProducer` implements for real.

`design-operators.md` gains [The release contract](src/interpreter/design-operators.md#the-release-contract), stating the promise and why every operator must obey it; what each one can forward stays in its table row.


dpmills:

- #58 ◀

src/interpreter/tile_operators/fanout.rs
```diff
@@ -505,3 +534,47 @@ impl TileProducer for MemoProducer {
...
+    use crate::interpreter::tile_operators::test_helpers::QuietSpy;
+    use crate::interpreter::{BaseType, ColumnValue, Extent};
+
+    /// A `Memo` releases its input universally as soon as the input hands over a
+    /// complete tile, and from then on the cache is the value: repeated pulls
+    /// answer the same single scalar.
+    ///
+    /// This has to hold in both build configurations, because only release builds
+    /// skip the upstream pull (see [`MemoProducer::get_impl`]). A debug build
+    /// re-pulls, and the input — honoring the release it was just handed — answers
```
sortalongo, line 550: Why have different behavior in release and debug builds? That seems like a smell.

sortalongo (approved):

src/interpreter/tile_operators/fanout.rs
dpmills: Having the extra unnecessary empty get catches more potential bugs in release logic.  It's a debug only check, and running the extra get could potentially be expensive.  I tried other ways of making sure this class of bugs is detectable in real programs, and this was needed to get full coverage (also tested in unit tests though)

PR #59, opened 2026-08-03
A variant type is spelled as its arms, so `Option` stops being the only one

dpmills:
`Option(T)` was the only variant type that could be written down. Anything else had to be inferred, which made the annotation path — and every rejection that depends on it — testable at exactly one shape.

A variant type is now written as its arms:

```python
x: some(Int) | none = .some(1)
y: ok(Int) | err(String) = .err("nope")
```

An arm is `tag(T)`, or a bare `tag` for one carrying no payload — the same `Unit` payload the nullary constructor `.tag` injects.

**This needed no parser work.** Annotations are parsed as ordinary expressions and *interpreted* as types by `lower_type_annotation`, and `|` already lexes and folds as logical-or, so `some(Int) | none` has always parsed — it just had no type reading. This adds that reading: a `|`-chain in type position flattens into a `Type::Variant`.

The two readings of `|` cannot collide, and it is the spec's existing capitalization rule that separates them rather than anything new: `Caps` is reserved for types, so a *lowercase* head in type position can only be a tag. `some(Int)` is the tag `some` carrying `Int`; `List(Int)` and `Option(Int)` keep working because they are capitalized; a capitalized arm is rejected as a type where a tag belongs. `|` in term position is untouched, which a test pins.

**Arms canonicalize into name order**, matching `Type::option_of` and the order the solver materializes a coalesced variant in. That is what makes `some(T) | none` *the same type* as `Option(T)` rather than a look-alike that happens to carry the same arms — so `match` dispatches over it identically, and an annotation written `none | some(Int)` compares equal to what inference produced. Without it, the tag-order mismatch that keeps `variant_param_accepts_subtype` ignored would have become reachable from source.

Rejections carry their own diagnostics: a capitalized arm, a tag named twice, and an arm with more than one payload type are each reported at lowering with the shape that was expected.

`chl-spec.md` gains the syntax and loses an inaccuracy in the same passage: it described `Option(T)` as abbreviating `{some: T, none: None}`, but braces are *record* type syntax — the shape it names is `some(T) | none`. The §3.15 Direction narrows to what is still open, which is *naming* a variant type (a declaration site for tags, and payload arity checked at the constructor), not writing one.

Tests cover a payload arm and a nullary arm, arm-order insensitivity, three arms with distinct payloads, a variant payload nested inside a variant type, `match` over an annotated variant type behaving as it does over `Option`, all five rejections, and `|` still meaning logical-or in term position.

Thanks for putting this together — but a heads-up before you go further: Cambra isn't accepting external pull requests right now. See [CONTRIBUTING.md](../CONTRIBUTING.md) for why, and for what to do instead.

If this is a bug fix, the fastest path is to open an issue describing (or attaching) the patch — small, clearly-correct fixes may get applied as a team commit with credit to you.

This PR will likely be closed unreviewed. That's about where the project is right now, not a judgment on the change.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -978,22 +978,38 @@ A leading dot is a constructor; a dot with an expression to its left
 stays attribute access (§3.9). A constructor is an ordinary atom, so it
 takes a postfix chain: `.some(r).x` is the attribute `x` of `.some(r)`.
 
-`Option(T)` abbreviates `{some: T, none: None}` and is a built-in type
-constructor in the same category as `List(T)` and `Map(K, V)` (§6.1) —
-not a distinguished kind of type. Nothing in the language privileges the
-spellings `some` and `none`; they are ordinary tags.
+A **variant type** is written as its arms, separated by `|`:
 
-> **Direction [Tentative].** Tags are undeclared today, so a *misspelled*
-> tag is caught only where it fails to meet its counterpart, and a
-> constructor's payload arity is not checked at the constructor. The
-> planned refinement is the structural type alias of §6.1 —
-> `Option(T) = some(T) | none`, `Result(T) = ok(T) | err(String)` — which
-> gives a tag a resolution site, checks the payload there, and would let
-> the disambiguating dot be dropped for a *declared* tag while keeping it
-> for an ad-hoc one. It would also replace the built-in `Option(T)` above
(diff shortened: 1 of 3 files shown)


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14
                    - #15
                        - #53
                            - #54
                                - #59 ◀

PR #60, opened 2026-08-04
Annotations stop standing in for binder structure, and stop outliving inference

dpmills:
Four places where a **pre-inference annotation** was doing a job that structure should do. An annotation is a raw type from lowering — never normalized, never coalesced — so every pass that pattern-matches one reads a shape from before inference ran, standing in for a fact the tree could have carried directly. This replaces the proxies with the facts and then makes the proxy unavailable: a mutable-variable introduction becomes a node, a binder slot records what the binder is bound at, `infer` clears every annotation on success, and `channelize` drops the pre-inference order it no longer runs in. No surface semantics change except the one the `Mut` discipline gives up below.

## A mutable-variable introduction is a node

`x := init` lowered to a `Let` carrying a `Mut(V, D)` annotation, so *"does this binder hold a mutable variable?"* had no structural answer. `TypedExprNode::MutDecl { binding, init, body }` is the missing marker — the declaring half of `:=`, paired with the `MutWrite` that was already the writing half — erased by `mut_elim`, since after elimination the variable *is* its seed: [ir.md](src/ccl/design/ir.md#mutdecl--the-mutable-variable-introduction).

**A `Let` can no longer bind a mutable variable**, which is what the node buys. `emit_let` reads through an initializer that is a mutable variable up front, so `b = a` binds `b` at `a`'s *value* — a snapshot, as `a + 1` already read it. That makes the mutable alias unrepresentable rather than rejected, retiring the discipline's **rule 3** and `InferError::UnannotatedMutBinding` with it; writing through the copy is still caught, by the write-target check.

## A binder's type is what it is bound at

The `let` binder slot recorded its **initializer's** type, which differs from the bound-at type in opposite directions for the annotated cases: `y: Int = x` off a mutable variable binds at the value while its initializer is a history; `x: Mut(V) := init` binds at the history while its initializer is a value. Nothing looked broken because three readers consulted `user_annotation`, which happened to answer correctly ([type-inference.md](src/ccl/design/type-inference.md#the-binder-slot-and-why-annotations-do-not-outlive-inference)).

With the slot honest, no post-inference reader is left, so `infer` clears every annotation and both post-inference walls assert the emptiness. The leak was real: `collect_type_errors` walks `ty` slots but not `user_annotation`, so `def f(a: Int, b)` ended inference with a raw `Type::Hole` inside its uncurried parameter annotation.

## `channelize` runs on an inferred tree

`channelize::run` took an `input_typed: bool` and skipped its type work when false. Both production call sites passed `true` — only the pass's own unit tests passed `false`, so the untyped order survived exclusively to keep hand-built fixtures compiling. The flag and its gates are gone and the fixtures type first, which found an **unreachable fan-out sub-branch** in the `Apply` arm: it read the argument as the iteration source, where a comprehension puts an *element*. That branch, its two unit tests, and the docs claiming two fan-out sites all go.

## One in-loop rule, at every spelling

A mutable variable introduced inside a `for` body is rejected uniformly. The gate keyed on the annotation, so the same construct got opposite answers — `y: Mut(Int) := 0` rejected, bare `y := 0` accepted — and the accepted spelling fell through to a per-iteration **shadowing `let`**, discarding every update at the iteration boundary. A variable's domain is fixed by where it is introduced, so an introduction cannot sit inside the context that would sequence it ([mutability.md](src/ccl/design/mutability.md) and the spec); the `+=` half of that same fallback is still open, in violation of a rule the spec already states.

## Enumerating node kinds by hand, twice

Adding a *binder* node exposed two passes doing it, both silently: `uniquify` left the binder's name `Raw` among `Unique` ones, so its writes never resolved to it, and `append_record_at_tail` missed it as a spine link, wrapping the sink record above the declaration and making a live cross-endpoint read return its seed. The binder half is closed structurally — `scope.rs`'s wildcard-free match and its corpus test refuse to build until the node declares its scoping rule — but the **spine** half is open: that walk is hand-rolled across `channelize`, `mut_elim`, `transact_phase` and lowering, and the copies disagree about whether `ExprStmt` is a link or the tail, which is what the second bug exploited. Naming the spine once is the follow-up.

Two more sites take the node with a decision attached: `collect_mut_binders`, where absence had read as "no objection" so the write-target predicate becomes `!= Some(&true)`, and `transact_phase`'s block-interior walkers, whose arms are `todo!` because lowering never binds a mutable variable inside `with begin():`.

Tests pin what replaced rule 3 (a plain copy reads, writing through it is rejected, `_` behaves identically), the binder slot in both annotated directions, that annotations do not survive inference, and that a `Mut` parameter still binds a mutable variable — pass-by-reference, which the initializer deref must never widen to cover.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60 ◀
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79
                            - #91
                                - #92
                                    - #93
                                        - #97

sortalongo (approved):

PR #62, opened 2026-08-04
Exact and bounded binder annotations: `x: T` fixes the type, `x <: T` bounds it

dpmills:
CHL gets two binder-annotation forms, meaning the same thing at every binder position — `=`, `:=`, and `def` parameters. `x: T` is **exact**: the binder's type *is* `T`, and nothing downstream sees more. `x <: T` is **bounded**: the type is inferred with `T` as an upper bound, so the value's own type flows through. Bounded is the behaviour both positions already had, so existing programs keep working under `<:` and the corpus needed no migration; what is new is `:` meaning what it says. That closes the asymmetry that motivated the work — `def f(v: {a: Int}): v.b` and `x: {a: Int} = (a=1, b=2); x.b` now agree — and it gives a program its one lever over specialization count. Spec: [Two annotation forms: exact and bounded](docs/chl-spec.md#two-annotation-forms-exact-and-bounded); design: [type-inference.md](src/ccl/design/type-inference.md#annotation-kinds-exact-and-bounded).

The forms coincide only where the value's type already *is* the annotation. They differ on **width** (`x: {a: Int} = (a=1, b=2)` makes `x.b` an error; `<:` binds at `{a: 1, b: 2}`) and on **literal singletons** (`x: Int = 5` discards the singleton, `x <: Int = 5` keeps `5` and still discharges `arr[x]`'s index obligation) — so a "simple" annotation is no guarantee that the two agree.

## `Type::BoundedHole(T)` is a marker in a type slot, not a type

In annotation position only; `normalize_annotation` erases it into a fresh variable bounded above by `T`. It is `Type::Hole` one rung up, and the two compose where a compound annotation is partly specified. It denotes nothing — a bound picks out no set of values — so **no typing rule may take one**: five solver sites assert that rather than inventing a rule, and only the structural walks that rewrite every slot uniformly pass through. Putting the bound *in the type* is forced by the multi-parameter encoding: `def f(x: A, y <: B, z)` uncurries to one tuple annotation, and `Tuple([A, BoundedHole(B), Hole])` expresses three modes inside one type.

## Neither binder rule needs a mode test

A parameter binds at `normalize(annotation)` — exact to `T` itself, bounded to a variable bounded by `T` — which retires `emit_lambda`'s bind-fresh-then-reconcile step, the reason an exact annotation used to behave as neither reading. A `let` binds at the same normalization of its **completed** annotation (`emit::complete_annotation` fills each `Hole` from the initializer's type, structurally rather than by constraint, because variables minted after the RHS's level is popped escape inference unresolved). Two special cases fall out as consequences: a deref-copy binds at the annotation because that is what exact means, and a bare `_` completes to the initializer's type.

A bound on a mutable variable bounds its **value** type, so `x <: Mut(V) := e` and `x <: V := e` are one declaration: lowering applies the binder's mode to the value it extracts from a `Mut(…)` annotation (`apply_annotation_mode`). The value position is also the only one available — the binder's slot must stay structurally a `History` for `mut_value_type`, `mut_elim` and `transact_phase` to dispatch on — so `normalize_annotation` asserts a bound never wraps a history rather than collapsing one.

## Two things a reader might expect and should not

- A bounded parameter's type is still the **meet** of its bound and its body's demands, so `def f(v <: {a: Int}): v.b` requires callers to supply both fields.
- `<:` is not writable in nested positions, because it describes a *binder*, not a type. That, and the bounded reading arguably being the more common intent while carrying the heavier spelling, is why the spec marks the **spelling** `[Open]`; the mode is a two-valued property lowering reads off the surface, so a respelling is a parser change.

Tests cover both differing columns, the two binder positions agreeing, `_` completion, a bounded variable's bound constraining seed and writes, and specialization count: `def f(v <: Int)` called at `f(1)` and `f(2)` produces two clones, `def f(v: Int)` one. Only four existing assertions changed, each asserting the old reading of `:` directly.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -1416,9 +1416,9 @@ rest of that iteration and gone at the next.
 The converse also holds: a mutable variable must be introduced **before**
 the loop that accumulates it. A `:=` inside a loop body whose target is not
 an already-declared accumulator is a lowering error, at every spelling —
-`y := 0`, `y: Int := 0`, and `y: Mut(Int) := 0` alike, since whether an
-introduction carries a type annotation says nothing about whether it
-introduces a mutable variable. A mutable variable scoped to one iteration would need
+`y := 0`, `y: Mut(Int) := 0`, and `y: Mut(Int, Txn) := 0` alike, since
+whether an introduction carries a type annotation says nothing about
+whether it introduces a mutable variable. A mutable variable scoped to one iteration would need
 the loop's own iteration extent as its sequencing domain, which is the
 nested-recurrence case below:
 
@@ -1657,18 +1657,21 @@ element `{T,}`), record type `{f: T}`, variant type
 An annotation at a binder answers one of two questions, and the two
 have different spellings because the answers differ.
 
-> **Note.** The *distinction* below is implemented and pinned by tests —
-> that a binder can either fix its type or bound it, and what each means
-> at every binder form. The **spelling** is **[Open]**: `:` versus `<:`
-> is not a syntax we are satisfied with, and it may change without the
-> semantics changing. Two things make it unsatisfying. `<:` reads as a
-> type operator but describes a *binder*, which is why it cannot be
-> written in a nested position — a restriction that falls out of the
-> implementation rather than out of anything a reader would expect from
-> the notation. And the more common intent is arguably the bounded one,
-> yet it carries the heavier spelling. Read the semantics as settled and
-> the two tokens as provisional; code written against them may need a
-> mechanical rename.
+> **Note — why these two spellings.** `<:` relates two *types* everywhere
+> else, and here it sits between a term and a type. The capitalization
+> rule (§6.1) is what makes that unambiguous rather than a pun: a
+> lowercase head means the left side is a term, so `a <: T` reads "`a` has
+> a type that is a subtype of `T`", while `A <: T` reads "`A` is a subtype
+> of `T`". Which reading applies is settled by the case of the name, not
+> by the operator.
+>
+> Exact gets the lighter spelling because it is the safer default, not
+> because it is the more frequent intent. An annotation is written where
+> the type is *not* obvious or where a contract is being fixed, and in
+> both of those cases the more precise reading is the one to have by
+> default. `<:` is then a deliberate opt-in: it loosens the contract and
+> accepts inference whose result is harder to predict from the annotation
+> alone.
 
 `x: T` is **exact**: the binder's type *is* `T`. The initializer (or,
 at a parameter, the argument) must be a subtype of `T`, and nothing
@@ -1681,8 +1684,12 @@ constrains what may reach the binder.
 Both forms are accepted wherever a binder is introduced — an
 assignment (`x: T = e`, `x <: T = e`), a mutable introduction
 (`x: Mut(V) := e`), and a `def` parameter — and mean the same thing in
-each. `<:` is **not** written in nested positions: it describes a
-binder, not a type, so `{a <: Int}` is not a type.
+each. `<:` does not appear **inside a type literal**: it says how an
+annotation is read, and what it annotates is a term, so `{a <: Int}` is
+not a record type. That is a restriction on type literals, not a claim
+that a binder is the only place an annotation can go — an inline
+annotation on an expression would carry both modes for the same reason a
+binder does.
 
 The two coincide only when the value's type already **is** the
 annotation — when there is nothing for the annotation to discard. They
@@ -1691,8 +1698,8 @@ more often than it sounds, because a CHL type carries more than a base:
 
 - **Width.** `x: {a: Int} = (a=1, b=2)` binds `x` at `{a: Int}`, so
   `x.b` is an error — the annotation is what discards the field.
-  `x <: {a: Int} = (a=1, b=2)` binds `x` at `{a: 1, b: 2}`, and `x.b`
-  is `2`.
+  `x <: {a: Int} = (a=1, b=2)` binds `x` at the record's own type, which
+  still has both fields, so `x.b` is `2`.
 - **Literal singletons** (§3.1). `i: Int = 0` binds `i` at `Int`;
   `i <: Int = 0` leaves it at `0`. Only the second still carries the
   fact a totality proof needs, so an exact annotation on an index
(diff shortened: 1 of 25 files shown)


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62 ◀
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79

docs/chl-spec.md
```diff
@@ -1641,6 +1652,81 @@ element `{T,}`), record type `{f: T}`, variant type
...
+A `_` position **declares nothing** and is completed from the
+initializer, so `x: _ = e` means exactly `x = e`. That holds nested
+too: `x: List(_) = [1, 2, 3]` binds `x` at `List(Int)`. Consequently
+`_` does not suppress the mutable-alias rule (§8.1): `b: _ = a` off a
+mutable `a` is the same error as a bare `b = a`.
+
+On a mutable introduction the mode applies to the **value type**, which
+is what the annotation names: `x <: Mut(V) := e` declares a mutable
+whose value type is inferred subject to `<: V`, and is the same
+declaration as `x <: V := e`. So `x <: Mut(Int) := 5` binds the value at
```
sortalongo: > `x <: Mut(V) := e` [...] is the same declaration as `x <: V := e`

This is troubling. I thought the upstack PRs got rid of `V <: Mut(V)`?

I object to being able to implicitly drop `Mut` annotations.

src/ccl/ty.rs
```diff
@@ -569,6 +570,34 @@ pub enum Type {
...
+    ///
+    /// **Annotation position only, and transient**: lowering writes it, inference
+    /// erases it, and no pass downstream may observe one. The annotation slots it
+    /// occupies do not survive inference at all
+    /// (`infer::api::debug_assert_annotations_cleared`), so a binder `ty` is the
+    /// only place a survivor could hide — flagged there by `collect_type_errors`
+    /// as [`InferError::UnresolvedBelow`](crate::ccl::infer::InferError::UnresolvedBelow).
+    ///
+    /// See `src/ccl/design/type-inference.md`, "Annotation kinds: exact and bounded".
+    Below(Box<Type>),
```
sortalongo: Strong preference for `BoundedHole` as the name of this thing. `Below` is pithy, but ambiguous in many contexts.

docs/chl-spec.md
```diff
@@ -1641,6 +1652,81 @@ element `{T,}`), record type `{f: T}`, variant type
...
+> at every binder form. The **spelling** is **[Open]**: `:` versus `<:`
+> is not a syntax we are satisfied with, and it may change without the
+> semantics changing. Two things make it unsatisfying. `<:` reads as a
+> type operator but describes a *binder*, which is why it cannot be
+> written in a nested position — a restriction that falls out of the
+> implementation rather than out of anything a reader would expect from
+> the notation. And the more common intent is arguably the bounded one,
+> yet it carries the heavier spelling. Read the semantics as settled and
+> the two tokens as provisional; code written against them may need a
+> mechanical rename.
```
sortalongo: I thought about it more, and I'm actually happy with the `<:` syntax now.

1. My argument that `<:` relates two types is true, but it ignores the disambiguation that our term/type spelling convention creates. It's easy to tell that `a <: T` is "`a` has a type which is a subtype of T" and `A <: T` is "A is a subtype of T" simply via their capitalization.
2. IMO, the more common intent should be an exact bound. Annotations generally happen in places where the type is not obvious, or a contract is being established. In both cases, it's safer to have a more precise annotation, so `:` should be the default. I think programmers should only use`<:` when they want to loosen the contract, and opt-in to less-predictable compiler behavior.

If you agree, let's mark this direction as the current decision and move on.

docs/chl-spec.md
```diff
@@ -1641,6 +1652,81 @@ element `{T,}`), record type `{f: T}`, variant type
...
+
+`x <: T` is **bounded**: the binder's type is *inferred*, with `T` as
+an upper bound. The value's own type flows through; `T` only
+constrains what may reach the binder.
+
+Both forms are accepted wherever a binder is introduced — an
+assignment (`x: T = e`, `x <: T = e`), a mutable introduction
+(`x: Mut(V) := e`), and a `def` parameter — and mean the same thing in
+each. `<:` is **not** written in nested positions: it describes a
+binder, not a type, so `{a <: Int}` is not a type.
```
sortalongo: "nested positions" is ambiguous here. I agree that `{x <: Int}` would be bad. But `(x <: Int) + 30` seems fine. Specify "nested inside type literals". Also don't over-constrain the `<:` syntax to only apply to binders. In the future, in-line type annotations are fair game.

docs/chl-spec.md
```diff
@@ -1641,6 +1652,81 @@ element `{T,}`), record type `{f: T}`, variant type
...
+
+The two coincide only when the value's type already **is** the
+annotation — when there is nothing for the annotation to discard. They
+differ whenever the value's type is a *strict* subtype of it, which is
+more often than it sounds, because a CHL type carries more than a base:
+
+- **Width.** `x: {a: Int} = (a=1, b=2)` binds `x` at `{a: Int}`, so
+  `x.b` is an error — the annotation is what discards the field.
+  `x <: {a: Int} = (a=1, b=2)` binds `x` at `{a: 1, b: 2}`, and `x.b`
```
sortalongo: > binds `x` at `{a: 1, b: 2}`

I said this before (so I expect it's in your queue to change it), but I really don't like overloading numeric literals to mean "type refined to be this literal". That notation suddenly makes the above look like a Python dict, when it really is a record type literal with refined fields.

If concise notation is useful, let's come up with something more intuitive, e.g. `{Int is 1}`

docs/chl-spec.md
```diff
@@ -1641,6 +1652,81 @@ element `{T,}`), record type `{f: T}`, variant type
...
+still differ. The annotation's own shape is not what decides it: `0` is
+a strict subtype of `Int`, so there is something to discard. What makes
+the forms coincide is the *value* knowing nothing beyond what the
+annotation says.
+
+A `_` position **declares nothing** and is completed from the
+initializer, so `x: _ = e` means exactly `x = e`. That holds nested
+too: `x: List(_) = [1, 2, 3]` binds `x` at `List(Int)`. Consequently
+`_` does not suppress the mutable-alias rule (§8.1): `b: _ = a` off a
+mutable `a` is the same error as a bare `b = a`.
```
```suggestion
```

cruft

sortalongo (approved):

docs/chl-spec.md
dpmills: This was stale from before this PR was on top of the mut changes.  After thinking about this a bit more, I also decided to block x <: Mut(V) entirely.  It doesn't mean what it naively looks like, and always means the same thing as x: Mut(V).

src/ccl/ty.rs
dpmills: Done

docs/chl-spec.md
dpmills: Agreed, done

docs/chl-spec.md
dpmills: Clarified; I only meant nested in type literals

docs/chl-spec.md
dpmills: I do think we need concise notation, but yeah, bare literals aren't great.  Maybe Int@5 or something similar.  Regardless, a followup

docs/chl-spec.md

PR #63, opened 2026-08-05
Four defects in how `Mut` is typed: message polarity, an unclosable refinement, and `op=` as a rebind

dpmills:
Four defects in how `Mut` is typed, found by probing what the base PR left. None changes the lattice; each is independently reviewable.

- **A mismatch named the demand as `found`** — `x: Mut(Int) := "s"` reported *expected String, found Int*. `constrain_subtype(lhs, rhs)` means `lhs <: rhs`, so the fields are `found` and `expected` now, with `expected` an `Option`: a missing field and an unaccepted tag are faults in a *single* type, and had been borrowing the second slot with a `Type::Hole` that rendered as a bare `_`.

- **A refinement depending on a mutable variable is a rejection, not a compiler bug.** `x := 2; ys = [i for i in [1, 2, 3] if i < x]` tripped the debug-only `check_scope_valid`, so release builds panicked at the pre-desugar wall instead. It is `InferError::MutableInRefinedType` now, whose docs record why this is *staging* — the term that would discharge the binder is minted by `mut_elim`, passes after closure is demanded. Reporting it needed `ExprStmt` to propagate the lifted type, which it did not.

- **`op=` inside a loop body is a mutable write.** Both loop-lowering paths fell back to a per-iteration shadowing `let` for a target that was not a pre-loop accumulator, and since `op=` reads the old value, `y = 0; y += i` recomputed `0 + i` rather than accumulating. Each path now threads the `MutWrite` or rejects — the `op=` half of the hole whose `:=` half the base PR closed.

- **A mutable variable's value type is invariant across a `Mut` parameter — now pinned.** Narrowing is the unsound direction and nothing tested it. Both directions are rejected today but *not* by the invariance rule, since the deref arm fires first at an argument position, so `Type::History`'s docs name the two edges that assemble the property instead.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -1416,10 +1416,17 @@ nested-recurrence case below:
 > supported: declare it before the loop (`Y := …`) so its updates carry
 > across iterations, or bind a per-iteration value immutably with `Y = …``
 
-Falling back to a per-iteration binding is deliberately *not* what happens,
-for the reason a plain `=` to an outer name is rejected: it would silently
-discard every update at the iteration boundary, which is the one thing `:=`
-exists to rule out.
+The same holds for `op=`, which is a mutable write and not a rebind: `x += e`
+inside a loop body requires `x` to be a mutable variable declared before the
+loop. A target that is not gets the matching error rather than a
+per-iteration binding.
+
+Falling back to a per-iteration binding is deliberately *not* what happens in
+either case, for the reason a plain `=` to an outer name is rejected: it would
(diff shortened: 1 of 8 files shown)


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63 ◀
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79
                            - #91
                                - #92
                                    - #93
                                        - #97

sortalongo (approved):

PR #64, opened 2026-08-05
A mutable variable read is an explicit operation, not a subtyping rule

dpmills:
A mutable variable was related to its value by *subtyping*, in both directions, and neither is a lattice fact: `Mut(V) <: V` puts a mutable variable below its value, `V <: Mut(V, D)` puts it above, and `Mut` is invariant in `V`. Both fire against a fresh inference variable — `cnt + 1` emits `Mut(Int, D) <: ?α` — so the relation cannot tell a **read** from a **handle being passed along**, since a pass-by-reference argument meets a fresh variable too. The deref moves to the rule that emits the operand (`emit::emit_value_read`), both arms are deleted, and the relation relates a mutable variable only to another mutable variable.

## Where a handle survives

Rules 1 and 2 of the second-class discipline make the handle positions enumerable — a **pass-by-reference argument** and a **write's target** — so a parent knows without inspecting the operand. Everything else derefs, tail positions included; a lambda's *result* deliberately does not, since that is where rule 2 catches an escape. See [mutability.md](src/ccl/design/mutability.md#a-mutable-variable-read-is-an-explicit-operation). Invariance now supplies both directions at a call, so `contribute_pbr_writes` is deleted.

## The three positions that made the upward coercion look necessary

Each is a place the *rule* should have read through the handle:

- the coalesce-time **lifted type** copied a tail's continuation type verbatim, re-stamping a statement or a mutable-variable introduction with the handle the read had just looked through: `a := a + 5; a` derived `Int` and recorded `Mut(Int, D)`. 394 of the arm's 395 firings across `compilation_pipeline` were the wall reconciling that;
- a **`Case` arm** was not a value position, so a conditional over two mutable variables denoted a `Mut` whose writer cannot be traced — the case rule 2 exists to prevent;
- a **`Mut` parameter given something that is not one** had no value edge, leaving an argument the discipline check rejects under-determined.

`mut_var_value_type` asserted the old behaviour and now pins the corrected one, two tests cover the fixed tail and `Case` positions, and the arm's unit test becomes the statement that a value does not satisfy a `Mut` demand.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64 ◀
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79
                            - #91
                                - #92
                                    - #93
                                        - #97

src/ccl/design/mutability.md
```diff
@@ -96,6 +96,55 @@ contributing loop's domain, free to reference the letrec's bindings. (They are o
 history by the same eliminator — it is only the *append* merge law, with no carry-forward, that
 lets a feed be a plain output rather than a cyclic binding.)
 
+### A register read is an explicit operation
```
sortalongo: Reminder: get rid of the "register" terminology when it just means "mutable variable"

src/ccl/infer/emit.rs
```diff
@@ -819,10 +832,51 @@ pub(super) fn emit_defer<C: Typing>(ctx: &mut C) -> Type {
...
+/// what makes that enumerable: rule 1 forces a `Mut`-typed value to be a bare `Var`,
+/// and rule 2 keeps `Mut` out of every composite, so a parent always knows whether the
+/// operand it is about to constrain is a handle position without inspecting it.
+///
+/// Dereffing *here* rather than inside the subtyping relation is the point. A rule
+/// `Mut(V) <: V` reads as "a register is a subtype of its value", which is a coercion
+/// wearing a subtyping rule's clothes: it makes `Mut(V)` sit below `V` while `Mut` is
+/// invariant in `V`, and it fires against a fresh inference variable, so it cannot tell
+/// a read from a handle being passed along.
+fn read_operand<C: Typing>(e: &mut Expr, ctx: &mut C) -> Result<Type, LocatedInferError> {
```
sortalongo: "operand" is a weird word here, esp. when this is a one-liner for "deref_mut". Maybe it should be "read_mut" or something like that? Or maybe reconsider all of the naming across both?

sortalongo (approved):

src/ccl/design/mutability.md
dpmills: Done across the whole stack and preexisting usage

src/ccl/infer/emit.rs
dpmills: Renamed.  This one becomes emit_value_read (it acts like all the other emit_* functions) and deref_mut becomes read_through

PR #65, opened 2026-08-05
Reserve http test ports with a held lock, not a released bind hint

dpmills:
The `http_server` and `programs` test binaries fail intermittently with `http_serve: failed to bind port 35545: Address already in use`: `free_port()` bound `:0`, read the port the kernel assigned, and *closed* it, leaving a hint nothing owns until `SharedHttpServer::new` binds it a compile-window later — long enough for a concurrently running test process to be handed the same port. `reserve_test_port` replaces it with a reservation held for the life of the process, making a collision impossible by construction rather than by luck. Measured at 24-way concurrency: 1 failing run in 240 before this change, 0 in 432 after.

The reservation is a per-port advisory file lock held until exit, allocation from a band below every `ip_local_port_range` floor, and a bind probe for a foreign listener; its doc comment carries the why, including the `flock`-vs-`fcntl` dependency that same-process exclusion rests on. Living in the library behind the default-off `test-helpers` feature also collapses the two copies of `free_port`.

### Sleeps

`wait_for_bind()` and the twelve fixed 50 ms waits before a first request are gone — `SharedHttpServer::new` binds before spawning its dispatcher, so the race they claimed to cover cannot happen. Both `drive_until` loops now wait on `recv_timeout` rather than `try_recv` plus `sleep`.

### Tests

`tests/reserve_port.rs` pins the two properties the guarantee rests on: exclusion between threads of one process, and a reserved port still bindable by its reserver.

### DIFF
--- a/src/interpreter/http_server.rs
+++ b/src/interpreter/http_server.rs
@@ -118,67 +118,6 @@ impl SharedHttpServer {
         Ok(Self { routes })
     }
 
-    /// Reserve a port for a test program's `http_serve`, held against every other
-    /// concurrent test process for the rest of this process's life.
-    ///
-    /// **Test-only.** Shared through the default-off `test-helpers` feature so the
-    /// integration-test crates hold one implementation rather than a copy each.
-    ///
-    /// The obvious allocator — bind `:0`, read the port the kernel assigned,
-    /// close — is unsound here, and flakily so.  Closing makes the port a *hint*
-    /// that nothing owns: the listener that will really own it is [`Self::new`]'s,
-    /// opened at the far end of `compile_program`, and until then `bind(:0)` in a
-    /// *concurrently running test binary* is free to hand the very same port out
-    /// again.  Both compiles then race to bind it and the loser fails with
-    /// `EADDRINUSE`, surfacing as a lowering error from `http_serve`.  Under
-    /// `cargo test`, where the http test binaries run alongside each other, that
-    /// is the whole of the observed flakiness (measured: 13 failures in 7200
-    /// allocations at 24-way concurrency, every one of them a port two processes
-    /// had been handed at once).
-    ///
-    /// So the reservation has to be held by something that outlives the hint.  An
-    /// advisory lock on a per-port file does it: exclusive across processes,
-    /// released by the OS on exit, so a crashed run cannot permanently blacklist
-    /// a port the way a leftover lock *file* would.  Sequential allocation from
-    /// `BASE` is then collision-free by construction rather than by luck — the
-    /// lock, not chance, decides who gets a port.  The bind probe covers the one
-    /// thing the lock cannot: a *foreign* process already listening there.
-    #[cfg(any(test, feature = "test-helpers"))]
-    pub fn reserve_test_port() -> u16 {
-        use std::{fs::File, net::TcpListener, sync::Mutex};
-
-        /// Below the lowest `ip_local_port_range` floor in practice (32768 on
-        /// Linux, higher on macOS and Windows) so the kernel's own allocator
-        /// never draws from this band for outbound connections, and clear of the
-        /// privileged range.
-        const BASE: u16 = 20_000;
-        const SPAN: u16 = 10_000;
-
-        /// Every lock this process holds.  The reservation must outlive the call —
-        /// the server binds later, inside `compile_program` — and a test has no
-        /// scope to keep a guard in, so the locks live until the process exits.
-        static RESERVED: Mutex<Vec<File>> = Mutex::new(Vec::new());
-
-        let dir = std::env::temp_dir();
-        for candidate in BASE..BASE + SPAN {
-            let Ok(lock) = File::create(dir.join(format!("cambra-test-port-{candidate}.lock")))
-            else {
-                continue;
-            };
-            if lock.try_lock().is_err() {
-                continue;
-            }
-            // Probe the same wildcard address `Server::http` will bind, so a
-            // listener on any interface — not just loopback — rules the port out.
-            if TcpListener::bind(("0.0.0.0", candidate)).is_err() {
-                continue;
-            }
-            RESERVED.lock().unwrap().push(lock);
-            return candidate;
-        }
-        panic!("no test port available in {BASE}..{}", BASE + SPAN)
-    }
-
     /// Register a new `(method, path)` route and return the `Receiver` for incoming requests.
     ///
     /// Requests matching this route will be sent on the returned channel.  A value of `None`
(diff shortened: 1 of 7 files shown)


dpmills:

- #65 ◀

src/interpreter/http_server.rs
```diff
@@ -118,6 +118,67 @@ impl SharedHttpServer {
...
+
+        let dir = std::env::temp_dir();
+        for candidate in BASE..BASE + SPAN {
+            let Ok(lock) = File::create(dir.join(format!("cambra-test-port-{candidate}.lock")))
+            else {
+                continue;
+            };
+            if lock.try_lock().is_err() {
+                continue;
+            }
```
groundlar: The File::create arm silently swallows every I/O error, and misdiagnoses the result:

```
let Ok(lock) = File::create(dir.join(format!("cambra-test-port-{candidate}.lock"))) else {
    continue;
};
```
A read-only TMPDIR, ENOSPC, EMFILE, or — the realistic one — a shared /tmp where the lock files belong to another user (sticky bit, mode 0644) all land here. I confirmed the multi-user case:

File::create -> Permission denied (os error 13)   <-- silently `continue`s

The loop then walks all 10,000 candidates and panics with no test port available in 20000..30000, which reads as port exhaustion and points the reader at concurrency limits or a leaked-reservation bug. That is the same failure shape this commit was written to eliminate: an error message that points away from the cause. Cheap fix — keep the last error and put it in the panic:

```
let mut last_err = None;
for candidate in BASE..BASE + SPAN {
    let lock = match File::create(dir.join(format!("cambra-test-port-{candidate}.lock"))) {
        Ok(f) => f,
        Err(e) => { last_err = Some(e); continue }
    };
    ...
}
panic!("no test port available in {BASE}..{}; last lock-file error: {last_err:?}", BASE + SPAN)
```

src/interpreter/http_server.rs
```diff
@@ -118,6 +118,67 @@ impl SharedHttpServer {
...
+    ///
+    /// So the reservation has to be held by something that outlives the hint.  An
+    /// advisory lock on a per-port file does it: exclusive across processes,
+    /// released by the OS on exit, so a crashed run cannot permanently blacklist
+    /// a port the way a leftover lock *file* would.  Sequential allocation from
+    /// `BASE` is then collision-free by construction rather than by luck — the
+    /// lock, not chance, decides who gets a port.  The bind probe covers the one
+    /// thing the lock cannot: a *foreign* process already listening there.
+    #[cfg(any(test, feature = "test-helpers"))]
+    pub fn reserve_test_port() -> u16 {
```
groundlar: `reserve_test_port` is now the single mechanism keeping two test binaries non-flaky, and nothing exercises it. Specifically, nothing would catch same-process exclusion regressing — if a future std switches a platform to `fcntl(F_SETLK)` (per-process, so a second lock in the same process *succeeds*), the flakiness returns silently and identically. This is a case where a test is cheap and the failure is expensive-to-diagnose by construction.

Both pass today:

```rust
//! tests/reserve_port.rs — coverage for the allocator the anti-flake guarantee rests on.
use std::{collections::HashSet, net::TcpListener, thread};
use cambra::interpreter::http_server::SharedHttpServer;

/// The reservation must be exclusive *within* the process too: `cargo test` runs
/// tests as threads, so two concurrent calls handed the same port would reproduce
/// exactly the `EADDRINUSE` this allocator exists to prevent.  It holds because
/// the lock is a `flock` on the open file description, not a per-process fcntl lock.
#[test]
fn reserve_test_port_is_exclusive_across_threads() {
    let ports: Vec<u16> = thread::scope(|s| {
        let hs: Vec<_> = (0..16).map(|_| s.spawn(SharedHttpServer::reserve_test_port)).collect();
        hs.into_iter().map(|h| h.join().unwrap()).collect()
    });
    let unique: HashSet<_> = ports.iter().copied().collect();
    assert_eq!(unique.len(), ports.len(), "duplicate ports handed out: {ports:?}");
    assert!(ports.iter().all(|p| (20_000..30_000).contains(p)), "{ports:?}");
}

/// A reserved port must still be bindable by the caller — the probe's listener has
/// to be closed, not leaked into the static that holds the locks.
#[test]
fn reserved_port_is_bindable_by_the_caller() {
    let port = SharedHttpServer::reserve_test_port();
    TcpListener::bind(("0.0.0.0", port)).expect("reserved port must be bindable");
}
```

Alongside it, the `flock`-vs-`fcntl` dependency deserves one sentence in the doc comment — it is the non-obvious *why* behind the whole design.

groundlar: nit: `reserve_test_port` never mentions `Self` (only prose references `Self::new`). Then at the consumer, tests/programs/common/mod.rs, a doc comment titled "Port allocation for the {PORT} placeholder in sink programs" is attached to a `pub use ...SharedHttpServer`. 

A free `pub fn` would let common re-export this.

groundlar (approved):

src/interpreter/http_server.rs
dpmills: Added, in `tests/reserve_port.rs` — both of your tests, near-verbatim.

You're right that the same-process case is the one that would regress silently, and it isn't vacuous: under `fcntl(F_SETLK)` semantics the second thread's `try_lock` on the same path succeeds, and the bind probe succeeds too (the first thread already closed its probe listener), so the duplicate reaches the assertion rather than being caught by a later failure.

The `flock`-vs-`fcntl` sentence is in the doc comment now, as the paragraph that says why exclusion holds within a process and what pins it.

dpmills: Done — it's a free `pub fn` in the module now. The consumer reads better for it: `common` re-exports the function itself, so the "Port allocation for the `{PORT}` placeholder" doc comment is attached to the thing it describes rather than to a server type, and the three program tests import `reserve_test_port` instead of a `SharedHttpServer` they never otherwise mention.

src/interpreter/http_server.rs
dpmills: Fixed, and thank you for confirming the multi-user case — that's the one that actually happens.

I took the fix with one change: rather than always appending a `last_err` that is `None` on the normal path, the two exhaustion causes now print distinctly.

```rust
match last_create_err {
    Some(e) => panic!(
        "no test port available in {BASE}..{}: could not create a lock file in {}: {e}",
        BASE + SPAN, dir.display(),
    ),
    None => panic!(
        "no test port available in {BASE}..{}: every port is reserved or already bound",
        BASE + SPAN,
    ),
}
```

The arm carries a comment saying why the distinction is load-bearing: every cause of a `File::create` failure here is per-directory, not per-port, so it rejects all 10 000 candidates — which means whenever one occurred, the exhaustion reading is not merely unhelpful but wrong.

PR #66, opened 2026-08-06
Design docs: the program-branching model and its operational configurations

groundlar:
Adds the branching design space to `src/ccl/design/`: `branching.md` (the model of record) and `branching-sequence-diagrams.md` (the same configurations drawn as sequence diagrams).

`branching.md` slices the design space of running two versions of a program at once over two axes — *evaluation* (which arms run at a position) and *effect ownership* (which arm's effects are external truth) — and derives a third, per-register *history* (**agreed**/**mixed**/**parallel**), from the evaluation setting plus divergence reachability. Five deployment patterns fall out of the grid.

Also indexes both docs in `src/ccl/design/README.md`, noting they describe a model rather than a pass, and points at them from the **Validation and program branching [Open]** bullet in `docs/design.md`.

These are the first ```mermaid blocks in the repo and nothing validates them, so `ci.sh` and `.github/workflows/ci.yml` carry TODOs for a `ci_mermaid` gate: what it should check, and why the syntax half needs a Node dependency the repo does not currently have. All 13 diagrams were verified to parse against Mermaid 11.16.1 by hand for this change.


src/ccl/design/branching.md
```diff
@@ -0,0 +1,488 @@
...
+
+Cambra can mirror both kinds. For an unreplayable source the mirrored window is bounded by a configurable **source cache**: the runtime retains what it fetched, so a non-owning arm consumes the same elements without a second advance. Exceeding the window is what closes the mirrored configurations to a program, per "Flows, not states".
+
+### Sinks
+
+**Mirrored arms require idempotency, suppression, or sandboxing.** Two arms cannot generally drive one sink, so the non-owning arm's effects must either be indistinguishable when happening twice, be directed somewhere isolated, or — where the runtime holds the endpoint — never be emitted at all, which is "Applying ownership at the boundary" below. Which of the three is available is a property of the sink, not of the deployment.
+
+**Promotable and non-promotable effects.** A sandboxed or suppressed effect is only useful if it can later be *made authoritative*, when a test arm becomes production:
```
dpmills: This claim needs clarification.  It's still useful for validation purposes

src/ccl/design/branching.md
```diff
@@ -0,0 +1,488 @@
...
+
+- **Promotable** — output files, a durable datastore. The sandbox can become the new source of truth, which is what makes shadow → canary → reverse shadow a *progression* rather than three unrelated setups.
```
dpmills: This doesn't seem quite right.  For non-promotable, shadow → canary → reverse-shadow is the right flow (assuming the output is partitionable).  For promotable, you can do shadow -> reverse-shadow -> promote directly, since you can fully validate the output

src/ccl/design/branching.md
```diff
@@ -0,0 +1,488 @@
...
+
+The gap that opens in the second case is easy to misname. It is *not* v1 against v0 — that difference is the entire purpose of the exercise. It is **v1-sandboxed against v1-in-production**: three programs, not two. The $8 charge might be declined where the $10 charge was not, by a fraud rule or a currency minimum, so v1's shadow history records a completed order where prod-v1 would record a cancelled one. The infidelity is the stub diverging from the external world, not v1 diverging from v0.
+
+Two mechanisms look like they should rescue this and do not:
+
+- **Idempotency does not apply.** It says the same call twice is harmless; here the two calls differ, so there is no "same call".
+- **The source cache does not extend.** It works by handing over bytes that were genuinely fetched, and for a request nobody sent there are none.
+
+By the correspondence under "Sinks", this gap exists exactly where the sink is non-promotable: a promotable sandbox is a real instance and answers truthfully, merely truthfully of a different world.
```
dpmills: This whole section feels incomplete.  What's the final story here?

src/ccl/design/branching.md
```diff
@@ -0,0 +1,488 @@
...
+
+An **unreplayable** source holds state, and advancing it changes the world: consuming from a Kafka partition moves an offset, and a source holding a receiver cannot hand the same element to two callers. Even where the advance is a separate call from the fetch, it is conceptually part of the read. So **reading can be an effect**, and effect ownership has to be defined at sources too, not only at sinks.
+
+Cambra can mirror both kinds. For an unreplayable source the mirrored window is bounded by a configurable **source cache**: the runtime retains what it fetched, so a non-owning arm consumes the same elements without a second advance. Exceeding the window is what closes the mirrored configurations to a program, per "Flows, not states".
+
+### Sinks
+
+**Mirrored arms require idempotency, suppression, or sandboxing.** Two arms cannot generally drive one sink, so the non-owning arm's effects must either be indistinguishable when happening twice, be directed somewhere isolated, or — where the runtime holds the endpoint — never be emitted at all, which is "Applying ownership at the boundary" below. Which of the three is available is a property of the sink, not of the deployment.
```
dpmills: I don't think idempotency is actually sufficient here.  idempotency only means that two identical side effects are the same as just one of them, but there's no guarantee the two mirrored branches will produce _identical_ side effects for the same logical input

groundlar: Good callout, this needs clarification--idempotency is only sufficient if the actions are identical.

src/ccl/design/img/execution-model.svg
dpmills, line 1: Does the reconvergence here make sense? How can we reason about a shared fragment when inputs to it are different?

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+> **Status: [Decided] for internal state, [Sketched] for anything touching sources and sinks.** The axes, the two laws, the per-register history derivation, and the four well-formedness conditions are settled and carry the table below. Sections on the frontier carry their own banners; "Promotion and its operations" and "Oracle" are **TODO**.
+
+## Vocabulary
+
+- **Arm** — one version of the program. Two arms are one tree with the differing sites selected, so they share everything they have in common. Where the diagrams say **prod** and **test**, those are v0 and v1. Two arms are emphatically *not* two running programs, and "The execution model" below is what that costs and buys.
+- **Position** — where a unit of work sits in a domain the program already has: a commit on the transaction clock, or an element of an input stream. Every decision below is a function *from* a position.
```
dpmills, line 8: "unit of work" or "piece of data"?

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+**What it buys structurally.** A divergent fragment has its own store by construction, so copy-on-write is a consequence of where the split sits rather than an overlay the runtime maintains and checks. The sandbox's *extent* is derived, not chosen — which is the claim "CoW applies to fragments, not programs" makes from the other direction.
+
+Whatever form the splitting operator takes, its semantics have to say where effect ownership sits, since the duplication it introduces is exactly the duplication the two laws constrain.
```
dpmills, line 34: "splitting operator" is a bit concerning; do we actually need an operator for this? I would hope not, because that means actively modifying an existing running program in order to do a shadow

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+**Program-defined *selection* is harder than program-defined *ownership*.** Under *total* evaluation an ownership predicate is ordinary agreed-region computation: both arms run regardless, so the predicate is evaluated by logic they share. An evaluation *selector* has a bootstrap problem — something must decide who runs before anything runs, so the predicate has to be evaluable by the runtime, or by an agreed prelude ahead of dispatch. This is the same shape as the argument under "Parallel histories require total evaluation" that mirroring buys its way out of generalized merging: another thing the cost of mirroring pays for.
+
+## History, per register
+
+The derived axis, defined here because the table below reports it. Three modes: **agreed** is decided by the compiler outright, and the choice between the other two follows from the evaluation setting alone.
+
+The available patterns are decided by **divergence reachability**, which under this model stops being a cost optimization and becomes the load-bearing analysis.
+
+- **agreed** — no divergence is upstream, so both arms write the register identically. One store.
```
dpmills, line 71: Is "agreed" useful to have in this breakdown? If we're in a non-divergent region, then maybe we shouldn't even think of there being two "arms" there

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+- **It does not partition.** Reads cross arms while the deployment runs, so compatibility is a precondition of running at all. With per-user state and a random v0/v1 shard, every request must see the *real* per-user state, so under compatibility v1's state should deliberately **not** be sandboxed.
+- **It does partition.** No read ever crosses arms, so nothing is required to run. Compatibility becomes the precondition of *reverting* a slice: the migration that moves tenant A back to v0 has to produce records v0 can read.
+
+It is also the precondition a merge of two parallel histories needs, per "Promotion and its operations" — the compiler can say the second case above is unsafe to merge before anyone tries. Note that CoW/sandboxed semantics are safe in *both* cases; compatibility is about rejoining, not about running.
```
dpmills, line 135: Do we ever want to merge state?

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+So a register's adoptability follows the promotability of the effect it records, which makes this the same correspondence as the one under "Sinks" rather than an independent rule.
+
+## The configuration space
+
+Five configurations, over *evaluation* × *ownership*, with history derived. Under *selected* evaluation ownership is forced, so the only remaining freedom is the selector's domain — the commit clock or a property of the input. Under *total* evaluation ownership is free across its own domains, less the one that is not a configuration at all: a clock-valued ownership function is a *move* between two configurations, so it appears in "Flows, not states" as an arrow rather than here as a row. That leaves constant v0, constant v1, and a property of the input.
+
+The **answers** column reports whose output becomes external truth for two users, Foo and Bar, Bar being the one who gets the new version — the experiment arm, or the canary. They are the same pair throughout [branching-sequence-diagrams.md](branching-sequence-diagrams.md) and in "Concretely: canary" below.
+
+| pattern | evaluation | ownership | answers (Foo / Bar) | history of diverged registers | rollback fidelity |
```
dpmills, line 164: This categorization assumes that ownership at the output layer is partitionable according to properties of the input.  That's not true in general, and I don't think we need to assume it.  We can frame it as partitioned based on properties of the output instead.  For the cases where the partitioning flows through unchanged, it's equivalent, and for cases where it doesn't, it still works

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+deactivate v1
+```
+
+Every mirrored configuration has this shape. The rest of [branching-sequence-diagrams.md](branching-sequence-diagrams.md) draws the others.
+
+## Authoritative state: promotion and rollback
+
+The table above enumerates *states*. This section is about **moving** between them, which always means moving ownership — making some arm's history **authoritative**. The two directions have different characters: **promotion** is constrained by what the boundary allows, since an effect that cannot be made true cannot be promoted, while **rollback** is available wherever a parallel history exists but is lossy by a measurable amount. Which moves are legal at all comes first.
+
+### Flows, not states
```
dpmills, line 236: The concept of Flow is spread out all over the doc.  I think we should define states and flows as early as possible, with lists of both.  This section makes it sound like deployment is the only type of flow, but then at the very end of the doc we start talking about testing etc.

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+"State compatibility" is the precondition: the merged store has to be one the surviving program can read, and the compiler can say a case is unsafe to merge before anyone tries. "External-effect-recording registers" is the other constraint and it is orthogonal — it decides *which* registers may travel at all, independently of how the ones that travel are combined.
+
+**Cumulative registers are the hard case, and overwrite there is wrong rather than merely imprecise.** A cumulative register is not partitionable — it is a fold over every position into one cell — so pointwise selection has nothing to select against. Worse, under a ramp *neither* arm's fold is authoritative: with v0 owning 95% of orders, v0's `revenue_total` folds v0's prices over **all** orders and v1's folds v1's prices over all orders, while the true figure is v0's prices on the 95% it owned plus v1's on the 5% it did. No arm computed that. Recomputation is the only correct answer whenever the owned contributing set changed during the window.
```
dpmills, line 354: Previously, we only talked about ownership and promotion of outputs, but this is about registers.  Is ownership actually broader than just side effects?  This seems to be a bit fuzzy in several places through the doc

groundlar:
# Vs Rainbow Deploys

Evaluating this doc against [rainbow deployments](https://brandon.dimcheff.com/2018/02/rainbow-deploys-with-kubernetes/): n versions of a service kept alive simultaneously, each identified by commit SHA, because long-lived stateful connections need 24–48h to drain naturally. New connections go to the newest version; old versions keep serving their existing connections until those close, then are deleted manually.

## Verdict

The model **accommodates** rainbow deployments but does not **support** them, and the doc never names the pattern. Every mechanism rainbow needs exists in some form, but three of them exist only in a two-arm reading, and rainbow's defining property is that n versions are live at once. The gaps are all in the same place: the doc treats "two arms" as the general case with n as a footnote, and rainbow is the workload that makes n load-bearing.

## Where rainbow lands in the doc's own table

Each connection is handled by exactly one version, so evaluation is *selected* by a property of the input, which derives **mixed** history and no restorable world — the **live experiment** configuration. The doc's advice for that configuration applies verbatim and is good advice: make version identity the *ownership* function rather than an evaluation selector, so every arm keeps evaluating everything and the old arm's store stays current. Reverting a cohort of connections becomes a rollback rather than a migration. That is a materially better story than the Kubernetes version, where a retired Deployment's state is simply gone.

The article's resource complaint — 128 pods to keep eight versions alive — is answered directly by the fragment execution model: what multiplies is the diff, not the program. Subject to the second gap below.

Worth separating two variants, because only one is hard. The **queue-pinned** form — pin the old code to queue A, drain it, migrate to queue B — makes the selector "which source did this element arrive on", which is envelope-defined and fits the model with nothing added. The **connection-pinned** form in the article is the one that exercises the gaps below.

## Gaps, in order of how much they'd change the doc

**The source cache window is ambiguous, and the ambiguity decides the whole verdict.** "Not every path is open" closes the mirrored configurations to a program that outruns its source cache. Read one way, the window must span the deployment's duration — rainbow's 24–48h drain means it can never shadow or canary, and it is stuck in mixed history with no fallback. Read the other way, the window need only cover the *skew* between arms consuming the same element, since under total evaluation both consume concurrently — and then connection lifetime is irrelevant and the mirrored path is wide open. The doc doesn't distinguish these. For rainbow the difference is everything.

**The divergence base for more than two arms is unstated.** "n arms cost one program plus n − 1 diffs" holds for a fan from one shared base. Rainbow's versions form a *chain* — each SHA diffed against its predecessor, eight generations deep. If reachability is measured against the oldest live arm, the divergence-reachable set grows monotonically with generations and by the eighth version approaches everything, which is this doc's own limiting case from "CoW applies to fragments, not programs" with the split at the source. The cost claim degrades toward n× precisely in rainbow's regime. The doc needs to say what the base is, and restate the cost for chained rather than fanned divergence.

**Ownership with affinity isn't in the domain list.** Ownership draws from a constant, the commit clock, or a property of the input. Rainbow's is none of those purely: it is "whichever arm was current when this connection opened, held constant for the connection's life." It *is* expressible — a connection-id-to-arm assignment written in the agreed region at connect time is program-defined ownership, and it satisfies the undiverged ownership predicate trivially. But that exercises something the doc never shows: an ownership function reading state written at an earlier position. Everything as written reads as though ownership is memoryless, and under the memoryless reading rainbow is inexpressible.

**No sink property governs the granularity at which ownership may vary.** The three static properties are whether the call yields a consumed value, whether it is promotable, and whether it is partitionable; `http.serve`'s response channel "consumes nothing and partitions per request". A long-lived connection partitions per *connection*, and a mediator that re-evaluated ownership per message would move a live WebSocket between arms mid-stream. Nothing in the model forbids it. Rainbow exists to prevent exactly that. This wants a fourth property, or a stated requirement that the ownership domain be constant over the partition unit.

**State compatibility is stated pairwise; rainbow needs it n-way.** "Each arm can read what the other wrote" — with eight live versions and a non-partitioning register (room membership, message history), the requirement is all-pairs mutual, and the failure that bites is non-adjacent: v0 against v7, when operators only ever checked adjacent expand/contract steps. This is the single most valuable claim the compiler could make about a rainbow deployment, and it is the one thing rainbow operators currently handle entirely by convention. The condition should quantify over all live arms.

**Retirement has no precondition.** The flow diagram carries `retire v0` as a bare arrow. Drain detection is the article's one explicitly unsolved problem — cleanup is manual, by operator judgement about acceptable reconnection impact. Cambra can answer it: with ownership held in the mediator over an agreed assignment register, "arm k is retirable" is "the preimage of k under the ownership function is empty", which the runtime evaluates rather than an operator guessing. That is a concrete differentiator sitting unclaimed. It also connects to the open question of whether promotion is itself a position on the commit clock — retirement needs the same answer.

Related and smaller: rainbow's ramp has no operator. The fraction on the newest version rises as new connections arrive and old ones decay; nobody sets it. The doc defines a ramp as a move an operator takes. The form fits, the agency doesn't.

**Rainbow is a deployment topology; the shared-fragment saving is an execution-topology claim.** "Executed once" assumes arms are co-resident, and "Implications for the implementation" concedes that arms deployed as separate processes "do not in fact share the agreed execution", bringing idempotency back as a concern. Rainbow is inherently one Deployment per version. So either Cambra co-locates arms — and then rainbow becomes *unnecessary*, which is the stronger pitch, since you get many live versions without n deployments — or arms are separate processes, the agreed fragment runs n times, the cost claim collapses to the conventional n×, and what is left is the static analysis. A reader arriving from Kubernetes will assume the second. Worth stating the position outright.

## Suggested shape

Rainbow probably belongs in "Configurations that are not deployments" as a reduction — the live experiment configuration with connection-latched selection, or better, the canary configuration with connection-latched ownership and n arms — with its extra requirements named: the assignment register must be agreed, compatibility quantifies over all live arms, and retirement is gated on an empty ownership preimage. That is four sentences plus whatever the n-arm cost claim turns out to be, and it would let the doc claim a pattern the deployment world currently has no good story for.

src/ccl/design/branching.md
groundlar: Fixed. The two examples are different kinds of thing — a commit is a unit of work, a stream element is a piece of data — so whichever noun went there was wrong for one of them. Dropped the noun instead:

> **Position** — an index in a domain the program already has: a commit on the transaction clock, or an element of an input stream.

src/ccl/design/branching.md
groundlar: Discussed in slack — capturing it here so the thread closes on the current position rather than the one that prompted the comment.

Your read there is out of date: avoiding inserted operators for lifecycle ops mattered less once hot reload landed, and whether a split is modelled as operator insertion or as a richer notion of data ports on the operators already there is an open question whose details don't bite yet.

Reworded to say that, and to answer the worry that prompted it — the split is derived from the diff, so neither version's source carries it and shadowing does not mean editing the program under test:

> How the runtime realizes a split is open: an inserted operator, or a richer notion of data ports on the operators already there. Either way it is derived from the diff rather than written into either arm, so neither version's source carries it. What the realization must say is where effect ownership sits, since the duplication the split introduces is exactly the duplication the two laws constrain.

src/ccl/design/branching.md
groundlar: Keeping the term, because two places need a name for the case: "Sandbox containment" makes a fabricated reply reaching an **agreed** or **mixed** register a hard error, and the configuration table reports only diverged registers, which is sayable only because the undiverged ones are *agreed* and stay in one shared store.

You're right about the framing, though — *agreed* is the absence of arms rather than a third peer mode, and the section only half-said so. Made it explicit in the bullet:

> **agreed** — no divergence is upstream, so both arms write the register identically. One store, written by the shared fragment. The two arms are two readings of a tree that coincides here, per "The execution model", so *agreed* names the absence of the question the other two modes answer rather than a third peer setting.

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
+How to slice the design space of running two versions of a program at once: what an operator sets, what the compiler derives, and which familiar deployment patterns fall out. This is the model of record; [branching-sequence-diagrams.md](branching-sequence-diagrams.md) draws the configurations concretely.
```
dpmills, line 1: This doc predates the hot-reload doc.  Run a pass to deduplicate content and make sure that terminology is consistent between them.  I think that will end up with substantial changes to this, so I've only left a few other comments so far

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+> **Status: [Decided] for internal state, [Sketched] for anything touching sources and sinks.** The axes, the two laws, the per-register history derivation, and the four well-formedness conditions are settled and carry the table below. Sections on the frontier carry their own banners; "Promotion and its operations" and "Oracle" are **TODO**.
```
dpmills, line 3: "register" is gone, use specific existing terminology like mutable variable

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+
+Two laws and four conditions, all static properties of the program, and all of one kind: they say which configurations mean anything. Everything below this section is read against them.
+
+### Ownership is grounded
+
+The owner at a position must be an arm that evaluates at that position — you cannot answer with logic you did not run. So under *selected* evaluation ownership is **forced** (only one arm ran, so only one candidate exists), and it is free only under *total* evaluation. This is why the mirrored patterns read as one design with a parameter while the selected ones read as separate designs.
+
+### One value per key per position
+
+Two arms may both write a key if and only if they provably agree — then it is one write proved twice, not two writes. They may disagree only with isolated stores (**parallel**, one store per arm). This is what makes things like a shared inventory safe under mirroring: both arms decrement it identically, so there is no double-selling.
```
dpmills, line 103: This feels like too strong of a condition.  In the case where we have "mixed" upstream, the state really might not be partitioned by key.  For a really simple case, we could just have a global counter incremented on every input, so there's no way to even define "agreement per key".  That scenario seems fine though

The title says "per position" but the text doesn't.  If it is per position, then my previous paragraph doesn't hold, but then I think this becomes vacuous because the two arms will only process the same position in the case that they have branched state anyways

src/ccl/design/branching.md
```diff
@@ -0,0 +1,576 @@
...
+# Unsafe: v0 enforces that names are capitalized and v1 relaxes it,
+# so v0 cannot safely operate on v1's data
+v0 writes {Name: "John", LastName: "OCallahan"}
+v1 writes {Name: "John", LastName: "oCallahan"}
+```
+
+Whether the selector partitions the register's key space decides **when** this bites, which is the one thing partitioning still decides after "Parallel histories require total evaluation". This is **partitioned evaluation** in the sense of "Vocabulary", and it takes two things: the register must be partitionable, and the selector must cut along its keys. The user-id hash above has the first and lacks the second; the tenant-id selector has both.
+
+- **It does not partition.** Reads cross arms while the deployment runs, so compatibility is a precondition of running at all. With per-user state and a random v0/v1 shard, every request must see the *real* per-user state, so under compatibility v1's state should deliberately **not** be sandboxed.
+- **It does partition.** No read ever crosses arms, so nothing is required to run. Compatibility becomes the precondition of *reverting* a slice: the migration that moves tenant A back to v0 has to produce records v0 can read.
```
dpmills, line 133: I thought we didn't want to partition in the shadowing case; that requires extra state migration logic

src/ccl/design/branching.md
groundlar: Agreed, and done. The move that made it coherent was to split the two axes rather than restate both: **evaluation stays a function of the input, ownership becomes a function of the output.**

- The axes table now gives ownership `a constant, the commit clock, or a property of the **output**`.
- The shared-domains paragraph is replaced. Evaluation reads the input because something has to decide who runs before anything has run; ownership reads the output because it is applied where an arm's output meets the endpoint, at the mediator.
- "Who computes the ownership function" is re-based on the output. Envelope-defined is now explicitly your "partitioning flows through unchanged" case — an envelope property is available before dispatch *and* carried to the output, so the two framings agree there. Program-defined becomes a function of what the arm produced, which is strictly broader than the old "parsed input, and possibly of state".
- The **canary** row reads `by output property`; **live experiment** keeps `by input property`, because that cell is an evaluation selector.

Breaking the axes' symmetry turned out to be a gain. The old paragraph justified it as "what makes the configurations below look like a grid", which is cosmetic, while the section below it had to *argue* that program-defined selection is harder than program-defined ownership. Under the output framing that asymmetry is structural and needs no argument.

One consequence worth flagging, since it changed a **[Decided]** condition. If ownership is a property of the output, and a divergent arm's output is divergent by definition, then "Undiverged ownership predicate" is violated on its face. The repair is to have the condition name what actually has to be agreed — **the position key the predicate reads**:

> Where v0 and v1 write the same key — a request id, a tenant, an order number — ownership partitions, and the value beside that key may diverge as freely as it likes. Where the arms write different keys there is no common position for the mediator to arbitrate, so ownership cannot be partitioned at that sink at all; that case is the **whole-sink** column of "What canary is for", reached from the other direction.

That is strictly more accurate than the input version — it names the key rather than the whole predicate — and it lands on the same distinction the sink `partitionable` annotation already encodes, which is some evidence it is the right restatement rather than a patch.

src/ccl/design/branching.md
groundlar: Added a **States and flows** section directly after "The axes", carrying the definitions and both lists, and shrank the opening of "Flows, not states" to a back-reference so the definitions live in one place.

Your second point — that deployment reads as the only kind of flow — turned out to be the more useful half, because the fix is a distinction the doc was already making without naming:

- **Lifecycle flows** move a version toward or away from being the only one running: entry, flip and ramp, promotion and rollback.
- **Terminal flows** enter a state, do a job, and tear down with no version becoming the only one running: ad-hoc query, testing and prod shadowing, ephemeral test environment.

Terminal is not a lesser case, and "What promotion requires" already said so without the vocabulary to land it — a program whose sinks are all non-promotable can run *only* these, and the job still gets done. That section calls indefinite shadowing "a legitimate terminal configuration rather than a stuck one", so the word is taken from the doc rather than coined here.

The states list is a bare orientation table (state, evaluation, ownership) pointing forward to "The configuration space" for the derived history, answers and fidelity columns, so the full table stays the single place those are set out.

`Replication` is deliberately off the terminal list — that section is an open question about which machinery earns its keep, not a configuration.

Note the states table says `by output property` for canary, so it reads against the change in the ownership-partitioning thread.

PR #67, opened 2026-08-06
mutability-conditionals 7/7: writer decision as a `commit | `abort variant; retire F_COMMIT

dpmills:
## Writer decision as a ``{`commit{𝑃} | `abort}`` variant

PR 7/7 of the mutability-conditionals stack. Replaces the writer decision record's `commit : Bool` field with a **decision variant** ``{`commit{𝑃} | `abort}``, where `𝑃` is the dense write/reply payload record `{writes, to_<defer>*}`. A committing position carries the payload; `` `abort `` is the nullary whole-transaction deny. Both the transaction writer and the induction writer build it through `ccl_utils::wrap_decision_variant`, and the runtime (`body_decision_at`) decodes the tag.

Making the grant/deny the *tag* rather than a sibling field leaves "denied yet real writes" unrepresentable — the payload only exists on the branch that commits.

The arm names are lowercase because an arm is a *constructor*, and `Caps` in CHL means a type — the convention `BaseType::keyword` already documents. `V_COMMIT` / `V_ABORT` are the only place the two strings live; `wrap_decision_variant`, `commit_project`, `commit_payload_ty` and the runtime decode all read them.

Three enum variants named `Commit` are **not** tags and keep their casing: `TxnBinding::Commit` (which letrec binding a group's binding is), `StoreReadKind::Commit` (which engine backs a store), and the `CommitEngine`/`CommitOperator`/`CommitTs` family. `Commit` is overloaded in this subsystem — the transaction *mechanism* and the decision *tag* are different things that happen to share a word, and only the second one is an arm.

### The decision variant needs no canonical tag position

With union arms keyed by tag (6/7), ``{`commit{𝑃} | `abort}`` does not have to be *built* in a particular order for the runtime to decode it, so the convention that made that work goes away: `COMMIT_TAG_INDEX`, the `variant_tag_index` lookup, the `debug_assert` that a decision type places `` `commit `` at the canonical index, and `commit_operator`'s `COMMIT_TAG` position. `commit_project` names the tag it projects and the decode asks `is_commit_tag`, so the two ends agree by name.

This removes a coupling that was live in both directions: the constant's own doc noted it "deliberately does **not** follow the `BTreeMap` tag order inference materializes inferred variants in", which meant a decision variant reaching a pass through inference rather than by hand could silently desync the read. There is nothing left to desync.

`type-inference.md` drops the caveat about hand-built variants fixing their own positional order, and the `commit_operator` fixtures name `` `commit ``/`` `abort `` instead of `0`/`1` — as does `mut_elim`'s recognition assertion, which reads ``variant_wrap(`commit)`` and so states what the recognizer actually requires.

### Pass-level tree walks grow the stack

The second commit is fallout from the first. The decision variant wraps each writer body in more nesting than the record it replaced, and that pushed `both_arms_feed_in_committing_block` past a test thread's stack.

The depth is the trigger, not the cause. Every pass that walks the tree recurses one frame per node, and each frame is sized for the union of a `match` over *every* node kind — so these walks have always been one deep-enough program away from overflowing. `lambda_elim` was alone in guarding itself with `stacker::maybe_grow`; the same guard now sits at the entry of the other per-node walks:

- `channelize::desugar` and `channelize::extract_for_defer`
- `infer::emit::emit_node` and `infer::solve::coalesce_node`
- `infer::check::check_node`
- `infer::solver::constrain::constrain_go` (structural descent over two types)
- `interpreter::operator_conversion::convert_impl`

Found by running the suite under `RUST_MIN_STACK` far below CI's stack, which names one walk at a time; each site here is one the stress run actually reached.

### Note on this PR's history

This branch previously had PR #16, which GitHub auto-closed when its base branch (the tag-keying branch, #53) was deleted after being squashed into #15. A closed PR whose head has since been force-pushed cannot be reopened, hence the new number. The branch has been rebuilt on the current #15: its two stale commits — the superseded `VariantProject`/`VariantWrap` 6/7 and the tag-keying commit — are both absorbed into #15's reworked 6/7 and were dropped.

Green under `./ci.sh`.


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14
                    - #15
                        - #67 ◀
                            - #17
                                - #68

src/ccl/design/mutability.md
```diff
@@ -828,15 +830,17 @@ Every fan-out shares one first-match encoding, `πᵢ = 𝑔ᵢ ∧ ¬⋁ⱼ˂
 `ccl_utils`, shared between channelize, the value fan-outs, and the transaction path walk).
 
 A **conditional induction write** takes a different, simpler route than the value fan-out: it rides
-the induction accumulator's own `{commit, writes}` **decision** rather than a union of restricted
-sources. `lower_loop_body_chain` lowers `if 𝑝: acc += e` to a statement-position `Case`
+the induction accumulator's own `[.Commit(⟨writes⟩) | .Abort]` **decision variant** rather than a union of
```
sortalongo: Is this square brackets the real syntax for variants? I expected it would be parens or braces. The double parens is also confusing—looks like noise. These notations seem to have permeated throughout the codebase, unfortunately.

tests/compilation_pipeline/transactions.rs
```diff
@@ -1927,3 +1927,76 @@ fn heterogeneous_multi_key_store_reads_string_key() {
...
+
+/// **Cost #1 — off-path partial op does not fault.** A guarded `//` in a
+/// conditional transaction write: `if d != 0: acc := acc // d`. The divisor `d`
+/// starts at 3 and is zeroed on the first commit, so the second request's guard
+/// (`d != 0`, read on the snapshot) is false — an `.Abort`. The write value `acc
+/// // d` rides the lazy `⧺ filter_values(d != 0) ≫ (acc // d)` arm the value-`Case`
+/// compiles to, so the division is evaluated only where its guard holds — never at
+/// the `d == 0` position. Were the partial op evaluated off-path it would panic on
+/// divide-by-zero. First commit: `acc := 100 // 3 = 33` (and `d := 0`); the second
+/// aborts, so `acc` stays 33.
```
sortalongo: What does "cost" refer to here? What's described sounds like correct behavior, so this label is confusing.

sortalongo (approved):

src/ccl/design/mutability.md
dpmills: Fixed to the new syntax

tests/compilation_pipeline/transactions.rs
dpmills: Yeah, this was stale.  Edited for brevity and to remove "cost" framing

PR #68, opened 2026-08-07
Say "final" for a completed history's last entry, "latest" for most-recent

dpmills:
`last` reads three ways at once in this codebase — the **final entry of a terminating history**, the **most-recent value so far**, and the **positional end of a slice**. The mutability work has all three live in the same paragraph: a read of a `Txn` register is an arbitrary as-of *latest*, while a read of a drained induction accumulator is a *final*. Naming both of them "last" is what made the register's missing terminal read easy to mistake for one that already exists.

This is a behavior-preserving rename splitting the two temporal senses apart:

- `ExtractLast` → `ExtractFinal` (`extract_last.rs` → `extract_final.rs`), including its internals.
- `TransactWriter::last_emit` → `latest_emit` — a most-recent-emit idempotency key, not a final one.
- Doc comments and design prose across `commit_operator.rs`, `transact_phase.rs`, `tile.rs`, `mutability.md`, `ir.md`, `design-operators.md`, and `chl-spec.md`: each "last" resolved to whichever it actually meant.
- Test programs no longer name a `Mut(_, Txn)` register `last` — it is `latest`, which is what an as-of read of it yields.

Left alone deliberately: the `last-write-wins` / `last-tag-wins` / `last-def-wins` merge laws (established terms where "last" is the point), the `split_last` / `is_last` / innermost-last slice and scope idioms, and "the last statement of a block" (the spec's own vocabulary, and positional in source rather than temporal).

No behavior change, so no new tests; `./ci.sh` is clean.


dpmills:

- #10
    - #11
        - #12
            - #13
                - #14
                    - #15
                        - #67
                            - #17
                                - #68 ◀

sortalongo (approved):

PR #69, opened 2026-08-07
CHL syntax decisions: variants, `where` refinements, destructuring; braces are always a product

dpmills:
Records four CHL surface-syntax decisions in `docs/chl-spec.md`, and implements the three of them that turned out to be small — one of which was covering a live bug.

## Decided, not implemented

**Refinements** (new §6.4) are written `{T where p(_)}`. `where` replaces the earlier `|` separator, and `_` *is* the refined value rather than a named binder. Moving off `|` is what frees that token for variant alternation, and dropping the binder is what keeps the data-refinement surface distinct from the function-contract surface (`assert`s, §6) — a dependent codomain refinement has to name what it closes over, so the anonymous `_` deliberately cannot express one. CCL's `{T | p}` rendering is unchanged; the surface `_` lowers to its `__elem` binder.

**Variants** (new §6.5) are backtick-tagged in every position, with the tag case fixed by the existing `Caps` means type rule: ``{ `some{Int} | `none }`` as a type, `` `some(1) `` as a term, `` case `some(x) `` as a pattern. Types use `{…}` and terms/patterns `( … )`, following the delimiter split (§2.4). A tag's `{…}` *is* its field list, not a nested type, which is the brace elision — `` `some{a: Int} ``, never `` `some{{a: Int}} ``.

**Destructuring patterns** (new §4.3.1) get the section they were missing. One grammar serves assignment targets, `for` binders, and `case` arms; patterns mirror the term-level constructors, annotations ride any node, and `:` binds tighter than `,` so an annotated bare tuple pattern needs no parens (`x: Int, y: Int = tuple`).

Two questions are flagged **[Open]** rather than papered over: the block syntax of a `match` inside a refinement predicate, where §1.4 suppresses `INDENT`/`DEDENT` and the arms have to be delimiter-derived instead — the concrete pressure behind a single-line `match` — and, unchanged, pattern matching's block syntax generally.

The north-star programs move to the decided syntax (`` `some ``/`` `none `` tags, `{Int where _ >= 0}` store refinements). Because the backtick is not lexed and lexing is a whole-file pass, it becomes the *first* blocker in `txn_kv`, `nonneg_inventory`, and both `storefront` versions, ahead of the `->` map entry and the `match`-block indentation failure that were pinned before; the pins and their blocked-at prose are updated to match.

## Implemented

**Braces in type position are always a product, never grouping.** A one-element product carries the trailing comma — `{T,}` — exactly as the term-level `(e,)`/`(e)` split does, and a comma-free `{T}` is a parse error pointing at `{T,}`. The brace parser now *captures* the trailing comma rather than `.allow_trailing()`-ing it away, since it is the token that distinguishes the form.

**`{}` is the unit type** — the product of zero types — and its only CHL spelling. `Unit` and `None` are not accepted in annotation position and are not *reserved* either: they are ordinary undefined identifiers, free for a program to bind. The one asymmetry that stays, documented at both ends: `BaseType::from_keyword` still accepts `Unit`, because it is the CCL primitive-name round-trip and CCL renders the type by that name; `name_type`, the CHL annotation reader, is a strict subset of it.

**The unit value is `()`.** `Token::None` and `ast::Lit::None` are gone. A borrowed `None` both collided with the term/type capitalization rule (§6.1) and did double duty as the type; a product constructor says what the value *is*, and matches its type `{}`.

**Empty tuples and records are not valid types.** `Type::Tuple([])` and `Type::Record([])` are now invalid representations: a product with no fields has no keys to distinguish positional from named keying, so both descriptions name the same one-inhabitant type, and that type is unit.

That last one was a live bug rather than a tidiness rule, and it is why `()` is safe to adopt as the unit value. Two spellings for one type fail to *reconcile*: the solver coalesced an empty record variable to `Tuple([])` while `product`'s vacuously-all-`Name` test rebuilt the same node as `Record([])`, and the post-inference consistency wall — which compares a node's recorded type against one rebuilt from its children — rejected the disagreement with the self-contradictory `Type mismatch for type of (): expected (), found ()`. The empty product term did not compile at all. The invariant is now structural rather than remembered: `Type::tuple` / `Type::record` map the empty case to `Unit`, every site building a product from a variable-length collection goes through them, and a `debug_assert` at the post-inference wall catches any path that names the variant directly.

`()` typechecks now. It still fails at runtime, on the pre-existing gap that a unit-valued program output has no column representation — the identical panic `x = None` produced before this branch. §12 records that as the runtime limitation it is, rather than leaving it invisible behind a value nobody wrote.

## Two pre-existing diagnostic gaps

Both surfaced because the new `{T}` error is the first *common* user error to travel that path, and both made it unusable:

- **`Rich::custom` messages were dropped on the floor.** `ParseErrorInfo` had nowhere to put them, so a rule that parses a well-formed token sequence and then rejects what it *means* rendered as a bare `found end of input` — `found`/`expected` are both empty for a custom error. They now ride `ParseErrorInfo::custom` and replace the derived text at both rendering sites. This also restores the message on the pre-existing mixed-brace error (`{a: Int, Bool}`), which no user had ever seen.
- **ariadne panicked on inverted context spans.** chumsky hands back an `.as_context()` span whose start is after its end for an error raised inside a `validate` (the context opens ahead of the error's own offset, giving e.g. `3..2`), and ariadne aborts the entire report on one. Every label span now goes through `label_range`, which collapses an inverted span to an empty one at its start — still pointing the "while parsing …" note at the right place.

## Notes

Nothing in the corpus used `Unit` or `None` in annotation position, and no `.cambra` program used `None` at all, so the migration was confined to the lexer, the `Lit` enum, and a handful of test cases.

`./ci.sh` is green — all four clippy configurations, the doc build, `doc_refs`, and every test.


dpmills:

- #69 ◀

docs/chl-spec.md
sortalongo: I think it's silly to try to make `5` mean both a term and a fully-refined type. Is there a reason to adopt such a confusion notation?

docs/chl-spec.md
```diff
@@ -1407,15 +1562,17 @@ marked one carries its status per "How to read this document".)
...
+surface form is `{T where p(_)}` (§6.4), not writable yet.
 
 > **Direction [Decided] — function contracts are asserts.** There is
 > no refinement syntax at a function definition site: a function's
 > contract is written as `assert` statements in its body, lifted to
 > refinement types in CCL. The two canonical shapes: precondition
 > asserts at the top of the body, referencing parameters, lift to
 > refinements on those parameters (`assert qty > 0` makes the domain
-> `{qty: Int | qty > 0}`); an assert on the result variable
+> `{qty: Int | qty > 0}` — the lift's results are shown here in CCL's
```
```suggestion
> `{qty: Int where qty > 0}` — the lift's results are shown here in CCL's
```

missed a rewrite. Worth another pass to make sure others weren't missed.

docs/chl-spec.md
```diff
@@ -1550,6 +1712,185 @@ Sketched at the same **[Tentative]** level:
...
+> block syntax is **[Tentative]** in any case (§1.6).
+
+Naming a refinement's subject is *not* how a function states a contract:
+a function's preconditions and postconditions are `assert` statements in
+its body, lifted to refinements over the parameter binders (the Direction
+note in §6). That lift needs the parameter's *name* in the predicate —
+`{qty: Int | qty > 0}`, in CCL's named-binder notation — which is exactly what
+the anonymous `_` cannot express, and exactly why the two surfaces are
+separate: `where`/`_` is for **data** types, `assert` is for **function**
+contracts.
```
sortalongo: This is an unnecessary distinction. Asserts could be lifted to use `_` like `{{qty: Int} where _.qty > 0}`

docs/chl-spec.md
```diff
@@ -1550,6 +1712,185 @@ Sketched at the same **[Tentative]** level:
...
+§6.3). `Result` is the same shape with `` `ok ``/`` `err ``.
+
+Variants are matched with `match`/`case` (**[Tentative]** as to block
+syntax — §1.6), or destructured directly against a single-tag variant type
+where the match cannot fail (§4.3.1).
+
+CCL has the sum type already (`Type::Variant`, a tag→payload list keyed by
+name or by position) and renders it `[.some(Int) | .none]`; that internal
+notation is unchanged. Nothing in CHL constructs one today — no backtick is
+lexed, and neither `match` nor a tag is parsed.
```
sortalongo: It's confusing to have a different notation for CCL types like this. Let's have the spec say that we'll converge the two over time.

docs/chl-spec.md
```diff
@@ -1550,6 +1712,185 @@ Sketched at the same **[Tentative]** level:
...
+constructors that map the empty case to `Unit`, and a `debug_assert` at the
+post-inference wall catches any path that bypasses them.
+
+This is an invariant and not a convention because two spellings of one type do
+not merely look untidy — they fail to *reconcile*. Inference records a node's
+type and a later consistency check rebuilds it from the node's children, so if
+those two steps can reach a zero-field product by different routes they disagree
+about a node that is perfectly well typed — and the disagreement is
+undiagnosable, since both spellings render `()`. One representation is what makes
+the unit value `()` typecheck at all.
```
sortalongo: This justification is confusing, and seems off-base. I think the real justification is subtyping: empty tuple and empty record are coherent ideas, but letting Unit be them leads to what appears to be silent data loss due to subtyping.

docs/chl-spec.md
```diff
@@ -1550,6 +1712,185 @@ Sketched at the same **[Tentative]** level:
...
+
+Both are therefore names a program may bind like any other identifier —
+`Unit = 1` is a legal binding. Nothing is reserved against a future use of
+either.
+
+There is **no empty type** (no uninhabited / void type) in CHL or CCL: the
+base types are `Int`, `UInt`, `String`, `Bool`, and `Unit`, and nothing else
+is atomic. `{}` is unambiguous because of this — the empty *sum* would be the
+void type, and there is none to spell. Introducing one would be new design,
+not a re-reading of `{}`.
```
sortalongo: rm irrelevant cruft

src/chl_parser/lexer.rs
```diff
@@ -53,8 +53,14 @@ pub enum Token {
...
     #[token("False", priority = 3)]
     False,
-    #[token("None", priority = 3)]
-    None,
+    /// The refinement predicate separator, `{T where p(_)}` (`docs/chl-spec.md`,
+    /// "6.4 Refinement syntax").
+    ///
+    /// Lexed so the name is **reserved**: refinements are not parsed yet, so
+    /// every use is a parse error today. That is the point — a program cannot
+    /// bind `where` as an identifier and then break when the syntax lands.
```
sortalongo: rm

src/chl_parser/parser.rs
```diff
@@ -315,30 +314,38 @@ where
...
+        //     record type nor a tuple type
         //
         // Parsing items as `expr (":" expr)?` keeps a single committed brace
         // parser, so classification is a total function of what was matched.
+        //
+        // The trailing comma is captured rather than silently allowed, because
+        // it *distinguishes* a one-element tuple type: `{T,}` is the one-tuple
+        // and a comma-free `{T}` is an error, mirroring the term-level
+        // `(e,)`/`(e)` split. Braces are not grouping in type position, so
+        // there is no reading of `{T}` left over to accept.
```
sortalongo:
```suggestion
        // The trailing comma on `{T,}` is required.
```

sortalongo (approved):

docs/chl-spec.md
dpmills: It makes complex CCL asts easier to read, but it doesn't belong in the spec.  I removed mentions of it here; we can decide about CCL rendering separately

docs/chl-spec.md
dpmills: CCL still renders with |, but it's not useful to discuss CCL in the spec like this, so removed this and other similar places

docs/chl-spec.md
dpmills: Removed this entirely

docs/chl-spec.md

docs/chl-spec.md
dpmills: rm

src/chl_parser/parser.rs
dpmills: Done

src/chl_parser/lexer.rs
dpmills: done

docs/chl-spec.md
dpmills: Rewrote this section to fix this and generally make it clearer

PR #70, opened 2026-08-07
CLAUDE.md: move PR-description guidance into a `pr-description` skill

dpmills:
PR descriptions here read as investigation transcripts: they open mid-mechanism and restate rationale a design doc already carries. This adds a `pr-description` skill — summary first (why / what / so what), then a review guide, a sub-linear length budget, three revision passes — and leaves `CLAUDE.md` a one-line pointer, so the guidance loads when a description is being written instead of riding in every session's context. Backtested by regenerating descriptions for five merged PRs (11 to 6,000 changed lines) from their diffs alone: 27–72% shorter than what was posted, cutting harder as the PR grows.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -185,30 +185,7 @@ When you are using compact, focus on test output and code changes.
 - When making changes, verify the freshness of the local repo by fetching and comparing the diff. The following commands do this, if there are differences, warn the user and ask the user whether they would like to pull or rebase.
    1. `git fetch origin`
    2. `git log master..origin/master --pretty=format:"%h%x09%an%x09%ad%x09%s"| head -n 20`
-
-### Commit descriptions are PR bodies — author them as Markdown
-
-PRs here are often created with `gh pr create --fill`, so the
-PR body is the commit description *verbatim*, the PR body squashes to become
-the `main` commit message. GitHub renders that body as GitHub-Flavored Markdown,
-so write commit descriptions as Markdown,not as traditional column-wrapped git prose:
-
-- **Do not hard-wrap inside a paragraph.** Let each paragraph be one
-  soft-wrapped line and separate paragraphs with a blank line — otherwise every
-  wrap point renders as a ragged `<br>` on GitHub.
-- **Backtick every code construct** — type/identifier names, file paths,
-  field and flag names (`` `CompiledProgram` ``, `` `dead_code` ``,
-  `` `src/ccl/lineage.rs` ``).
-- Use real Markdown for structure where it helps: `-` bullet lists for
-  enumerations, blank lines between paragraphs.
-
-### Updating PR descriptions
-
-`gh pr edit --body` fails with exit code 1 due to a Projects (classic) deprecation warning, even when the body update would otherwise succeed. Use the REST API instead:
-
-```bash
-gh api repos/OWNER/REPO/pulls/PR_NUMBER --method PATCH --field body="..." --jq .number
-```
+- Writing a PR description — or a commit description, which `gh pr create --fill` copies into the body verbatim — is the `pr-description` skill's job (`.claude/skills/pr-description/SKILL.md`): summary first, then a review guide, sub-linear length, the Markdown mechanics, and the `gh api` workaround for editing a posted body. Load it rather than working from memory.
 
 ### Stack Management & Rebasing (jj vs git-spice)
 
--- a/.claude/skills/pr-description/SKILL.md
+++ b/.claude/skills/pr-description/SKILL.md
@@ -1,0 +1,60 @@
+---
+name: pr-description
+description: Write or revise a pull-request description, or the commit description that becomes one (`gh pr create --fill` copies it verbatim). Use when creating a PR, writing a commit message that will become a PR body, or editing an existing PR body.
+---
+
+# PR descriptions
+
+A PR body is a map for the reviewer.
+
+## The rules
+
+**Open with a summary that stands alone** — one paragraph, before the first heading: **why** (the problem in terms the reviewer can understand), **what** (a concise description of the change), **so what** (the impact of the change). A reader who stops there should be able to say what the PR does and why it exists. An opening that starts mid-mechanism fails that test; that sentence is the second paragraph's job.
+
+**The description is a review guide.** Order sections the way the change should be read: the core mechanism of the change first, then its consequences, then test coverage and docs changes. One idea per section, and name a section for the idea rather than for the files it touches — a section per changed file is noise.
+
+**Focus on the *changes*. Push long-lived information to the docs.** Some content belongs only in a PR description: what connects the old and new versions of the code — why the old shape was wrong, what a reviewer should be suspicious of, how to read the diff. Everything else belongs in the codebase itself, with the description carrying a short summary of it. Terms and concepts the PR introduces get a one-line gloss so the body still reads on its own.
+
+**Where the body and a doc would say the same thing, cite and summarize in a sentence.** One sentence handing off to a doc or a code pointer beats a restatement: the restatement drifts from the doc by the next change, and then two places disagree. Use repo-relative Markdown links, which GitHub resolves in a PR body, and point at a heading rather than a section number — `§3.9` rots at the next reordering, which is why the repo's reference conventions require a heading. `./ci.sh doc_refs` does **not** check descriptions (nor anything under `.claude/`), so confirm by hand that the heading exists and the anchor spells it correctly.
+
+This applies hardest to the diff's *own* comments and docs. A change that explains itself in a new code comment, doc comment, or design-doc section has already published its reasoning; the body states the change and points at where the reasoning lives, in a clause rather than a paragraph. Restating a comment the same diff adds is the most common way a small PR ends up with a long description.
+
+**Limit the size of the description.** It scales *sub-linearly* with the size of the change: a description is the first few levels of a tree, each level pointing at the next and carrying exponentially more detail than the last. The larger the PR, the larger the fraction of self-describing content that has to live in the codebase instead.
+
+Compute the budget before drafting, from the diff's changed lines (`additions + deletions`) — about 10 × √lines:
+
+| Diff | Description |
+|---|---|
+| 1 line | ~10 words (one sentence) |
+| 10 lines | ~30 words |
+| 100 lines | ~100 words |
+| 1,000 lines | ~300 words |
+| 10,000 lines | ~900 words |
+
+Under about 100 words the description is the summary paragraph and nothing else — no headings, no `## Tests` section. Headings start earning their keep when there is a reading order to impose.
+
+**Cut what neither motivates nor orients** — investigation narrative (wrong hypotheses, the order in which things were discovered) and re-narration of what the diff already shows. Compression must not cost accuracy, though: re-check every name, count, and list that survives against the diff.
+
+## Author the body as Markdown
+
+PRs here are often created with `gh pr create --fill`, so the PR body is the commit description *verbatim*, and this repo squash-merges, so the body becomes the `main` commit message. GitHub renders it as GitHub-Flavored Markdown, so write commit descriptions as Markdown, not as traditional column-wrapped git prose:
+
+- **Do not hard-wrap inside a paragraph.** Let each paragraph be one soft-wrapped line and separate paragraphs with a blank line — otherwise every wrap point renders as a ragged `<br>` on GitHub.
+- **Backtick every code construct** — type/identifier names, file paths, field and flag names (`` `CompiledProgram` ``, `` `dead_code` ``, `` `src/ccl/lineage.rs` ``).
+- Use real Markdown for structure where it helps: `-` bullet lists for enumerations, blank lines between paragraphs.
+
+## Revise before posting
+
+Write three revisions before posting. A description is read by every reviewer and then survives as the commit message, so it earns the extra passes:
+
+1. **Draft** from the diff.
+2. **Check against the rules.** Does the opening stand alone? Is the section order a reading order? Is anything durable stated only here — or restated here from a comment the diff itself adds? Then count the words and compare against the budget; if the draft is over, the cut comes from what the codebase already says, not from the reviewer's map.
+3. **Check against the diff.** Re-verify every name, count, and list. A compression pass introduces factual errors more readily than a draft does — an over-cut description that says a pass is affected when the diff exempts it is worse than a long one.
+
+## Posting and updating
+
+`gh pr edit --body` fails with exit code 1 on a Projects (classic) deprecation warning even when the update succeeds. Use the REST API:
+
+```bash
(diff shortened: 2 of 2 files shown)


CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
    1. `git fetch origin`
    2. `git log master..origin/master --pretty=format:"%h%x09%an%x09%ad%x09%s"| head -n 20`
 
-### Commit descriptions are PR bodies — author them as Markdown
+### PR descriptions: summary first, then a review guide
```
groundlar: Nit: might consider moving to a skill for less permanent context injection

groundlar (approved):

CLAUDE.md
sortalongo: Agreed

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+A PR body is a map for the reviewer, not a write-up of the investigation that produced the change.
```
sortalongo:
```suggestion
A PR body is a map for the reviewer.
```

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+**Open with a summary that stands alone** — one paragraph, before the first heading: **why** (the problem, as a symptom the reader already cares about, not the internals producing it), **what** (the change as the new rule or invariant, not a list of edits), **so what** (what is now true, possible, or gone). A reader who stops there should be able to say what the PR does and why it exists; an opening that starts mid-mechanism fails that test.
```
sortalongo:
```suggestion
**Open with a summary that stands alone** — one paragraph: **why** (the problem in terms the reviewer can understand), **what** (a concise description of the change), **so what** (the impact of the change).
```

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+**Structure the rest as a review guide.** Each section is a stop on the reader's path: it names the file, symbol, or pass it covers, says what changes there, and says what to check. Order them the way the change should be read — the core mechanism, then one section per area that had to move (a file a reviewer would be surprised to see in the diff earns a sentence, not silence), then `## Tests` and `## Docs`.
```
sortalongo: I think this instruction is too fine-grained. It will make Claude want to have a section for every file that's changed. That's not useful.


```suggestion
**The description is a review guide.** Order sections the way the change should be read, with early sections describing the core mechanisms of the change, then describing test coverage, and docs changes.
```

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+**No information may live only in the PR description.** A body is squash-merged into a commit message, so afterwards it is reachable through `git log` and nowhere else. Every durable claim — an invariant, a rejected alternative, a cost argument — must already exist in the code or a design doc; the body is a view onto that material, never the only copy. A paragraph with nowhere to point is a doc comment you have not written yet: write it, then point at it. A complicated change handled well leaves behind more design-doc text and a *shorter* body.
```
sortalongo:
```suggestion
**Focus on the _changes_. Push long-lived information to the docs.** Some content belongs only in a PR description: that which connects the old and new versions of the code. All other content should live in the codebase itself, with the PR description providing only short summaries. Important terms and concepts introduced by this PR should be introduced briefly in the description to facilitate readability.
```

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+**Where the body and a doc would say the same thing, cite and summarize in a sentence** rather than retelling it condensed — a retelling drifts from the doc by the next change, and then two places disagree. One sentence handing off to [type-inference.md](src/ccl/design/type-inference.md#keying-a-specialization) beats a section reconstructing the argument it already makes; a section that opens "worth reading the module docs for" and then restates them has broken this rule, not followed it. Use a repo-relative Markdown link, which GitHub resolves in a PR body; `./ci.sh doc_refs` does **not** check descriptions, so confirm the heading exists and the anchor spells it correctly before posting.
```
sortalongo:
```suggestion
**Where the body and a doc would say the same thing, cite and summarize in a sentence**. One sentence handing off to a doc or code pointer is better than a re-statement of the doc. Use repo-relative Markdown links, which GitHub resolves in a PR body.
```

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+
+**A section is one idea, about 150 words.** The limit is on the section, never the body: a change with ten moving parts earns ten short sections, while the same change as four 400-word walls is where a reviewer stops reading. A section that will not fit was two ideas.
+
+**Cut what neither motivates nor orients** — investigation narrative (wrong hypotheses, the order of discovery) and re-narration of what the diff already shows. Compression must not cost accuracy, though: re-check every name, count, and list that survives against the diff.
+
```
sortalongo:
```suggestion

Write 3 revisions of the PR description before posting. With each edit pass, check for compliance with the above guidelines, and remove irrelevant and repeated content.

```

PR descriptions are particularly important for improving our productivity, and deserve elevated attention.

CLAUDE.md
```diff
@@ -186,7 +186,23 @@ When you are using compact, focus on test output and code changes.
...
+**Where the body and a doc would say the same thing, cite and summarize in a sentence** rather than retelling it condensed — a retelling drifts from the doc by the next change, and then two places disagree. One sentence handing off to [type-inference.md](src/ccl/design/type-inference.md#keying-a-specialization) beats a section reconstructing the argument it already makes; a section that opens "worth reading the module docs for" and then restates them has broken this rule, not followed it. Use a repo-relative Markdown link, which GitHub resolves in a PR body; `./ci.sh doc_refs` does **not** check descriptions, so confirm the heading exists and the anchor spells it correctly before posting.
+
+**A section is one idea, about 150 words.** The limit is on the section, never the body: a change with ten moving parts earns ten short sections, while the same change as four 400-word walls is where a reviewer stops reading. A section that will not fit was two ideas.
```
sortalongo: Ironically, this PR's description dramatically fails a pretty obvious test: it's longer (675 words) than the PR diff itself (~500). While not limiting description length makes sense abstractly, I think AI's verbosity makes not having any length guide at all a recipe for absurdly-long descriptions.

```suggestion
**Limit the size of the description.** The size of a PR description should scale sub-linearly with the size of the change. As PRs get larger, a larger fraction of self-descriptive content should live in the codebase, keeping descriptions to a manageable size.

Rules of thumb: 
- a 10 word (1 line) PR should have a 10 word description.
- a 1000 word (100 line diff) PR should have a ~100 word description.
- a 10,000 word (1k line diff) PR should have a ~300 word description.
- a 100,000 word (10k line diff) PR should have a ~900 word description.

If you need more words to properly describe the PR, that content belongs in the codebase.

```

sortalongo (changes requested):
I made a bunch of suggestions that better fit my idea of good PR descriptions. Of course, this is just speculation, so I'd appreciate if you could try them out to see if they actually improve the descriptions.

CLAUDE.md
dpmills: Done — the guidance now lives in `.claude/skills/pr-description/SKILL.md`, and `CLAUDE.md` keeps a one-line pointer in **Git Conventions** (the pointer earns its place: the skill also owns the `gh api` workaround for editing a posted body, which is easy to hit without loading the skill).

Discovery held up in the backtest: across 20 headless runs the model reached the skill every time — 19 via the `Skill` tool, and one run that read `SKILL.md` directly after following the `CLAUDE.md` pointer. It also stayed loaded long enough to be applied, not just read: one run computed its own diff's `additions + deletions` with `awk` to get the length budget.

CLAUDE.md
dpmills: Taken verbatim.

CLAUDE.md
dpmills: Taken verbatim. I kept one sentence of the original after it: a reader who stops there should be able to say what the PR does and why it exists, and an opening that starts mid-mechanism fails that test. That failure test was the part that measurably changed behavior when I first wrote this, so I did not want to drop it.

CLAUDE.md
dpmills: Adopted — you were right that the per-file version pulls the wrong way. I kept a half-sentence making the point explicit rather than merely not-asked-for, since the failure mode is what a model reaches for by default: *name a section for the idea rather than for the files it touches — a section per changed file is noise.* None of the 20 backtest runs produced a per-file section.

CLAUDE.md
dpmills: Adopted. Backtesting it turned up one addition worth making, so this rule is your text plus a paragraph:

> This applies hardest to the diff's *own* comments and docs. A change that explains itself in a new code comment, doc comment, or design-doc section has already published its reasoning; the body states the change and points at where the reasoning lives, in a clause rather than a paragraph. Restating a comment the same diff adds is the most common way a small PR ends up with a long description.

That came out of the measurement, not from taste. The first draft of the rule said "docs", and on small PRs the generated descriptions kept overshooting the length budget by 3-4× — every time, the overshoot was a paragraph restating a code comment **the same diff had just added**. #49 is the clean case: an 11-line diff whose 8 added lines are a comment explaining why bare `set -e` cannot work there, and the description re-explained it in 60 words. Adding the paragraph above took #49 from 147 to 94 words and #35 from 191 to 144.

CLAUDE.md
dpmills: Adopted. Two things kept from the original, both operational rather than motivational: `./ci.sh doc_refs` does not check PR descriptions (so anchors have to be confirmed by hand before posting), and a citation names a heading, not a section number.

The section-number half was added because of the backtest — a run on #48 emitted `docs/chl-spec.md §3.9`, which is exactly what **Referencing docs and sections** in `CLAUDE.md` rules out. With the clause in place no run emitted a `§` reference.

CLAUDE.md
dpmills: Adopted, as a `## Revise before posting` section, with your framing that descriptions deserve elevated attention. I gave the three passes distinct jobs, because the passes are where the earlier measurements said the risk concentrates:

1. **Draft** from the diff.
2. **Check against the rules** — including whether anything here is restated from a comment the diff itself adds — then count the words against the budget. When over, the cut comes from what the codebase already says, not from the reviewer's map.
3. **Check against the diff** — re-verify every name, count, and list, because a compression pass introduces factual errors more readily than a draft does.

Pass 3 is not ceremony: the earlier round of this experiment had a run that, while cutting to a word budget, listed a compiler pass among those affected when the change explicitly exempted it. Compression is exactly when accuracy breaks.

CLAUDE.md
dpmills: Adopted verbatim, including the four rules of thumb, and the irony is fixed: this PR's description is now 96 words against a 92-word budget for its own 85-line diff, and per the "no headings under ~100 words" rule it is a single paragraph.

Your curve turns out to be almost exactly `10 × √lines`, which is in the skill as the interpolation rule, since three of the four anchors need interpolating in practice.

**How it measured.** Five merged PRs spanning 11 to 6,072 changed lines, each regenerated in a fresh headless session from a sandbox holding the post-change tree with **no `.git`** — so the diff is the only account of the change and the real description cannot leak. 2-4 reps each; "posted" is the description actually merged.

| PR | diff lines | budget | posted | with the skill | reps | cut |
|---|---|---|---|---|---|---|
| #49 | 11 | ~33 | 190 | **94** (2.8×) | 97, 97, 81, 101 | −51% |
| #35 | 74 | ~86 | 261 | **144** (1.7×) | 145, 136, 132, 165 | −45% |
| #48 | 407 | ~202 | 458 | **333** (1.6×) | 385, 318, 295 | −27% |
| #50 | 1,650 | ~406 | 1,173 | **498** (1.2×) | 462, 444, 588 | −58% |
| #19 | 6,072 | ~779 | 1,998 | **562** (0.7×) | 596, 440, 649 | −72% |

The sub-linear shape is the part that clearly works, and it works in the direction you argued for: the bigger the PR, the harder the rule cuts — −27% at 400 lines, −72% at 6,000 — and at 6,072 lines it lands *under* budget. #19's 1,998-word description becoming ~560 words is the result I would point at.

**Where it does not comply: the low end.** Under ~100 lines it settles at 1.7-2.8× budget. The residue is not padding; it is the *why*. #49 is an 11-line diff, but the 33-word budget cannot state why bare `set -e` fails inside `ci_all`, and the runs would not go below ~90 words while keeping that.

**A loosening I tried and reverted.** I replaced the two sub-100-line rows with prose — "at most a paragraph, and only as long as the *why* costs" — expecting the same result with an honest rule. It got measurably worse, 3 reps each:

| PR | numeric anchors | prose |
|---|---|---|
| #49 | 97, 97, 81 (mean 92) | 122, 130, 139 (mean **130**) |
| #35 | 145, 136, 132 (mean 138) | 195, 156, 205 (mean **185**) |

40% longer at #49, 34% at #35, no overlap between the arms. **A number binds and prose does not** — even a number that gets missed by 3× beats a qualitative version of the same instruction, which is the same result your original review predicted for the qualitative advice in the first draft. So your table stayed as written. Read the low-end rows as a gradient the model pushes against rather than a line it hits, and the top three rows as live.

One structural note, out of scope here: `./ci.sh doc_refs` skips `.claude/`, so the skill's own prose is unchecked. Removing `.claude` from `SKIP_DIRS` finds one real defect (a bare `CLAUDE.md` citation is ambiguous — there are three) plus two false positives in `staff-rust-review/SKILL.md`, where a `.md` path followed by a closing quote parses as an unterminated citation title. Worth a follow-up; it needs a checker fix, not just an edit.

dpmills:
Tried them out. All seven suggestions are in, six of them verbatim; the guidance moved to `.claude/skills/pr-description/SKILL.md` per @groundlar, with a one-line pointer left in `CLAUDE.md`.

Method: five merged PRs from 11 to 6,072 changed lines, each description regenerated in a fresh headless session against a sandbox that holds the post-change tree with no `.git`, so the diff is the only account of the change. 2-4 reps each. Results are [in the length-limit thread](https://github.com/cambra-dev/cambra/pull/70#discussion_r3751283915) — descriptions came out 27-72% shorter than what was actually merged, cutting harder as the PR grows, and #19 went from 1,998 words to ~560.

Two findings worth your attention there: the length rule does not reach budget under ~100 lines (1.7-2.8× over, and the residue is the *why*, not padding), and when I tried replacing those two rows with prose it got 34-40% **worse**, so your table stayed as written. Your "focus on the changes" rule also grew one paragraph from the measurement — restating a comment the same diff adds was the single largest source of overshoot on small PRs.

This PR's own description is now 96 words against a 92-word budget.

sortalongo (approved):
so much better!!!

PR #71, opened 2026-08-10
Freshening asks a different question than `type_level`, so it needs its own

dpmills:
A specialization clone must share no live inference state with the definition it was cloned from. That invariant breaks when a quantified variable is reachable only through a **refinement predicate**: `freshen_above` short-circuits on `type_level`, which is predicate-blind (it reports a `Refinement`'s base level), so a type whose only above-cutoff content sits in a nested predicate is returned verbatim — still pointing at the definition's own variables. The symptom is a duplicate specialization, the clone reaching one generalized `let` both through freshened variables and through the original's.

The fix is a second level function, not a change to `type_level`, whose predicate-blindness is load-bearing for its other callers (extrusion, bound-recording scope, `should_generalize`); `freshen_level`'s doc comment states the two questions. No measured cost.

### DIFF
--- a/src/ccl/infer/solver/scheme.rs
+++ b/src/ccl/infer/solver/scheme.rs
@@ -13,8 +13,6 @@ use crate::ccl::subst::Subst;
 use crate::ccl::ty::{FunKind, FunKindVar, FunKindVarId};
 use crate::ccl::{Bound, InferVar, InferVarId, Level, Refinement, Type, TypedExpr};
 
-use super::type_level;
-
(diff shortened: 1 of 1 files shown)


dpmills:

- #71 ◀
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79

sortalongo (approved):

PR #72, opened 2026-08-10
`SharedHole`: let lowering state that two type positions are the same

dpmills:
A `groupby`'s key type *is* its key function's codomain, and the lowered shape relates them nowhere: `__gb_k`'s only occurrence is as an operand of the partition predicate's `==`, so the key's type could only arrive backwards through that comparison — making group-by key inference depend on an operator's internals.

`Type::SharedHole(id)` lets lowering state the relation outright: a `Hole` with an identity, every occurrence of one id normalizing to the same inference variable. It is transient exactly as `Hole` is, and groups with it everywhere outside `normalize_annotation`.

`groupby` stamps one id on the key application and on its own `data_fun` annotation's domain — the positions the claim is about, not the binder this desugaring invented. That also makes the edge directional for free: `bind_annotation` records `inferred <: ann` and a function is contravariant in its domain, so keys flow *into* the domain rather than being forced equal to it.

`symbolic` does not render annotations, so `test_lower_groupby` cannot show the relation; `test_groupby_key_type_comes_from_the_key_function` pins it.

### DIFF
--- a/tests/type_check.rs
+++ b/tests/type_check.rs
@@ -585,6 +585,31 @@ fn test_collection_union_heterogeneous_rejected() {
 // GroupBy + aggregate tests
 // ---------------------------------------------------------------------------
 
+// A group-by's key type is its key function's codomain, and the lowering says so
+// **directly** rather than leaving it to be recovered through the partition
+// predicate's `==`.
+//
+// `__gb_k`'s only occurrence in the lowered shape is as an operand of that
+// comparison, so without a stated relation its type can only arrive backwards
+// along the operand requirement that relates a comparison's two sides — making a
+// group-by's key inference depend on an operator's internals. One
+// `Type::SharedHole` states it, carried by the key application and by the domain of
+// the group-by's own `data_fun` annotation; these cases pin that the key resolves
+// to the key function's result type and not to the collection's element type.
+//
+// The relation is **not** visible in `test_lower_groupby`'s snapshots, because
+// `symbolic` does not render annotations. These are the tests that cover it.
+#[rstest]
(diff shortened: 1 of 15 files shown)


dpmills:

- #71
    - #72 ◀
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79

tests/type_check.rs
```diff
@@ -585,6 +585,31 @@ fn test_collection_union_heterogeneous_rejected() {
...
+// group-by's key inference depend on an operator's internals. One
+// `Type::SharedHole` states it, carried by the key application and by the domain of
+// the group-by's own `data_fun` annotation; these cases pin that the key resolves
+// to the key function's result type and not to the collection's element type.
+//
+// The relation is **not** visible in `test_lower_groupby`'s snapshots, because
+// `symbolic` does not render annotations. These are the tests that cover it.
+#[rstest]
+#[case("groupby([1, 2, 3], \\x -> x)", int())]
+#[case("groupby([(a=1, b=\"w\"), (a=2, b=\"e\")], \\r -> r.b)", string())]
```
sortalongo, line 605: Does freshening handle this automatically? We need a test for a polymorphic function that uses `groupby` with different key types to make sure.

sortalongo (approved):

tests/type_check.rs
dpmills: Yes, it works as expected since we end up with Infer vars like normal.  Added some tests to verify

PR #73, opened 2026-08-10
Traits: state what an operator requires of its operands, not that they are the same

dpmills:
A polymorphic operator's requirement cannot be stated as a signature. The only relation a signature can put between two operands is that they *share a variable*, which also forces every other lattice dimension to agree — so one operand's refinement became a requirement on the other, and a refinement both operands carried survived onto a value the operator computed rather than selected. Arithmetic, comparison and negation now state a **trait** instead: unrelated inference variables per operand, plus an obligation recorded beside the graph. Because an operator's result is an ordinary variable rather than a marker standing for a computation, information still flows backwards through it — so a function typechecks without its call sites, and `(1 + 2) and True` is an ordinary bound conflict. The refinement strip in `apply_binary_scheme` goes away rather than being fixed: a syntactic peel does nothing for an operand still holding a `Type::Infer`.

`src/ccl/infer/solver/traits.rs` opens with the vocabulary; [type-inference.md](src/ccl/design/type-inference.md#traits) carries the design. One term to fix here, because it reads two ways: an **obligation** is a single claim with two halves, and neither alone is "the obligation" — *the operands are types some implementation accepts*, **and** *each associated position is what that implementation associates*.

## Traits associate types; they are not functions

A trait names **associated types** (Rust's term) rather than having *an* output, and a type is associated only when it **depends** on the types satisfying the trait. Three shapes result — unary with an `Output` (`Negatable`), binary with one (`Addable`), binary with **none** (`Equatable`/`Orderable`) — each exercised by ordinary programs (`each_trait_shape_types_a_real_program`).

## What an obligation determines

A candidate set shrinks as base types arrive, and an associated type is deposited once every survivor agrees on it — onto **associated positions only**, since the obligation is their sole source of information while an operand always has the program's own `left <: A` edge. How much is determined is therefore a property of the table, and shrinks as it grows *including for an associated type*: `\x -> x + 1` has result `Int` only because `Int` in the second position leaves one row, and `Addable(Float, Int) ⇝ Float` would open the result just as the parameter already is. Hence agreement, not uniqueness, as the deposit condition.

Each variant's doc states its shape — how many types it is over, and what it associates — and `Trait::arity`/`assocs` read that shape off the table so the prose cannot drift from the rows (`every_trait_has_a_consistent_shape`).

## Delivery

One invariant carries the mechanism — a concrete type reaching an operand variable must reach the obligation watching it — and the bound closure does *not* provide it. A variable's lower bounds are written in four places, delivery is wired into each, and all four are load-bearing: deleting one fails exactly its own case in `a_concrete_operand_reaches_its_obligation`. [The design doc](src/ccl/design/type-inference.md#delivery-the-watch-follows-the-edge) argues it; `verify_narrowing_is_complete` checks it on every program rather than trusting it.

## Consequences

`test_lambda_unapplied` asserts an open parameter with the codomain per case; `test_collect_multi_conflict` and `test_unary_neg_wrong_type` see `NoTraitImpl` naming the trait instead of a mismatch against a hardcoded domain. `a_register_that_reads_itself_still_gets_a_type` ports #51's recurrence cases, which pin something stronger here: narrowing consumes bounds as they are recorded rather than resolved types, so an obligation never enters the recurrence — the base it needs sits on the register's seed.

**`max`'s comparability tripwire flips**, closing `type-checker-traits-comparability`. `Comparable(γ)` on its codomain is the fourth shape — unary, associating nothing, since the scheme already returns an element of what it consumes.

### DIFF
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -684,7 +684,7 @@ The reason is that a literal knows more about itself than its base does, and tha
 
 What this changed is instructive, because refinements were rare enough before that several rules assumed their absence. Each was already wrong for a user-written refinement; literals are merely the first thing that makes them reachable.
 
-* **An operator does not *inherit* its operands' refinements**, and does not have to be *made* not to. A refinement is a fact about a value, so an operator that computes a new value cannot carry one over: `𝑥 + 𝑥` where `𝑥` is `2` produces `4`. Arithmetic, comparison and negation state their requirement as a [trait](#traits), over a variable per operand and per associated type — all unrelated — which leaves no path for an operand's refinement to reach the result by sharing. The remaining monomorphic operators (`and`, `++`, `not`) keep an ordinary scheme and pass their operands verbatim — nothing is shared with the result, so a refined operand simply flows into a concrete domain. Aggregates likewise keep theirs, since their operand is a *collection* whose refinements describe its domain and the rule must see them.
+* **An operator does not *inherit* its operands' refinements**, and does not have to be *made* not to. A refinement is a fact about a value, so an operator that computes a new value cannot carry one over: `𝑥 + 𝑥` where `𝑥` is `2` produces `4`. Arithmetic, comparison and negation state their requirement as a [trait](#traits), over a variable per operand and per associated type — all unrelated — which leaves no path for an operand's refinement to reach the result by sharing (`an_operator_result_carries_no_operand_refinement`). The remaining monomorphic operators (`and`, `++`, `not`) keep an ordinary scheme and pass their operands verbatim — nothing is shared with the result, so a refined operand simply flows into a concrete domain. Aggregates likewise keep theirs, since their operand is a *collection* whose refinements describe its domain and the rule must see them.
 
   Inheriting is not the same as **computing**, and only the first is ruled out. `{Int | __elem == 2} + {Int | __elem == 3}` genuinely *is* `{Int | __elem == 5}`, and a trait implementation is where such a rule would live, since it determines the output type rather than forcing it to be a position the operands already occupy. Today every implementation computes a base and stops — a property of the table, not of the mechanism. Two things would have to change to lift it: an implementation would need the operands' *types* rather than their bases, and the deposit would have to move to a point where those types are final. Eager deposit is sound for a base because a base never weakens, while a refinement set only shrinks as further lower bounds arrive — so a refinement computed from a partial view is too strong. A rule computing from resolved operands then meets recurrences (`x := x + 1` resolves its operand through its own output), where it must already be sound at the cut; and anything beyond constant folding and interval arithmetic needs predicate *implication*, which the lattice deliberately does not have (refinements match structurally — see this file's module-level note in `src/ccl/infer/solver/mod.rs`).
 * **A mutable register takes no refinement** from its initializer or from any single write. A register is not one value but the sequence its writes produce, so its value type is the join over all of them; taking one contribution's refinement would assert it never changes, which is what declaring it mutable denies. The rule holds at every place a register's value type is *built*, not just at the `:=`/`+=` rule: the `Transact` carrier's keys (where the seed is the value type's only lower bound, so an unstripped seed would resolve the register — and every read of it — to the seed's singleton), the recognition that builds that carrier, and the phase that reads the value type back off the seed binding.
@@ -1134,12 +1134,12 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
 
 ## Traits
 
-The constraint lattice can state that two positions are **equal** or **related by subtyping**. That is everything an operator needs when its result *is* one of its operands — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is a shared lattice position. It is not enough for an operator whose result is **computed from** its operands, and `+` is the smallest example: the sum of two values is neither of them. (Sharing one variable across operands and result — the signature this replaced — states the requirement in the one place it is wrong in both polarities at once; `an_operator_result_carries_no_operand_refinement` pins the consequence.)
+The constraint lattice can state that two positions are **equal** or **related by subtyping**. That is everything an operator needs when its result *is* one of its operands — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is a shared lattice position. It is not enough for an operator whose result is **computed from** its operands, and `+` is the smallest example: the sum of two values is neither of them.
 
 ### Vocabulary
 
-* A **trait** is a named requirement a type may satisfy — `Addable`, `Orderable`, `Comparable`. **A trait is not a type**: no `Type` variant, no lattice point, no subtyping edge, and the type grammar and `constrain_go`'s rules are untouched. Types *satisfy* traits.
-* An **implementation** is one row of a trait's table: the types it accepts, and the types it associates with them.
+* A **trait** is a named requirement a list of types may satisfy — `Addable`, `Orderable`, `Comparable`. **A trait is not a type**: no `Type` variant, no lattice point, no subtyping edge, and the type grammar and `constrain_go`'s rules are untouched. Types *satisfy* traits.
+* An **implementation** is one row of a trait's table: the types it accepts, and the types it associates with them. Written `Addable(Int, Int ⇝ Int)` — accepted types, then `⇝`, then the associated ones.
 * An **associated type** is a type a trait *names* — `Output`, the type an arithmetic operator's result takes. A trait is a requirement rather than a function, so it associates any number, **including none**. A type is associated only when it *depends* on the types satisfying the trait: a comparison's `Bool` is the same for every pair `Equatable` accepts, so it belongs to the operator's signature and `Equatable` associates nothing — recording it as an association would claim the trait determines something it does not.
 * An **obligation** is one recorded instance of a trait at specific type positions: one **operand position** per argument the trait takes, and one **associated position** per type it names. It is a single claim with two halves, and neither alone is the obligation: *the operand positions are types some implementation accepts*, **and** *each associated position is what that implementation associates*. Every position is an ordinary inference variable, unrelated to the others.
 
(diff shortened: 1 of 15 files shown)


dpmills:

- #71
    - #72
        - #73 ◀
            - #87
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79
                            - #91
                                - #92
                                    - #93
                                        - #97

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1048,67 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
 
 ---
 
+## Traits
+
+The constraint lattice can state that two positions are **equal** or **related by subtyping**. That is everything an operator needs when its result *is* one of its operands — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is a shared lattice position. It is not enough for an operator whose result is **computed from** its operands, and `+` is the smallest example: the sum of two values is neither of them. (Sharing one variable across operands and result — the signature this replaced — states the requirement in the one place it is wrong in both polarities at once; `an_operator_result_carries_no_operand_refinement` pins the consequence.)
```
sortalongo:
```suggestion
The constraint lattice can state that two positions are **equal** or **related by subtyping**. That is everything an operator needs when its result *is* one of its operands — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is a shared lattice position. It is not enough for an operator whose result is **computed from** its operands, and `+` is the smallest example: the sum of two values is neither of them.
```

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1048,67 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+### Vocabulary
+
+* A **trait** is a named requirement a type may satisfy — `Addable`, `Orderable`, `Comparable`. **A trait is not a type**: no `Type` variant, no lattice point, no subtyping edge, and the type grammar and `constrain_go`'s rules are untouched. Types *satisfy* traits.
```
sortalongo:
```suggestion
* A **trait** is a named requirement a list of types may satisfy — `Addable`, `Orderable`, `Comparable`. **A trait is not a type**: no `Type` variant, no lattice point, no subtyping edge, and the type grammar and `constrain_go`'s rules are untouched. Types *satisfy* traits.
```

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1048,67 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
 
 ---
 
+## Traits
```
sortalongo, line 1135: This implementation of traits is tightly scoped, which is good. But there's an obvious omission: the ability to define different function implementations according to the trait instance. e.g. string `+` and Int `+` are very different functions. That distinction is not captured by saying that `+` is either `{String, String} => String` or `{Int, Int} => Int`. That's handled by dynamic dispatch in the interpreter right now, but it's not what we want in the fullness of time.

It's worth being explicit about that here, because this description makes it sound like a trait is _only_ an association between types. What it should say is that a trait is _at least_ an association between types, but it can also associate those types with function definitions. We just don't support the latter yet.

One way to think about it is:
- A trait is an (N + A + F)-ary relation between (N + A) types and F functions, where the N (solving) types functionally determine the A (associated) types and the F functions.
- A trait instance is one hyper-edge (i.e. tuple) in that relation
- Trait instance solving consists in finding a hyper-edge that satisfies the types given by the inference process.
- We will eventually want to define a CHL language for defining these relations in an expressive, but ergonomic way.
- This implementation only supports N ∈ {1, 2}, A = 1, F = 0, with instances built-in to the compiler.

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1048,67 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+* An **obligation** is one recorded instance of a trait at specific type positions: one **operand position** per argument the trait takes, and one **associated position** per type it names. It is a single claim with two halves, and neither alone is the obligation: *the operand positions are types some implementation accepts*, **and** *each associated position is what that implementation associates*. Every position is an ordinary inference variable, unrelated to the others.
+
+An operator's signature is therefore `𝐴₁ → … → 𝐴ₙ → 𝑅` plus the obligation, for the trait's arity `𝑛`, where `𝑅` is either one of the associated positions or a type the operator fixes. The three shapes the operators take:
+
+| operator | signature | obligation |
+|---|---|---|
+| `+` | `∀ 𝐴 𝐵 𝑂. 𝐴 → 𝐵 → 𝑂` | `Addable(𝐴, 𝐵)`, `Output` at `𝑂` |
+| `==` | `∀ 𝐴 𝐵. 𝐴 → 𝐵 → Bool` | `Equatable(𝐴, 𝐵)`, nothing associated |
+| unary `-` | `∀ 𝐴 𝑂. 𝐴 → 𝑂` | `Negatable(𝐴)`, `Output` at `𝑂` |
```
sortalongo: I'd write the last column with associated types _in_ the trait name:

- `Addable(A, B ⇒ O)`
- `Equatable(A, B)`
- `Negatable(A ⇒ O)`

That makes the relation's structure more obvious.

src/ccl/design/type-inference.md
```diff
@@ -1046,6 +1048,67 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+Mechanism: `src/ccl/infer/solver/traits.rs`.
+
+Because an associated position like `𝑂` is an ordinary variable rather than a marker standing for a computation, information flows *backwards* through an operator's result and misusing that result is an ordinary diagnostic — `(1 + 2) and True` fails as a bound conflict. A computed-type marker cannot have this property: the solver cannot compare an unreduced computation against anything, so a function could not be typechecked without seeing its call sites.
+
+### Refinements are transparent
+
+`{𝑇 | 𝑝}` satisfies a trait exactly when `𝑇` does. This holds *by construction*: satisfaction is judged on each bound contribution as it arrives, with refinements peeled at that moment — when the base actually exists. An operand is usually still an inference variable at emission, so a peel performed *there* would have nothing to work on.
```
sortalongo: Is this a temporary, or permanent restriction?

sortalongo (commented):
Review in progress. Stopped at `## Discharge is incremental` for now

src/ccl/design/type-inference.md
dpmills: Added this discussion

src/ccl/design/type-inference.md
dpmills: Done

src/ccl/design/type-inference.md
dpmills: I think it should be permanent.  Allowing restrictions to participate in traits makes things significantly more complicated and much less likely to be decidable

src/ccl/design/type-inference.md
dpmills: Done

src/ccl/design/type-inference.md
dpmills: Done

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
 ---
 
+## Traits
+
+The constraint lattice can state that two positions are **equal** or **related by subtyping**. That is everything an operator needs when its result *is* one of its operands — `max(xs)` returns an element, `x.f` returns the field — because "is one of" is a shared lattice position. It is not enough for an operator whose result is **computed from** its operands, and `+` is the smallest example: the sum of two values is neither of them.
+
+### Vocabulary
+
+* A **trait** is a named requirement a list of types may satisfy — `Addable`, `Orderable`, `Comparable`. **A trait is not a type**: no `Type` variant, no lattice point, no subtyping edge, and the type grammar and `constrain_go`'s rules are untouched. Types *satisfy* traits.
+* An **implementation** is one row of a trait's table: the types it accepts, and the types it associates with them. Written `Addable(Int, Int ⇝ Int)` — accepted types, then `⇝`, then the associated ones.
```
sortalongo: Naming: I know "implement" is a well-established term for this, but I think "implementation" is overloaded to mean too many things (plus it's a mouthful). To simplify, I'd prefer to call these "trait instances", or "instance" for short.

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+* An **associated type** is a type a trait *names* — `Output`, the type an arithmetic operator's result takes. A trait is a requirement rather than a function, so it associates any number, **including none**. A type is associated only when it *depends* on the types satisfying the trait: a comparison's `Bool` is the same for every pair `Equatable` accepts, so it belongs to the operator's signature and `Equatable` associates nothing — recording it as an association would claim the trait determines something it does not.
+* An **obligation** is one recorded instance of a trait at specific type positions: one **operand position** per argument the trait takes, and one **associated position** per type it names. It is a single claim with two halves, and neither alone is the obligation: *the operand positions are types some implementation accepts*, **and** *each associated position is what that implementation associates*. Every position is an ordinary inference variable, unrelated to the others.
```
sortalongo: I don't understand what this term is supposed to represent. Implementations/instances are already defined above. But this paragraph says "one recorded instance of a trait". That sounds wrong given what follows. Maybe it should read "the set of requirements imposed on the instance resolved at a specific _usage_ of the trait"?

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+An operator's signature is therefore `𝐴₁ → … → 𝐴ₙ → 𝑅` plus the obligation, for the trait's arity `𝑛`, where `𝑅` is either one of the associated positions or a type the operator fixes. The three shapes the operators take:
```
sortalongo: This is overly constraining. 

```suggestion
A function can require a trait in its type signature: `f: T ⇒ U requires MyTrait(T, U)`, which means the compiler has to find a trait instance satisfying the given obligation.

The three shapes the current set of operators take are:
```

src/ccl/design/type-inference.md
sortalongo: I think that's true, but it also means not allowing refinements to propagate through trait-implemented operands. For example, knowing that `+: {Nat, Nat} ⇒ Nat` is monotonic would be useful, but it's not possible to do that unless the `Addable` associated type can have a refinement.

So I think we'll need to strike a balance. One idea is that refinements don't participate in trait instance resolution (i.e. different refinements on operand types never point to different instance "definition" (we need a syntax of instance definitions to make this statement meaningful, but I think it holds together)), but they are visible in the trait relation and help determine associated types.

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+
+A contribution arriving at a position is one of three things, and each has its own outcome:
+
+| contribution | example | outcome |
+|---|---|---|
+| a **base** | `Int` | narrows the candidate set |
+| **not determined yet** | a variable, a hole, a `Feed` handle whose payload arrives separately | nothing to say |
+| **determined and not a base** | a tuple, record, variant, function | rejected — no implementation can accept it |
+
+The third is a rejection and not silence, because "no base here" is true of both it and the second. A tuple that merely failed to narrow would leave `(1, 2) == (3, 4)` well-typed: a comparison has no associated position to strand, so nothing downstream would object either.
```
sortalongo: I think it's worth expanding on the topic of what is Equatable. tuples, records, and functions not being Equatable make sense. Variants I think could be Equatable. 

Either way, I'd split that discussion from how instance resolution works. The fact that only base types have traits right now is purely coincidental, not a fundamental aspect of resolution.

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+### Refinements are transparent
+
+`{𝑇 | 𝑝}` satisfies a trait exactly when `𝑇` does. This holds by construction: satisfaction is judged on each bound contribution as it arrives, and refinements are peeled at that moment, when the base exists. Peeling at emission instead would have nothing to work on, an operand usually being still a variable there.
+
+Transparency follows from incremental discharge and is permanent. A candidate set only ever shrinks ([Discharge is incremental](#discharge-is-incremental)), and a refinement is one of the things a bound can deliver late; if `{𝑇 | 𝑝}` could satisfy a requirement `𝑇` does not, a refinement arriving after the base would have to re-admit a dropped candidate, and order would start to matter.
+
+Growing `𝐹` above zero would not change this. Choosing between two functions by `𝑝` is dispatch on a fact about a *value*, which a table keyed on types cannot express — and which no refinement survives in any case: `𝑥 + 𝑥` where `𝑥` is `2` produces `4`.
+
+### Discharge is incremental
+
```
sortalongo: "discharge" is heavily overloaded, and there's an established term for this: trait/type class/instance "resolution".

src/ccl/design/type-inference.md
```diff
@@ -1130,6 +1132,113 @@ Recorded so a reader can tell a deliberate boundary from an oversight.
...
+| contribution | example | outcome |
+|---|---|---|
+| a **base** | `Int` | narrows the candidate set |
+| **not determined yet** | a variable, a hole, a `Feed` handle whose payload arrives separately | nothing to say |
+| **determined and not a base** | a tuple, record, variant, function | rejected — no implementation can accept it |
+
+The third is a rejection and not silence, because "no base here" is true of both it and the second. A tuple that merely failed to narrow would leave `(1, 2) == (3, 4)` well-typed: a comparison has no associated position to strand, so nothing downstream would object either.
+
+A position the program never determines is not a rejection: it is reported as an unresolved variable rather than as a missing implementation.
```
sortalongo: > A position the program never determines

I don't know what this means.

sortalongo (approved):

src/ccl/design/type-inference.md
dpmills: Yeah, something like that seem reasonable.  I don't think we have enough info to write anything about this yet; at some point we'll need a comprehensive plan around smarter refinement usage

src/ccl/design/type-inference.md
dpmills: Renamed throughout: `TraitImpl` → `TraitInstance`, `Trait::impls` → `Trait::instances`, `InferError::NoTraitImpl` → `NoTraitInstance` (its message now reads "No Addable instance for …"). The vocabulary entry defines *instance* as one row of a trait's table.

src/ccl/design/type-inference.md
dpmills: Agreed, and "one recorded instance" was doubly wrong — it also collided with the row sense of *instance*. It now reads: an **obligation** is what one *use* of a trait records — the demand that some instance fit the type positions at that use, one operand position per argument the trait takes and one associated position per type it names.

src/ccl/design/type-inference.md
dpmills: Taken. The general form is stated first — `𝑓 : 𝐴 ⇒ 𝐵 requires MyTrait(𝐴 ⇝ 𝐵)` — with the operator table demoted to the shapes the *current* operators take.

Worth noting obligations already ride variables into schemes, so a user-written function carries its operators' requirements today: `λ 𝑎 𝑏 → 𝑎 + 𝑏` is `∀ 𝐴 𝐵 𝑂. 𝐴 ⇒ 𝐵 ⇒ 𝑂 requires Addable(𝐴, 𝐵 ⇝ 𝑂)`. What is missing is only the *surface* for writing `requires`.

src/ccl/design/type-inference.md
dpmills: Renamed. "Discharge" was already spent in this repo — `Subst::discharge` for Pi binders, refinement and assert discharge, the mutability phases — so a trait sense of it would have been a third meaning rather than a reuse. The section is now "Resolution is incremental", which is also what the literature calls it.

src/ccl/design/type-inference.md
dpmills: Reworded to say what it meant: "A position that stays in the second row for the whole program — nothing ever determines it — is not a rejection either. Its obligation simply never narrows, and the variable is reported as unresolved rather than as a missing instance."

src/ccl/design/type-inference.md
dpmills: Split out — "What the tables hold" is now its own subsection, separate from how resolution works:

> Every instance in every table accepts **base types only**, and every one is homogeneous — `Addable(Int, Int ⇝ Int)`, never `Addable(Int, String ⇝ …)`. Both facts are the tables' content, not properties of resolution: nothing in narrowing or deposit assumes either. So `Equatable` rejecting a tuple, a record or a variant is what these rows happen to be, and not a judgement that such types are incomparable.

PR #74, opened 2026-08-11
`constrain_go` asserts the uniquely-keyed invariant it rests on

dpmills:
`constrain_go`'s record and variant arms look a key up with `iter().find(..)`, so on a duplicate-keyed product the verdict depends on which copy comes first — and disagrees with the function's own trivial-equality short-circuit, which accepts `t <: t` where find-first demands `Int <: Bool` between the duplicates. Nothing in `Type` enforces uniqueness; the builders merely happen not to violate it. This adds a debug-only check ahead of the short-circuit, since `t <: t` is the case the two answer differently. No program in the suite trips it, so a builder that starts violating the invariant now fails loudly instead of silently making subtyping depend on incidental structure.

### DIFF
--- a/src/ccl/infer/solver/constrain.rs
+++ b/src/ccl/infer/solver/constrain.rs
@@ -344,6 +344,58 @@ fn bridge_holder_gap(lo: &Subst, hi: &Subst) -> (Subst, Subst) {
     );
 }
 
+/// The **uniquely-keyed** invariant the record and variant arms rest on, checked
+/// where they rest on it.
+///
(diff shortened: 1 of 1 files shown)


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74 ◀
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79

sortalongo (approved):

PR #75, opened 2026-08-11
Update the value-`Case` fan-out's prose to the `DisjointJoin` it builds

dpmills:
Three prose sites still described the value-`Case` fan-out as it worked before the `CollectionUnion` split (#86), reconciling a coproduct against the `Case`'s type by `is_index_partition_of` — a helper #86 deleted. Two are test comments, one is `mutability.md`'s data-typed-selection bullet; all three now state what `lambda_elim::build_value_case_fanout` builds, a `DisjointJoin` over the one domain the arms share. `Σ` stays as the recorded design for arms at differing domains, which inference rejects today. Comments and doc prose only, no code changes.

PR #76, opened 2026-08-11
Canonicalize each arrow's Pi binder as the solver flattens types

dpmills:
A dependent function type's binder name leaks into two identities that are meant to be blind to it. `spec_key` excludes the binder from the key, but a dependent refinement predicate references it, so `(𝑥: 𝐷) ⤇ {Int | __elem == 𝑥}` and its `𝑦`-twin key apart and split a specialization two uses should share. Merging α-variant bounds keeps the first arrival's binder while unioning both copies of one predicate, coalescing to an arrival-order-dependent type whose second predicate references a binder that is no longer there. Both flattening walks now rename each arrow's binder, and every codomain reference to it, to the reserved `__pi{depth}` name, so α-equivalence coincides with structural equality on the flattened form.

`Subst::canonical_pi_binder` owns the rule; its doc comment names the three walks that apply it and what a silent divergence between them costs. `Type::alpha_normalized` is that scheme as a pure function, which `lambda_elim`'s debug assertions compose with `without_pi_names` to compare canonical solver output against a type rebuilt from the term's own binders.

A regression test in `compact.rs` and one in `spec_key.rs` state the repaired findings.

Why the rewrite sits at the flattening layer rather than at emission or in a comparison, and what the four alternatives break on, is recorded in [type-inference.md](src/ccl/design/type-inference.md#canonicalizing-a-pi-binder-needs-a-position-so-it-happens-at-flattening); `canonical_pi_binder` cites it.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76 ◀
                                            - #77
                                                - #78
                                                    - #79

dpmills:
Superseded by the scoped-inference-variables branch (`dmills/scoped-infer-vars`): the flatten-time canonicalization is replaced by the locally-nameless coordinate — indices assigned once at abstraction instead of canonical names assigned across three walks with an unchecked birth-to-flatten window. This PR's regression tests carried over as that design's acceptance tests (see `type-inference.md`, "Scoped inference variables: stored fragments close against a telescope"). #77 and up are restacked onto the new branch.

PR #79, opened 2026-08-11
Add a Lean model of the CCL type system and fuzz the solver against it

dpmills:
`constrain_go` implements a subtype relation that nothing in the repo states, and no test covers the solver's order-independence claims about `KindMerge`'s transitive force propagation and one-sided var-var bound propagation. This PR adds `formal/`, a Lean 4 `lake` project that states the ground fragment of the type system declaratively and proves its metatheory, an oracle binary that answers subtype verdicts, and two `#[cfg(test)]` Rust fuzz harnesses that diff the solver against the model and against itself under permuted constraint order. A semantic divergence between `constrain`/coalesce and the stated relation becomes a test failure instead of a doc claim.

The diff is additions only. Outside `formal/` and the two new test modules it is four lines of `mod` declarations in `src/ccl/infer/solver/mod.rs`.

## Reading order

[`formal/design.md`](formal/design.md) is the plan of record: the oracle stance (the model checks, it does not reproduce), milestones M0–M6, every adjudicated rule, and the findings the model has produced so far. Read it before the Lean. [`formal/README.md`](formal/README.md) has the build commands.

Then the library:

- `Ty.lean`, `Sub.lean` — the ground grammar with the uniquely-keyed invariant `Ty.WF`, and `Sub`, the declarative relation. `Sub.lean`'s module doc lists its two departures from `constrain_go`: no trivial-equality short-circuit, no partition-collapse arm.
- `Equiv.lean`, `Props.lean`, `Decide.lean` — decidability (`sub_of_subCheck`, `subCheck_of_sub`), `Sub.refl` as a theorem rather than a rule, and the `#guard` spec examples pinning each adjudicated rule.
- `Transitivity.lean` — `sub_trans` / `sub_trans_id`: transitivity for the canonical (`__pi{d}`) fragment, over six independent identity-acting rename environments.
- `Term.lean`, `Safety.lean` — the pure-core term calculus and the safety battery: `progress` (value, steps, or filter-blocked), `preservation`, `refinement_soundness`, `case_binder_sound`.
- `Merge.lean` — the ground fragment of `compact.rs`'s bound merging: `eqv` an equivalence, `merge_comm`, `merge_idem`, `merge_congr`, `merge_assoc_cf`, `foldMerge_perm`.

No file carries a `sorry`.

## The harnesses

`differential.rs` generates biased ground type pairs with a seeded xorshift (deterministic, no new dev-dependency), serializes them to the wire schema `Json.lean` defines, and diffs `constrain_subtype`'s verdict against the `subverdict` oracle case by case (`differential_ground_subtype_vs_lean_model`). It skips loudly when `formal/.lake/build/bin/subverdict` is absent, so a machine without a Lean toolchain stays green. `transitivity_chain_fuzz` needs no oracle: it builds chains the solver accepts and fails on any direct edge the solver then rejects. Knobs: `CAMBRA_DIFF_SEED`, `CAMBRA_DIFF_N`.

`confluence.rs` applies one constraint set in permuted orders against fresh variables and asserts the coalesced outcomes agree (`bound_order_permutation_fuzz`), plus two pinned cases — `refinement_claims_are_arrival_order_independent` and `vintage_claims_render_alike_but_do_not_dedup`. The outcome excludes which constraint trips the rejection of an unsatisfiable set, since that is order-relative under record-then-sweep; a set flipping between accepted and rejected across orders is a hard violation.

## What to be suspicious of

- `./ci.sh` is untouched, so nothing gates the Lean half. `lake build` would put a Lean toolchain on everyone's gate; `design.md` records that as a pending decision rather than an oversight.
- The pair generator excludes duplicate record/variant keys — the one type shape where the model and `constrain_go` are known to disagree, outside `Ty.WF`.
- `merge_not_assoc` is an open finding: the mixed-kind merge arm makes arrival order decide accept-versus-reject, and whether a real program can put those three bounds on one variable is a Rust-side question the confluence generator's vocabulary does not reach.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #62
                                - #74
                                    - #75
                                        - #76
                                            - #77
                                                - #78
                                                    - #79 ◀

dpmills:
Superseded: the formal model was rebuilt directly on `dmills/scoped-infer-vars` over the index coordinate — the `Ren` rename machinery this edition carried is deleted there, transitivity is proved with no fragment restriction, and the differential oracle re-ran green (150k pairs) after catching a real capture in the solver's Fun/Fun opening. This branch keeps the model's development history; the confluence harness rides `dmills/refinement-set`'s lineage where its assertions hold.

dpmills:
Follow-up: the rebuilt model now lands as #109, at the top of the restructured stack (#108 → #77 → #78 → #109), where every finding its design doc records is about the code beneath it.

PR #82, opened 2026-08-11
The allocation probe measures a thread, not a wall-clock window

dpmills:
`tests/scope_walk_allocations.rs` measured its zero-allocation claim with a process-global counter, so an allocation by *any* thread inside the window was charged to the walk — and `libtest`'s main thread makes exactly four one-time allocations right after it spawns the test thread. A loaded CI runner that preempts it there loses the race, and the walk is blamed: the intermittent "the scoped walk allocated 4 times". Counting per-thread charges only the thread that armed the window, which is what the assertion means; the module doc carries the mechanism. A second test checks the probe can still see an allocation.

### DIFF
--- a/tests/scope_walk_allocations.rs
+++ b/tests/scope_walk_allocations.rs
@@ -8,33 +8,69 @@
 //!
 //! `is_free_in_value` is the right probe — unlike `is_free` it threads no
 //! `visited` set, so nothing but the walk itself can allocate.
+//!
+//! # The window is per-thread, and has to be
+//!
+//! "Allocations during a window of wall-clock time" is not the property under
+//! test; "allocations *this thread* made during the walk" is. A process-global
+//! counter conflates them, and the difference is not hypothetical: libtest runs
+//! each test on a spawned thread and its main thread keeps working right
+//! afterwards, doing exactly four one-time allocations before it blocks —
+//! `running_tests.insert` (first insert into an empty `HashMap`),
+//! `timeout_queue.push_back` (first push into an empty `VecDeque`), and two
+//! inside the first blocking `rx.recv_timeout`. If the spawned test thread wins
+//! the race to the walk, a global counter charges all four to the walk. That is
+//! a real observed failure ("the scoped walk allocated 4 times") — rare, because
+//! the main thread normally finishes those four before a freshly spawned thread
+//! is scheduled at all, but reachable whenever the main thread is preempted
+//! right after the spawn, which is what a loaded CI runner does.
+//!
+//! Counting per-thread costs nothing in strength. The walk is a plain recursive
+//! fold that spawns nothing, so every allocation it could make is made by the
+//! thread that armed the window; the counter still sees all of them and now
+//! sees only them.
 
 use std::alloc::{GlobalAlloc, Layout, System};
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::cell::Cell;
 
 use cambra::ccl::TypedBinding;
 use cambra::ccl::ccl_utils::is_free_in_value;
 use cambra::ccl::{Name, TypedExpr};
 
-static ALLOCS: AtomicUsize = AtomicUsize::new(0);
-static COUNTING: AtomicUsize = AtomicUsize::new(0);
+thread_local! {
+    /// Allocations this thread has made since it armed its window.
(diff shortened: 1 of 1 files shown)


dpmills:

- #82 ◀

tests/scope_walk_allocations.rs
```diff
@@ -8,48 +8,85 @@
...
+/// measurement the assertions want.
 fn allocations(body: impl FnOnce()) -> usize {
-    ALLOCS.store(0, Ordering::Relaxed);
-    COUNTING.store(1, Ordering::Relaxed);
+    ALLOCS.set(0);
+    COUNTING.set(true);
     body();
-    COUNTING.store(0, Ordering::Relaxed);
-    ALLOCS.load(Ordering::Relaxed)
+    COUNTING.set(false);
```
groundlar: opt: `body()` panicking means this doesn't get reset. Could add a drop guard / RAII. Add an inline struct implementing Drop which resets `COUNTING`

tests/scope_walk_allocations.rs
```diff
@@ -85,3 +122,19 @@ fn a_freeness_query_over_a_letrec_spine_does_not_allocate() {
          collecting it into a Vec"
     );
 }
+
```
groundlar, line 146:
```suggestion
/// A global counter charges another thread's allocations to this window; a
/// per-thread one does not. Everything that allocates on *this* thread
/// (spawning, joining) stays outside the window, and the handshake spins on
/// atomics so it allocates nothing of its own.
// 
// The naive version doesn't work, and I tried it first:
//
//     let allocs = allocations(|| {
//         std::thread::scope(|s| { s.spawn(|| { /* allocate */ }); });
//     });
//
// That reported 6 even with the fix. Spawning a thread allocates on the spawning 
// thread — std has to heap-allocate the closure and the result slot to hand to the 
// new thread. Joining does bookkeeping too. Those are the measuring thread's 
// own allocations, correctly counted, and they drown the signal.
//
// So the window has to bracket only the worker's allocating loop, with the spawn 
// and join outside it. That's what the two AtomicBools do.
#[test]
fn another_threads_allocations_are_not_charged_to_this_window() {
    use std::sync::atomic::{AtomicBool, Ordering};
    let go = AtomicBool::new(false);
    let done = AtomicBool::new(false);
    std::thread::scope(|s| {
        s.spawn(|| {
            while !go.load(Ordering::Acquire) {
                std::hint::spin_loop();
            }
            for _ in 0..1000 {
                std::hint::black_box(Vec::<u8>::with_capacity(64));
            }
            done.store(true, Ordering::Release);
        });
        let allocs = allocations(|| {
            go.store(true, Ordering::Release);
            while !done.load(Ordering::Acquire) {
                std::hint::spin_loop();
            }
        });
        assert_eq!(
            allocs, 0,
            "another thread's allocations were charged to this window: {allocs}"
        );
    });
}

```

opt: I think this actually covers the bug? Opt because this is claude-generated overkill.

groundlar (approved):

tests/scope_walk_allocations.rs
dpmills: Done. Arming now goes through `Window::open()`, which returns a guard that resets `COUNTING` on drop — so `allocations` can't arm the counter without also scheduling its close, rather than merely remembering to.

tests/scope_walk_allocations.rs
dpmills: Taken — and you're right that it covers the bug. I checked rather than assuming: swapping the counters back to process-global statics leaves both existing tests green and fails only this one, at 1002 vs 0 — the worker's thousand plus two the harness itself made. That number is in the test's doc comment now as the concrete counterexample.

Two changes to your version. I trimmed the naive-attempt block down to the mechanism it implies — the spawn heap-allocates the closure and result slot on the *spawning* thread and the join does its own bookkeeping, so the window has to bracket the worker's loop and nothing else, hence the two atomics. Describing an approach that doesn't work is fine per CLAUDE.md, but the comment has to read as an explanation of the code that's there rather than as a record of what was tried, and the `let allocs = allocations(...)` snippet was the part that read as history.

I also moved it below `the_probe_counts_allocations_made_by_the_measuring_thread`, since the two are a pair: the probe sees its own thread's allocations, and only those.

PR #83, opened 2026-08-11
A sum's body may be another sum

dpmills:
A sum's body may be another sum

Two generators nest two sums — `Σ σ₄ ∈ 𝐾₄. Σ σ₇ ∈ 𝐾₇. ((σ₄, σ₇) ⤇ Int)` — and
every rule between formation and the post-inference wall assumed a sum was **one
binder deep**, with a body `𝑤 ⤇ 𝑉`. Formation already builds the nested type
(`two_conditional_sources_keep_their_witnesses_apart` pins it); consuming one hit
`SigmaType::body_residue`'s `unreachable!` on the first rule that looked.

None of this needed a new rule. Each site states the rule it already had in terms
of the body it actually has:

- **Consumption** (`demand_domain_range`) related the consumer's domain to the
  *bare witness*. It now relates it to the **body's** domain, which at one binder
  deep *is* that witness and nested is the index `Tuple`. Tuples need no rule of
  their own — the ordinary structural walk reaches the `Type::WitnessRef` leaves.
- **Peeling** is one arm: a sum whose body is a sum publishes its range and hands
  the body to the same rule, so nothing has to know how deep the nesting goes.
- **`Check`'s function destructuring** and **`Type::fun_like`** each unwrapped or
  rebuilt exactly one wrapper; both now go through as many as there are.
- **Adopting the context's name** was whole-domain ("a consumer whose domain *is* a
  witness reference was written against a consumption that already happened"). It
  is now positional, so each component of an index tuple adopts its own.

The distributing arm is **restricted to witness-bodied sums**, which is the shape
its own comment already describes (`Σ 𝐷 ∈ 𝐾. 𝐷`) rather than a new condition.
Distribution *instantiates* the witness, so a body still mentioning it afterwards
has been pinned to one candidate — the silent narrowing the `Type::WitnessRef`
arms refuse by name, and what put a concrete `[0, 1]` where a witness belonged. A
demand that still names the witness is domain-*preserving* and takes the naming
rule, the same split the `(Σ, Fun)` arm makes, stated for the shapes it does not
cover.

`identical_rendering_hint` reported "0 more characters of structure" when two
types were *equal* as debug strings — the `find` returning `None` means either a
prefix or equality, and it assumed prefix. Equality is the more useful answer,
because it says the disagreement is not in the types at all and points at the rule
that rejected them.

End to end, `two_conditional_sources_compile` is still ignored, and its blocker is
now a different one: past `lambda_elim`, an application's argument edge meets the
collection **opened** — `(σ@4, σ@7)`, witnesses free — against the parameter's
**closed** sum. The two render identically only because `coalesce_for_error` closes
a free witness back into its sum for display. That is bound-versus-free, not depth,
and its fix is not a `𝑈 <: Σ` arm — whatever produced the opened type is what has to
bind it back.

### DIFF
--- a/src/ccl/infer/api.rs
+++ b/src/ccl/infer/api.rs
@@ -515,9 +515,19 @@ fn identical_rendering_hint(type_a: &Type, type_b: &Type) -> Option<String> {
         .find(|((_, x), y)| x != y)
         .map(|((i, _), _)| i)
     else {
-        // One structure is a *prefix* of the other, so no character disagrees and a window
-        // would quote the same text twice. The difference is depth: something wraps one
-        // side and not the other.
+        // No character disagrees. Either one structure is a *prefix* of the other — the
+        // difference is depth, something wrapping one side and not the other — or the two
+        // are structurally identical, in which case the types are not what disagree at all
+        // and the rule that rejected them is where to look.
+        if a.len() == b.len() {
+            return Some(
+                "note: these are structurally identical, not merely identical as printed — \
+                 so the disagreement is not in the types. Look at the rule that rejected \
+                 them: it failed on something the types do not carry (a witness range read \
+                 from the index, a kind obligation, a scope) rather than on their shape."
+                    .to_string(),
+            );
+        }
         let (shorter, longer, which) = if a.len() < b.len() {
             (a.len(), b.len(), "found")
         } else {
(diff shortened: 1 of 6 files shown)


dpmills:

- #10
    - #11
        - #45
            - #83 ◀

dpmills:
Folding into #45 — this belongs in the dependent-sum PR, not stacked on top of it as a separate review.

PR #84, opened 2026-08-11
A flat merge keys on the domain value, not a position

dpmills:
A flat merge reassembles the disjoint arms of one fed collection, and it keyed them on `usize` — so it rejected any fed stream whose index set is not a range of positions. A `match` over a conditional-element comprehension is exactly that: the comprehension is a `Copair`, so its domain is `Variant({Index(i): {D | π̂ᵢ}})` and the merge met `Union { tag, inner }` where it wanted a position. It now keys on the domain **value**, which needs nothing new — reassembly wants only a total order and equality, and `Value` already compares a tagged pair lexicographically by tag then payload.

The types were never wrong, which is why the fix is here and not in the producer: the join's operands share the *element* variant, and the consumer accepts the coproduct index set. Emitting a `DisjointJoin` from the comprehension instead would need an inference rule for a pre-inference birth site, whose domain variable carries only lower bounds while a domain sits in negative position — the same missing construction as [The domain join is a Σ](src/ccl/design/type-inference.md#the-domain-join-is-a-σ).

Incomparable keys now name the real error: arms fed *different* domains are a copairing, not a disjoint join. Mechanism in [design-operators.md](src/interpreter/design-operators.md); `test_match_over_conditional_collection` is no longer `#[ignore]`d.

Thanks for putting this together — but a heads-up before you go further: Cambra isn't accepting external pull requests right now. See [CONTRIBUTING.md](../CONTRIBUTING.md) for why, and for what to do instead.

If this is a bug fix, the fastest path is to open an issue describing (or attaching) the patch — small, clearly-correct fixes may get applied as a team commit with credit to you.

This PR will likely be closed unreviewed. That's about where the project is right now, not a judgment on the change.


dpmills:

- #57
    - #54
        - #84 ◀

dpmills:
Folded into #54 — the flat-merge change is not standalone: it only becomes reachable with the `match` surface that #54 adds, and #54's previously-`#[ignore]`d `test_match_over_conditional_collection` is what covers it.

PR #86, opened 2026-08-12
Split CollectionUnion into two distinct node types

dpmills:
Combining collections has been one IR node, `CollectionUnion`, carrying two different operations, with op-conversion deciding which one a node meant from whether a fed input was present — a property that correlates with the distinction without being it. This splits the node in two, gives each its own typing rule, and lets every construction site state which operation it built. The payoff is a deletion: a subtyping rule whose whole content was undoing a claim the producer had no reason to make.

They are not modes of one operation: signature, result domain, and definedness all differ. Copairing lands on the coproduct and is always defined — `x ++ x` gives `A + A`, not `A` — while a disjoint join stays on the operands' shared domain and is defined only where they are disjoint. That is also why a mode flag would be the wrong shape: a flag means one node with two typing rules, and the rules are exactly what differ. Both are stated in [ir.md](src/ccl/design/ir.md#copair-and-disjointjoin--two-collection-combining-operations-not-one).

"Union" named neither — a union type is *untagged*, while one of these is a tagged sum and the other a merge of partial maps. The surface `BinOp` for `++` keeps its name; this is IR vocabulary.

`lambda_elim`'s `Case` fan-outs build the join, so `constrain_go`'s bridge arm and its `is_index_partition_of` helper are gone: the two domains are now the same type and the ordinary arms apply.

Three things to look at. A **fed copairing** is rejected rather than flat-merged, since a flat merge reassembles one domain and nothing builds a fed copairing today. `insert_iterate_recurse` gained a `DisjointJoin` arm, without which the fan-out's gates never become filters. And a `Case` arm's `Compose` is now typed from its own chain (`arm_compose`) rather than from the enclosing `Case`, because an arm's result can sit strictly below the arms' join.


dpmills:

- #57
    - #86 ◀
        - #54
            - #99
                - #100

sortalongo (approved):

PR #87, opened 2026-08-12
Restrict opposite-polarity fallback to only undetermined positions

dpmills:
`compact_go`'s opposite-polarity fallback collapses `∀α ⊒ 𝐿` to its bound — the only way an under-determined position gets a type at all (removing it fails 163 tests). But it fired at *any* variable on a bound chain, including ones out in the call site, so a specialization's domain could come out as the join of everything done with any use's result. It now fires only at the position the walk entered, guarded by a one-line `fallback_allowed` (`parents.is_none()`). Reasoning: [The collapse happens at the position](src/ccl/design/type-inference.md#the-collapse-happens-at-the-position).

Confining it needs the relations it stood in for stated instead. Traits supplies one; the other is here — an unfiltered single-generator comprehension now says it iterates *exactly* its source's domain, adopting `groupby`'s key `SharedHole` when the source names one.

Also: the fallback's result no longer merges into the polarity-correct walk's leftover, whose empty refinement set is absorbing under the positive intersection.

Nearly half the diff is the two regression tests; both fail without the change.


dpmills:

- #71
    - #72
        - #73
            - #87 ◀

src/ccl/infer/solver/compact.rs
```diff
@@ -544,6 +544,32 @@ impl ParentPath<'_> {
...
+/// Letting it fire further along makes a function's domain the join of everything
+/// the program does with any use's result: an arm binder's upper-bound chain runs
+/// out through the `match` result and the codomain into the call site's operands,
+/// and the join sitting there comes back as the domain's demand. See
+/// `design/type-inference.md`, "The collapse happens at the position".
+///
+/// A structural boundary starts a new position, which is why [`compact_go`] resets
+/// `parents` to `None` at every structural child.
+fn fallback_allowed(parents: Option<&ParentPath<'_>>) -> bool {
+    parents.is_none()
```
groundlar: This appears unexercised, if changed to 

```
fn fallback_allowed(parents: Option<&ParentPath<'_>>) -> bool {
    let _ = parents;
    true
}
```
CI is still green.

Is this a noop or missing test coverage? My hunch is that the new tests have insufficient reach.

src/ccl/infer/solver/compact.rs
```diff
@@ -806,19 +836,41 @@ fn compact_go(
             // opposite polarity from where the lambda is coalesced. This is the
             // coalesce-time read of monomorphization; it is sound because every
             // var reaching coalesce is monomorphically determined (one type or
-            // an `IncompatibleBounds` error). See the rationale above.
+            // an `IncompatibleBounds` error). See the rationale above, and
+            // [`fallback_allowed`] for why it happens only at a position.
             let no_concrete = bound.as_ref().is_none_or(|b| {
                 b.atoms.is_empty() && b.rec.is_none() && b.fun.is_none() && b.history_slot.is_none()
             });
```
groundlar, line 952: nit: this is fairly subtle, would be more readable (and future proof) with an exhaustive match like:
```suggestion
let no_concrete = bound.as_ref().is_none_or(|b| {
    let CompactType {
        atoms, rec, fun, history_slot,
        // A variant shape deliberately does NOT count as concrete: the
        // fallback must still fire so a domain resolves to its own
        // argument rather than the arms its body can handle.
        var: _,
        // Not shape — carried across explicitly below.
        vars: _, refinements: _,
    } = b;
    atoms.is_empty() && rec.is_none() && fun.is_none() && history_slot.is_none()
});
```

groundlar (approved):

src/ccl/infer/solver/compact.rs
dpmills: Reproduced, and you were right about the reach — but it isn't a noop.

I instrumented the walk. The guard suppresses a fallback at **25,708** sites across the suite, at **25,051** of which the suppressed fallback would have produced a concrete shape. Nothing asserted any of the differences. Where one is checkable by eye the guard is the better answer:

| | guard | `true` |
|---|---|---|
| `test_def_param_annotation_enforced` | `(Int ⇒ 1)` | `(1 ⇒ 1)` — annotation collapsed to the one call |
| `feeds_cases::case_19/20/21` | `(10 ⇒ feed(chan(x) ⇒ Int))` | `(Int ⇒ …)` |
| `test_higher_order_dependent_application_discharges_key` | `((__arg: 0) ⇒ ({… == __arg} ⤇ Int)) ⇒ …` | `(0 ⇒ ({… == 0} ⤇ Int)) ⇒ …` — binder substituted away |

So: missing coverage. The reason the two `Case` tests don't catch it is that they are *also* repaired by the other half of this change — the `recovered` accumulator, which stops the fallback's result merging into the polarity-correct walk's leftover. Bisected: both fail at the parent commit, and both pass with the guard neutralized and the accumulator in place. Neither isolates `fallback_allowed`.

Isolating it needs a program with no variant and no refinement set in the frame, so nothing else can move the answer:

```
let g = λ x → x in let f = λ t → (t.0 ▷ g, t.2 ▷ g) in (0, 1, 2) ▷ f
```

`f`'s domain is `(0, 1, 2)` with the guard and `(0, 1, Int)` without it: the two `g` results meet in `f`'s result tuple, `0 ⊔ 2 = Int`, and the unrestricted collapse hands that join back as the demand on the position it started from. Added as `a_domain_does_not_read_back_the_join_its_results_meet_in`, verified failing under exactly your `let _ = parents; true` edit.

---

**Since this thread was written**, the branch has been rebased onto current main and three further things landed here, all downstream of the question you asked:

**The pin and this guard were claiming the same variable.** `pin_unobservable_arm_payload` (#57, which landed after this branch was written) gates on `resolve_var_type`, which runs `compact_type` — so the guard changes the pin's own input. In ``λ x → match x { `a(v) → v; `b(w) → w }``, `w` had been resolving to `Int@1` by walking to the arms' result join and reading that join's *other* contribution. That is the sibling-join step this PR rules out, so `w` correctly became undetermined, fell to the pin, and the pin chose `Unit` — which collided with `Int` in the arms' join. The pin's stated rule was already "a type the payload's requirements accept"; it only implemented the *trait* half. A subtyping upper bound is a requirement too, so `payload_flow_target` now reads it, resolving each bound **as its own position**. #57's `read_arm_payload_is_not_pinned` was resting on the unsound propagation and is renamed for what now happens.

**Two more, surfaced by PR 54 and verified reproducible here on IR alone.** A variant shape read *positively* is the value's own tags, so counting it "not concrete" at every polarity let the collapse fire past a value and hand back its own upper bound — a bounded annotation reading as the binder's type. And the pin's gate asked what *determines* a position rather than what *reached* it, which an upper bound alone can satisfy; `compact_type_polarity_only` asks the other question. Porting just these two into a 54+87 tree takes its 10 failures to zero.

**On the resolution losses I flagged above** — they are not losses. Post-inference IR is byte-identical for every `(? ⤇ Int)` test; those coalesce results are transient intermediates. Across all 482 pipeline programs only 11 differ, and the last group of those (`generators_udf_poly`) turns out to be the guard doing its job rather than costing anything: the domain variable's upper bounds already carry a concrete `Int` from the trait write-back, so the collapse never fires there at all. The `Int@1` main produced came from collapsing the `==` operand variables as a side effect of resolving the domain — the sibling chain again. Nothing left outstanding.

src/ccl/infer/solver/compact.rs
dpmills: Adopted — a new field should force a decision rather than defaulting to "not shape".

I checked your `var` comment before taking it rather than copying it on faith: adding `var.is_none()` to `no_concrete` fails both `Case` tests, so the omission is load-bearing exactly as you describe.

It turned out to be load-bearing at **one polarity only**, which the exhaustive form is what made visible. Read negatively, a variant is the arms a body can *handle* — your reasoning, and a domain must still fetch the argument the collapse finds. Read positively it comes off the lower bounds and is the value's own tags, so the collapse has nothing to add and firing past it replaces the value with its own upper bound. That is a real defect this PR introduced: with `x <: {`some{Int} | `none} = `some(1)` the binder came out as the annotation, `none` tag and all, instead of `{`some{Int@1}}` — a bounded annotation no longer behaving like a bound. So the field is now bound rather than discarded, and the predicate reads `pol && var.is_some()`, with the two readings written out. Records and atoms need no such split, being the same claim from either side.

Regression test is `a_bounded_variant_annotation_does_not_become_the_binders_type` in `tests/inference_variants.rs`.

PR #89, opened 2026-08-12
Typecheck a never-called function before dropping it

dpmills:
Monomorphization drops a generalized definition that no use demanded — and dropped it *unresolved*, so a type error only resolution finds needed a call site to appear: `f = \a -> (a.0, a.foo)` asks `a` to be both a tuple and a record, and type-checked as long as nobody called `f`. `coalesce_generalized_let` now resolves an undemanded definition before dropping it, keeping only the diagnostics. Soundness and the residual gap: [type-inference.md, "Typechecking a never-called definition"](src/ccl/design/type-inference.md#typechecking-a-never-called-definition).

## Two distinctions to review

**`SpecializeFrame::demanded`, not `specs.is_empty()`, gates the walk** — a use can fail to resolve before minting one, or decline to register inside dead code. Gating on the memo re-walks a definition that is used, double-reporting its body's conflict. A related gap is *recorded, not fixed*: a failed use is left naming a binding the rebuild drops — which wants the "could not be built" node, not a rescue here.

**Registering a specialization and splicing it are separate decisions.** A use in dead code still specializes — that is what checks the call — and still registers, so the memo shares its clone; `Specialization::referenced` carries splice-liveness and the rebuild filters on it, so nothing dead re-enters the program. Which frames that applies to is recorded per frame when it is pushed (`SpecializeFrame::inside_discarded`), not derived from a scope depth, which the re-entrant clone walk truncates.

## Behaviour

A/B over a 67-case matrix: 19 previously-accepted programs are now rejected, and nothing else moved (error counts and specialization counts unchanged). Three rejections are `Unsupported` domain-join limits reached through dead code; each fails identically when called.

## Tests

15 rejection cases over how the demand reaches the definition, 16 over-rejection guards, 4 source cases, two unit tests for splice liveness. One defect stays one diagnostic however many specializations of an enclosing definition contain it, held fixed across a varying count. `dead_code_specialization_cost` bounds the cost by allocation count rather than wall-clock, so the guard is hard rather than flaky; its probe moved to `tests/alloc_probe/`, shared with `scope_walk_allocations` and carrying its own self-tests. Programs use `indoc!`, a convention recorded in `CLAUDE.md`.

### DIFF
--- a/src/ccl/infer/solve.rs
+++ b/src/ccl/infer/solve.rs
@@ -148,21 +148,22 @@ pub(super) struct CoalesceCtx {
     /// [`coalesce_node`] on both exit paths. The same discipline `emit_node` and
     /// `check_node` use; seeded with the tree's root at construction.
     current_node: NodeId,
-    /// While a **discarded** subtree is being walked — a dead generalized
+    /// Whether a **discarded** subtree is being walked — a dead generalized
     /// definition, resolved for its diagnostics and then dropped
-    /// ([`typecheck_discarded_definition`]) — the `scope` depth at which that
-    /// walk began; `None` outside one. Frames at or above the boundary were
-    /// pushed *inside* the discarded subtree and go away with it, so a
-    /// specialization minted on one is registered as usual. A frame *below* it
-    /// belongs to a binding that survives, and registering there would splice a
-    /// specialization into the surviving program whose only use is about to be
-    /// discarded — so [`specialize_use`] typechecks such a use and drops it
-    /// (see [`SpecializeFrame::specs`], "The remaining gap", for the same
-    /// liveness question on the sharing path).
+    /// ([`typecheck_discarded_definition`]).
     ///
-    /// The *outermost* discarded walk owns the boundary: everything a nested one
-    /// could register is already inside a subtree being dropped.
-    discard_boundary: Option<usize>,
+    /// A specialization minted while this holds serves a use that is about to be
+    /// dropped, so it is registered — the memo is what stops the walk re-cloning —
+    /// but marked unreferenced ([`Specialization::referenced`]) unless its own
+    /// frame is being dropped too ([`SpecializeFrame::inside_discarded`]).
+    ///
+    /// Deliberately *not* a `scope` depth. A depth answers "is this frame outside
+    /// the discarded subtree" only while the stack grows monotonically, and
+    /// [`specialize_use`] truncates it (`split_off(frame_idx)`) for the re-entrant
+    /// clone walk: frames the clone's own body pushes then land at indices below
+    /// the mark and read as surviving, though they die with the clone. Asking each
+    /// frame what it is, at the moment it is created, is immune to that.
+    discarding: bool,
     /// Every read the walk performed, for the end-of-pass ordering-invariant
     /// check ([`assert_reads_stable`]). Debug builds only.
     #[cfg(debug_assertions)]
@@ -460,14 +461,28 @@ struct SpecializeFrame {
     /// The binding's polymorphism level — the freshen cutoff: variables
     /// deeper than this are the quantified ones.
     cutoff: Level,
+    /// Whether this binding is itself inside a subtree being discarded — recorded
+    /// from [`CoalesceCtx::discarding`] when the frame is pushed, the one moment
+    /// the answer is unambiguous.
+    ///
+    /// A specialization minted on such a frame is spliced as usual: the `let` it
+    /// splices into is going away with the subtree, so liveness there is moot. It
+    /// is the frames that **outlive** a discarded walk whose splices need filtering.
+    inside_discarded: bool,
     /// Whether the body demanded this binding at all — set by [`specialize_use`]
-    /// on entry, before anything can go wrong.
+    /// on entry, before anything can go wrong, and so **including uses that sit in
+    /// dead code**.
+    ///
+    /// That is deliberate: such a binding is not discard-walked, and does not need
+    /// to be. It was already checked through the clone the dead use pinned, which
+    /// is a *stricter* reading than the generic one a discard walk would take — a
+    /// specialization only adds bounds. Walking it again would report whatever that
+    /// clone already reported (`through_a_call_in_dead_code` covers the case).
     ///
     /// This is *not* `!specs.is_empty()`, and the difference is exactly the case
     /// where a use exists but produced no specialization: a use whose
-    /// instantiation fails to resolve reports and returns before minting one, and
-    /// a use inside a discarded subtree deliberately does not register (see
-    /// [`CoalesceCtx::discard_boundary`]). Only *this* flag answers "is the
+    /// instantiation fails to resolve reports and returns before minting one. Only
+    /// *this* flag answers "is the
     /// definition dead code", which is what decides whether it is typechecked on
     /// the way out ([`typecheck_discarded_definition`]) — asking `specs` instead
     /// re-walks a definition whose uses merely failed, and reports its body's
@@ -565,8 +580,22 @@ struct Specialization {
     name: Name,
     /// The specialized, fully-coalesced definition. Spliced as a `let`
     /// binding around the generalized `let`'s body once that body's walk
-    /// completes.
+    /// completes — if [`referenced`](Self::referenced).
     def: Expr,
+    /// Whether a use that **survives** refers to this specialization.
+    ///
+    /// A use inside a discarded subtree still needs a clone — pinning and
+    /// coalescing it is what typechecks the call — but the `let` it would splice
+    /// outlives the subtree its only use is in, so splicing would leave the
+    /// program a definition nothing references. Splitting that from the memo is
+    /// what lets the clone be *shared*: declining to register instead made every
+    /// dead use re-clone and re-coalesce its callee, which compounds through a
+    /// call chain (see [`SpecializeFrame::specs`], "The remaining gap", where the
+    /// same split is what reference-liveness filtering asks for).
+    ///
+    /// False at mint for a discarded use; a later surviving use that *hits* this
+    /// entry sets it, because sharing the clone is exactly what makes it live.
+    referenced: bool,
 }
 
 /// Find the scope entry a free use of `name` refers to: scanning innermost-
@@ -604,7 +633,7 @@ pub(super) fn coalesce_pass(expr: &mut Expr) -> Vec<LocatedInferError> {
         current_node: expr.node_id(),
         errors: Vec::new(),
         pred_memo: PredMemo::new(),
-        discard_boundary: None,
+        discarding: false,
         #[cfg(debug_assertions)]
         reads: Vec::new(),
     };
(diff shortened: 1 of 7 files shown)


src/ccl/infer/solve.rs
```diff
@@ -1468,16 +1514,36 @@ pub(super) fn coalesce_generalized_let(expr: &mut Expr, level: Level, ctx: &mut
...
+    // code, so it is typechecked before it is dropped.
+    if !frame.demanded {
+        debug_assert!(
+            frame.specs.is_empty(),
+            "specialization without a demand: `{}` registered {} specialization(s) \
+             though no use reached `specialize_use`",
+            frame.name,
+            frame.specs.len(),
+        );
+        typecheck_discarded_definition(&mut frame.def, level, ctx);
```
groundlar, line 1587: **Important — one defect's diagnostics multiply by the enclosing definition's specialization count.**

A dead definition nested inside a *live* generalized definition sits inside each clone, so it is discard-walked once per clone:

```
def g(x):
    h = \y -> (y.0, y.foo)
    x
live1 = g(1)
1
```
→ **3** errors.

```
def g(x):
    h = \y -> (y.0, y.foo)
    x
live1 = g(1)
live2 = g("s")
1
```
→ **6** errors, same single defect. A dead helper inside a library function used at five types gives fifteen diagnostics for one bug.

This is not pre-existing clone-walk behaviour: making `h` *live* (body `h(x)`) gives exactly **1** error at both one and two uses, so the live path already collapses this and only the new discard walk multiplies.

`SpecializeFrame::demanded` exists precisely to avoid "one defect, four diagnostics" — this is the same failure via the enclosing clone, and the design note doesn't cover that route. `a_suppressed_specialization_does_not_get_its_definition_re_walked` pins the count at 3; this deserves the same, under a *varying* number of enclosing specializations. `nested_in_a_live_definition` in the rstest matrix has exactly one live use of `g`, which is why the matrix doesn't see it — adding a second `live = g("s")` line is a one-line repro.

Cheapest mechanism is probably deduping the walk per definition rather than per clone; deduping the final error vector on `(InferError, label)` would also work, since the label is `symbolic` of the node and is identical across clones.

src/ccl/infer/solve.rs
```diff
@@ -1417,6 +1454,14 @@ pub(super) fn specialize_use(use_expr: &mut Expr, frame_idx: usize, ctx: &mut Co
 
     use_expr.node = TypedExprNode::Var(spec_name.clone());
     use_expr.ty = clone.ty.clone();
+    // This use lives in a subtree that is about to be dropped, while the binding
+    // it specializes outlives that subtree: the clone was built for its
+    // diagnostics, which the walk above has already collected, and registering it
+    // would leave the surviving program a definition nothing references. Drop it
+    // instead — see [`CoalesceCtx::discard_boundary`].
+    if ctx.discard_boundary.is_some_and(|depth| frame_idx < depth) {
```
groundlar: **Important — this early return also loses the sharing memo, so dead code re-clones.**

It returns before `frame.specs.push(...)`, and the memo lookup above is the same `Vec`. So every use in a discarded subtree re-clones and re-coalesces the callee where a live use would have shared.

Measured, release build, on a chain `a{i} = \x -> a{i-1}(x) + a{i-1}(x)` (scalar-typed, so the *result type* does not grow), `f = \q -> a{d}(q)`, with and without a `live = f(1)` line:

| depth | live | dead | ratio |
|---|---|---|---|
| 3 | 22.7ms | 23.6ms | 1.0× |
| 4 | 111ms | 192ms | 1.7× |
| 5 | 539ms | 1.37s | 2.5× |
| 6 | 2.41s | 8.33s | 3.5× |
| 7 | 10.8s | 47.5s | 4.4× |

The base exponential (~4.6×/level even when live) is pre-existing — the `SpecKey` over-splitting already documented under `SpecializeFrame::specs`, "The remaining gap" — not this PR's doing. What *is* this PR's doing is that dead code now pays it, plus an extra ~1.2×/level from the missing memo, where before it paid nothing.

Root cause is structural: `specs` is now doing two jobs — the *sharing memo* (`specs.iter().find(|s| s.key == key)`, whose purpose is to not re-clone) and the *splice list* (`for spec in frame.specs.into_iter().rev()`, whose purpose is what ends up in the program). The discard boundary's intent is only "do not splice this", but this return suppresses both. Always populate the memo and decide splice-liveness at rebuild: that's strictly less code than this return, and it's the *same* abstraction "The remaining gap" already contemplates under "reference-liveness filtering at the splice". Per CLAUDE.md's recurring-shapes rule, a second independent arrival at the same missing distinction is the signal to name it rather than wait for the third.

Separately: the new doc section is exhaustive on soundness and on what the walk can't reach, and silent on cost — while the walk runs unconditionally in release on every uncalled definition. A paragraph in "Typechecking a never-called definition" noting the cost characteristic, and that the pre-existing specialization blowup is now reachable from code nobody calls, belongs there.

src/ccl/infer/solve.rs
```diff
@@ -148,6 +148,21 @@ pub(super) struct CoalesceCtx {
...
+    /// specialization minted on one is registered as usual. A frame *below* it
+    /// belongs to a binding that survives, and registering there would splice a
+    /// specialization into the surviving program whose only use is about to be
+    /// discarded — so [`specialize_use`] typechecks such a use and drops it
+    /// (see [`SpecializeFrame::specs`], "The remaining gap", for the same
+    /// liveness question on the sharing path).
+    ///
+    /// The *outermost* discarded walk owns the boundary: everything a nested one
+    /// could register is already inside a subtree being dropped.
+    discard_boundary: Option<usize>,
```
groundlar: **Nit — as a scope *index*, this isn't quite the predicate the comment states.**

"Frames at or above the boundary were pushed *inside* the discarded subtree" holds at the top of a discard walk, but not inside a re-entrant clone walk: `specialize_use` does `ctx.scope.split_off(frame_idx)` with `frame_idx < discard_boundary`, so frames the clone's own body pushes land at indices *below* the boundary and are read as surviving. Their specializations get suppressed, and `coalesce_generalized_let` then drops those bindings while their uses were already renamed — the dangling-reference shape you documented, reached without any error being pushed, so the "unobservable while an inference error discards the tree" argument doesn't cover this instance.

It *is* benign, because that whole clone is itself discarded. But the honest encoding of "does this frame outlive the discarded subtree" is a `bool` on `SpecializeFrame`, set at push time from `discard_boundary.is_some()` — immune to truncation and reading as what it means. A sentence in this comment would do if you'd rather not change the representation.

src/ccl/infer/solve.rs
```diff
@@ -1290,6 +1320,13 @@ fn freshen_clone_node_ids(expr: &mut Expr) {
 // (matching the solver's module-level allow).
 #[allow(clippy::mutable_key_type)]
 pub(super) fn specialize_use(use_expr: &mut Expr, frame_idx: usize, ctx: &mut CoalesceCtx) {
+    // Record the demand before anything below can fail or decline to register: a
+    // definition is dead code only if no use ever reached this function.
+    let ScopeEntry::Generalized(frame) = &mut ctx.scope[frame_idx] else {
+        unreachable!("lookup_generalized returns indices of Generalized entries only");
+    };
+    frame.demanded = true;
```
groundlar, line 1358: **Nit — worth stating that this makes a binding demanded *only from dead code* never discard-walked.**

Set before the boundary check below, so such a binding counts as demanded and is checked only via the clone pinned to a dead-code instantiation. That's sound — a specialization only adds bounds, so it's at least as strict as the generic resolution — and `through_a_call_in_dead_code` exercises it. But a reader will stop here and have to derive it; one sentence in the doc comment saves them the trip.

groundlar (commented):
Staff review — the soundness argument holds and the doc section is the repo's benchmark for this kind of change. Debug tests green locally (type_check + lib: 1245 passing), `./ci.sh doc_refs` clean, and I could not trip `assert_reads_stable` with dead code that constrains earlier live bindings (six programs, including `xs = [1,2,3]` / `f = \a -> [x for x in xs if x > 0]` / `xs`, which still resolves to `([0, 2] ⤇ Int)`).

Four comments inline: two I'd want addressed before merge, two smaller. Both of the first two are cheaper to fix now than after the Σ work lands on top.

groundlar (approved):
Great find!

src/ccl/infer/solve.rs
dpmills: Changed the representation — you're right that it isn't a nit. I reproduced the truncation case: with `split_off(frame_idx)`, a frame the clone's own body pushes lands below the mark and reads as surviving, so its splices are suppressed while its uses were already renamed. Benign, as you say, but reached with no error pushed, so my "unobservable while an inference error discards the tree" argument didn't cover it.

`SpecializeFrame::inside_discarded` now records the answer when the frame is pushed, and `surviving_use` reads it. The unit test is renamed to `splice_liveness_follows_the_frame_not_the_scope_depth`, since that is what it is actually pinning.

src/ccl/infer/solve.rs
dpmills: Added, in the `demanded` doc comment: the flag is set on entry and so includes uses in dead code, which is deliberate — such a binding is not discard-walked and does not need to be, having been checked through the clone that use pinned, which is a *stricter* reading than the generic one (a specialization only adds bounds). It names `through_a_call_in_dead_code` as the case that exercises it.

src/ccl/infer/solve.rs
dpmills: Taken as diagnosed — `specs` was doing both jobs and the early return suppressed both, where it only meant to suppress splicing. Split rather than patched: `Specialization::referenced` carries splice-liveness, a discarded use registers its clone like any other so the memo shares it, and the rebuild filters on the flag. A later surviving use that *hits* such an entry sets it, since sharing the clone is what makes it live. That deleted the early return instead of adding to it.

Reproduced your measurement and the fix on the same chain, release, depth 6: **1894ms → 380ms** dead, against **662ms** live. Dead now comes in under live, which is what you'd expect when it does the same specialization work and splices none of it.

Guarding it took two attempts worth recording. Counting clones via diagnostics doesn't work — an ill-typed callee fails resolution *before* specializing, and an emit-time error is fail-fast, so neither path reaches the memo. `tests/dead_code_specialization_cost.rs` instead counts allocations, which is deterministic enough for a hard bound rather than a timing threshold: a dead chain must allocate under half what the same chain costs live. Calibration is in the comment — sharing gives `0.28 ×`, a lost memo `0.94 ×` at that depth and over `1.0 ×` one level deeper. It fails on the pre-fix behaviour, and runs in 0.6s.

Also took the cost paragraph: the design section now says the walk runs unconditionally in release, that the pre-existing specialization blowup is reachable from code nobody calls, and what bounds it.

While adding the test I picked up #82 on rebase and found I had copied the *old* process-global probe — the exact flake you fixed there. The probe now lives in `tests/alloc_probe/`, shared with `scope_walk_allocations`, with its self-tests alongside it so a binary that adopts the probe adopts the evidence it works.

src/ccl/infer/solve.rs
dpmills: Confirmed and fixed. Your repro reproduces exactly: 3 / 6 / 9 diagnostics for one / two / three live uses of `g`, with the control (`h` made live) staying at 1 — so the multiplication is the discard walk's, not pre-existing clone-walk behaviour.

I kept the walk per clone rather than deduping it per definition: `h`'s body can reference `g`'s parameter, so it can be ill-typed at one instantiation and fine at another, and walking once would miss that. What was wrong is the *reporting*, so a new error that repeats one already reported is dropped — compared only against what preceded the walk, never within it, so a single walk's shape still matches what live code produces (one report per node whose type carries the defect). Now 3 / 3 / 3.

Pinned by `one_defect_does_not_multiply_by_the_enclosing_specialization_count`, which is the count held fixed while the number of enclosing specializations varies, as you suggested — the `nested_in_a_live_definition` matrix case stays as it is, since it is about the demand route rather than the count.

PR #91, opened 2026-08-13
Split transactional commit stores by connected components of transaction blocks

dpmills:
Two `for` loops over `:=` accumulators have always compiled to two induction stores, one letrec each. Two `with begin():` loops over unrelated `Mut(_, Txn)` variables compiled to **one** commit store, because `transact_phase` unioned every writer site's footprint into a single key set. That coupled two things that should be independent: their commit clocks, and their completion. The second is observable — a store is terminal only when *every* writer has drained, and a fed-out read is an `AsOf`, non-terminal until its store is — so a finite variable's trailing read could not settle if any unrelated variable had a live-source writer. This partitions the keys, giving the transaction path one store per set of variables some block actually relates.

## Planning and placement come apart

`build_letrec` becomes `plan_store` plus `splice_stores`. A store sits below everything its bindings need and above everything reading its keys; each spine statement gets a **level** — 0 if it reads no store, else one past the last store it reads, transitively — and rides inside that store. A statement cannot read two stores at once, which is what makes the level well-defined: reading two variables together is a block, and the partition put those keys in one store precisely so the read has one snapshot. `CrossDomain` wraps the whole nest, since an accumulator a decision reads must be outside every store reading it.

Both mutability paths ended in the same four steps — trailing reads, feed hoists, causality assert, `LetRec` — now `mut_elim::close_recurrence_group`. It puts the assert where neither caller can forget it, and states as load-bearing an ordering both had right by coincidence: a read hoisted over a feed would break `channelize`'s outermost-first collection silently.

## Tests

Nine, four of them asserting the *planned graph* rather than only values — two programs can agree on every number and differ in store count. `a_finite_mut_var_completes_despite_a_live_unrelated_writer` is the payoff, checked against the base rather than assumed: there the read yields empty forever. Recognition and op-conversion needed no changes.

### DIFF
--- a/src/ccl/design/mutability.md
+++ b/src/ccl/design/mutability.md
@@ -503,6 +503,73 @@ mutable variables and channels they target. Generators survive inlining because 
 leaves nothing to lose — a generator body is `For` + `Feed` nodes against an implicit result feed,
 substituted wholesale.
 
+### How many commit stores a program has
+
+**One per set of mutable variables a `with begin():` block relates** — not one per program, and
+not one per mutable variable. `transact_phase::partition_keys` computes the partition; each part
+becomes its own `LetRec`, its own `Transact{domain: Txn}`, and its own `CommitOperator`. The
+induction path partitions the same way: `mut_elim::transform_loop` emits one letrec per loop.
+
+A block is what forces mutable variables together, for two reasons that coincide on the block's
+footprint. The store is not a user-facing concept: the
+[CHL spec's ordering model](../../../docs/chl-spec.md#85-ordering-and-concurrency) says only that
+two blocks are ordered when they mention a mutable variable in common, and that blocks sharing no
+variable are unordered. The partition is how that is realized — a program never has to reason
+about it.
+
+- **Atomicity.** A writing block produces one commit record, so every key it *writes*
+  advances at one tick, and every key it *reads* is read at that tick's snapshot.
+  `{reads ∪ writes}` is therefore one store — which is why the `limit` a guard consults is
+  in the same store as the `total` it guards, though nothing writes `limit`.
+- **Snapshot consistency.** A read-only block's reads are latched at one frontier, which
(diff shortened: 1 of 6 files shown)


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #91 ◀
                                - #92
                                    - #93
                                        - #97
                                        - #102
                                            - #106

src/ccl/design/mutability.md
```diff
@@ -503,6 +503,73 @@ mutable variables and channels they target. Generators survive inlining because
 leaves nothing to lose — a generator body is `For` + `Feed` nodes against an implicit result feed,
 substituted wholesale.
 
+### How many commit stores a program has
+
+**One per set of mutable variables a `with begin():` block relates** — not one per program, and
+not one per mutable variable. `transact_phase::partition_keys` computes the partition; each part
+becomes its own `LetRec`, its own `Transact{domain: Txn}`, and its own `CommitOperator`. The
+induction path partitions the same way: `mut_elim::transform_loop` emits one letrec per loop.
```
sortalongo, line 511: As discussed live, I think this is still too constraining. e.g.

```
x := 0
y := 0

for i in finite_source():
  with begin():
    x += 1
    y += 1

for j in infinite_source():
  with begin():
    y -= 1


z = await_final(x)
```

Then `z` should terminate when `finite_source` does, not be blocked on `infinite_source` forever.

I'm not sure we need to implement this analysis right now, but our docs should acknowledge the gap and cite it as future work. Among that future work is a more rigorous model of when variables become coupled or don't. "One per set of mutable variables a `with begin():` block relates" is not sufficient.

sortalongo (approved):

src/ccl/design/mutability.md
dpmills: There's a callout in the next PR up (92) which implements await_final, and there's an actual fix in #102 (still a draft; a bit more to do on it before it's ready for review)

PR #92, opened 2026-08-13
Build the `await_final` terminal read as a completion-point sample, and give the as-of read its own term

dpmills:
`await_final(x)` — the terminal read of a transactional mutable variable — was specified but not built, so the batch transaction tests read their results through a trailing `with begin(): out << x` block: an arbitrary as-of sample that landed on the drained store only because `AsOfProducer` deferred a terminal trigger's latch until the store drained. This builds the primitive, gives the fed-out as-of read a term of its own, and drops that latch-timing gate. The batch suite now reads with `await_final`, so its expected values are semantic rather than scheduling artifacts.

## The read is a sample, not a reduction

A `Txn` read takes a key's carried value at some commit position, and a terminal read's position is where that key's own writers finish. `Builtin::FinalRead` is that term, and it compiles to a `StoreFinalRead` tile operator taking the same sample `AsOf` takes, through the same `store_current` — what differs between them is only what fixes the position, a trigger's arrival there and the store's closure here.

Two things follow from the read being a sample. It carries **no seed operand**, because tick 0 of every store is its keys' seeds — the fact `as_of_read` already relies on for having none. And `final_or_default` goes back to being the induction reducer alone, taking `final_or_default_read` with it as dead code.

Completion is still per store rather than per key, so an await settles once every writer of the *store* has drained. That gap is documented and is `dmills/per-key-completion`'s subject; closing it needs one fact and one disjunct rather than a second projection of the history. Mechanism: [src/ccl/design/mutability.md, "`await_final`"](src/ccl/design/mutability.md#await_final).

## Surface and typing

`lower_call`'s `await_final` arm → `Builtin::AwaitFinal` → `transact_phase::resolve_await_finals` is the whole path. Lowering resolves the operand by name, as it does a write target, so the mention is a handle position rather than a value read — the third such position — and the scheme `∀ν. Mut(ν, Txn) ⇒ ν` makes awaiting an induction accumulator a type error, with lowering rejecting it first and pointing at the trailing read that already gives a loop accumulator its final value.

A key **no writer site writes** never reaches the operator: `resolve_writer_free_awaits` replaces its await with its seed, its write history being statically empty. That covers a variable no block mentions and a variable some block only *reads* — a footprint key, since `{reads ∪ writes}` is one store, and so one whose runtime completion would otherwise wait on writers that cannot write it.

## The as-of read becomes its own term

`StorePlan::reads` mints `Builtin::AsOfRead` — an as-of read whose position is not yet supplied — where it minted a `final_or_default`, and `rewrite_as_of_reads` supplies the position by pairing the read with its reading loop. Position would not separate the two reads on its own: `channelize` copies a bound await's read inside the channel it closes, landing it in exactly the shape the rewrite matches. An unpaired read is now a compile error rather than one waiting on a store nothing drains, so `run` and `rewrite_as_of_reads` return `Result`; op-conversion keeps a backstop arm for each builtin. The pass and its helpers are renamed off "live", which named a distinction no path draws.

## Reviewing

`src/interpreter/tile_operators/map.rs` carries a separate fix `await_final` is what makes reachable: applying a data function before it has answered every key panicked, because the sealed path ignored the `domain_predicate` that distinguishes a key it will never answer from one it has not answered yet. Reasoning is in the comment there.

The test file is the largest hunk, most of it the batch suite switching to `await_final`.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #91
                                - #92 ◀
                                    - #93
                                        - #97
                                        - #102
                                            - #106

docs/chl-spec.md
```diff
@@ -2309,33 +2309,71 @@ Consequences a program may rely on:
...
+or every transaction denies, the await is `x`'s initializer.
+
+> **Current limitation.** The implementation is coarser: variables that some
+> `with begin():` block mentions together report completion as a group. Two
+> consequences, both defects against this section rather than intended
+> restrictions. `await_final(x)` does not settle while a variable `x` shares a
+> block with still has a live writer. And the cycle rule above is enforced against
+> that group, so a writer of a *store-mate* of `x` may not depend on
+> `await_final(x)` either, though no block writing `x` follows the await.
+> Variables that no block relates are unaffected by either.
```
sortalongo, line 2738: An idea I had for solving this problem: don't do `final_or_default` on the txn domain. Instead, do an `as_of` read against it with a timestamp which takes the max of the txn timestamps of all blocks that write to the variable. That lets us partition txn domains using connected components, but lets each variable go free as soon as its writers are done!

sortalongo (approved):
Woo, sound(er) semantics!

docs/chl-spec.md
dpmills: Took part of this idea and used it to improve and simplify how the final reads work.  The part that actually solves this problem entirely is upstack in #102

PR #93, opened 2026-08-13
Feed both writers' decision bodies through tiles, retiring `WriterBuffer`

dpmills:
Both mutation writers fed their decision body by pushing rows onto a `WriterBuffer` that the body's input leaf also held, so a value computed by one producer reached another without traversing an edge: the dependency was real but invisible to the producer graph, and the ordering it needed came from the driver's own code rather than from the dataflow. This splits each writer so the body's input is an ordinary tile, and gates the rule in CI — it was already documented in `src/interpreter/CLAUDE.md`, and broken anyway.

## The split

Each writer becomes a **drive** that produces the body input and a **store**/**writer** that consumes decisions. `InductionDrive` reads the changelog back through a `FanOut::new_cyclic` and holds no part of the recurrence: the store's decided frontier *is* the next position to iterate, and the previous accumulator is that key's value at it, so a row is a pure function of the store tile and the source tile. `TransactDrive` is the same shape over the transaction source, and owning that source makes it the writer's completeness signal. `WriterBuffer`, `BodyInputBuffer` and `BodyInputSource` are deleted; the mechanism is in [design-operators.md](src/interpreter/design-operators.md#the-commit-operator-interpretercommit_operatorrs). A shared `DriveWindow` holds everything the two drives do the same way, leaving each with only its own rule for which row to emit next.

## One position per outer pull

The cyclic fan serves a snapshot taken before the traversal began, so a position decided *during* a pull is invisible until the next. Induction drove every arrived position in one pull because it owned the engine; it now runs at the rate every other cyclic operator here does, which the doc argues as a retention problem rather than a rate one. `StoreDenseRead` gains a decided-positions filter so a `Memo` above it cannot latch a fold the store has not decided.

## The gate

`./ci.sh shared_state` flags shared mutable state under `src/interpreter/` unless its inner type is a known-legitimate kind or the site says why. It cannot prove an absence, only stop one being added silently — the checker's own docstring argues the rest. `CycleSlot` collapses the late-wiring cell both stores hand-rolled and is the shape the gate recognises; `shared_consumer` / `forwarding_consumer` give the notification wiring one spelling.

## Tests

The mutability, transaction and async-source suites are the gate and needed no behavioural changes. `a_contended_item_keeps_the_drive_window_flat` is new: the first operator-level test of the real drive / writer / commit-operator cycle, with six writers contending on one key so an item loses five times before winning. Two alternating writers make the loser retry once, so the end-to-end suite cannot reach that depth.

### DIFF
--- a/src/interpreter/CLAUDE.md
+++ b/src/interpreter/CLAUDE.md
@@ -24,12 +24,15 @@ Concretely:
   reads the prev-acc directly from a `Rc<RefCell<...>>` instead of
   through `get`") is a violation, even if it works on the test cases at
   hand.  Cyclic graphs go through `FanOut::new_cyclic` and the
-  re-entrancy machinery on `FanOutShared` — see
-  [`Recurse`]'s docs for the contract — but the data on the wire is
-  still a [`Tile`].
-- Constructor-time wiring (e.g., `Recurse::recursive_input_setter`)
-  passes `TileOperator` handles, not raw values.  The operator graph is
-  static; values flow through it at `get` time.
+  re-entrancy machinery on `FanOutShared` — see [`FanOutReentrancy`]'s
+  docs for the contract — but the data on the wire is still a [`Tile`].
+  A cyclic pull is served the fan's cached snapshot rather than
+  re-entering the inner producer, which is why such a cycle advances one
+  step per outer pull.
+- Constructor-time wiring (a [`CycleSlot`], filled through its
+  `setter` once the rest of the cycle exists) passes `TileOperator`
+  handles, not raw values.  The operator graph is static; values flow
+  through it at `get` time.
 - Side effects (I/O, sinks, notifications) live at the boundary —
   `compile_program` wires a `SinkConsumer` to the final operator, and
   `Scheduler::check_for_notifications` drives them.  Operators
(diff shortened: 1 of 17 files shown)


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #91
                                - #92
                                    - #93 ◀
                                        - #97
                                        - #102
                                            - #106

src/interpreter/CLAUDE.md
```diff
@@ -38,6 +41,38 @@ Concretely:
 If you find yourself reaching for a back-channel to avoid a tile
 shape that's awkward to express, fix the tile shape instead.
 
+**This rule is gated**, because a back channel is invisible to the
```
sortalongo: We don't need to put this in every session. Let Claude run into it and then learn about it.

src/interpreter/design-operators.md
sortalongo, line 1: Is "drive" really the right word? Or should it be the "driver"?

sortalongo (approved):

src/interpreter/CLAUDE.md
dpmills: Removed the whole section, pointer included.

Nothing is lost by letting it be discovered on the first failure: `check_shared_state.py`'s output already names the invariant, cites the CLAUDE.md heading, gives the `// shared-state-ok:` form, and makes the `EXPECTED_EXCEPTIONS` ratchet argument ("a justification is not a cost-free annotation — it is a permanent hole in the invariant").

What is left in this file's diff is only what had to change: `Recurse` was deleted, so those two bullets now point at `FanOutReentrancy` and `CycleSlot`.

src/interpreter/design-operators.md
dpmills: Renamed — `InductionDriver` / `TransactDriver`, plus `DriverWindow` / `DriverRow` / `DriverInputs`.

`driver` was already taken three times in this directory, and checking each one showed the blocker was itself a misuse: `operator_conversion` called the **writer** "a re-entrant / per-proposal driver" at two `Memo` sites. That is the thing which *pulls*, not a name for it, and the fact the `Memo` actually rests on is that the transaction writer re-pulls the body once per proposal — which is what both comments now say. `design-operators.md`'s heading "Convergence: the writer drives" was the same loose gloss; it now reads "the writer re-arms", which is what that section claims.

The other two uses are a different kind of thing and stayed: the `Units(1)` driver **row** in C-form arm dispatch (`map.rs`), and the outer loop that calls `check_for_notifications` (`scheduler.rs`).

PR #94, opened 2026-08-13
Read a value's trait requirements together, and write back what they determine

dpmills:
Currently, satisfying a trait obligation is delivery-driven: it learns only what *arrives*, one contribution at a time. That rejects a bad call, not a bad definition. In `f = \a -> (a + 1, a + "s")` each obligation narrows through its **other** operand — one to `{Int}`, one to `{String}` — so neither set empties and nothing ever compares them: no argument satisfies `a`, yet uncalled the definition type-checks. 

This PR adds a pass between emission and coalesce that reads each value's requirements together, rejects a value no type satisfies, and writes back what they determine as an additional constraint

This lets us correctly reject uncalled functions with trait bounds that are unsatisfiable, and also lets us give principal types to at least some more uncalled functions (e.g. \x -> x + 1 can now be typed as Int => Int, at least until we add heterogenous Addable).  It's also a requirement for any future work that wants to correctly assign principal ForAll types to functions that are legitimately polymorphic

### DIFF
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -1370,11 +1370,13 @@ This is the gap [Typechecking a never-called definition](#typechecking-a-never-c
 
 #### The unit is a place, not a variable
 
-A **place** is a variable plus a path of field selections; the empty path is the variable itself. Places are found by following **upper** bounds: `𝑣 <: 𝑈` means `𝑣`'s value reaches `𝑈`, so a requirement on `𝑈` is one on `𝑣`. A variable bound stays at the same place; a structural one descends, so in `𝑣 <: (𝑈₀, 𝑈₁)` the requirements on `𝑈₀` belong one field deeper and not to `𝑣`.
+Every position of a requirement is an ordinary inference variable, but a requirement is *about* a value, and one value is generally several variables. A **place** is that value. It is named by a root variable plus the path of field selections reaching it — each element of the path a **step** — and the empty path names the root's own value. `places_under` returns, per place, the variables standing at it together with the requirements they carry; it is one *place*'s requirements that are read together.
 
-A variable is the wrong unit because `λ 𝑎 𝑏 → …` passes its parameters through a tuple: each occurrence of `𝑎` is a separate variable, and none of them carries both of `𝑎`'s requirements. Curried, one variable carries both. Only the spelling differs.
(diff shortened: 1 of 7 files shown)


src/ccl/design/type-inference.md
```diff
@@ -1237,11 +1238,46 @@ A position that stays in the second row for the whole program — nothing ever d
...
-How much is determined follows from the table, associated positions included. `λ 𝑥 → 𝑥 + 1` is `∀𝐴 𝑂. (𝐴 : Addable(𝐴, Int ⇝ 𝑂)) ⇒ 𝐴 → 𝑂` with `𝑂` resolving to `Int`, because `Int` at the second position leaves only `Addable(Int, Int ⇝ Int)`. Adding `Addable(Float, Int ⇝ Float)` would leave two rows whose outputs disagree, and `𝑂` would be as open as `𝐴` already is — which is why a deposit waits for agreement rather than firing on a unique candidate.
+This is the gap [Typechecking a never-called definition](#typechecking-a-never-called-definition) names. The two are complementary. That walk resolves a dead definition's recorded bounds, which catches `λ 𝑎 → (𝑎.0, 𝑎.foo)`; this pass needs no delivery, and does not depend on whether anything calls the definition.
+
+#### The unit is a place, not a variable
+
+A **place** is a variable plus a path of field selections; the empty path is the variable itself. Places are found by following **upper** bounds: `𝑣 <: 𝑈` means `𝑣`'s value reaches `𝑈`, so a requirement on `𝑈` is one on `𝑣`. A variable bound stays at the same place; a structural one descends, so in `𝑣 <: (𝑈₀, 𝑈₁)` the requirements on `𝑈₀` belong one field deeper and not to `𝑣`.
```
sortalongo: But isn't a specific field of a structural bound itself a variable? I think this is introducing terminology that should just be explained with prose. "trait constraints come together on _interior variables_" or something like that

Actually, what kind of variable are we talking about? This text talks about type vars. The example below talks about lambda params.

src/ccl/design/type-inference.md
```diff
@@ -1237,11 +1238,46 @@ A position that stays in the second row for the whole program — nothing ever d
...
+
+A variable is the wrong unit because `λ 𝑎 𝑏 → …` passes its parameters through a tuple: each occurrence of `𝑎` is a separate variable, and none of them carries both of `𝑎`'s requirements. Curried, one variable carries both. Only the spelling differs.
+
+Which positions are steps is decided per type former, by an exhaustive match — a new `Type` variant that can hold a type fails the build rather than shrinking what the pass reaches. The rules are one comment per former at `places_under`, in `src/ccl/infer/solver/traits.rs`.
```
sortalongo:
```suggestion
Which positions are steps is decided per type former, by an exhaustive match. The rules are one comment per former at `places_under`, in `src/ccl/infer/solver/traits.rs`.
```

sortalongo (approved):

src/ccl/design/type-inference.md
dpmills: It ends up being a bit more complicated than that, and I think some new terminology really is warranted (and is in code too).  I cleaned it up to make things clearer and remove some unneeded terminology

src/ccl/design/type-inference.md
dpmills: Done

PR #97, opened 2026-08-13
Add an end-to-end walkthrough of mutability, from CHL source to operators

dpmills:
Mutability is the one part of Cambra with no path *through* it. `src/ccl/design/mutability.md` and the spec are organised by mechanism and answer questions you already know to ask; someone working on the system who does not know the induction/transaction layer deeply has nowhere to start. This adds [docs/mutability-walkthrough.md](docs/mutability-walkthrough.md) — a conceptual tour that follows two programs from CHL source to operators, keeping the design docs as the authority and linking out rather than restating.

Written for the reader who knows Cambra but not this layer: the model (a mutable variable *is* a function from a sequencing domain to a value), why induction and transactions are asymmetric, what a programmer writes, `Mut` as a type, the pipeline phase by phase, two worked examples, the engines, and the edges.

## Reviewing it

The stage listings are **real `RUST_LOG=debug` output**, regenerated against this branch's tip rather than written by hand — so read them for drift from the compiler, not for prose. If they are stale, the fix is to re-run the program, and the two programs are in the doc verbatim. Both still answer as the doc says (`6` and `35`) now that `await_final` compiles to a sample of the key's carried value rather than a reduction of its history, and the listings carry that shape: `pool ▷ final_read`, no seed operand.

§8's operator diagrams are the exception and the part most likely to be wrong: they are hand-drawn, because the graph renderer stops at a cyclic `FanOut` rather than walking back into the store it wraps. They describe the drive/store split as `InductionDrive` and `TransactDrive` implement it.

## The design-doc fixes it carries

Writing the tour surfaced four defects in `src/ccl/design/mutability.md`, fixed here rather than left for later: a variable's domain comes from the context that *writes* it, not from where it is introduced; the block-introduction case has no gate of its own and is rejected as `Unbound variable`, a diagnostic gap now stated as one; lowering makes one classification choice, not two, and the loop is where it decides nothing; and `transact_phase` was missing from the pipeline block. `type-inference.md` gets a matching one: its "three sites" that contribute to a variable's value type are three *rules*, and the third — a `Transact` key's seed — is the same rule shape one variable over, not a third input to the same join.


dpmills:

- #93
    - #102
        - #106
            - #97 ◀

docs/mutability-walkthrough.md
```diff
@@ -0,0 +1,903 @@
...
+                   ▲               │                         ▲                  │
+                   │               └─────────────────────────┘                  │
+                   │                   the ack branch                           ▼
+                   └────────────────── store fan (cyclic) ◀────────────────────┘
+```
+
+The drive's tile reaches the writer twice: once through the body, which turns it into a
+decision, and once directly. The second edge carries nothing the writer needs — it exists so
+the writer has something to **release**, and that release is the ack. That is why the writer
+holds an input it barely reads.
```
sortalongo: I don't understand this. Would like to discuss live.

sortalongo (approved):

docs/mutability-walkthrough.md
dpmills: Short answer is we need a signal backwards that stuff is committed.  It's awkward as a tile but fits pretty nicely as a release.  Release doesn't flow smoothly through intervening Memos though, so we need a direct edge.  Can discuss more live if you want

PR #98, opened 2026-08-14
Check NodeId uniqueness at every pipeline boundary, and fix what that surfaces

groundlar:
`Clone` was derived on `Expr`, so it copied `node_id`: a subtree reaching the output at two positions left both nodes sharing one identity, which makes the pane projection ambiguous and collapses the pair into a single entry in every `NodeId`-keyed walk. Only one pipeline boundary — `post-lowering` — checked for that. This adds the other seven (`post-inline`, `post-transact`, `post-letrec-run`, `post-desugar`, `post-live-read`, `post-lambda-elim`, `post-planning`), fixes the two violations they surface, and then **removes the class of defect entirely by making `Clone` freshen**.

## The two violations the new boundaries surfaced

**`simplify`'s zip-distribute** placed the left operand on both legs with bare clones, so the legs shared an identity. It survived because the follow-on product-beta consumes one copy per leg whenever the arms read different slots (`⟨.0, .1⟩`); only same-slot arms (`⟨.0, .0⟩`) leave both live. The first leg is now the survivor, the second a freshened sibling.

**`transact_phase::subst_var_with`** freshened the whole replacement, satisfying uniqueness by *deleting* the occurrence's id — and the read site is user-written, so its span went too. It now root-carries, like the two `subst_env`s below it.

`planning::groupby`'s key lift was flagged by the same audit and is already safe: it leaves the predicate domain through `lambda_elim::run`, which re-mints every node. `groupby_recognition_lifts_the_key_without_aliasing` pins that, so an elim that started preserving ids fails there rather than at a pane boundary.

## `Clone` freshens, and the opt-out is named

A clone is a **sibling**, not the same node. The hand-written `Clone` mints a new `NodeId` for every node it copies and reports each `(origin, fresh)` pair through `on_copy`. The recursion is the derive's — `node.clone()` clones the children, each reaching the same impl — so the freshen is deep by construction and **fused** into the copy rather than being the second walk `fresh_copy` made. `fresh_copy`, `freshen_node_ids_deep`, `freshen_interior_node_ids`, `freshen_node_id` and monomorphization's `freshen_clone_node_ids` are all deleted; 34 call sites become a plain `clone()`.

**The decision inverts rather than disappearing.** The safe default is now automatic, and a copy that is *not a new node* says so through `clone_preserving_ids`. Two shapes qualify — a snapshot taken for rollback or comparison, and a test comparing trees across a pass — plus the moves-out-of-a-borrow, where Rust forces a clone, the source is dropped, and the copy takes its position. Those last were *already* id-preserving here; only the spelling flips, since `clone()` used to preserve and `fresh_copy()` opted into freshening.

What it is **not** for is silencing a leak. An `Unexplained` or `ParentUnknown` means a copy was made with no frame open, or against an origin the table never recorded — a *recording* gap, whose honest fix is a bracket.

### Measured, because the objection was about cost

Release, min-of-5, against the unchanged parent, on the instrumented top of the stack (so the pane gates were live):

| program | compile time | peak RSS | ids |
|---|---|---|---|
| `stress_10_10_3_10` | 1.067 → **1.030 s** (0.97×) | +2.2% | 2.09× |
| `stress_10_10_5_12` | 3.702 → **3.472 s** (0.94×) | +1.3% | 1.74× |
| `stress_10_10_8_12` | 21.631 → **20.569 s** (0.95×) | +0.5% | 1.36× |

Compile time *improves*, consistently and with a mechanism: one fused walk replaces copy-then-walk. Isolating capture with `CAMBRA_LINEAGE=0/1`, the lineage table goes 5.2 MB → 10.3 MB on a 1 GB compile; with capture off the two arms agree to 0.05%, so the freshening itself costs no memory. Even the unmitigated arm — freshening everywhere, no opt-outs — was faster than baseline at 2–3× the ids.

## Lowering records its predicates

`collect_tree_ids` now reaches refinement predicates, so the lowering fold has to explain them. It could not: lowering builds a predicate from already-lowered sub-expressions plus the nodes minted and copied to join them up, seals it into a `Refinement` via `refined_data_fun`, and those assembly nodes live in a type slot outside the `walk_children` domain. 1331 of them across 80 pipeline tests. **Nearly half are copies that exist only because `Clone` freshens** — they used to alias already-tagged main-tree ids — which is why this rides in the same commit.

`LoweringContext::tag_predicate` sweeps the finished predicate at the three sites that build one. A copy-frame cannot do this job: `flush_into_lowering` deliberately asserts a lowering frame captures no mints (a frame-flushed step carries no span, so a mint routed through one would be silently unresolvable), and a predicate is a mixed mint/copy region.

**The sweep skips nodes that are already recorded, and that is the load-bearing part.** The fold is last-write-wins, so a blanket sweep leaves every node perfectly *explained* while silently replacing its real span and label with the sweep's coarse ones — 318 nodes on the corpus, with every gate green. `the_predicate_sweep_skips_already_recorded_nodes` pins it, because no leak class can. `assert_unique_node_ids` stays narrow: it answers uniqueness, not explanation, and predicates legitimately alias main-tree ids at inline's blind spot.

## What this PR cannot prove on its own

This commit gates on the lowering fold and uniquify's id-stability tripwire, and that is all — the `NodeId`-keyed table and the pane boundaries arrive in #107. The pane snapshots, `check()`'s scratch copy, the register-init stash and the type-domain discharges carry no local test; they are verified on the instrumented stack and land here on that evidence. Two markers record what the next commit must do:

- **`TODO(mono-frame)`** in `specialize_use`: when the recorder arrives, its frame must open **before** the clone, since the clone is what fires `on_copy` — a frame entered after it leaves the whole specialization `Unexplained`.
- **`TODO(predicate-domain)`** on `lineage::preserving_ids`: one legitimate caller (`PredMemo`'s rebuild), slated for removal once predicate mints are recorded. Inference's predicate producers are deliberately out of scope here — there is no pass recorder over inference yet, so a bracket would be a no-op with no test that could fail.

Full reading, including the eight distinct intents that were hiding under one `clone` and the exhaustive compiler-oracle sweep of all 118 production clone sites: `freshening-clone-report` in the vault.

---

## Rebased onto `main` (`7a9d1201`)

Restacked onto current `main`, which brought #91 (commit stores split by connected
components of transaction blocks) and #92 (`await_final` terminal reads).

**One new commit rides here:** `ccl(transact): a dropped as-of read moves its body
rather than freshening it`. `rewrite_as_of_reads`' dead-binding sweep
(`drop_dead_as_of_reads`) does `*e = (**body).clone()` to drop a `let` whose as-of
read nothing reads. That was a move when `clone()` preserved ids; with the
inversion above it re-mints the whole continuation instead, stranding every id
below the dropped binding — and on a reply chain that subtree is the rest of the
program. It now `mem::take`s the body, which copies nothing at all.

The site arrived upstream *after* the clone audit in this PR ran, so it was never
one of the 118 sites that sweep classified. Read that claim as "118 as of the
pre-rebase tree"; this is the 119th, and it is the one case the sweep's own
argument predicts — a `clone()` that was a move, at a call site nobody re-read
when the default flipped.

**The second placement of a register init is gone.** The trailing
`final_or_default(reg_x, init)` read that placed a key's seed a second time is now
an `as_of_read` with no seed operand (#92), so the freshening this PR added there
had nothing left to freshen and was dropped rather than ported.

**Boundary rename:** `assert_unique_node_ids(&expr, "post-live-read")` is now
`"post-as-of-read"`, following upstream's rename of `rewrite_live_reads` →
`rewrite_as_of_reads`.

**Stale numbers.** The compile-time / peak-RSS / id-ratio table above was measured
before the restack and has **not** been re-run. The mechanism it reports — one
fused walk replacing copy-then-walk — is unchanged by the rebase.

### DIFF
--- a/src/ccl/channelize.rs
+++ b/src/ccl/channelize.rs
@@ -2120,7 +2120,10 @@ fn extract_for_defer_impl(
                     let mut fvs = HashSet::new();
                     collect_free_vars(feed, &mut fvs);
                     if fvs.contains(&binding.name) {
-                        let placeholder = Expr::new(TypedExprNode::Lit(Lit::Unit));
+                        // A `mem::take` slot, overwritten below — mint nothing for
+                        // it (`NodeId::PLACEHOLDER`), or the recorder logs a birth
+                        // for a node that never reaches the tree.
+                        let placeholder = Expr::throwaway(TypedExprNode::Lit(Lit::Unit));
                         let original = std::mem::replace(feed, placeholder);
                         // stamp the wrap at construction —
                         // the let's type is its body's, closed over the binder
--- a/src/ccl/channelize.rs
+++ b/src/ccl/channelize.rs
@@ -284,7 +284,10 @@ fn try_extract_fanout_feed(body: &Expr, defer_name: &Name) -> Option<Vec<(Expr, 
 /// the lifted body, with stale Feed/Define target names renamed to `y`.
 fn try_lift_defer(binding_name: &Name, bound_expr: &Expr, body: &Expr) -> Option<(Expr, Name)> {
     let mut prefix: Vec<Expr> = Vec::new();
-    let mut current = bound_expr.clone();
+    // A move out of a borrow, not a duplication: the walk below destructures
+    // `current` and rebuilds it at its own ids, and what it yields *replaces*
+    // `bound_expr` in the output rather than standing beside it.
+    let mut current = bound_expr.clone_preserving_ids();
     loop {
         let cur_id = current.node_id;
         match current.node {
@@ -744,7 +747,11 @@ fn erase_chan_domains(expr: &mut Expr, map: &mut HashMap<Name, Type>) {
         erase_chan_domains(bound_expr, map);
         erase_chan_domains(body, map);
         // §6.2 Let-closing on the substitution content (see fn docs).
-        let discharge = crate::ccl::subst::Subst::discharge(&binding.name, (**bound_expr).clone());
+        // A type-level discharge. The term keeps its ids because it is a
+        // *template*, cloned again at every read; that read is where the sibling
+        // is minted.
+        let discharge =
+            crate::ccl::subst::Subst::discharge(&binding.name, bound_expr.clone_preserving_ids());
         for dom in map.values_mut() {
             *dom = discharge.apply_type(dom);
         }
@@ -2138,9 +2145,8 @@ fn extract_for_defer_impl(
                         // `bound_expr` in the body, and each extracted feed that
                         // captures the binder gets its own re-binding of the same
                         // definition, so every wrap is a copy.
-                        *feed =
-                            Expr::let_bind(binding.name.clone(), bound_expr.fresh_copy(), original)
-                                .with_ty(let_ty);
+                        *feed = Expr::let_bind(binding.name.clone(), bound_expr.clone(), original)
+                            .with_ty(let_ty);
                     }
                 }
                 new_body
@@ -2226,8 +2232,7 @@ fn extract_for_defer_impl(
                     let channel_lambda = Expr::lambda(&param.name, param.ty.clone(), v);
                     // Each companion channel applies the same source, which also
                     // stays on the rebuilt `Apply` below, so each gets its own copy.
-                    let channel =
-                        Expr::apply(new_argument.fresh_copy(), channel_lambda).with_ty(v_ty);
+                    let channel = Expr::apply(new_argument.clone(), channel_lambda).with_ty(v_ty);
                     feeds.push(channel);
                 }
                 let new_function = TypedExpr {
@@ -2389,7 +2394,7 @@ fn extract_for_defer_impl(
                                     // each arm's copy must carry its own ids —
                                     // a bare clone would put one identity at N
                                     // live positions.
-                                    let mut refined_prefix = source_prefix.fresh_copy();
+                                    let mut refined_prefix = source_prefix.clone();
                                     refine_source_domain(&mut refined_prefix, refinement_struct);
                                     let channel_lambda =
                                         Expr::lambda(&param.name, param.ty.clone(), value);
@@ -2450,7 +2455,7 @@ fn extract_for_defer_impl(
                             // The prefix stays in `new_elts` for the rebuilt
                             // compose, so each companion channel takes its own copy.
                             let mut channel_elts: Vec<Expr> =
-                                new_elts.iter().map(Expr::fresh_copy).collect();
+                                new_elts.iter().map(Expr::clone).collect();
                             channel_elts.push(channel_lambda);
                             // A single-element "compose" is just that
                             // element; otherwise build a Compose.
--- a/src/ccl/channelize.rs
+++ b/src/ccl/channelize.rs
@@ -747,9 +747,8 @@ fn erase_chan_domains(expr: &mut Expr, map: &mut HashMap<Name, Type>) {
         erase_chan_domains(bound_expr, map);
         erase_chan_domains(body, map);
         // §6.2 Let-closing on the substitution content (see fn docs).
-        // A type-level discharge. The term keeps its ids because it is a
-        // *template*, cloned again at every read; that read is where the sibling
-        // is minted.
+        // A discharge template is not a tree node: it is cloned again at every
+        // read, and that read is where the sibling is minted.
         let discharge =
             crate::ccl::subst::Subst::discharge(&binding.name, bound_expr.clone_preserving_ids());
         for dom in map.values_mut() {
@@ -2127,9 +2126,8 @@ fn extract_for_defer_impl(
                     let mut fvs = HashSet::new();
                     collect_free_vars(feed, &mut fvs);
                     if fvs.contains(&binding.name) {
-                        // A `mem::take` slot, overwritten below — mint nothing for
-                        // it (`NodeId::PLACEHOLDER`), or the recorder logs a birth
-                        // for a node that never reaches the tree.
+                        // A `mem::take` slot, overwritten below: minting for it
+                        // would log a birth for a node no tree ever holds.
                         let placeholder = Expr::throwaway(TypedExprNode::Lit(Lit::Unit));
                         let original = std::mem::replace(feed, placeholder);
                         // stamp the wrap at construction —
@@ -2141,10 +2139,6 @@ fn extract_for_defer_impl(
                         let let_ty =
                             crate::ccl::subst::Subst::discharge(&binding.name, bound_expr.clone())
                                 .apply_type(&original.ty);
-                        // The `Let` this walk is rebuilding keeps the original
-                        // `bound_expr` in the body, and each extracted feed that
-                        // captures the binder gets its own re-binding of the same
-                        // definition, so every wrap is a copy.
                         *feed = Expr::let_bind(binding.name.clone(), bound_expr.clone(), original)
                             .with_ty(let_ty);
                     }
@@ -2230,8 +2224,6 @@ fn extract_for_defer_impl(
                     // (the argument matches `param.ty`). Typed at construction.
                     let v_ty = v.ty.clone();
                     let channel_lambda = Expr::lambda(&param.name, param.ty.clone(), v);
-                    // Each companion channel applies the same source, which also
-                    // stays on the rebuilt `Apply` below, so each gets its own copy.
                     let channel = Expr::apply(new_argument.clone(), channel_lambda).with_ty(v_ty);
                     feeds.push(channel);
                 }
@@ -2390,10 +2382,6 @@ fn extract_for_defer_impl(
                                         .with_ty(Type::Base(BaseType::Bool));
                                     let refinement_struct =
                                         Refinement::born(Rc::new(pred_on_source));
-                                    // One refined source per feeding arm, so
-                                    // each arm's copy must carry its own ids —
-                                    // a bare clone would put one identity at N
-                                    // live positions.
                                     let mut refined_prefix = source_prefix.clone();
                                     refine_source_domain(&mut refined_prefix, refinement_struct);
                                     let channel_lambda =
@@ -2452,10 +2440,7 @@ fn extract_for_defer_impl(
                             // handle's rigid `ChanDom` domain, closed by the
                             // final `erase_chan_domains` substitution.
                             let channel_lambda = Expr::lambda(&param.name, param.ty.clone(), v);
-                            // The prefix stays in `new_elts` for the rebuilt
-                            // compose, so each companion channel takes its own copy.
-                            let mut channel_elts: Vec<Expr> =
-                                new_elts.iter().map(Expr::clone).collect();
+                            let mut channel_elts = new_elts.clone();
                             channel_elts.push(channel_lambda);
                             // A single-element "compose" is just that
                             // element; otherwise build a Compose.
--- a/src/ccl/channelize.rs
+++ b/src/ccl/channelize.rs
@@ -267,7 +267,38 @@ fn try_extract_fanout_feed(body: &Expr, defer_name: &Name) -> Option<Vec<(Expr, 
     any_feed.then_some(arms)
 }
 
-/// Attempt to apply the defer-returning lift to a `Let` binding.
+/// The defer-returning lift's output.
+struct DeferLift {
+    /// `let y = Defer in …` — the two defer scopes merged into one.
+    expr: Expr,
+    /// The inner defer binder, which the lift renamed to the outer name.
+    inner_name: Name,
+    /// The channel domain of the inner handle the lift replaced, when its type
+    /// records one.
+    inner_chan_dom: Option<(Name, crate::ccl::ChanLevel)>,
+}
+
+/// Whether `bound_expr` has the defer-returning lift's shape: any `ExprStmt`
+/// prefix, then `let x = Defer in body_x` with `body_x` defer-returning.
+///
+/// [`lift_defer`] consumes its input, so the shape is decided here first. A
+/// matcher that failed partway through would have to rebuild what it had already
+/// taken apart, or work on a copy.
+fn is_lift_shape(bound_expr: &Expr) -> bool {
+    let mut current = bound_expr;
+    while let TypedExprNode::ExprStmt { body, .. } = &current.node {
+        current = body;
+    }
+    matches!(
+        &current.node,
+        TypedExprNode::Let { binding, bound_expr: inner_be, body }
+            if matches!(inner_be.node, TypedExprNode::Defer)
+                && is_defer_returning(body, &binding.name)
+    )
+}
+
+/// Apply the defer-returning lift to a `Let` binding whose `bound_expr` has
+/// passed [`is_lift_shape`].
 ///
 /// Pattern: `let y = (let x = Defer in body_x) in body_y` where
 /// `body_x` is *defer-returning* (ends in `Var(x)` after walking
@@ -282,14 +313,10 @@ fn try_extract_fanout_feed(body: &Expr, defer_name: &Name) -> Option<Vec<(Expr, 
 /// Also handles an optional `ExprStmt` prefix on `bound_expr`: the
 /// heads of any leading ExprStmts become a prefix that's prepended to
 /// the lifted body, with stale Feed/Define target names renamed to `y`.
-fn try_lift_defer(binding_name: &Name, bound_expr: &Expr, body: &Expr) -> Option<(Expr, Name)> {
+fn lift_defer(binding_name: &Name, bound_expr: Expr, body: &Expr) -> DeferLift {
     let mut prefix: Vec<Expr> = Vec::new();
-    // A move out of a borrow, not a duplication: the walk below destructures
-    // `current` and rebuilds it at its own ids, and what it yields *replaces*
-    // `bound_expr` in the output rather than standing beside it.
-    let mut current = bound_expr.clone_preserving_ids();
+    let mut current = bound_expr;
     loop {
-        let cur_id = current.node_id;
         match current.node {
             TypedExprNode::ExprStmt {
                 expr: head,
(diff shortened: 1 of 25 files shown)


src/ccl/design/provenance.md
```diff
@@ -71,6 +71,65 @@ shares a few incidentally (`pred_sources = gen_sources.clone()`), but the nodes
...
+order, so a reordered pass carries its check with it. It also means a pass is only
+implicated at a boundary that looks at it: a clean run is evidence about the
+gates, not about the passes between them.
+
+Two primitives, and the choice between them is about what the copy *denotes*:
+
+- **`Expr::fresh_copy`** — duplication. The copy is a *sibling*: same value,
+  distinct identity, `annot(p) = annot(o)`. Use it wherever one subtree reaches
+  the output at more than one position.
+- **Root-carry** (`freshen_interior_node_ids` + `re_root`) — substitution. The
```
dpmills: Is there any harm in changing the root node_id too in this case?

src/ccl/design/provenance.md
```diff
@@ -71,6 +71,65 @@ shares a few incidentally (`pred_sources = gen_sources.clone()`), but the nodes
 actually blamed for guard errors are minted fresh in predicate position and have
 no main-tree twin.
 
+### Duplication discipline
+
+**Every pass yields unique `NodeId`s.** `Clone` is derived, so it copies
```
dpmills: This reads like an accident causing us to build excess machinery.  We could impl clone instead of derive it so that ids are always refreshed.  That prevents getting this wrong in the future automatically

src/ccl/design/provenance.md
```diff
@@ -71,6 +71,65 @@ shares a few incidentally (`pred_sources = gen_sources.clone()`), but the nodes
...
+original id is a *fate* question, and keep-first guesses it wrong exactly when
+position 0 is the copy that later dies. Freshening needs no knowledge of
+downstream fates: the original either survives in place and keeps its id, or dies
+and is consumed by the rewrite that dropped it. (Lowering's `fan_out_copy` is
+keep-first on purpose — it is the pass that *mints* the originals, so position 0
+is the source image by construction.)
+
+**Freshen at placement, not at construction.** Most passes build intermediate
+structures — guard vectors, path conjunctions, per-branch environments — whose
+entries are aliased and then copied into the output. Freshening where a node is
```
dpmills: Is this actually important? It seems like it complicated the model without a clear payoff, especially if we are going to the live-map model instead of a log in the future

src/ccl/design/provenance.md
groundlar: Measured this instead of arguing it, and two of my objections didn't survive.

**It is implementable.** I enumerated the sites that reach `TypedExpr::clone` only *through* a container (`Subst`, `Mapping`, `Box<Expr>`, `Vec<Expr>`, `Rc::make_mut`, `Cow::to_mut`) — 39 lines — by dropping `Clone` from `TypedExpr` and adding an inherent shadowing `clone`, so only indirect paths error. Then I patched each one to freshen and ran the suite: **5 lines must preserve, 31 are indifferent (nothing detects the freshen), 3 never execute, 0 are unreachable.** All 5 are in `src/ccl/subst.rs` (`Subst::then`, `licensed_correspondence_view`) — our code, so an explicit `dup()` is writable at each. I had thought `Cow::to_mut` and `Rc::make_mut` were unfixable because std calls `Clone::clone`; both were tested freshening and left the suite green. Migration is those 5 lines plus the ~108 direct sites rustc points at.

**And you're right that this PR is detection, not prevention.** Chasing it turned up three live bare-clone duplications the asserts don't flag, because no test walks the path. Two are in `lambda_elim`'s per-arm scrutinee placement, one in `mut_elim`'s per-feed loop source. Mint-by-default would have made all three unwritable.

**What still argues against it is tracking cost and validation asymmetry.**

Cost: mint-on-clone raises ids minted per compile by 2.6× / 3.7× / 7.5× / 34.6× on four corpus programs, 6.1× suite-wide, measured with a counting `Clone` against the two production `NodeId::fresh()` callers. The table is keyed by minted id and `NodeId`s go out on the wire, so that is 1–2 orders of magnitude on rows and payload. "The node-heavy passes already freshen everything" does not cover it: inference/mono holds 67% of clone events and mints 2.4% of ids today, so +123× there, against +0.74× in `lambda_elim` and +1.35× in `planning`. It is a large multiple of a small number, in the phase the argument misses.

Validation: "a mint was performed needlessly" is not a checkable property, while duplicate ids are an `O(nodes)` tree walk needing no knowledge of intent. Tested directly — freshening where an id must be preserved, at `subst.rs:425`, was exercised 53,469 times and left every pane's leak counts byte-identical. Inside an open lineage frame a wrong freshen rows a well-formed `Copy`, and nothing checks how many live positions an origin has. Outside a frame it is caught (the pane snapshots fail loudly), but every in-pass rewrite is inside one.

So: today's default has a silent-but-checkable failure mode; mint-by-default has one that takes effort to write and cannot be checked once written, at 1–2 orders of magnitude in tracking.

**Three findings from the rebase onto `main`, since #54 changed the picture.**

1. `match` is real syntax now, so the two `lambda_elim` scrutinee sites are reachable from source. Both are fixed here.
2. No test accompanies that fix, deliberately. I built the triggering program (a `def` matching on `p.v` with two arms), reverted the freshen, and instrumented: the arm fires with `scrut_is_id=false nbranches=2` and both placements carry the same id — a real duplication — yet the compile passes, because the catch-all arm re-mints every pass-through node and launders it before any boundary walk. A source-level regression would pass either way.
3. The rebase surfaced a duplication of the same shape in #54's own `build_scrutinee_case_cform`: `scrut_stream` (the `const` lift) is built once and composed into every arm, with all arms live in the union. 27 `variants::` tests failed at `post-lambda-elim` with `x ▷ const` duplicated across the two arms of the `⊎`. Fixed here by freshening at the placement.

That third one is the assert earning its keep, and it is also why fixing the `TODO(preserve)` re-mint is worth doing: it currently masks `ccl_utils::synthesize_arm_predicate`'s guard clones too, which duplicate in real compiles (253 hits across the suite). Whoever takes that change should expect to fix these together.

src/ccl/design/provenance.md
groundlar: No harm — you're right, and the doc's reasoning was wrong.

For anyone reading cold: root-carry is how substitution replaces a `Var(𝑥)` occurrence with a compound term. The replacement's root takes the occurrence's own id, and only the interior is freshened.

I had assumed freshening the root would re-attribute the node to the definition site instead of the use site. It wouldn't, because of ordering: the carry happens first, so by the time a freshen fires the root already holds the occurrence's id — the recorded copy names the occurrence as its origin, and attribution resolves to the same use-site span one hop out. There is no phantom `Copy` either; the produced id is the one the output node ends up holding.

What is left is cost, not correctness. Carrying is free: a clone mints nothing and re-rooting records nothing, so the output root gets no row at all — no id spent, no hop to walk, and the occurrence stays in the live set. Freshening buys a row and a hop for no new information, drops the occurrence out of the live set so it appears in the death report (reported, not gated), and inside refinement predicates adds an edge to an id the window has no row for, which is gated. Supporting that this is bookkeeping rather than semantics: nothing reads a substitution template's ids before placement — every read of the replacement term is by name or structure, and structural equality ignores ids.

Fixed here rather than in a follow-up, since it is one correction in two places: the doc bullet no longer implies the span depends on carrying, and the comment at the substitution site no longer describes a freshen-then-carry ordering the code doesn't use.

src/ccl/design/provenance.md
groundlar: Largely not, and you're right to push. I had conflated two shapes with very different stakes. Freshening before placement guesses the copy's multiplicity, and there are three outcomes:

- **Moved** into the output — fine. The output node holds the fresh id and its parent is the original.
- **Copied again** into the output — a real defect, and the gate catches it. The intermediate generation is stranded, so the placed copy's `Copy` names an origin no node holds and a `parents` walk for a span dead-ends. Demonstrated: freshening the `Subst`-resident templates in `Subst::then` / `licensed_correspondence_view` yields `parent_unknown: 246` on `generator_pipeline` and fails `pane_boundaries_fold_with_no_structural_leaks` and `monomorphization_explains_every_node_it_produces`.
- **Dropped** — pure waste, not a correctness problem. One id spent from the counter and one row nothing will ever query. Nothing incorrect happens: no duplicate ids, no surviving node with damaged attribution. And nothing detects it — both leak checks enumerate from the tree, there is no produced-side check, and a node absent from the tree is outside `assert_unique_node_ids`. Two tests pin this shape as tolerated on purpose (`born_copied_discarded_template_composes_without_leaks`, `lowering_born_copied_discarded_template_composes_away`), so we already have sites doing it.

So as a correctness rule it is close to vacuous: the harmful case is already caught by the gate, and the uncaught case only costs an id. What the prose is good for is telling an author how to avoid tripping that gate — worth keeping as guidance, not as an invariant sitting among the real ones. The old justification was also stale, phrased in the retired declared-model's terms ("declared dead by whoever claims the region").

The paragraph is rewritten around those three outcomes, with the drop case labelled as waste. If you would rather mechanize the waste case than document it, the check exists and nothing calls it: `LineageTable::deaths(&live)` is exactly `recorded_rows ∖ live` — though gating it at zero needs those two tolerated template sites cleaned up first.

src/ccl/expr.rs
groundlar: (note this is removed in an  upstack PR)

src/ccl/design/provenance.md
dpmills: How is "pass an identity primitive"?

src/ccl/design/provenance.md
```diff
@@ -36,40 +36,137 @@ prose describes code you can go read.
...
-The cost, accepted: an inference error blamed on a predicate-interior node
-resolves to no span (the guard of `[x for x in xs if x > "a"]` reports without a
-caret) because the id is not in the lowering projection. Fixing that means
-seeding predicate-position nodes into the fold as live roots and making
-`output_ids` predicate-inclusive — deferred; see the lineage-redesign doc's
-decisions 16-17. Sharing ids with the main tree is *not* a fix: lowering already
-shares a few incidentally (`pred_sources = gen_sources.clone()`), but the nodes
-actually blamed for guard errors are minted fresh in predicate position and have
-no main-tree twin.
+**Two questions, two domains.** Keep them apart, because the answers differ:
```
dpmills: This claims there are two "domains" that are kept "apart", but what are the domains? We just have NodeIds.  This whole section is titled "The ID domain" but it never says what type of domain we're talking about.

I think we should rewrite this section entirely.  It should be roughly
- Properties of the NodeId (uniqueness, not part of Eq, always new IDs when creating/copying exprs, uniqueness exception for refinements)
- Simple explanation of how the uniqueness is maintained
- Operations we can do on NodeIds, like collecting them all, asserting uniqueness, etc

Each of these just needs to be stated once; the current docs are very repetetive

After rewriting, point a fresh agent at the docs and have it check that they are understandable without having the entire later design in context already

src/ccl/design/provenance.md
```diff
@@ -36,40 +36,137 @@ prose describes code you can go read.
...
+- **`clone_preserving_ids`** — the copy *is the same node*, so it keeps its ids.
+  Sound because the copy is never reachable from a tree beside its source, and
+  narrow: a `Subst` discharge template, which is not a tree node and is copied
+  again at every read; a retained snapshot or rollback copy, which replaces or
+  shadows its source; and a test comparing trees across a pass.
+
+  **Not a way to silence a leak.** An `Unexplained` or a `CopyOfUnknown` means a
+  copy was made with no step open, or against an origin the log never recorded.
+  That is a *recording* gap, and its fix is to record the copy.
+- **Root-carry** (`clone().re_root(id)`) — substitution. The replacement for a
```
dpmills: I thought we had gotten rid of this one; do we still need it?

src/ccl/lower/comprehension.rs
```diff
@@ -353,12 +354,15 @@ pub(super) fn lower_list_comp(
...
+    let copy = if *used {
         use crate::ccl::lineage::copy_frame;
         let _frame = copy_frame(label);
-        copy.freshen_node_ids_deep();
-    }
+        origin.clone()
+    } else {
+        // Keep-first: this placement *is* the original, so it keeps its ids and
+        // records nothing. Only the second and later arms are siblings.
+        origin.clone_preserving_ids()
```
dpmills: This seems unnecessary, we can just unconditionally add the copy_frame.  This is creeping back to the per-clone-decision stuff we're trying to avoid

src/ccl/channelize.rs
```diff
@@ -284,7 +284,10 @@ fn try_extract_fanout_feed(body: &Expr, defer_name: &Name) -> Option<Vec<(Expr,
 /// the lifted body, with stale Feed/Define target names renamed to `y`.
 fn try_lift_defer(binding_name: &Name, bound_expr: &Expr, body: &Expr) -> Option<(Expr, Name)> {
     let mut prefix: Vec<Expr> = Vec::new();
-    let mut current = bound_expr.clone();
+    // A move out of a borrow, not a duplication: the walk below destructures
+    // `current` and rebuilds it at its own ids, and what it yields *replaces*
+    // `bound_expr` in the output rather than standing beside it.
+    let mut current = bound_expr.clone_preserving_ids();
```
dpmills: Worth seeing if there's a way to refactor to avoid this; this is a real clone, not a snapshot etc like clone_preserving_ids is supposed to be used for

src/ccl/channelize.rs
```diff
@@ -2131,6 +2141,10 @@ fn extract_for_defer_impl(
                         let let_ty =
                             crate::ccl::subst::Subst::discharge(&binding.name, bound_expr.clone())
                                 .apply_type(&original.ty);
+                        // The `Let` this walk is rebuilding keeps the original
+                        // `bound_expr` in the body, and each extracted feed that
+                        // captures the binder gets its own re-binding of the same
+                        // definition, so every wrap is a copy.
```
dpmills: Looks like there are a ton of unnecessary comments due to the historical progression of this PR.  Cloning gives new IDs by default and basically every part of compilation is supposed to work that way, so we don't need to comment on each one.  It's only worth commenting on the preservation case since that's the exception

src/ccl/channelize.rs
```diff
@@ -2432,7 +2452,10 @@ fn extract_for_defer_impl(
                             // handle's rigid `ChanDom` domain, closed by the
                             // final `erase_chan_domains` substitution.
                             let channel_lambda = Expr::lambda(&param.name, param.ty.clone(), v);
-                            let mut channel_elts = new_elts.clone();
+                            // The prefix stays in `new_elts` for the rebuilt
+                            // compose, so each companion channel takes its own copy.
+                            let mut channel_elts: Vec<Expr> =
+                                new_elts.iter().map(Expr::clone).collect();
```
dpmills: Revert

src/ccl/expr.rs
```diff
@@ -863,6 +900,63 @@ impl TypedExpr {
...
+    ///
+    /// # 1. A `Subst` discharge template
+    ///
+    /// A [`Subst`](crate::ccl::subst::Subst) discharge payload is never a tree
+    /// node. `Mapping::as_expr` clones it afresh at every read, and that read is
+    /// where each sibling is minted, so copying the template itself must mint
+    /// nothing. Every site of this shape is either the argument to
+    /// `Subst::discharge` or `Mapping`'s own `Clone` propagating one.
+    ///
+    /// # 2. A copy that replaces or shadows its source
```
dpmills: This bullet doesn't match the text.  Maybe "A throwaway copy" or something like that.  The current heading licenses too much, and that slipped into the PR in the cases I noted earlier

src/ccl/lambda_elim.rs
```diff
@@ -380,21 +385,30 @@ fn build_value_case_cform(
...
     if arms.len() == 1 {
+        // The single arm is discarded, so this body reaches the output once and
+        // keeps the branch's own ids.
         return Ok(default_body);
     }
+    // Past here the last branch's body reaches the output *twice*: as its own
+    // gated arm, and as `final_or_default`'s default. The arm is the copy that
+    // actually fires, so it keeps the source ids and the unreachable type anchor
+    // is the freshened sibling.
+    let default_body = default_body.clone();
```
dpmills: This comment doesn't make sense and the clone is probably unneeded

src/ccl/subst.rs
```diff
@@ -651,20 +705,10 @@ impl Subst {
             let occurrence_ty = e.ty.clone();
             *e = repl.as_expr_preserving(e.node_id, &occurrence_ty);
             if !matches!(e.node, TypedExprNode::Var(_)) {
```
dpmills: This is now an empty block with a comment; probably irrelevant

dpmills (approved):

src/ccl/channelize.rs
groundlar: Refactored in `ccl(channelize): the defer lift matches before it consumes`. `is_lift_shape` decides the shape through a borrow, and `lift_defer` consumes its input and is total, with the pattern it relies on stated as a `debug_assert!`. The `clone_preserving_ids` is gone, and so is the hand-rolled re-root and its `TODO(preserve)` — the spine loop puts the node back on the expression it moved it out of.

`find_defer_chan_dom` went with it: `lift_defer` reads the inner handle's channel domain off the binding it replaces, which is what that function re-walked the tree to recover.

src/ccl/channelize.rs
groundlar: Pruned. Four in this function — the feed re-binding wrap, the companion-channel `Apply`, the refined-source-per-arm block, and the compose prefix below — and the same pass over the rest of the diff. What is left sits on preservation sites.

src/ccl/channelize.rs
groundlar: Reverted: back to `new_elts.clone()`, comment gone.

src/ccl/expr.rs
groundlar: Now `# 2. A throwaway copy`, and it gained the line the old heading left out: a copy that *reaches the output* is not this shape, even when the source is dropped on the way.

The three sites the wider wording licensed are gone too — lowering's comprehension fan-out, its chained-comparison operands, and the defer lift.

src/ccl/expr.rs
groundlar: Correction to my own note: it is removed in *this* PR, not upstack. `re_root` is deleted, and substitution root-carries through `Mapping::as_expr_preserving`, which builds the copy's root at the carried id instead of minting one and overwriting it.

Where that method itself goes is the `provenance.md` root-carry thread.

src/ccl/lambda_elim.rs
groundlar: Both gone. The `body.clone()` in the `i == last` branch already hands the default its own copy, so the second clone was redundant.

src/ccl/subst.rs
groundlar: Removed — the arm carries the occurrence's id onto the replacement and returns.

src/ccl/lower/comprehension.rs
groundlar: `fan_out_copy(origin, label)` is unconditional now: one `copy_frame`, one `clone()`. The `used: &mut bool` is threaded out of `lower_list_comp`, `float_comp_source_case` and `fan_out_element_case`, and the keep-first paragraph is out of the doc comment.

src/ccl/design/provenance.md
groundlar: It isn't one. `Pass` moved to "The lineage model", which is where it lives in the data, and the section is now "Node identity".

src/ccl/design/provenance.md
groundlar: Rewritten to that outline. "The ID domain" is gone. The sections are "Node identity" — four properties of a `NodeId`: it is not part of a node's value, two live nodes never share one, construction and copying both mint, and uniqueness is asserted on the main tree only — then "Maintaining uniqueness", "Walking the ids" (the three walks and the question each answers, as a table), and "Duplication". The repetition is cut; the identity half of the doc is a net 234 lines shorter.

The fresh-eyes pass ran and found some inconsistencies--good idea!

src/ccl/design/provenance.md
groundlar: Per our Slack thread: `re_root()` and `clone_at()`  go away, and the root-carry inlines at its one call site in `rewrite_expr_go`, so the pattern stops being an API affordance.

PR #99, opened 2026-08-14
Carry the FunKind through compilation instead of reconstructing it

dpmills:
Data-vs-compute (`FunKind`) was only correct up until inference, because it's only use site was there.  Multiple phases didn't propagate it correctly, and this resulted in not being able to retire all hacks that needed the distinction and the post-inference checks running kind-blind `ConstrainCache` to tolerate that. The latter also switched off **data-domain invariance**, hiding real defects.

The bulk of the change is mechanical plumbing; lots of transformations were losing the kind information and they all need to preserve it.

This allows things that should be dependent on FunKind to actually depend on it:
- Inlining now only inlines compute functions
- Monomorphization only applies to compute functions
- Iteration can validate that it only tries to iterate data functions

This had several consequences:
- The MapResult operator needs to be able to take a Scalar and look it up in a CurriedFunction to get a SealedFunctionResult
- Planning needs to be slightly smarter about groupby handling to recognize non-inlined, non-monomorphized shapes


dpmills:

- #99 ◀

src/ccl/design/optimization.md
```diff
@@ -27,20 +27,9 @@ Performing this before `lambda_elim` prevents the let-in-lambda rule from hoisti
 
 ### What is inlined (UDF step)
 
-A `Let { binding, bound_expr, body }` is inlined when `should_inline(bound_expr.ty)` returns `true`:
+A `Let { binding, bound_expr, body }` is inlined when `bound_expr.ty` is a **capability** — a `Fun` whose [`FunKind`](type-inference.md#46-data-vs-compute-functions) is not `Data`. That is the whole rule.
```
sortalongo, line 30: This isn't what we want long term, though. Data function inlining corresponds to loop fusion, which is preferable more often than not. What's the reason to turn it off completely? Either way, leave a note saying that data function inlining is an optimization we'll introduce in the future.

src/ccl/design/type-inference.md
```diff
@@ -124,7 +124,7 @@ def identity(x):
...
...**, integrated into the coalesce walk (`infer::specialize_use`): a use of a generalized binding is specialized at first visit — clone + `freshen_expr_type_slots`, a two-way pin against the use's *live* instantiation type, and a re-entrant coalesce of the clone — and the binding's `let` rebuilds itself as the chain of demanded specializations. Specialization is keyed on the use's **instantiation identity** (a `SpecKey`, not a resolved type — see [Keying a specialization](#keying-a-specialization)), so uses that instantiate the definition identically **share** one definition. So `def f(x): x == x; f(1); f("foo")` type-checks and runs, a generator used at two element types compiles to two cached specializations (see F2), and a generalized UDF used only inside *another* generalized definition (poly-calls-poly) specializes by plain recursion — its use becomes concrete inside each wrapper clone's re-entrant walk. Levels are live (extrude fires on a genuine level mismatch). A `let` bound to a **collection** is not generalized, whatever its RHS looks like: a data function's domain *is* its data, so specializing one duplicates the data rather than the recipe, and every use would rebuild it.
```
sortalongo: (worth line splitting here, btw)

>  A `let` bound to a **collection** is not generalized, whatever its RHS looks like: a data function's domain *is* its data, so specializing one duplicates the data rather than the recipe, and every use would rebuild it.

Hm... IIUC, the "specialization" we're talking about is refinements, right? So before this, we'd monomorphize a collection to each filter that was applied to it and recompute each one separately. If so, can you give that as an example here?

Furthermore, it's not obvious that we _never_ want to monomorphize in this case. Indeed, doing so in the case of refinements is just filter pushdown. If you have selective filters, you typically want to do it. If you have lots of readers with unselective or overlapping filters, you don't. It's not one or the other.

Even if we don't want to implement the machinery to decide this right now (IMO we don't), please leave a note on the need for such decisions in the future.

sortalongo (approved):
The correspondence between inlining and fusion, and monomorphization and filter pushdown, make me think... do these create opportunities for more intuitive compiler hints? For example, using `:` vs `<:` already determines whether a function is polymorphic. If monomorphization is filter pushdown, does using `:` act as a hint to the compiler to not do a filter pushdown? Is there an analogously clean way of expressing inlining? (aside from the classic `@inline` annotation)

src/ccl/design/optimization.md
dpmills: Two things worth separating: this PR doesn't turn data-function inlining off — it was never on. The rule it replaces is `is_iterable_domain`, a structural test for "can the operator graph enumerate this domain", and it declined `UIntRange`/`DataSource` domains just the same. What changed is the *signal* (carried kind vs. reconstructed shape), not the policy.

On the substance you're right, and the text I wrote was worse than the old one about it — "that is the whole rule" reads as though there were nothing left to decide. Added `#### Inlining a collection is loop fusion`, which says what you said: substituting a collection fuses the use site's pipeline into the source's chain exactly as an inlined capability's body fuses into its caller's, so refusing it is a policy and not a structural bar. What keeps it refused is that the trade is one materialization against N recomputations of the source — worth it when the source is cheap or singly-used, not when it's expensive with many uses. Deciding per binding needs a cost model we don't have (use count is a walk away, source cost isn't), so the rule takes the bounded-worst-case side and fusion waits behind that model.

src/ccl/design/type-inference.md
dpmills: Split — and the block moved out entirely. It's now `### Generalizing a collection is filter pushdown` under 4.6, alongside "The domain join is a Σ", which is where `FunKind` is defined and where the test is meaningful. The roadmap bullet keeps one sentence and a link; its siblings are one-liners and this had grown to a paragraph.

On the mechanism: yes, and more directly than what I'd written. The shape that actually reaches `should_generalize` is `groupby`, whose type at that point is still

```
(__gb_k: ?k) ⤇ ((__gb_i: {?i | __elem ▷ xs ▷ key == __gb_k}) ⤇ ?v)
```

so the domain refinement *is* the dependent group-key predicate from 4.5, and a `SpecKey`'s negative read follows precisely the `arg <: domain` edge that pins `__gb_k`. Specializing a grouping per use is therefore one copy of the source filtered to each reader's key — filter pushdown, as you say.

Your second point is in as the reason the blanket refusal is a placeholder rather than an answer, with `sum(g(1)) + sum(g(2)) + sum(g(3))` as the losing case: three specializations rebuild the whole partition three times where one `Memo` serves all three. Selectivity against reader count is the decision, and inference is the wrong layer to make it — it runs before planning knows an extent. Cross-linked with the inlining note, since it's the same missing cost model showing up on the term side.

PR #100, opened 2026-08-14
A grouping is a collection: bind it once, share it

dpmills:
A `groupby` was inlined at every use, rebuilding the whole scan-and-partition per key, and iterating one twice did not compile at all. The cause was `should_inline` asking whether a binding's *domain* was finitely enumerable — a question about what planning can compile, not about what the value is — and a grouping's key domain is an unbounded `Int`. It now asks the [`FunKind`](src/ccl/design/type-inference.md#46-data-vs-compute-functions): a capability is inlined because there is nothing behind it to share, a collection is bound once and shared behind a `Memo`. `is_iterable_domain` is deleted with its tests.

Read the commits in order; each is needed by the next.

## The kind has to survive elimination first

Elimination rebuilt arrows through combinator constructors that mint the capability kind, and its desugaring arms re-entered on a rewritten body, so a `groupby` reached planning as `(k: Int) ⇒ …` having been `⤇` at inference. `elim_lambda` now takes the kind.

## Then two holes close

The group-by recognizer only matched a grouping at the head of a `Compose` — the consumed-in-place shape — so a `let`-bound one fell through to the generic path. And `MapResult` had no arm for a `Scalar` input against a `CurriedFunction`: that is the single-key lookup, the same walk the stream case does at one key. Its absence is why the lookup had to be inlined to work.

## Generalization asks the kind too

A collection is a value binding however it is spelled, and `groupby` lowers to a `Lambda`. Sharing it then needs its domain *named* rather than merely ordered, which is the comprehension change — ported from `dmills/collapse-at-position`, which needs it for its own reasons. Whichever lands second drops the duplicate.

One `IterateExtent` now, however many lookups or iterations.


dpmills:

- #57
    - #86
        - #54
            - #99
                - #100 ◀

dpmills:
Folded into #99 — the two were the same change split at an arbitrary point.

PR #101, opened 2026-08-14
Collapse the phase-local substitution copies onto one Subst engine

groundlar:
The mutability-elimination phases each carried their own substitution: a `subst_env` in `mut_elim`, another in `transact_phase`, and a single-name `subst_var_with` beside it. `Subst` already is a simultaneous map, so these were a second engine reimplementing it — and drifting, since a property fixed in one copy did not reach the others. This collapses all three onto `Subst::discharge_env_in_place` / `discharge_env`, and fixes two properties only one caller had.

## Two properties the copies got wrong

**Simultaneous, not a fold.** The env form is a constructor over `Subst`, not a loop of single-name discharges: a sequential fold re-substitutes into a replacement that mentions another key, sending `{a ↦ b, b ↦ 0}` to `0` instead of `b`. Caller ranges are key-free today, but that is a property of the callers, not of the operation.

**Root-carry.** Each replaced occurrence keeps its own `NodeId` — the replacement's root is built at the occurrence's id, inheriting the read site's span, and only the interior is freshened. A whole-subtree freshen discards both.

## The arrow species

`Compose` has its ends recomputed from the rewritten elements: substituting a `Var` whose type was an unresolved placeholder can concretize them, and the `Compose.ty == Fun(first_domain, last_codomain)` invariant must follow. `FunKind` and the Pi binder belong to the composition rather than to its elements, so they survive — a data arrow (`⤇`) is no longer downgraded to a compute one (`⇒`).

`ccl_utils::free_among` answers the free-variable question for a whole candidate set in one traversal; the per-name `is_free` made selecting a live subset of a wide environment cost `|candidates| × |expr|`.

## Rebased onto #98's freshening `Clone`

#98 now makes `Clone` mint a fresh `NodeId` for every node it copies, which changes how the root-carry above is *achieved* without changing what it guarantees.

Previously the replacement was cloned (sharing ids), then its interior was freshened explicitly via `freshen_interior_node_ids`, leaving the root free to take the occurrence's id. Now the clone freshens root and interior alike, and `re_root` overwrites the root with the occurrence's id immediately after. So `freshen_interior_node_ids` is gone from these sites — it is what `Clone` does — and the sequence is simply `rep.clone().re_root(occurrence_id)`.

The guarantee is unchanged: each replaced occurrence keeps its own `NodeId` and its read site's span, and N reads still give N distinct roots. The one cost is that the clone mints a root id which the re-root then discards — one stranded id per substituted occurrence, which folds as a death rather than a defect, and which the occurrence itself never leaves the live set for. That trade was taken deliberately over adding a `clone_at(node_id)` constructor, to keep #98's diff minimal.

---

## Rebased onto `main` (`7a9d1201`)

**The arrow-species fix landed upstream independently**, as #113 — the identical
`Type::fun` → `Type::fun_like` one-liner in `subst.rs`. So this commit no longer
carries that change. What it still carries on that point is the part upstream did
*not* land: the two tests that pin the property
(`a_compose_keeps_its_fun_kind_and_binder_across_a_discharge`,
`a_data_compose_is_not_downgraded_to_a_compute_arrow` — #113's whole `subst.rs`
diff is 5 insertions and no test), the failure mode spelled out at the call site,
and the `rewrite_expr` doc. "The arrow species" section above should be read as
what the commit now *asserts* rather than what it fixes.

The claim was re-verified against upstream's `ty.rs` rather than assumed, because
#113 is titled *"Relate the two function kinds by equality instead of ordering
`Data` below `Compute`"* and could have moved it: `Type::fun` still stamps
`None`/`Compute`, `Type::pi` still stamps `Some(n)`/`Compute`, `data_fun` still
stamps `Data`, and `fun_like` copies `name` and `kind` off its exemplar. What #113
changed is how the two kinds *relate in the solver* — they are now incomparable
and pinned rather than ordered, and `FunKind::Var` joined the enum — none of which
touches what these constructors stamp.

**The substitution collapse itself is unchanged**, and upstream's rewritten
`transact_phase` (#91/#92) brought new call sites for the engines this PR deletes;
they are converted to `Subst::discharge_env` the same way. `free_among`,
`discharge_env_in_place`'s one-traversal selection, and root-carry are as
described above.


src/ccl/subst.rs
```diff
@@ -630,6 +634,84 @@ impl Subst {
...
+    /// names — and the asserts are what makes that an enforced precondition
+    /// instead of a comment, which is what the hand-rolled predecessors had.
+    /// `env` names the callers' concrete container rather than an
+    /// `IntoIterator` of borrowed pairs, which looks like the more decoupled
+    /// choice and is not. Two passes are needed — select the live keys, then
+    /// build from the surviving entries — so the iterator form needs a `Clone`
+    /// bound, and the mutability phases thread their environment as
+    /// `&mut HashMap` and pass it directly, which satisfies neither that bound
+    /// nor the borrowed-pair item type. It would buy the removal of one import
+    /// at the cost of a reborrow at every call site.
```
dpmills: This comment is massive, and definitely not all important.

src/ccl/subst.rs
```diff
@@ -630,6 +634,84 @@ impl Subst {
...
+            .iter()
+            .filter(|(name, _)| live_names.contains(*name))
+            .map(|(name, term)| (name.clone(), Mapping::Discharge(Box::new(term.clone()))))
+            .collect();
+        Subst(live).rewrite_expr(e);
+    }
+
+    /// [`Self::discharge_env_in_place`] by value, for the callers that own (or
+    /// clone) the subtree and want the result back.
+    pub fn discharge_env(mut e: TypedExpr, env: &HashMap<Name, TypedExpr>) -> TypedExpr {
```
dpmills: This doesn't match the terminology of the existing subst code.  The distinction between discharge vs discharge_in_place is that discharge returns a Subst object that can perform the given discharge.  Looks like nothing calls the mut version of discharge_env_in_place anyways, so we can probably just call this one that and not have two

dpmills (approved):

src/ccl/subst.rs
groundlar: Cut from 58 lines to 26. What went, and where:

- **Root-carry** was a near-verbatim restatement of `provenance.md`, "Duplication discipline", which specifies it — now one sentence and a citation. The same duplication sat in `rewrite_expr_go`'s inline comment and at the `transact_phase` env-construction site; both are citations now too.
- **The selection's cost argument** moved onto `free_among`, which owns it, including the measured number so the claim is checkable where it is made.
- **The `&HashMap` vs `IntoIterator` signature rationale** is gone — it argued an objection nobody raised.

What stayed is the two properties a caller cannot read off the signature: simultaneous rather than a fold, and the post-inference α-uniqueness the two release-live asserts enforce.

src/ccl/subst.rs
groundlar: Collapsed to one function. `discharge_env_in_place` had no caller but its own wrapper, so it is gone and the by-value `discharge_env` absorbed the body — the shape all fourteen call sites already used.

The constructor form is not available as a third option: the live-key selection reads `e` (`free_among(env.keys(), &e)`), so a `Subst`-returning `discharge_env(env)` would have to either drop the selection and clone every environment value, or take `e` anyway. Flagging it because it means the env form cannot be made to parallel `discharge` exactly — the asymmetry is now down to one function instead of two.

src/ccl/transact_phase.rs
```diff
@@ -340,19 +341,30 @@ fn build_zip_read(
...
     } else {
         for r in used {
             let field = Expr::new(TypedExprNode::Proj(ProjKey::Field(r.field.clone())))
                 .with_ty(Type::fun(snap_ty.clone(), r.value_ty.clone()));
+            // Every field read embeds a bare `snap_expr.clone()`, so these
+            // environment values share ids with each other. An environment value
+            // is not a tree position — none of them enters the output, and each
+            // occurrence the discharge replaces gets its own root-carried copy —
+            // so the duplication discipline does not reach them
+            // (`src/ccl/design/provenance.md`, "Duplication discipline").
```
dpmills: There are some stale comments like this in this PR about cloning.  Make sure when you restack the bottom changes up you clean them up

src/ccl/subst.rs
```diff
@@ -705,10 +755,19 @@ impl Subst {
...
+                // so the predicate interiors the clone reaches stay shared with
+                // its source.
+                //
+                // The root is the one place this costs anything: the clone mints
+                // an id for it and the carry above immediately discards that id
+                // in favour of the occurrence's. So the occurrence stays in the
+                // live set (the carry precedes nothing that could displace it),
+                // and what is stranded is a single minted id that folds as a
+                // death rather than a defect — one row and one id per
+                // substituted occurrence, no extra hop.
```
dpmills: rm this block if not already done downstack

src/ccl/subst.rs
```diff
@@ -684,6 +688,52 @@ impl Subst {
...
+    /// entry.
+    ///
+    /// Two hard `assert!`s ride this path, release-live rather than debug
+    /// tripwires: [`assert_preserves_typedness`] fires if an environment value is
+    /// untyped against a typed occurrence, and [`Self::assert_no_capture`] fires
+    /// if the environment's range mentions a binder the walk passes under. The
+    /// mutability phases satisfy both — they run post-inference over α-unique
+    /// names — and the asserts make that an enforced precondition rather than a
+    /// comment.
+    pub fn discharge_env(mut e: TypedExpr, env: &HashMap<Name, TypedExpr>) -> TypedExpr {
```
dpmills: This should be discharge_env_in_place to match discharge_in_place.  The non _in_place version returns a Subst for delayed substitution instead of running immediately

dpmills (approved):

src/ccl/subst.rs
groundlar: Renamed, keeping the by-value signature: the axis you named is immediate vs. delayed, and this one runs immediately.

The `Subst`-returning slot stays unoccupiable for the env form — the live-key selection reads `e` (`free_among(env.keys(), &e)`), so a delayed constructor would have to either drop the selection and clone every environment value, or take `e` anyway. Fourteen call sites and three doc references renamed with it.

src/ccl/subst.rs
groundlar: Gone. The `Var` branch is root-carry then `return`; the block and its comment appear nowhere in the branch.

src/ccl/transact_phase.rs
groundlar: Deleted, and swept the whole PR for the same shape. Three went: this one, the "a bare clone would be neither" claim in the engine test's doc, and the "only the interior is freshened" restatement of `as_expr_preserving`.

The first of those was false rather than merely redundant — `TypedExpr::clone` mints at every node via `copy_id`, so a bare clone *is* unique; it only fails the identity half. That sentence was true under the `re_root`/`clone_at` concept this commit used to carry.

What stayed is the cost claims and the one-sentence identity contract with its citation to `provenance.md`.

PR #102, opened 2026-08-15
Close a commit store's keys individually, so an await waits only for its own variable's writers

dpmills:
`await_final(x)` waits for `x` to be finished with, but completion was a property of the whole commit store — one `terminal` flag, set once every writer had finished. Two variables share a store when some `with begin():` block mentions them together, so `await_final(a)` could not settle while a live writer kept committing `b`, even though nothing could write `a` again and its final value was fixed. That is a defect against the ordering model, which promises completion per variable. A key now closes once every writer whose *static* write footprint contains it has drained.

`Tile::Store` gains `closed_keys` alongside `terminal`, and neither derives from the other: the tile does not know its own key universe, so "every key listed is closed" is not "the store is closed". `CommitOperator` is now built with the per-writer write footprints that closure is computed from. The static footprint is the right input — a conditional write keeps its key in the block's footprint, so a block that *might* write the key keeps it open.

The terminal read consults that fact directly: `StoreFinalRead`'s gate becomes `closed_keys.contains(key) || terminal`. No projection of the history closes on `closed_keys`, and none needs to — every projection gains a position at a tick the store allocates, so `terminal` alone bounds their positions. Reasoning in [src/ccl/design/mutability.md, "`await_final`"](src/ccl/design/mutability.md#await_final).

`a_finite_mut_var_completes_despite_a_live_writer_it_shares_a_block_with` is the broken case, asserting the store count first so it cannot pass by the two variables being partitioned rather than by the key closing.


dpmills:

- #71
    - #72
        - #73
            - #52
                - #60
                    - #63
                        - #64
                            - #91
                                - #92
                                    - #93
                                        - #102 ◀
                                            - #106

tests/compilation_pipeline/transactions.rs
sortalongo, line 1: Please add a test for the still-too-coarse scenario, which is expected to yield a compilation error.

sortalongo (approved):

tests/compilation_pipeline/transactions.rs
dpmills: Added `a_block_writing_only_a_read_only_store_mate_of_an_await_rejected`, in the widest form of the limitation: no block writes both variables, and they are store-mates only because one block **reads** `a` while writing `b` (the partition is over `reads ∪ writes`). `a`'s only writer drains before the await, so per-key closure settles `await_final(a)` and the last block writes `b` alone — the per-variable rule permits the program and the per-store check refuses it.

Two things fell out of writing it, fixed in the same commit:

- The diagnostic claimed the block "would await a history it has not finished writing". The offending block does not write that history at all. It now names the recurrence, and points at what puts two variables in one store.
- `writer_body_depending_on_its_own_store_s_await_rejected` justified itself with "the store would have to finish before it could finish", which per-key closure makes false.

One consequence worth deciding separately: `check_await_final_linearity` already rejects reaching the stated rule through `x` itself, so **every** program `check_store_acyclicity` rejects is now one the stated rule permits — the check's whole domain is over-approximation, not just the margin the spec's note describes. Narrowing it means splitting a store at a key's closure point. Recorded in `mutability.md` for now.

PR #104, opened 2026-08-18
Add specification-prose rules to CLAUDE.md, with a `spec-prose` skill and a reminder hook

dpmills:
An agent writing a doc section imitates the prose in front of it, and this repo's older design docs are essays. This change states the specification-prose rules and puts a pointer to them at the moment of writing. No existing prose is converted here.

Read CLAUDE.md's new "Prose style: specifications, not blog posts" section first — it is the rule list. `.claude/skills/spec-prose/SKILL.md` is the keyboard-level companion: before/after pairs drawn from prose currently in `provenance.md` and `mutability.md`, the voice failures to avoid, the invented-vocabulary case (`wall` and `gate` — one metaphor over three mechanisms) with a report-it-in-session protocol, and a read-back list. `.claude/hooks/prose-check.py` prints the reminder after an `Edit`/`Write` that adds at least 25 words of prose to a `.md` or `.rs` file, once per file per session; it scores nothing and never blocks. `.claude/settings.json` registers it as a `PostToolUse` hook.

### DIFF
--- a/.claude/skills/spec-prose/SKILL.md
+++ b/.claude/skills/spec-prose/SKILL.md
@@ -1,0 +1,178 @@
+---
+name: spec-prose
+description: Write or revise prose in a design doc, a doc comment, or a code comment so it reads as specification rather than as a blog post. Use when adding or editing a `*/design*.md` section, a `///`/`//!` doc comment, a module-level comment, or any explanatory prose in this repo.
+---
+
+# Specification prose
+
+The reader arrives to look one thing up and already knows they need it. Lead with the claim and
+stop.
+
+The rules are in CLAUDE.md, "Prose style: specifications, not blog posts". This skill is what to do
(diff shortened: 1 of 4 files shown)


dpmills:

- #104 ◀

.claude/skills/spec-prose/SKILL.md
sortalongo, line 1: let's hard-wrap this, in line with our new convention for md files.

sortalongo (approved):
I hope this works! Our docs are becoming quite hard to read, despite their comprehensiveness.

.claude/skills/spec-prose/SKILL.md
dpmills: Done, and added a note about hard-wrapping docs; that wasn't actually written down anywhere yet

PR #105, opened 2026-08-18
Render a singleton type as `Int@7` rather than as the bare literal

dpmills:
A singleton refinement printed as the literal it pins, so `{Int | __elem == 7}` rendered `7` — a type spelled as a term, with its base nowhere in the string. It now renders `Int@7`: the base, then the value pinned to it. Display only; parsing, CHL surface syntax, and the `Type::Refinement` shape are unchanged.


sortalongo (approved):
I'm pretty sure there are more uses of numeric literal types in design docs than this. Do a pass to find more. Seems like the kind of thing Haiku might be good for?

PR #106, opened 2026-08-18
Retire the `reg` and `Recurse` names, and pin the out-of-block read rule

dpmills:
The mutability subsystem renamed its concept to "mutable variable", but the identifiers carrying it still said register — the fresh name was `__reg`, a key's history binding was `reg_k`, locals were `reg_read` / `reg_vtys` — and `__reg` reaches symbolic output and test assertions, so the old word outlived the rename exactly where a reader meets it. This retires it (the record is `__hist`, its fields `hist_k`, and the prose calls it a **history record**) along with a second name in the same state: `Recurse`, the removed recurrence operator behind an induction accumulator, still cited in twenty-five comments across thirteen files as though it were something to go look at. Each of those names what it now means, since the old one stood for the induction-store engine in some places and a loop's own cycle in others.

## The out-of-block read is a rule, not a caution

`mutability.md` presented the non-atomic pre-check `if balance > 0: with begin(): balance -= req` as compilable but unwise, and asked for a lint. The guard reads a `Txn` variable outside a block, which lowering rejects and always will, so the shape does not compile and there is nothing to lint; the spec's reads bullet is marked `[Decided]` to match. The rule had no test where it is easiest to lose — following a `Mut(_, Txn)` parameter into a callee — and now has one.

## Prose

`mutability.md` also drops the voice its own rules call out: stress-italicised copulas and negations throughout, `deliberately` seven times, `Footgun` as a heading, and `strict wall` for the strict `typecheck`, a metaphor already colliding with two other mechanisms.


dpmills:

- #93
    - #102
        - #106 ◀
            - #97

docs/chl-spec.md
```diff
@@ -2572,10 +2572,12 @@ around `reserve` + `quote` + the feed).
 - **Trailing induction read** — after a `for` loop, a bare reference to an
   induction accumulator is its final value (or the pre-loop value if the
   source was empty). The loop has ended, so "latest" is unambiguous.
-- **A `Txn` mutable variable is read only inside a `with begin():` block.** A bare
-  read outside one is an error. Reading inside a block pins a
-  **snapshot-consistent** view: several mutable variable reads in one block see
-  one commit snapshot — the reason the block is required.
+- **A `Txn` mutable variable is read only inside a `with begin():` block
+  [Decided].** A bare read outside one is an error, and stays one: the block is
```
sortalongo: Is this only "decided"? I thought it was implemented, and should thus be unqualified.

sortalongo (approved):

docs/chl-spec.md
dpmills: Right — dropped.

The legend at the top of the spec makes it unambiguous: `[Decided]` is "a design decision that has been made and recorded ... but **not implemented**", while unmarked body text is what the compiler implements today. Lowering rejects the out-of-block read now, pinned by `bare_txn_read_outside_tx_rejected` and `bare_read_of_a_mut_param_outside_a_block_rejected`. The "and stays one" phrasing already carries the permanence the marker was doing duty for.

The same reading applies to two neighbours, so those went too: the terminal-read bullet's `await_final(x) [Decided]`, and the `### 8.6 await_final [Decided]` heading — both describe implemented, tested behaviour. The three inbound links to that anchor moved with it.

PR #107, opened 2026-08-18
Replace the lineage log with a `NodeId`-keyed provenance table and gate the pane folds

groundlar:
On `main` the provenance substrate has no users: `src/ccl/lineage.rs` holds an append-only `RewriteStep` log, no phase records into it, and the pane folds plus their leak gate are marked *Planned* in the design doc. This change makes the recording real — a `ProvenanceTable` keyed by the node each rewrite produced, a phase scope around every phase between the three retained panes, and recording sites inside them — and turns the two pane-pair folds and their leak gate on. A node in the `post-channelize` pane now resolves through recorded rows to the source span the user wrote, and a rewrite that mints with nothing recording fails a test instead of silently degrading a diagnostic.

## The table replaces the log

A row is keyed by the node a rewrite **produced** and carries three columns: `parents` (the ids consumed), `blame` (upstream ids the node is about but did not consume), and an interned `RewriteTag {via, nature, label}`. No span column and no fate column — spans come from walking `parents` back to the lowering projection, and deaths are `input_ids ∖ output_ids` between two panes, returned by the fold as a product rather than a leak class. `collapse` becomes `fold`, returning a `ProvenanceMap` whose edges carry an `EdgeLabels` set closed weakest-link (one blame hop makes the whole path blame), the output pane's `SourceProjection`, the deaths, and the leaks. The leak taxonomy collapses to two classes: `Unrecorded` from walking the output tree, `DanglingParent` from walking the table. The fold is one ascending-`NodeId` sweep with no fixed point, resting on invariants asserted at `ProvenanceTable::record` and at the sweep — see [The provenance model](src/ccl/design/provenance.md#the-provenance-model) and [The fold is order-free](src/ccl/design/provenance.md#the-fold-is-order-free).

Reading the diff: `src/ccl/lineage.rs` is deleted and its contents live in `src/ccl/provenance.rs`, so roughly 2,200 added lines are a move. `Pass` becomes `Phase` and moves to `src/ccl/context.rs`, where declaration order is pipeline order because `ProvenanceTable::deaths` takes a phase range.

## What the phases record

`TableSession` installs the table for a compile, `PhaseScope::enter(phase)` supplies the ambient `via`, and a site calls `provenance::enter(id, label, nature)` and declares nothing else: every id minted while that guard is innermost rows against the named node, with `also_consumes` and `blame` as inherent methods on the guard. Sites landed in `infer/solve` (`mono.specialize`, `mono.coalesce_let`), `infer/emit`, `inline`, `mut_elim`, `transact_phase` (fourteen, one per rewrite), `channelize`, and two shared helpers that record under whichever scope is open — `subst` and `PredMemo::rebuild`. `simplify` needed no rule-body edits at all: one `ruled` combinator wraps all thirteen rule invocations, which works because a recording declares nothing, so wrapping every *attempt* rather than every firing costs a push and pop when a rule declines.

The semantic change to read closely is `PredMemo`. A rebuild now defaults to **deriving** — the rebuilt predicate freshens and rows against the term it was built from — where before every rebuild preserved ids silently. `PredMemo::replacing()` is the opt-in that keeps the original's ids, and `uniquify` is its only caller: it reaches every occurrence, and now asserts that 1:1 correspondence (N distinct terms in, N out, same ids) on every debug compile.

## Uniqueness now covers the predicate domain

`assert_unique_node_ids` walked children only. It now runs two walks over the same domain `collect_tree_ids` enumerates: `duplicate_node_ids` stays `Rc`-blind over the main tree, where a second sighting of an id is a collision outright, and `predicate_id_collisions` dedups by `Rc` pointer first, then reports an id carried by two distinct predicate terms or by both a predicate and the main tree. The `Rc` dedup is the whole point — one term riding many type slots is load-bearing sharing, and a raw id-set walk would report it as a collision. Both stay behind `cfg!(any(debug_assertions, test))`, and `post-inference` joins the boundary list, since inference mints singleton predicates and clones a definition per instantiation before the pane is taken off it. Rationale: [Walking the ids](src/ccl/design/provenance.md#walking-the-ids).

## The gate

`gate_leaks` asserts a whole `Leak` vector empty; it runs at the lowering handoff as before, and now at the pane pairs. `pane_pair_folds_have_no_structural_leaks` folds **both** pairs for every program in `corpus()` and asserts both classes zero. `CAMBRA_PROVENANCE_GATE=1`, which CI sets on the test step at about +4%, gates only `gated_pane_pairs()` — the second pair — but over every program the caller compiles. The first pair is excluded because it spans inference, whose predicate producers do not record, so its residue there is a constant `DanglingParent` count rather than a regression; `gated_pane_pairs` and the doc both say to add it in the commit that makes inference record. The wider corpus is what found two gaps the eleven corpus programs miss: `transact_phase` calling `mut_elim::fold_induction_loop` unrecorded, and `flatten_spine`'s value-position writer hoist. `ProvenanceAudit` spans (`CAMBRA_PROVENANCE_AUDIT=full|letrec|mutelim|planning`) measure instead of gating and reach the sites recording below the last pane. Endpoints and coverage: [Gate and audit coverage](src/ccl/design/provenance.md#gate-and-audit-coverage).

## Tests and docs

Nine tests in `src/ccl/context.rs` cover the recording, eight of them over a shared `corpus()`. Three carry pinned lists rather than counts, so that a corpus edit cannot leave a green tautology behind: the pane pairs where the fold reads a row at all, the ones where blame reaches the map, and the programs that specialize. `provenance_pane_perf` is `#[ignore]`d, a measurement rather than an assertion. `src/ccl/design/provenance.md` is rewritten around the shipped shape and records what still blocks panes below `post-channelize` in [Known prerequisites for panes past `post-channelize`](src/ccl/design/provenance.md#known-prerequisites-for-panes-past-post-channelize). The `desugar` → `channelize` rename finishes across the surviving prose and API (`check_pre_desugar` → `check_pre_channelize`, `post_desugar_ir` → `post_channelize_ir`).


src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+[Known prerequisites for panes past `post-desugar`](#known-prerequisites-for-panes-past-post-desugar).
+
+## Terms
+
+Five words do most of the work here, and two pairs of them are easy to run
+together. Pinned once, used consistently below.
+
+| term | what it is |
+|---|---|
+| **pane** | One of the three retained AST snapshots on `CompiledProgram`: `pre_inference_ir`, `post_inference_ir`, `post_desugar_ir`. A pane is a *thing the inspector displays*. There are three, and adding one costs a retained full-tree clone. |
```
dpmills: The definition doesn't need to specify where we happen to do the snapshots today; we don't want that to be a contract

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+| **pane boundary** | The fold between two **adjacent panes** — `pre → post-inference` and `post-inference → post-desugar`. Produced by `materialize_panes`, which restricts the table by the boundary's passes. This is the **durable, gated** artifact: the leak classes are asserted here. |
```
dpmills: This term is confusing.  It sounds more like a diff than a boundary?

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+the recorder, the passes' adoption of it, the always-on lowering projection, the
+`NodeId`-keyed table that is the sink
+([The node table](#the-node-table-srcccllineagers)), and the pane-boundary fold
+over it — is in tree. Everything a **Planned** marker introduces is designed but
+not yet built: the inspector's consumption of the panes. A reader can tell the
+two apart by the marker alone; unmarked prose describes code you can go read. Adoption is complete through `post-desugar`: every pass inside
+the two pane boundaries brackets, and the gate holds at zero on both. It now
+extends further — `lambda_elim` and `planning` bracket too, though no pane
+boundary folds them yet. Operator conversion is the one pass below the span still
+unadopted; see
```
dpmills: This section seems contradictory; should it just say complete through op-conversion?

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+## Terms
+
+Five words do most of the work here, and two pairs of them are easy to run
+together. Pinned once, used consistently below.
+
+| term | what it is |
+|---|---|
+| **pane** | One of the three retained AST snapshots on `CompiledProgram`: `pre_inference_ir`, `post_inference_ir`, `post_desugar_ir`. A pane is a *thing the inspector displays*. There are three, and adding one costs a retained full-tree clone. |
+| **pane boundary** | The fold between two **adjacent panes** — `pre → post-inference` and `post-inference → post-desugar`. Produced by `materialize_panes`, which restricts the table by the boundary's passes. This is the **durable, gated** artifact: the leak classes are asserted here. |
+| **window** | A set of passes a fold restricts the table by, and by extension the `LineageAudit` measurement span that opens one at an arbitrary point and folds at another, reading the *live tree* at each end rather than a retained snapshot. An audit is env-selected, and a **measurement, never a gate**. |
```
dpmills: Having not read the rest of the doc yet, this is incomprehensible

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+| **slot** | The node a bracket names — `lineage::enter(slot_id, …)` — read off the tree *before* the rewrite runs. Always a main-tree node, because that is the only kind of node the id domain contains. |
```
dpmills: What's a bracket?

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+| **predicate interior** | A `NodeId` carried by a `TypedExpr` inside a `Type::Refinement`'s predicate. Real ids from the same counter, but **outside the id domain** — carried, never checked. See "The id domain". |
```
dpmills: Why does this need a dedicated term?

src/ccl/design/provenance.md
```diff
@@ -133,58 +159,259 @@ is that nothing aliased arrives, and each site pins it with a test.
...
+Four properties, each load-bearing:
+
+- **No `span` column.** Spans are derived by walking `parents` back to a node
+  the lowering projection covers, which is a handful of hops and does not
+  lengthen as programs grow. Storing spans per row would denormalize that onto
+  every row ever minted.
+- **"No row" is a legitimate state.** The key space is `NodeId`, and a `NodeId`
+  can be addressed without ever having been recorded — every
+  refinement-predicate interior is (see [The id domain](#the-id-domain)). Reads
+  answer empty / `None`; nothing panics.
```
dpmills: It seems like not handling refinements is causing distortions in the model; should we try to sort that out now?

src/ccl/design/provenance.md
```diff
@@ -133,58 +159,259 @@ is that nothing aliased arrives, and each site pins it with a test.
...
-`specialize_use` clone `Copy`s and the `coalesce_generalized_let` wrapper
-`Transform`), `inline` (beta/alias discard `Transform`s + fan-out `Copy`s), and
-`channelize` (cluster/feed-union/lift/drop). Each boundary in `compile_program`
-wraps the pass in a `RecorderSession` and retains its drained `LineageLog` on
-`CompiledProgram::pass_lineage`, in pipeline order: `[(Mono, …), (Inline, …),
-(Transact, …), (Letrec, …), (Desugar, …)]`.
-
-Today the only production step frames are lowering's two `Copy`-capture frames
-(below); every other rewrite runs unrecorded.
+An ambient thread-local frame stack (`STEP_STACK`) + an installed sink,
```
dpmills: What's the difference between a frame and a bracket?

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
+over it — is in tree. Everything a **Planned** marker introduces is designed but
+not yet built: the inspector's consumption of the panes. A reader can tell the
+two apart by the marker alone; unmarked prose describes code you can go read. Adoption is complete through `post-desugar`: every pass inside
+the two pane boundaries brackets, and the gate holds at zero on both. It now
+extends further — `lambda_elim` and `planning` bracket too, though no pane
+boundary folds them yet. Operator conversion is the one pass below the span still
+unadopted; see
+[Known prerequisites for panes past `post-desugar`](#known-prerequisites-for-panes-past-post-desugar).
+
+## Terms
```
dpmills, line 62: In addition to the early terms section, it would be helpful to have a very high level summary of the whole architecture near the beginning of the doc

src/ccl/channelize.rs
```diff
@@ -1182,6 +1183,19 @@ fn desugar_inner(expr: Expr, ctx: &mut DesugarCtx) -> Result<Expr, DeferError> {
...
+            // `let d = Defer` node it replaces. The inner defers of a cluster are
+            // consumed by the same rewrite but are not named: they are absent
+            // from the output tree, so the boundary difference reports them, and
+            // naming them here would be a fate assertion this pass is in no
+            // position to make (a defer whose handle survives in a type is not
+            // dead).
+            //
+            // Opened *after* the body recursion so a nested cluster's products
+            // attach to their own `let`, not to this one.
+            let _g = lineage::enter(node_id, "channelize.cluster", lineage::Nature::Expansion);
```
dpmills: Here and elsewhere, do these lineage::enter calls really need comments at all?  If they need comments this big, that smells like the API is wrong.  If it's just claude logorrhea, delete it

src/ccl/design/provenance.md
```diff
@@ -7,11 +7,37 @@ channelize rewriting defers, lambda-elim synthesizing combinators, planning
...
-of the recorder, the pane-boundary folds, and the inspector's consumption of
-them. A reader on `main` can tell the two apart by the marker alone; unmarked
-prose describes code you can go read.
+the recorder, the passes' adoption of it, the always-on lowering projection, the
+`NodeId`-keyed table that is the sink
+([The node table](#the-node-table-srcccllineagers)), and the pane-boundary fold
+over it — is in tree. Everything a **Planned** marker introduces is designed but
+not yet built: the inspector's consumption of the panes. A reader can tell the
+two apart by the marker alone; unmarked prose describes code you can go read. Adoption is complete through `post-desugar`: every pass inside
+the two pane boundaries brackets, and the gate holds at zero on both. It now
```
dpmills: "post-desugar" shouldn't be a term.  There's nothing named "desugar" anymore

src/ccl/lineage.rs
```diff
@@ -172,86 +144,209 @@ impl Nature {
...
+///
+/// The two labels are the two relations a row records, closed transitively:
+///
+/// * **descent** — the downstream node *descends from* the upstream one, i.e.
+///   every hop between them consumed the node before it to produce the next.
+///   A row's `parents` are its descent hops.
+/// * **relatedness** — the downstream node is *related to, but not consumed
+///   from*, the upstream one. A row's `blame` are its relatedness hops, and
+///   they may name a node that is still alive elsewhere in the output tree,
+///   which is exactly why relatedness is not an ancestry claim.
```
dpmills: Why do we need parallel terms here? We already have "parent" and "blame"

src/ccl/design/provenance.md
```diff
@@ -133,58 +159,259 @@ is that nothing aliased arrives, and each site pins it with a test.
...
-structs, not one generic. Its log is folded **once** at the lowering boundary by
+a pass runs, every node it *produces* gets a row in the `LineageTable`:
+
+| column | what it holds |
+|---|---|
+| `parents` | the ids the rewrite consumed to produce this node — a bracket's slot, or a fusion's whole consumed set |
+| `blame` | the upstream ids the node is *related to but did not consume* — **not** the same as `parents` |
+| `rule` | an interned `RewriteTag` — the `{via, nature, label}` triple |
+
+A row is **silent on its parents' own fate**. That silence is what makes
```
dpmills: I thought this was the distinction between blame and parent? Why do we need this for parent?

src/ccl/design/provenance.md
groundlar: Done in `fb2eb53d`. The row is now just "A retained AST snapshot the inspector displays, materialized after a set of passes. Each one costs a retained full-tree clone." The three field names moved up into the status paragraph, where they read as current state rather than as part of the definition.

src/ccl/design/provenance.md
groundlar: Dropped the term rather than reworded it. You're right that it named the fold and not a boundary — and the tree already spelled that thing three ways: `pane relation` (`context.rs`), `pane-to-pane relation` (`lineage.rs`), `inter-pane relation` (`provenance.md`). Now one name, **pane relation**: what folding the passes between two adjacent panes produces.

`boundaries()` → `pane_relations()`, `gated_boundaries()` → `gated_pane_relations()`, `gate_leaks`' parameter → `relation`. The drop had to reach bare "boundary" too, since the lineage subsystem spelled the same concept that way in ~50 more places; converting only the qualified spelling would have kept the term alive under a shorter name. Bare "boundary" keeps its unrelated senses — the lowering handoff, an inference level, and the pass boundary `assert_unique_node_ids` asserts at (`context.rs:897`). One word was covering four things.

src/ccl/design/provenance.md
groundlar: It was contradictory: it claimed `lambda_elim` and `planning` bracket their rewrites while also claiming adoption complete through `post-desugar`. `lambda_elim` records nothing at all, and that false claim was one of the rotted ones `bf41c2d3` mined out.

Not "complete through op-conversion" either — op-conversion is the one pass that structurally cannot record, having no identity to record against (`TileOperator` carries none, there is no `OperatorId`). So the section now names the state per pass instead of summarizing it: adopted across the five passes the two pane relations span; three sites record but reach no table because `compile_program` opens no `PassScope` around them (`simplify`, `planning/iterate`, `transact_phase`'s as-of-read rewrite); `lambda_elim` records nothing; op-conversion has no identity.

src/ccl/design/provenance.md
groundlar: Agreed, and the term is gone rather than reworded. A "window" was the set of passes a fold reads — `MONO_PASSES` is a single pass, so it was never a window with endpoints — and the word was already taken twice in the tree by `ProposalWindow` and the interpreter's streaming windows, the same collision that retired "frame".

`collapse` now takes `passes`; `LineageAudit` has a `span` with two chosen endpoints. `MONO_WINDOW`/`DESUGAR_WINDOW` → `MONO_PASSES`/`CHANNELIZE_PASSES`, `WINDOW_VIA` → `AUDIT_VIA`, `lineage_audit_window()` → `lineage_audit_span()`. No flag changes — `CAMBRA_LINEAGE_AUDIT` already named spans.

The three paragraphs that used to sit between the glossary and the gate collapsed to one, because most of them were disambiguating the term. And there's now a `## Mechanism at a glance` section *ahead* of the terms, so no glossary row has to explain the design on its own.

src/ccl/design/provenance.md
groundlar: Nothing, now. That was the finding — two bespoke words for one idea, and "frame" was already spoken for four other ways in the tree (stack frames, solver specialization frames, scope frames, type-morphism frames). `bf41c2d3` deletes both: a rewrite **records**, what it opens is a **recording**, and the RAII value is a **guard**. `grep -c bracket src/ccl/design/provenance.md` is 0; `frame` survives only as the two item spellings left unrenamed to keep that diff reviewable (`FrameGuard`, `copy_frame`).

src/ccl/design/provenance.md
groundlar: It earns the row differently than it did when you asked, which is worth saying plainly. At the revision you reviewed it named ids that were *carried but outside* the id domain — a term for an exclusion, which is a fair thing to object to. Since `14878ab8` they are in the domain, and the term now marks the one place two questions come apart: the fold must **explain** a predicate interior (`collect_tree_ids` enumerates them), but `assert_unique_node_ids` deliberately walks the main tree only, because a predicate interior may legitimately carry a main-tree id at inline's blind spot.

The row is rewritten to state exactly that and nothing else. If you'd still rather not have a glossary entry for it, "Walking the ids" is the only section where the divergence matters and could carry it alone — say so and I'll drop the row.

src/ccl/design/provenance.md
groundlar: Right, and the commit directly above the one you were reading does it: `14878ab8` puts predicate ids **in** the id domain (`collect_tree_ids` reaches them through type slots, `user_annotation` and `Cast` targets) and makes `PredMemo` record rebuilds.

What you caught is that the doc still carried the pre-`14878ab8` model in four places, contradicting its own glossary in the same file. That is the distortion, and all four are fixed in `fb2eb53d`:

- "a `NodeId` can be addressed without ever having been recorded — every refinement-predicate interior is" — false; lowering's projection covers every id `collect_tree_ids` enumerates.
- "**Planned** is admitting predicate interiors as rows" — done, not planned.
- "that set is `collect_tree_ids`, the `walk_children` domain, which excludes predicate interiors" — `collect_tree_ids` is the predicate-*inclusive* one; `collect_main_tree_ids` is the narrow walk. The produced-side check it was arguing against is still not wanted, but for the two legitimate shapes named in the next paragraph, not for this reason.
- "a predicate interior being nothing a pane enumerates".

What legitimately remains unrecorded is narrower than the doc implied and already named: inference's predicate producers (`specialize_use` cloning a definition per instantiation), and planning raising a predicate back into the main tree. The first is why the gate covers `gated_pane_relations()` rather than both.

src/ccl/design/provenance.md
groundlar: No difference — that was the finding. One idea with two bespoke words, and "frame" was already spoken for four other ways in the tree. `bf41c2d3` collapses both onto **recording** (the scope) and **guard** (the RAII value); `bracket` is gone from the doc entirely.

src/ccl/design/provenance.md
groundlar: Added in `fb2eb53d` as `## Mechanism at a glance`, between the status markers and the terms. Five paragraphs: identity and what preserves it; what a recording claims and the two things it deliberately does not declare; the table and what folding a set of passes turns it into; the two consumers (a pane relation, gated on every compile, and a `LineageAudit`, which measures); and where source spans enter, which is lowering and nowhere else.

The glossary shrank as a result — it can be five definitions now instead of definitions doing double duty as an explanation.

src/ccl/channelize.rs
groundlar: Both, and the API was the bigger half. I measured before cutting: 36 `enter` sites, 21 with comments, 8 of those at 11–18 lines.

The bulk was the API contract restated per site. Four claims recurred almost verbatim — nothing declares a death, blame is not consumption, the hooks need an open recording to write to, don't wrap a recursion in one. Those are properties of `lineage::enter`, and its own doc already stated the first. So they're stated once there (plus one addition on `FrameGuard::blame`) and cut from the sites, which now keep only the judgment that is theirs: which node is the slot, and why.

Sites at ≥11 lines: 8 → 4. Largest: 18 → 12. One long comment stayed on purpose, `mut_elim.rs:442` — its middle paragraph justifies a drop-path test that was *deleted*, and no API doc can carry a rationale for absent code.

src/ccl/design/provenance.md
groundlar: Renamed in `60d68549` — code first, so the doc has something greppable to track. It was 243 occurrences across ~20 identifiers: `Pass::Desugar` → `Pass::Channelize`, `post_desugar_ir` → `post_channelize_ir`, the `"post-desugar"` label → `"post-channelize"`, and inside the module the recursive worker `desugar` → `channelize_expr` (the entry point was already `run`), with `DesugarCtx`, `check_pre_desugar`, `PreDesugar`, `CompileError::DesugarDefers`, `has_pre_desugar_artifacts`, `desugar_leaks` and `desugar_map` following.

One scoping call: the word survives where it is not this pass. Lowering desugars `x op= e` and a chained comparison, lambda elimination desugars `BinOp` to applied-combinator form, and `optimization.md` has a desugaring *identity*. Renaming those would claim channelize does them — and that the same word covered both jobs is what made the pass name ambiguous in the first place.

src/ccl/lineage.rs
groundlar: Collapsed onto one vocabulary in `fb2eb53d`, with one deviation from "parent and blame" that I want to flag rather than bury.

`relatedness` is gone outright — it was a synonym for the column it closes, so `EdgeLabels::RELATED` is `BLAME` and `relates()` is `has_blame()`.

`descent` became `ANCESTRY`, not `PARENT`, for the direction reason you raised. The closure is reflexive: a surviving node carries a self-edge. Label that `PARENT` and the doc says a node is its own parent twelve lines after saying "a node cannot be its own parent — an in-place rewrite that keeps its id is a *preserve*". `ancestry` is the standard reflexive-transitive closure of parenthood, so it stays in the `parents` family without competing with the column, and `context.rs:1945` already said "reads ancestry off the label it finds there". Both coined words are gone either way.

Accessors are `has_ancestry()` / `has_blame()` rather than `is_`, since `EdgeLabels` is documented as a set and one pair can carry both. One `sed` if you want the literal `PARENT` spelling instead.

src/ccl/design/provenance.md
groundlar: Because the fate axis is orthogonal to *both* columns, and the doc buried that by hanging it off `parents` as though it were a parents-specific caveat.

Three independent things: `parents` is a consumption claim about the product; `blame` is an attribution claim about the product; and neither says anything about whether a named id survived — deaths come only from `input_ids ∖ output_ids`. It read as a parents caveat because `parents` is the column whose name invites the death reading, and the identical point was then repeated under `blame` in different words ("`blame` says nothing about any node's fate").

`fb2eb53d` states it once as **No column records a fate**, ahead of the column table, and drops both per-column disclaimers. The adopt-a-live-subtree paragraph follows from it instead of introducing it.

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+enumerates onto the span of the CHL it came from, and every later span is derived
+by walking `parents` back into that projection.
+
+## Terms
+
+| term | what it is |
+|---|---|
+| **pane** | A retained AST snapshot the inspector displays, materialized after a set of passes. Each one costs a retained full-tree clone. |
+| **pane relation** | What folding the passes between two adjacent panes produces: an id-to-id relation with labelled edges. The **durable, gated** artifact — the leak classes are asserted here. |
+| **recording** | The scope `lineage::enter` opens over one rewrite, held as a `FrameGuard`. Every node minted while it is the innermost open one takes the node it names as a parent. Prose here says "a recording" for the scope, "the recording site" for the code location, and "records against X"; the guard is the RAII value that closes it. |
```
dpmills: Sounds like FrameGuard should be RecordingGuard

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+| **slot** | The node a recording names — `lineage::enter(slot_id, …)` — read off the tree *before* the rewrite runs. Normally a main-tree node. A predicate interior *may* be one, but work **on** a predicate is usually recorded against the predicate's own root, and work that *produces* one against the main-tree node whose type will carry it. |
```
dpmills: How is "slot" different from "parent"?

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+| **predicate interior** | A `NodeId` on a `TypedExpr` inside a `Type::Refinement`'s predicate. Ordinary ids from the same counter, and inside the id domain a fold must explain: `collect_tree_ids` enumerates them. They are the one place explanation and uniqueness come apart — `assert_unique_node_ids` walks the main tree only, because a predicate interior may legitimately carry a main-tree id. See "Walking the ids". |
```
dpmills: Defining "predicate interior" to mean a type of node ID is a confusing definition.  Does this really need a term? Just writing "NodeIds inside refinements" when those need to be referenced feels just as good and doesn't involve new terminology

src/ccl/design/provenance.md
```diff
@@ -25,199 +100,432 @@ prose describes code you can go read.
...
-which is where ids could be dropped or re-minted.
-
-`Rc` sharing is load-bearing and constrains how recording may be done. One predicate term rides many
-type slots as a shared `Rc`, and planning's compile memo is `Rc`-keyed, so splitting the sharing
-compiles one predicate once per occurrence. Recording is therefore idempotent per id rather than per
-slot the id is reached through: `lowering_predicate_leaf` skips an id already recorded, which also
-stops a sweep replacing precise attribution with a coarse label. For the same reason a duplication
-path may share a predicate `Rc` with its source rather than rebuild one (see
-`design/type-inference.md`, "Sharing is an invariant, not an optimization detail").
+**Two questions, two domains.** Keep them apart, because they used to have the
```
dpmills: Looks like this is reverting a bunch of edits you made downstack

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+| **recording** | The scope `lineage::enter` opens over one rewrite, held as a `FrameGuard`. Every node minted while it is the innermost open one takes the node it names as a parent. Prose here says "a recording" for the scope, "the recording site" for the code location, and "records against X"; the guard is the RAII value that closes it. |
+| **slot** | The node a recording names — `lineage::enter(slot_id, …)` — read off the tree *before* the rewrite runs. Normally a main-tree node. A predicate interior *may* be one, but work **on** a predicate is usually recorded against the predicate's own root, and work that *produces* one against the main-tree node whose type will carry it. |
+| **predicate interior** | A `NodeId` on a `TypedExpr` inside a `Type::Refinement`'s predicate. Ordinary ids from the same counter, and inside the id domain a fold must explain: `collect_tree_ids` enumerates them. They are the one place explanation and uniqueness come apart — `assert_unique_node_ids` walks the main tree only, because a predicate interior may legitimately carry a main-tree id. See "Walking the ids". |
+
+An audit's endpoint is **chosen**, because a span running past the last
```
dpmills: All four of these paragraphs are misplaced.  They aren't defining terms, and this is the terms section.  Looks like this is stuff about what automated verification we have; it probably belongs in a section at the end, or maybe alongside whichever section introduces the invariant that is being checked.

Also, it looks like the content might be stale.  It references exceptions for refinements

src/ccl/design/provenance.md
```diff
@@ -25,199 +100,432 @@ prose describes code you can go read.
...
+identically on every leak class. Sites are placed correctly because the
+attribution *is* the product; the number will not tell you when they are wrong.
+
+`compile_program` opens one `PassScope` per pass — `Mono`, then `Inline`,
+`Transact`, `Letrec`, `Channelize` — inside the single `TableSession` that spans the
+whole compile, and retains the drained table as
+`CompiledProgram::lineage_table`. A pass that rewrites nothing on a given program
+writes no rows, which is the preserve case and not a gap.
+
+### The node table (`src/ccl/lineage.rs`)
```
dpmills: LineageTable is already defined and discussed in "## The lineage model".  This section is mostly duplicate info; fold any important pieces into the former section

src/ccl/subst.rs
```diff
@@ -82,12 +82,16 @@ pub enum Mapping {
...
+    /// **Considered and deferred: `Rc<TypedExpr>`.** The payload is a
+    /// *template* — never a tree node, cloned afresh at every read
+    /// ([`Mapping::as_expr`]) — so sharing it is sound, and the solver copies
+    /// substitutions constantly (`Bound::render_subst`, [`Subst::then`],
+    /// `compact`, `constrain`). With a `Box`, every one of those ~28 sites deep-
+    /// copies the payload tree; with an `Rc` they are refcount bumps. That cost
+    /// is **pre-existing** — the derived `Clone` paid it too — so it is an
+    /// improvement over both arms rather than anything the freshening `Clone`
+    /// introduced, and it wants its own before/after rather than riding along
+    /// here.
```
dpmills: This comment change feel unnecessary

src/ccl/subst.rs
```diff
@@ -772,6 +785,21 @@ impl Subst {
...
+                // clone shares with its source stay shared.
+                //
+                // The root is the one place this costs anything: the clone mints
+                // an id for it and the carry above immediately discards that id
+                // in favour of the occurrence's. So the occurrence stays in the
+                // live set (the carry precedes nothing that could displace it),
+                // and what is stranded is a single minted id that folds as a
+                // death rather than a defect — one row and one id per
+                // substituted occurrence, no extra hop.
+            }
```
dpmills: This is also reverting changes made downstack.  This is the second one I've caught, so I think we need a full sweep over the PR to look for things that were broken during rebasing

src/ccl/provenance.rs
```diff
@@ -118,37 +119,41 @@ impl serde::Serialize for NodeId {
 pub enum Pass {
```
dpmills: This enum should probably be in the top level context.rs file; it doesn't feel provenance-specific

dpmills: We already have Phase as the name for the pieces of compilation, don't need Pass too.  That rename can happen later if you want (and we could settle on either pass or phase; I don't really care)

src/ccl/provenance.rs
```diff
@@ -118,37 +119,41 @@ impl serde::Serialize for NodeId {
...
     /// ([`crate::ccl::mut_elim::run`]): folding direct-mirror `For`/`MutWrite`
     /// loops into guarded `LetRec` induction histories. Runs between the
-    /// post-inference and post-desugar snapshots (after `Transact`, before
-    /// `Desugar`).
+    /// post-inference and post-channelize snapshots (after `Transact`, before
+    /// `Channelize`).
     Letrec,
     /// Monomorphization: cloning a generalized definition's subtree once per
     /// distinct resolved type (during inference).
     Mono,
```
dpmills: Why is 'Mono' a phase? It doesn't even run at a specific time, we do multiple separate rounds inside inference.  Also, Infer isn't in this list; probably just need to replace Mono with Infer (and move it up to the right spot)

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
-
 **Status markers.** The substrate — the identity primitives, the lineage model,
-the recorder, and the always-on lowering projection — is in tree. Everything a
-**Planned** marker introduces is designed but not yet built: the passes' adoption
-of the recorder, the pane-boundary folds, and the inspector's consumption of
-them. A reader on `main` can tell the two apart by the marker alone; unmarked
-prose describes code you can go read.
+the recorder, the passes' adoption of it, the always-on lowering projection, the
+`NodeId`-keyed table that is the sink
+([The node table](#the-node-table-srcccllineagers)), and the fold over it — is in
```
dpmills: "fold" is used all over the place and never defined.  Later on, there is also discussion of "collapse".  Probably clearer just to define LineageTable and LineageMap early and use those terms.

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+Source spans enter at one place. Lowering projects every id `collect_tree_ids`
+enumerates onto the span of the CHL it came from, and every later span is derived
+by walking `parents` back into that projection.
+
+## Terms
+
+| term | what it is |
+|---|---|
+| **pane** | A retained AST snapshot the inspector displays, materialized after a set of passes. Each one costs a retained full-tree clone. |
+| **pane relation** | What folding the passes between two adjacent panes produces: an id-to-id relation with labelled edges. The **durable, gated** artifact — the leak classes are asserted here. |
```
dpmills: This is LineageMap, right? We could just call it a lineage map

src/ccl/design/provenance.md
dpmills: Why are provenance and lineage two separate things? The contents of provenance.rs aren't even "provenance" in the colloquial sense; it's just node identity.  Can we drop one of these two terms entirely?

src/ccl/design/provenance.md
```diff
@@ -6,17 +6,92 @@ at lowering, monomorphization cloning subtrees, inline fanning UDF bodies out,
...
+rest is derived from it.
+
+A rewrite **records**: it names the node it is about to rewrite, and every node
+minted while that recording is the innermost open one takes the named node as a
+parent. A site declares nothing else. It does not declare what it produced — the
+produced side is discovered through the mint and copy hooks — and it does not
+declare what it destroyed, because a death is the difference between the ids live
+before and after.
+
+Those rows accumulate in one `NodeId`-keyed table per compile, one row per node:
```
dpmills: What happens if you create multiple recording guards against the same nodeID? For example, maybe we create something blamed to a node, then later transform that same node into new nodes? Do we maintain the proper edges?

src/ccl/lineage.rs
```diff
@@ -349,210 +462,312 @@ impl RewriteTag {
...
+    /// An input id that is absent from the output pane — **the death report**,
+    /// not a defect.
+    ///
+    /// Deaths are the set difference `input_ids ∖ output_ids`, which is the
+    /// whole of it: nothing declares a fate, so nothing can over-claim one, and
+    /// this class fires for *every* node that dies across the fold. It is data the
+    /// inspector reads, never something a gate asserts against. The class that
+    /// *is* a defect on this side is [`Unexplained`](Leak::Unexplained) — an
+    /// output node no capture explains.
+    Died { input: NodeId },
```
dpmills: If this isn't a leak, why is it in the Leak enum?

src/ccl/lineage.rs
```diff
@@ -349,210 +462,312 @@ impl RewriteTag {
...
+    /// produced it and the input pane does not hold it. The node's
+    /// lineage stops at an id that describes nothing.
+    ///
+    /// **One class for both edge shapes**, deliberately. The sole parent of a
+    /// freshened copy and one consumed id of a fusion are the identical
+    /// condition — an edge to an id outside the fold — and telling them apart
+    /// would mean recording the *shape* of the rewrite, which the `parents`
+    /// column does not and should not carry: its cardinality already expresses
+    /// 1:1, 1:many and many:1, and nothing else about the shape was ever read.
+    ParentUnknown { parent: NodeId },
```
dpmills: How is ParentUnknown different from Unexplained? Seems like you would hit them the same way: when you see a row in the LineageTable where the parent doesn't exist.  Is the difference whether we discovered the node by looking at the LineageTable or by walking the AST and comparing against the LineageMap? That doesn't seem like a particularly important distinction, but if it is, these need better names since that is very non-obvious from either the names or the comments.

If we don't care about the distinction, then this entire Leak enum can be deleted

src/ccl/lineage.rs
```diff
@@ -764,176 +921,485 @@ pub(crate) fn collapse_lowering(
...
-/// An in-flight step accumulating the ids born and copied within its dynamic
-/// extent. Finalized (flushed to a [`RewriteStep`]) by its [`StepGuard`]'s
-/// `Drop`. The `produced`/copy sides are *captured*, not declared: this is what
-/// makes recording a byproduct of construction rather than a parallel
-/// annotation.
+/// One in-flight recording, accumulating the ids born and copied while its guard
+/// is the innermost one open. Finalized when the [`FrameGuard`] drops. The
+/// produced side is captured from the construction hooks rather than declared,
+/// which is what makes a row a byproduct of the rewrite.
 struct OpenStep {
```
dpmills: If this is an inflight recording, should it be called InflightRecording?

src/ccl/design/provenance.md
groundlar: Done. Dropped "lineage". `lineage.rs` is gone — its contents moved into `provenance.rs`, which is now the single module for node identity, the phase tag, the recorder, the table, and the folds. `grep -rn 'lineage\|Lineage' src/ccl/` returns nothing. `LineageTable`/`LineageMap`/`LineageAudit` are `ProvenanceTable`/`ProvenanceMap`/`ProvenanceAudit`.

src/ccl/design/provenance.md
groundlar: Done. "Mechanism at a glance" now names both up front with the row shape (`NodeId ↦ ([Parents], [blame], {Phase, RewriteLabel, ...})`) and defines the fold as the operation over the table that produces a `ProvenanceMap`. "Collapse" is gone as a term — the two remaining uses are ordinary English ("collapse to a single entry") and the real rule label `channelize.defer_collapse`.

src/ccl/design/provenance.md
groundlar: Yes, and it's a supported shape — "Where to open a recording" covers it under *One entry serving arms of different `Nature`*: open a second recording on the same node. Rows are keyed by the node the rewrite **produced**, not by the node the recording names, so two recordings naming one parent write disjoint row sets and no key collides.

Your second case (a node blamed to X, then X itself rewritten into new nodes) is the *The named node may be one the phase itself just minted* bullet — the fold reaches the original in two hops.

src/ccl/design/provenance.md
groundlar: Done. The "pane relation" row is deleted; the doc now says the provenance map joining two panes is the durable, gated artifact, so there's one name for it.

src/ccl/design/provenance.md
groundlar: Done. `RecordingGuard` in `provenance.rs`, and the terms row matches.

src/ccl/design/provenance.md
groundlar: Done — it isn't, so "slot" is gone from the recording vocabulary. The node a recording names is the parent of everything minted under it. `RawSite.slot` → `parent`, `CrossDomain.slots` → `parents`, `StorePlan::carrier_slot` → `carrier_parent`, and the comments at every recording site in `mut_elim`, `transact_phase`, `channelize`, `planning/iterate`, `infer/solve`, `infer/emit`, and `subst` now say "the recording names X" or "the parent".

"Slot" keeps only its ordinary senses — a type slot, a binder slot, a tuple slot, a thread-local cell.

src/ccl/design/provenance.md
groundlar: Done. The terms row is gone, the doc no longer uses the phrase anywhere, and the one residue now reads "the `NodeId`s inside refinement predicates are outside the uniqueness walk".

It survives as descriptive phrasing in five code comments (`context.rs` ×3, `provenance.rs`, `infer/solver/traits.rs`) rather than as a defined term — across `src/ccl` that's 6 uses on `main` down to 5, with the design doc at zero. Say the word if you want those swept too.

src/ccl/design/provenance.md
groundlar: Done. Moved to `### Gate and audit coverage` at the end of "The provenance model", next to the leak classes they constrain; the terms table is down to two rows.

The stale content is rewritten against what the code does: the first pane pair is ungated because inference's predicate producers still don't record, so its residue is entirely `DanglingParent` and gating it would report a constant — matching `gated_pane_pairs()`, which returns one pair.

src/ccl/design/provenance.md
groundlar: Done. "Walking the ids" keeps all of main's content — the `uniquify::collect_node_ids` third walk, the `Rc`-sharing paragraph, and the `tag_predicate` entry crossing — and adds the three crossings on top rather than replacing it.

src/ccl/design/provenance.md
groundlar: Done. The separate "The node table" section is gone; its content is folded into "The provenance model" (the four load-bearing table properties) and "The fold".

src/ccl/subst.rs
groundlar: Done. Reverted to main's wording.

src/ccl/subst.rs
groundlar: Done, and swept. The empty `if !matches!(...)` block is gone and the code matches main.

I diffed the whole PR against `origin/main` and checked every hunk in `subst.rs`, `infer/`, `expr.rs`, `uniquify.rs`, `ty.rs`, `names.rs`, `planning/`, and the three other design docs: no further reverts of #98 or #101 — the rest is the `desugar`→`channelize` and `lineage`→`provenance` renames, the new recording sites, and `clone_preserving_ids`. The sweep did turn up a garbled paragraph in the status-markers block, now fixed, along with the hard-wrap and trailing-whitespace violations that came with it.

src/ccl/provenance.rs
groundlar: Done. `pub enum Phase` now lives in `context.rs`, beside the `PhaseScope`s `compile_program` opens and the `INFER_PHASES`/`CHANNELIZE_PHASES` ranges that read it.

groundlar: Done. Settled on `Phase`. `Pass`, `PassScope`, and `pass`-worded prose are all gone.

src/ccl/provenance.rs
groundlar: Done. `Mono` is off the axis and `Infer` is declared after `Uniquify` in pipeline order, with a doc comment saying monomorphization is what mints inside it. The pane-pair fields it fed followed: `mono_map`/`mono_deaths`/`mono_leaks` are `infer_map`/`infer_deaths`/`infer_leaks`.

src/ccl/lineage.rs
groundlar: Done. `Died` is off the enum; `fold` returns deaths as their own product (`input_ids ∖ output_ids`), so every `Leak` variant is a defect.

src/ccl/lineage.rs
groundlar: Done. The enum is down to two variants, renamed to say which side they check. `Unrecorded { output }` is the **capture** gate — an output-pane node no recording covered, i.e. a driver that missed a mint. `DanglingParent { parent }` is the **record-integrity** gate — a row naming a parent no row produced and no input pane holds.

Both edge shapes (a freshened copy's sole parent, a fusion's consumed id) fold into `DanglingParent` deliberately: telling them apart would mean recording the rewrite's shape, which `parents` doesn't carry.

src/ccl/lineage.rs
groundlar: Done, renamed to `OpenRecording` rather than `InflightRecording` — "the innermost **open** one" is the phrase the terms table and the `enter` docs already use for the nesting rule, so this keeps one word for it. Happy to switch to `InflightRecording` if you prefer that spelling.

src/ccl/lineage.rs
dpmills: This didn't address whether it's actually important to name the distinction that remains.  My impression is that the distinction is not important; it's going to be essentially random whether some other transformation in a phase touches a node that was incorrectly created without provenance, so which type of leak we see will be arbitrary (or we'll end up with duplicate leaks).  Evidence for this is the fact that nothing in this PR makes use of this distinction; we treat all leaks equally as errors

dpmills (approved):

src/ccl/lineage.rs
groundlar: You're right on the mechanism, and the comments were wrong. `fold` detects the two classes in independent loops, so one unrecorded mint fires either or both: `z` minted unrecorded, consumed by a recorded rewrite, and also surviving into the output pane yields `[DanglingParent { parent: z }, Unrecorded { output: z }]`. The claim that they "fail for different reasons" is deleted from `Leak`, `gate_leaks`, `materialize_panes` and `design/provenance.md`, and `one_unrecorded_mint_can_fire_both_classes_for_the_same_id` pins the behaviour.

The duplicate-leaks parenthetical was a real bug too. A dangling parent was discovered once per row naming it, so a count reported rows rather than broken ancestry targets — `test_udf_containing_filter` reported 32 entries for 8 targets. Both folds now sort and dedup, and `a_dangling_parent_that_many_rows_name_is_one_leak` pins one entry per target.

Where I'd push back is "nothing in this PR makes use of this distinction". `gated_pane_pairs()` holds the first pane pair out of the gate on exactly this basis, and I re-verified it: returning both pairs and running `CAMBRA_PROVENANCE_GATE=1` over `tests/compilation_pipeline` fails six programs on 46 leaks (190 before the dedup), every one `DanglingParent` and none `Unrecorded`. The mechanism is why that is a constant rather than a coin flip: inference's own rewrites do sit inside a recording scope, so every post-inference node carries a row and `Unrecorded` is zero; what dangles is the parents those rows name. The classes therefore separate two remediation targets across a whole fold — `Unrecorded == 0` says every mint had a scope open, and a `DanglingParent` count over that says which upstream producers still don't row. Collapsed to one class, "this residue is a constant, not a regression" cannot be stated, and pair 1 reads the same as a `channelize` site that missed a mint.

So: two variants kept, both comments rewritten to say the split is the id's role at detection rather than a different defect, and the measurement moved into the `gated_pane_pairs` rationale.

One caveat on that rationale, since I am leaning on it here: its attribution to `specialize_use` is wrong, as a later commit in this stack records — the actual causes are inference freshening a refinement predicate per scheme instantiation, and `collect_tree_ids` omitting binder-declared type slots. The class signature above is unaffected, being measured rather than derived from that story.

Pushed as e6823c0c.

PR #108, opened 2026-08-18
Scope inference variables with telescopes, and identify binder references by position

dpmills:
A refinement predicate may reference an enclosing binder, and the bound holding it travels away from that binder during inference: group-by's lowering puts the dependent refinement in a cast target while the function that binds `__gb_k` is minted separately, so the refinement lands on a variable whose position has no enclosing function. Nothing checked it there — `check_scope_valid` runs on coalesced node types after inference ends, never on bounds — and the sites that decide identity compare stored types structurally, which is α-sensitive. This PR gives every inference variable the **telescope** of binders in scope at its creation, rejects a recorded bound that does not close against it, and identifies every binder reference by **position** rather than by name. α-variant types become structurally identical wherever identity is decided, and an escaping bound is a record-time internal error rather than a latent wrong type.

The design of record is [type-inference.md, "Scoped inference variables"](https://github.com/cambra-dev/cambra/blob/dmills/scoped-infer-vars/src/ccl/design/type-inference.md#scoped-inference-variables-a-stored-bound-closes-against-a-telescope).

### Where the binder sits decides how it is stored

Two positions, two mechanisms — [the two forms](https://github.com/cambra-dev/cambra/blob/dmills/scoped-infer-vars/src/ccl/design/type-inference.md#a-binder-reference-is-stored-in-one-of-two-forms) and [interior binders](https://github.com/cambra-dev/cambra/blob/dmills/scoped-infer-vars/src/ccl/design/type-inference.md#interior-term-binders-stay-named-and-compare-by-position).

A **function's** binder sits outside the refinement, so a reference to it must be readable off the reference alone: `Name::PiBound` holds a de Bruijn index, assigned when the function is constructed. A reference to a **telescope entry** stays a uniquified name, which is what keeps two refinements about different enclosing binders apart. `close_pi_binder` and `open_pi_binder` own the conversions and run at four sites ([where](https://github.com/cambra-dev/cambra/blob/dmills/scoped-infer-vars/src/ccl/design/type-inference.md#where-the-conversions-run)): construction closes, landing closes, descent opens at the binder name, application opens at the argument. Landing runs in two walks — `compact_go` and `key_go` — which have to enter and leave the same crossings, or a compacted type and the `SpecKey` beside it would spell one refinement two ways; `subst::RefinementScope` is the state both thread, owning the enclosing-binder stack and the closing memo together.

A **predicate's own** lambda binder sits inside the refinement, so the compared terms carry it and the correspondence is local: `eq_refinement_predicate` threads a pairing and `hash_refinement_predicate` threads the same scope, both resolving such a reference by position. The name stays the stored form, so a predicate is still a term planning can compile.

Positions and indices are both needed. Every refinement routes its element through a function, so it carries an interior lambda, and a filter written twice mints that binder twice — with indices alone, a conditional over two character-identical filtered comprehensions is rejected as two domains that do not join.

### The record-time check

`Telescope` is a persistent cons list on `InferVar`, extended by `scoped`/`scoped_let` in both `InferCtx` and `CheckCtx` through `TelescopeWalk::under_binder`, which owns the save/restore so no entry can leave a later variable recording a scope it does not sit in, and threaded through `normalize_annotation`, so a variable minted inside a dependent annotation carries its Pi binders. `observe_bound_scope` checks a recorded bound's free names against the holder's telescope and the edge substitutions.

Whether a gap is an error depends on the derivation, and `Derivation` names three: `LiveSolve` and `PostPass` enforce, and `SubTree` — `debug_typecheck`'s per-operation probe — does not, because the context it was cut from binds those references. The same value gates the `Fun`/`Fun` codomain opening, which carries the same tripwire as `subst::open_codomain`: an unnamed function whose codomain references it by index has no binder to open at, so the reference reaches the codomain edge and compares against the other side's name. `CAMBRA_TELESCOPE_LOG` enables the observation log. One shape that would reach the solver as a reference to a binder nothing binds — a parameter annotation naming a parameter, `(a: Int, c: {Int where _ >= a})` — lowering rejects instead (`reject_annotation_references`); supporting it needs a telescope the tupled parameter list does not carry.

### Reading, not storing

`Display for Type` threads the functions it descends through, so a reference to one prints as that function's binder name. A diagnostic blames a *fragment* — `coalesce_compact_go`'s domain-join conflict reports the domains of a function it is half-way through assembling — and has none to descend through, so `PiRef` also carries the binder's spelling. Identity reads its index alone.

### Tests

Every test that infers from CHL source now α-uniquifies first, as the pipeline does before `infer`; `type_check.rs` had eight copies of that prefix and now has `lower_uniquified`. That change falsified `test_case_with_filtered_comprehension_arms_passes_consistency_wall`, which α-invariant refinement identity then fixed.

`alpha_variant_bound_merge_is_canonical` (`compact.rs`) and `spec_key_shares_alpha_variant_dependent_types` (`spec_key.rs`) each have a `keeps_distinct_binders_distinct` sibling. `alpha_variant_predicates_are_one_restriction` covers the three binding forms that reach the comparison, asserting hash equality alongside `eq`.

### Vocabulary

`arrow` for a function type was current when these commits were written and retired the next day by #113, which left only the rendered-glyph sense (`FunKind::arrow`). Those uses stay; the type-level ones say `function`, and the one that meant the kind says `FunKind`. `frame` already meant `SpecializeFrame` and the provenance recorder's copy frames, so the position an index counts to is a `function` too, and the walk state tracking them is `enclosing`. `coordinate` named the axis `closed`/`open` already names — and those are what the conversions are called — so it is gone, and the design heading is now the sentence that had defined it.

`claim` was a second name for `Refinement`. On `main` the word appears only in its ordinary sense — an assertion, as in "a refinement is a claim about the value at a position" — and this branch had turned it into the count noun for the object, 32 uses in the design section plus `ClaimScope`, `claim_closing_tests`, and `escaped_claim` in code. The object is a `Refinement` and is called one; the assertion sense stays where it reads as English. `family` and `fiber` were undefined and are gone — a dependent function is called that.


dpmills:

- #108 ◀
    - #77
        - #78
            - #115
                - #116
                    - #117
                        - #118
                            - #109

src/ccl/design/mutability.md
```diff
@@ -1090,7 +1090,7 @@ Both a **finite** loop and an **async** (streaming) source drive an induction ac
 model treats a finite domain as a stream that terminates (§Liveness) — and every induction accumulator
 uses *one* realization: the changelog `InductionStore`. Plain, conditional, and feed-carrying loops
 over finite or async extents all route through it. The
-drive reads its source by absolute domain position (async domains arrive unordered), reclaims the
```
sortalongo, line 1093: Looks like someone went HAM on the s/claim/refinement

src/ccl/design/type-inference.md
```diff
@@ -944,6 +944,260 @@ The expected binder is **always globally fresh** (proposal §5.2 verbatim; the 
...
+Two sites assemble a `Fun { name: Some(_), .. }` and close nothing, one on each side of the
+construction boundary. `emit_lambda`'s codomain is the live in-solve type, which stays name-spelled:
+its refinements accumulate behind an inference variable, and closing them is landing's job, so
+closing the reachable ones here would put an index-spelled refinement and a name-spelled one at a
+single position — what landing-close exists to prevent. `coalesce_compact_go`'s refinements arrive
+already closed, and the compact view it assembles from mirrors `Fun`s one-to-one, so a landed index
+counts to the function being built; closing is the identity there, and the `Fun` literal says so.
+
+**Landing closes.** The compact and key walks close a refinement against the enclosing functions the
+walk is inside of, as it lands in the view, in the same two arms that force the edge substitutions
```
sortalongo: "land" is an odd term, esp. when contrasted with "descent". Just call it "ascending past a binder"?

src/ccl/design/type-inference.md
```diff
@@ -944,6 +944,260 @@ The expected binder is **always globally fresh** (proposal §5.2 verbatim; the 
...
+check's `apply` rule is this site's second caller: it re-derives an application's type from the
+closed function the tree records, so it opens the codomain at the argument before comparing against
+the stored, already-discharged type.
+
+Each identity site therefore compares like with like: whether a type is closed follows from which
+side of the construction boundary it sits on, not from when it arrives.
+
+#### Rendering opens what it descended through
+
+An index is a stored form, not a read form. `Display for Type` threads the functions it descends
```
sortalongo: s/"render"/"display"

sortalongo (approved):
I left 👨 comments in GH UI.

> 🤖 **AI-generated review** (Claude Opus 5, via Claude Code). Findings are unverified by a human; treat each as a question to check rather than a defect to fix.

## Verdict

Merge-ready on substance. `./ci.sh` is green on the branch head as of `407e49a0` — all four clippy configurations, doc, and the full suite, with the record-time closure check enforcing throughout (it is `debug_assertions`-gated and tests run in debug, so the suite exercises it). The design section, the tripwires, and the negative tests are all present and load-bearing.

Two structural notes and a rebase are worth resolving first.

## Architectural

**1. One type now has two non-interchangeable spellings, and only convention keeps them apart.** After this PR a `Fun` with a named binder is either index-closed (built through `Type::pi` / `pi_kinded` / `fun_like`, which close) or name-spelled (built field-wise — `emit.rs:669`'s `emit_lambda` is the live-solve form). `Type::without_pi_names` deliberately does not reconcile them, and the PR says so at `ty.rs:1350`. But `Type::Fun { name: Some(_), .. }` stays publicly constructible without closing, so the distinction is enforced nowhere: the two tripwires (`subst::open_codomain`, `constrain.rs`'s `open`) fire only once a reference has already escaped its binder. Worth considering whether the construction path should be the only path — private fields behind the closing constructors, or a pass-boundary `debug_assert` that a *stored* function's codomain carries no free reference to its own binder.

**2. `spec_key.rs`: the cycle guard and the memo now key differently.** `visit_key = (uid, pol)` at `spec_key.rs:484`, `memo_key = (uid, pol, enclosing)` at `:485`. Reaching one variable at two enclosing-binder stacks within a single descent is therefore cut as a cycle, producing a coarser key — which is the "two uses reach one clone whose interior was resolved against the other's argument" hazard `closing_keeps_distinct_binders_distinct` exists to prevent. The truncation counter keeps the memo from being poisoned, so this degrades rather than corrupts. If it is unreachable because the shape needs a recursive type the solver rejects, that belongs at the site: the memo's doc comment was updated for the new key and the guard's was not.

## Medium

**3. `reject_annotation_references` has a documented gap with no test.** The `Mut`-curried arm of `uncurry_params` skips the check, justified by "that path declares only the `Mut` parameters' types and drops every other annotation". That reads true today — the arm never calls `lower_type_annotation`. But it is the invariant the rejection rests on, and if the path ever starts declaring sibling annotations the reference reaches the solver as a debug panic on user input instead of a diagnostic. A test pinning the drop, or a `debug_assert` in that arm, would hold it.

Adjacent and pre-existing, surfaced by that comment: a non-`Mut` parameter's annotation *is* silently dropped in the `Mut`-curried path — the same class of bug the single-parameter arm's comment says was fixed ("Without this the annotation was silently dropped and the param inferred purely from its body"). Separate change; flagging so it is not lost.

## Nits

**4. `constrain_go`'s `Fun`/`Fun` arm gained release-active whole-subtree scans.** `references_enclosing_function` runs up to twice per function-function edge (`constrain.rs:605`, `:612`) and allocates a `BTreeSet` once refinements are present, plus `type_contains_infer` in the dependent case. Constant-factor on the solver's hottest path. The PR measures the record-time check (~4% of debug inference); this path has no number.

**5. Two `Vec<Option<Name>>` clones per memo probe** — `RefinementScope::close`'s `(PredicateId, Vec<Option<Name>>)` and `spec_key`'s `(uid, pol, Vec<Option<Name>>)`. An interned scope id would drop both allocations.

**6. Three spellings of one predicate.** `subst::well_formed` (`subst.rs:1363`) is dead outside its own unit test; `check_scope_valid` (`solve.rs:891`) inlines exactly its body; and the new `bound_scope_gaps` (`infer_var.rs:389`) is the same shape a third time, plus the edge substitution domains. `well_formed` returning the gap set would leave one.

**7. `Name::base()` answers `"#"` for a hint-less `PiBound`.** `Display` renders `#0` correctly, but a diagnostic reaching for `.base()` prints a bare `#`.

## Confirmed good

- The α-invariance work keeps `Eq`/`Hash` in the safe direction — hash never finer than eq — and `assert_same_restriction` asserts both halves at every case rather than only `eq`. The capture case (`λ p → p > 1` vs `λ q → p > 1` with `p` free) falls out of `paired_refs_match`'s `(Some, None) => false` arm; worth a case in `alpha_variant_predicates_are_one_restriction` since it is the arm that makes the relation sound rather than merely coarse.
- `TelescopeWalk::under_binder` owning the save/restore is the right shape. `scoped` still drives `scopes.push_scope()` / `pop_scope()` beside it, so two structures track scope in lockstep at one call site — fine as is, worth watching if a third appears.
- The negative tests carry their weight: `closing_keeps_distinct_binders_distinct`, `a_ground_pair_does_not_open_at_a_shared_spelling`, `a_free_reference_still_compares_by_name`, and the lowering rejection's three-case accept list.

## Rebase

The branch predates `main`'s `#107` (`Replace the lineage log with a NodeId-keyed provenance table`). The two touch 25 files in common, including `subst.rs` (108 lines there vs 669 here), `ty.rs`, `names.rs`, `lower/functions.rs`, `planning/groupby.rs`, `solver/scheme.rs`, `solver/constrain.rs`, `lambda_elim.rs`, and `tests/type_check.rs`. Rebase and re-run `./ci.sh` before merge.

src/ccl/design/mutability.md
dpmills: Fixed, and it reached further than this line. The substitution ran over the verb forms as well as the noun: `reclaims` → `rerefinements` at 16 sites (this line, `design-operators.md`, `commit_operator.rs`, `ccl_utils.rs`, `uniquify.rs`, `solve.rs`, `solver/mod.rs`, `infer/api.rs`), `claiming` → `refinementing` in a lowering comment, and 14 further prose uses that lost the ordinary sense — "every claim about what runs", "the capabilities Cambra claims", "an arm that names no payload claims the tag carries nothing", "it carries both halves of the claim".

All 31 are reverted across 25 files. What stays converted is the type-system noun and the locals holding a `Refinement`; I classified each of the 8 survivors individually rather than reverting by pattern.

It came from the bottom commit of the branch rather than the terminology commit at the tip. The rebase surfaced one as a conflict, since #107 rewrote the `provenance.md` paragraph where it had produced "every id a row **refinements** to produce".

src/ccl/design/type-inference.md
dpmills: Renamed — though not to "ascending", because the close happens on the way *down*. `compact_go` pushes the binder with `st.scope.enter(name)` as it descends into a codomain, and `st.scope.close(&r)` fires at the `Type::Refinement` arm itself, against the stack that descent accumulated. Nothing closes on the return path.

Rather than swap one noun for another I dropped the noun. The heading is now the rule:

> **A refinement closes against the enclosing functions of the walk carrying it.**

and the body names `compact_go` and the `SpecKey` walk rather than a concept. Two intermediate names did not survive review: "entering a view" (`view` already means a polarity's read of a variable, and a declared domain read as a view, both in this file) and "storing closes" (`stored` is this section's own word for the closed form — "references to its own functions **stored** as indices" — so the heading read as near-tautology and appeared to subsume `Construction closes`).

Cleaning up the code sites turned up an error of mine. `planning/groupby.rs` had "**Landing** the lifted ids as they are would put one id-set on two live terms", which is about node ids arriving in the term tree and has nothing to do with refinement closing; a mechanical first pass had converted it with the rest. That and four other ordinary uses of "lands" are back to their originals. The five that were genuinely the site name now read "closing against the walk's enclosing binders", and `subst.rs` keeps "a refinement-closing walk". No identifiers were involved.

src/ccl/design/type-inference.md
dpmills: Done. The heading is now "Display opens what it descended through" and the two body uses follow. Both inbound citations are updated so `doc_refs` stays green — `ty.rs` and `tests/type_check.rs`, where the neighbouring "renders it as the key binder's name" is now "displays".

Left alone: "rendering morphism" (`edge_render_subst`) and "one polarity's rendering" earlier in the file. Different concepts, neither of them `Display`.

dpmills:
Rebased onto `b5b866c7` and pushed, with the review findings addressed. All nine heads pass `./ci.sh` (all four clippy configurations plus doc) and both ends pass `DEEP_TYPECHECK=1 ./ci.sh test`.

## Rebase

Four conflicts and two fallouts, all against #107:

- `provenance.md` — #107 rewrote the section this branch had mangled with the `s/claim/refinement` above; took #107's text.
- `context.rs` — `argument.clone()` → `clone_preserving_ids()`, merged with the `observe_bound_scope` restructure.
- `api.rs` — `check_pre_desugar` → `check_pre_channelize`; the `_sub_tree` / `_as` variants follow.
- `planning/iterate.rs` — #107's provenance guard over #77's `peel_refinements` rewrite.
- `uniquify.rs` and `context.rs` predicate walks now iterate `RefinementSet`.
- `subst.rs` `PiWalk::refinement` cloned a predicate raw with nothing recording — 24 `DanglingParent` in the `group_by` pane pair once `Clone` mints. It is a derive rather than a replace (`preserving_ids` trips the `predicate-vs-predicate` half of `predicate_id_collisions`, because the stored closed form stays live beside the opened one), so it now runs under a `predicate.coordinate` recording, mirroring `PredMemo::rebuild`.

## Addressed

**1 — the construction boundary, asserted from the outside.** Took the second option. `subst::name_spelled_stored_binders` rejects a stored function whose codomain still references its own binder by name, checked at `check_scope_valid`, which already visits every coalesced node. It asks about the codomain alone: `type_free_vars` on the whole `Fun` binds the Pi name over the codomain, which is the reference in question.

It fired exactly once across 1231 tests, on `scope_check_accepts_enclosing_binder` — which was hand-building the `Fun` literal that skips closing, so the hazard was reachable from a test rather than a pass. That test now builds through `Type::pi`; its comment also claimed the Pi binder *name* is what keeps the type in scope, which telescopes had made stale.

**2 — the cycle guard and the memo.** The guard should stay coarser: it is what bounds the walk, and a cycle growing the binder stack each turn would never repeat the finer key. The two agree because reaching one variable twice at different enclosing binders needs the variable to occur under a function in its own bound chain — a recursive type, which `coalesce_compact` rejects — while the cycles that do occur are var-to-var and cross no binder. Both the reasoning and a debug `visiting_scopes` check now sit at the guard, so the coarsening is loud rather than silent if that reasoning is wrong.

**3 — `reject_annotation_references`: deleted, both arms.** #128 landed two tests while this branch was stale, and they contradict it: a parameter's annotation is read in the scope *enclosing* the binder, so `x = "Hello"` ahead of `def test(x: {Int where _ >= x})` names the outer `String`. The check matched on the parameter's name, so it reported the wrong cause there — and its own accept list already blessed that shape under a non-colliding name (`outer_let`).

Its stated purpose was to keep the reference from reaching the solver as an internal error. It does not: at all four shapes it covered, the solver reports `Unbound variable`. Those cases moved to `tests/compilation_pipeline/type_annotations.rs`, where the pipeline can assert what they report. So the documented gap is gone rather than tested.

Your adjacent note stands and is untouched by this: the `Mut`-curried arm gives a non-`Mut` parameter `Type::Hole` and never calls `lower_type_annotation`, so `def f(a: Int, m: Mut(Int))` still drops `a: Int` silently. Separate change.

**6 — three spellings of one predicate.** `subst::well_formed` was `type_free_vars(ty).is_subset(ctx)` with no caller but its own unit test, `check_scope_valid` inlined exactly that, and `bound_scope_gaps` was the same plus the edge domains. One `scope_gaps` now, taking the scope as a predicate rather than a set so the telescope-plus-two-domains caller builds nothing to ask.

**7 — `Name::base()`.** `PiRef::bare` is the only hint-less constructor and is documented test-only, so the bare `#` is latent rather than live. A `debug_assert` makes it loud and points at `Display`.

**The capture case** is added — both directions, and only equality asserted, since `Eq`/`Hash` runs one way and a coarser hash is sound.

## Measured, not changed

**4 — `references_enclosing_function`.** Two corrections. The call sites are mutually exclusive: `debug_assert!(n.is_some() || !refs(c))` short-circuits when the binder is named, and the match guard runs only when it *is* — so at most one scan per `open`, two per function-function edge, and in release only for named functions.

For the number: ~543 calls per program over the eleven-program corpus, and running the scan twice and differencing wall clock puts the whole set at **~1.9% of a release compile**, against the ~4% measured for the record-time check. Left alone at that.

**5 — the two `Vec<Option<Name>>` clones.** `RefinementScope::close` runs 4074× over the corpus and returns early 91% of the time at the all-`None` check, building no key and cloning nothing. The 376 that do clone average **1.00** elements; `spec_key`'s 130 memo probes average 0.74. An interner would remove ~500 one-element vec clones per full-corpus compile, so I left it — happy to do it if you want it regardless.

Neither number came from a bench harness; the repo has none, and doubling the call and differencing was the honest measurement available without building one.

PR #109, opened 2026-08-18
Model the type system in Lean, and check the solver's subtyping, polar merge, and materialization against it

dpmills:
This PR is a first pass at modeling the type system in Lean and proving theorems about it.  Only concrete types (no inference variables) are included for now, and there are some exceptions to what is covered.

The model lives in a new formal/ directory, and formal/README.md is a good entry point for understanding what is modeled and proved and how to read the rest of the work


dpmills:

- #109 ◀

formal/README.md
```diff
@@ -0,0 +1,168 @@
...
+| Materialization (`coalesce_compact`) | `coalesce` | totality; the result is well-formed; the merge materializes to a bound of both operands where one exists, and to the least such type |
+| Terms and typing (a small pure core) | `Term`, `HasTy`, `Step` | progress (a well-typed term is a value, steps, or is filter-blocked), preservation (a step keeps the term's type), refinement soundness |
+
+A type is **concrete** when no inference-time unknown occurs anywhere in it — what a checked program
+exhibits, and the fragment everything above is stated over. The model has no node for any of the
+unknowns; [design.md](design.md#the-concrete-type-grammar) lists which `ccl::ty::Type` variants
+those are.
+
+Three things the model does not do: infer a type — it checks one it is given — reproduce the
+solver's choice where several types are valid, or say anything about those transients.
```
sortalongo, line 28: An interesting thought... if we can prove that the type system is sound, and that our type _checking_ implements that system, then is it actually that important to prove correctness of our inference algorithm? If inference yields an incorrect type, type checking will catch it.

sortalongo (approved):

formal/README.md
dpmills: Hmm, that's enough to prove that we can infer consistent types on the corpora of tests that we have, but not that we can infer correct types.  It also doesn't let us prove properties like decidability of inference.  In the future, it would also be nice to be able to prove stuff like whether finding principal types for polymorphic functions is decidable.
In terms of importance though, yeah, probably not actually the next most important thing

PR #113, opened 2026-08-19
Relate the two function kinds by equality instead of ordering `Data` below `Compute`

dpmills:
`Data <: Compute` let a collection be read as a capability, and once it is one the compute lattice's contravariant domain meet applies and narrows the domain — but a data function's domain *is* its rows, so the narrowing drops them with nothing in the type recording it. The two kinds are now incomparable: subtyping on them is equality, and each is decided where the function is built rather than inferred. A program that previously lost rows is rejected instead.

## The edge lived in two places

`constrain_kind` admitted `Data <: Compute` outright. It now accepts only reflexivity and rejects either mismatch as `ConstrainError::KindMismatch` (renamed from `ComputeWhereDataRequired`), returning the kind the edge equated its two sides to so the caller picks the invariant or contravariant domain rule without re-inspecting them.

`CompactFun::merge` carried the same ordering in the lattice operations: the join read `Data ⊔ Compute` as a capability over the *met* domains, and the meet took `Data` whenever either side had it. Both are `KindMerge::Conflict` now.

## `FunKindVar` becomes a write-once cell

An equality edge needs no propagation graph, so `link` and the `uppers`/`lowers` lists are gone: a concrete kind on either end pins the variable, and two variables meeting record nothing. `KindMerge::Unknown` separates "nothing required a kind" from "required a capability", which the old `(false, false) => Compute` conflated; the capability default now applies at coalesce instead of during the merge.

## Lowering decides

The two elimination demands mint the only kind variables (`Type::pi_eliminated`, `Type::fun_eliminated`), since applying a value and destructuring a function are one node for both kinds. Everything else is stamped where it is built: lowering marks the shapes it mints with a `data_fun` annotation, `emit_list` and `emit_disjoint_join` read the recorded type's kind rather than assuming `Data`, `emit_node` debug-asserts that a `Compose` arrives with a kind decision, and `lambda_elim` threads the eliminated lambda's (`arrow_kind` → `fun_kind`). `Type::with_kind_of` re-stamps a kind through refinements where a pass re-kinds a node.

Read `ty.rs` and `emit.rs` first; the rest follows from them. Background is in [type-inference.md](src/ccl/design/type-inference.md), "4.6 Data vs compute functions", rewritten and net shorter.

## Behaviour change

```
xs = [1, 2]
f = \x -> xs(x)
f if True else xs
```

inferred `[0,1] ⇒ Int` and now reports `CoalesceError::KindConflict`. Three tests cover the routes that stay: a parameter's kind solved from its argument, destructuring a collection, and this join.

The second commit is terminology only, giving each concept one name. `arrow` named the rendered symbol, a `FunKind`, and a `Type::Fun` at once. `equate` was a second name for `pin` while already meaning "an edge relates two types both ways" elsewhere in the solver, so `equate_compute`/`equate_data` become `pin_compute`/`pin_data`. `consumption rule` was a coined name for an elimination.


src/ccl/lambda_elim.rs
```diff
@@ -1479,6 +1518,20 @@ fn elim_lambdas(ctx: &mut ElimContext, expr: Expr) -> Result<Expr, LambdaElimErr
 }
 
 fn elim_lambdas_impl(ctx: &mut ElimContext, expr: Expr) -> Result<Expr, LambdaElimError> {
+    if let TypedExprNode::Cast { target, .. } = &expr.node {
+        let k = |t: &Type| match t.peel_refinements() {
+            Type::Fun { kind, .. } => format!("{kind:?}"),
+            _ => "-".into(),
+        };
+        if k(&expr.ty) != k(target) {
+            eprintln!(
```
groundlar: maybe assert instead?

src/ccl/ccl_utils.rs
```diff
@@ -808,6 +808,30 @@ pub fn sync_cast_targets(expr: &mut Expr) {
...
+/// independently, `emit_cast` reads it off `target` to type the node, and so the
+/// two copies have to agree or the node contradicts itself.
+///
+/// The caller is a rewrite that hands a node over to one of its own
+/// sub-expressions (`simplify`'s collapse rules). Such a rewrite writes the
+/// position's type onto the survivor, and where the survivor is a cast that
+/// re-kinds it — `⟨id, const 𝑥⟩ ≫ apply` collapsing to a `𝑥` that is a collection
+/// standing in a morphism position. Only the kind moves; the claims stay the
+/// cast's.
+pub fn sync_cast_target_kind(expr: &mut Expr) {
```
groundlar: nit: `pub(crate)`?

groundlar (approved):

src/ccl/design/type-inference.md
```diff
@@ -946,73 +946,21 @@ The pipeline passes downstream of inference treat function types structurally an
...
-> var that ends with `forced_compute ∧ forced_data` is the same `Compute <: Data`
-> error, surfaced at coalesce as `CoalesceError::ComputeWhereDataRequired`; a var
-> with only `forced_data` resolves to `Data` (a parameter used only as a collection).
-> The same arm carries the data-domain invariance guard — see
-> [Data domains are invariant](#data-domains-are-invariant) below.
+Functions can either represent collections or a computation that can be run.  This
+is stored as a `FunKind` on `Type::Fun` and is either `Data` or `Compute`.  The kind
+is chosen at lowering time based on the CHL construct used.  For example, UDFs are
+`Compute` if they use `return` and `Data` if they use `yield`, list comprehensions
+are always `Data`, etc.
```
groundlar: The `yield` half of the UDF example is wrong, and this section is the citation target for six code comments.

`lower/functions.rs` builds every UDF with a bare `Expr::lambda` (lines 131, 164, 215) and stamps no kind, and `Expr::lambda` is concrete `Compute`. A `yield` UDF's *result* is `Data`; the UDF itself is a capability taking a collection to a collection. That is what this PR's own new comment in `inline.rs` says — "the UDF around them is a capability taking one to another" — and what `a_parameter_iterated_as_a_collection_resolves_data` rests on when it says `inline` only inlines a capability.

"The kind is chosen at lowering time" also has an exception the section does not state: a parameter has no lowering-time stamp, and `Type::fun_eliminated` mints a `FunKind::Var` for it. That new test's own doc comment cites this heading for exactly that fact, so the section contradicts the test pointing at it.

Separately, the rewrite dropped facts other files still point here for. `subst.rs`, `lambda_elim.rs`, and `planning/mod.rs` cite this heading for the `fun_like` discipline (a domain/codomain-only rewrite may not flip the kind); `emit::stamp_kind_from` cites the `data_fun` provenance stamp. Neither survives in the section. `./ci.sh doc_refs` passes because the heading still exists — that is the rot the check cannot see.

Style: the paragraph reads as blog prose rather than spec prose — "etc.", "should be preserved", "nothing should implicitly convert", and "a computation that can be run" where the codebase already says "capability". Worth a pass through the `spec-prose` skill.

src/ccl/infer/solver/spec_key.rs
```diff
@@ -250,6 +250,8 @@ impl fmt::Display for KeyView {
                 KindMerge::Data => "⤇",
                 KindMerge::Compute => "⇒",
                 KindMerge::Conflict => "⇒!",
+                // The capability default this resolves to at coalesce.
+                KindMerge::Unknown => "⇒",
```
groundlar: `fun` is keyed by `KindMerge` (`BTreeMap<KindMerge, _>`, line 150), so `Unknown` is a distinct key from `Compute`. Before this change an unpinned kind var read `Compute` through `KindMerge::of` and shared a clone with a concrete compute function at the same position; now the two split into separate monomorphizations. Not incorrect, but it is specialization nobody asked for.

And because both render `⇒`, the two `SpecKey`s print identically while comparing unequal — so the split is invisible in the diagnostic that would otherwise explain it.

Either resolve `Unknown` to `Compute` when building the key (the default is decided at coalesce anyway, so keying by the resolved kind is the honest form), or render it distinctly, e.g. `⇒?`.

src/ccl/infer/solver/constrain.rs
```diff
@@ -183,48 +182,68 @@ pub fn constrain_subtype(
...
-            false
+        (Data, Var(v)) | (Var(v), Data) => {
+            v.pin_data();
+            Ok(PinnedKind::Data)
         }
+        // Two variables meeting record nothing. What the pair resolves to is not
+        // known at this edge, and deciding it from pins that arrive later
+        // would make typing depend on constraint order; each is pinned by the
+        // concrete kind that reaches it, if one does. See [`FunKindVar`].
+        (Var(_), Var(_)) => Ok(PinnedKind::Undetermined),
```
groundlar: Dropping the var-var link — with `freshen_kind_var`'s mirroring — also drops `kind_var_link_then_late_force_propagates` and `freshening_mirrors_kind_var_links_onto_instantiation`, which were the guards on it. The argument here (the concrete value pins the var directly, so a var-var edge carries nothing) holds for a freshly minted `fun_eliminated` demand, which sits in negative position.

Where I could not convince myself is a var kind reaching *positive* position through a generalized scheme. `f = \xs -> xs` with `xs` destructured coalesces to roughly `(?d ⇒^κ ?c) ⇒ (?d ⇒^κ ?c)`, so a freshened `κ'` occurs positively and can meet a second elimination's `κ₂` at a var-var edge; both then take the `Compute` default at coalesce even if the argument pins `κ'` to `Data`. I could not build a program where that miskind is observable (`inline`'s capability test is the sensitive consumer), and the suite is green — so this is a request to confirm, not a claimed defect.

If it is provably unreachable, the reason belongs in §4.6, or as a `debug_assert!` at coalesce that no positive-position slot materializes at `Unknown`. Right now the deleted tests are replaced by nothing.

groundlar (commented):
Three findings from a review of the two commits. The kind-equality change itself reads right to me — `Data` and `Compute` genuinely are incomparable, and `FunKindVar` collapsing to a write-once cell follows from that. `./ci.sh fast`, all three remaining clippy passes, and `doc_refs` are green locally at 22fe666.

src/ccl/lambda_elim.rs
dpmills: Deleted rather than converted — it was leftover `eprintln!("DBGIN …")` instrumentation from my own investigation, not on `main`. Good catch.

Replaced with the assertion it was probing for: at `elim_lambdas_impl`'s entry, a `Cast`'s node kind and its `target` kind must agree, which is the invariant `sync_cast_target_kind` exists to maintain. Verified it holds — 2144 tests pass with it live. It reads `expr.ty.fun_kind() == target.fun_kind()` off a new `Type::fun_kind` helper.

src/ccl/ccl_utils.rs
dpmills: Done.

src/ccl/design/type-inference.md
dpmills: All four correct; section rewritten.

Confirmed the `yield` claim before touching it, since the section is the citation target:

```
scalar-udf : (Int ⇒ Int)
yield-udf  : ((?101 ⇒ Int) ⇒ feed(chan(__result_0) ⇒ Int))
```

The UDF is the capability; its *result* is the collection — exactly what the `inline.rs` comment says. Fixed.

Added the parameter case as its own sentence: nothing for lowering to read, so it takes a kind variable the argument pins. That is what `a_parameter_iterated_as_a_collection_resolves_data` cites the heading for.

Restored both dropped facts — the `data_fun` annotation stamp and the `fun_like` rebuild discipline (a domain- or codomain-only rewrite cannot flip the kind) — so `subst.rs`, `lambda_elim.rs`, `planning/mod.rs`, and `stamp_kind_from` point at something that is there again.

Style pass done: "etc.", "should be preserved", and "nothing should implicitly convert" are gone, and "a computation that can be run" is now "capability".

Your last point is the one I would generalize. `doc_refs` was green the whole time because the *heading* survived while the facts under it did not — the check cannot see that by construction, and a citation naming a heading is exactly as strong as the heading's stability, not its content.

src/ccl/infer/solver/spec_key.rs
dpmills: Agreed, and took both halves of the suggestion.

`key_go` now resolves `Unknown` to `Compute` when building the key — coalesce applies that default anyway, so keying by the resolved kind is the honest form and the split disappears. Added the missing assertion to `fun_kind_is_part_of_the_key`: an unpinned var must share a key with a concrete compute function, which is the regression you identified and nothing was checking.

Also took the second suggestion, since it costs nothing and the failure mode you describe is nasty: `KeyView` renders `Unknown` as `⇒?`. It no longer reaches a key, but a diagnostic can never again print two unequal keys identically.

src/ccl/infer/solver/constrain.rs
dpmills: Measured it rather than argued it, and the answer is sharper than "benign": **the arm never fires.**

I instrumented the `(Var, Var)` case to report every edge and whether its two ends were pinned differently, then ran the full suite — zero hits. I checked the instrumentation fires on a synthetic var-var edge first, so the zero is real rather than an artifact of the probe.

So your scenario's precondition does not occur in the suite. That is evidence, not proof, and it is also precisely your point: the deleted tests were replaced by nothing, and an unreached arm is one nobody would notice becoming reachable.

Added a `debug_assert!` in the arm that the two sides' pins agree — the exact condition under which carrying nothing is unsound. Being explicit about what it does not cover: it catches a disagreement present *when the edge is drawn*, not a pin landing on one side afterwards. I did not take the coalesce formulation, because an unconstrained generic legitimately materializes at the default there, so "no positive-position slot at `Unknown`" would fire on `a_never_called_generic_function_is_accepted`.

Recorded in §4.6, "Deliberately incomplete here", including the zero measurement and the assertion's blind spot.

PR #114, opened 2026-08-19
Implement parse and lower for refinements and function-def output types

octalsrc:
This PR implements parsing and lowering for refinement types (like `{Int where _ >= 0}`) and parsing+lowering for output type annotations on function-defs (like `def add(x: Int, y: Int) => Int: ...`).

The refinement syntax was "decided, unimplemented" in the CHL spec, and the `=>` output type annotation was "tentative"; the docs have been updated. The function type expression syntax (`T => U`) remains "planned".


tests/programs/arithmetic/mod.rs
```diff
@@ -1,7 +1,7 @@
 //! Bind a variable, then use it in a second expression — smoke-test for
 //! sequencing and reference resolution.
 
-use super::common::expect_scalar;
+use super::common::{expect_compile_error, expect_scalar};
```
dpmills, line 5: revert the unused import

docs/chl-spec.md
```diff
@@ -1148,12 +1059,18 @@ def f(x: Int, y):
...
-inferred and bounded above by `T` (see
-[Two annotation forms: exact and bounded](#two-annotation-forms-exact-and-bounded)).
-The two forms may be mixed across a parameter list. Annotations on the
-function's *return type* are not yet supported.
+Annotations are arbitrary expressions evaluated in the surrounding
+scope; they refine the inferred parameter type. A return-type
+annotation, introduced by `=>`, refines the result type the same way:
+type inference checks that the body's value satisfies it, so a body
+whose inferred type conflicts with the annotation is a compile-time
+error. The annotation may itself be a refinement whose predicate refers
```
dpmills: Worth being explicit that return type annotations require the return to be a subtype of the annotated type, so they are analogous to `<:`, not `:` for a param

src/ccl/lower/stmts.rs
```diff
@@ -1256,11 +1077,32 @@ fn type_ctor_head(head: &Spanned<ChlExpr>) -> Result<&str, LoweringError> {
...
+/// [`LoweringContext::in_refinement_predicate`] set (save/restore, so a nested
+/// annotation cannot inherit it) so that a bare `_` resolves to the reserved
+/// refinement binder `__elem` rather than a variable named `_`. The result is a
+/// bare boolean [`Expr`] in which `__elem` is free — exactly the shape
+/// [`crate::ccl::Refinement`] stores.
+fn lower_refinement_predicate(
+    predicate: &Spanned<ChlExpr>,
+    ctx: &mut LoweringContext,
+) -> Result<Expr, LoweringError> {
+    let saved = ctx.in_refinement_predicate;
```
dpmills: Prefer a simple raii guard over explicitly setting the context value

tests/programs/refinement/mod.rs
```diff
@@ -0,0 +1,90 @@
+use super::common::{expect_compile_error, expect_scalar};
+
+#[test]
+fn refinement() {
```
dpmills, line 4: I don't we need this many example programs for refinements.  Having one to demonstrate the concept is fine, but the rest should be integration tests in tests/compilation_pipeline

dpmills (approved):
Nice! Just some minor things to fix before merging

src/ccl/lower/stmts.rs
octalsrc: ~What do you mean by a "rail guard" here?~

Fixed with helper method on `LoweringContext`.

src/ccl/lower/stmts.rs
```diff
@@ -1112,20 +1126,36 @@ pub(super) fn lower_type_expr(annotation: &Spanned<ChlExpr>) -> Result<Type, Low
...
+        // Refinement type `{T where p}` (`docs/chl-spec.md`, "6.4 Refinement
+        // syntax"): the base type refined by a `Bool` predicate over the
+        // anonymous subject `_`. The subject lowers to the reserved binder
+        // `__elem` (`REFINEMENT_BINDER`), so the stored predicate is a bare
+        // `Bool` term with `__elem` free — the same shape a singleton or a
+        // `groupby` refinement carries. Discharge of the predicate is the
+        // solver's structural refinement subsumption; this only builds the type.
+        ChlExpr::BraceRefinement { base, predicate } => {
+            let base_ty = lower_type_expr(base, ctx)?;
+            let pred = ctx.with_in_refinement_predicate(|ctx| lower_expr(predicate, ctx))?;
```
octalsrc, line 1153: I moved the set/restore logic to a helper method on `LoweringContext`.

PR #116, opened 2026-08-20
Apply a function slot's domain rule once its kind is resolved, not at each merge

dpmills:
`CompactFun::merge` chose how to combine two domains from the slot's `KindMerge` — union the alternatives for `Data`, take the contravariant meet otherwise — but a slot's kind is not settled until the last bound has merged, so the choice came from an answer that did not exist yet and a later bound could not undo it. Association therefore decided the outcome: an undetermined pair took the meet, `{a} ⊓ {b}` is `{a, b}`, and that deduplicated against a `Data` bound's own `{a, b}`, so a join of three collections over `{a, b}`, `{a}` and `{b}` typed as one collection over `{a, b}` in one association and was rejected in the other. The accepting association is the wrong answer: the kinds join to `Data`, all three domains *are* the data, and `union_domains` states the rule the meet broke — narrowing a data domain drops rows.

A positive join now accumulates the alternatives whatever the kind, and `coalesce_compact_go` applies the resolved kind's rule once: it meets the alternatives for a `Compute` reading and for a kind variable nothing pinned, and requires a single survivor for `Data`. The kind itself joins in the flat semilattice `Unknown < {Data, Compute} < Conflict`, which is the same operation at both polarities, so `merge`'s four kind branches collapse to one. A negative merge still meets its domains directly — that rule never read the kind.

`undetermined_kinds_join_without_deciding_the_domain_rule` pins the exhibit in every association, and `two_undetermined_kinds_join_to_a_capability_over_the_met_domain` pins the join that must keep working: two undetermined-kind bounds over `{a}` and `{b}` are a capability over `{a, b}`, which a rule that merely rejected an undetermined join would have refused. Instrumenting the arm showed no compiled program in the suite reaches a positive undetermined join, and no inferred type changes: the suite passes forward, under `DEEP_TYPECHECK=1`, and under `CAMBRA_REFINEMENT_ORDER=reverse`.

### DIFF
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -1310,16 +1310,21 @@ coalesce as alternatives and are reported there
 redundant, and Σ has to satisfy both.
 
 **Mechanics.** The decision is split across two phases, and the split is forced. At
-the compact merge, a positive `Data ⊔ Data` accumulates its domain **alternatives**
-(`CompactFun::domains`, unioned and deduplicated by `union_domains` — never met).
-It cannot decide there whether the alternatives are really two domains: a compact
-domain still carries inference-variable identity that `simplify_type` may merge
-afterwards, so two structurally identical domains can arrive as two alternatives.
-Coalesce materializes each alternative to a `Type` and deduplicates *again*, and
-that second comparison is the one that decides — one survivor is a plain data
-function, two or more is the rejection. A `Data ⊔ Compute` collision is a third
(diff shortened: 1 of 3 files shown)


dpmills:

- #108
    - #77
        - #78
            - #115
                - #116 ◀
                    - #117
                        - #118
                            - #109

sortalongo:
I suspect the rework we discussed yesterday will supersede this PR, so skipping it. Ping me if otherwise!

sortalongo (approved):
Nvm, just realized that I thought this was talking about the wrong kind of `Kind`

PR #117, opened 2026-08-20
Give a function slot's domain alternatives an order-insensitive set

dpmills:
`CompactFun.domains` was a `Vec` serving a set, so the merge was not commutative under `CompactType`'s own equality. `CompactFun` derives `PartialEq`, which compares a `Vec` positionally, while the union appended — so two data-function bounds over distinct domains merged to `[d₁, d₂]` or `[d₂, d₁]` depending on which arrived first, and the two results compared unequal. Structural equality is load-bearing where types are identities (the trivial-equality short-circuit, cache keys, `SpecKey`, the recorded-versus-recomputed walls), which is the same reason a refined type's refinements are a `RefinementSet` rather than a `Vec`: this is that defect, one slot over.

Nothing observed it yet — a `Data` slot with two alternatives is a `DomainJoinConflict` either way and a `Compute` slot's meet-fold normalizes — so no inferred type changes. `DomainSet` carries the alternatives with set-semantic `PartialEq` and a `union` that deduplicates, and `union_domains` becomes that method. `a_positive_join_of_two_domains_is_order_insensitive` pins it.

Deduplicating at compact time does not make coalesce's second dedup redundant, and one test said otherwise by construction: `coalesce_duplicate_data_domains_dedup_to_plain_fun` built two *structurally equal* alternatives, which `union_domains` could never produce and `DomainSet` now collapses on the way in. It builds the real shape instead — two alternatives differing in inference-variable identity, which `simplify_type` may merge away, so they are unequal at compact time and materialize to one `Type`. That is the case coalesce's dedup exists for.

`simplify_type` rebuilds the slot through `FromIterator`, which deduplicates: merging variables can make two previously-distinct alternatives equal, and collapsing them there is what keeps the invariant true downstream.

The negative merge's contravariant meet reads its single domain through the guard that establishes there is one, rather than through an accessor that answers `Option` and needs a fallback for the arm no caller reaches — a fallback that, as an empty `DomainSet`, would have been a function slot with no domain at all in a release build.

### DIFF
--- a/src/ccl/infer/solver/compact.rs
+++ b/src/ccl/infer/solver/compact.rs
@@ -9,6 +9,7 @@
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::rc::Rc;
 
+use itertools::Itertools;
 use smol_str::SmolStr;
 
 use crate::ccl::subst::{RefinementScope, Subst};
@@ -101,7 +102,7 @@ pub enum KindMerge {
     /// accumulated meet contravariantly when coalesce reads them.
     Compute,
     /// A collection: its domain *is* the data, so the alternatives never meet —
-    /// exactly one must survive (`union_domains`).
+    /// exactly one must survive ([`DomainSet`]).
     Data,
     /// A data/compute or multi-domain collision; deferred to coalesce.
     Conflict,
@@ -151,7 +152,7 @@ impl KindMerge {
 
 /// A merged function shape. `domains` holds the alternatives a positive join
 /// accumulated — one entry unless two distinct domains met — and coalesce
-/// applies the resolved [`KindMerge`]'s rule to them (see `union_domains` and
+/// applies the resolved [`KindMerge`]'s rule to them (see [`DomainSet`] and
 /// [`CompactFun::merge`]). A negative merge takes the contravariant meet
 /// directly, since that rule does not read the kind.
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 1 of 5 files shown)


dpmills:

- #108
    - #77
        - #78
            - #115
                - #116
                    - #117 ◀
                        - #118
                            - #109

src/ccl/infer/solver/simplify_type.rs
```diff
@@ -399,7 +401,7 @@ mod tests {
 
         let simplified = simplify_type(graph);
         let cf = simplified.term.fun.unwrap();
-        let dom_s = &cf.domains[0];
+        let dom_s = cf.domains.iter().next().expect("one domain");
```
sortalongo: `itertools::only()`

src/ccl/infer/solver/compact.rs
```diff
@@ -1294,10 +1358,12 @@ mod tests {
...
                 name: None,
                 kind,
-                domains: vec![CompactType {
+                domains: [CompactType {
                     rec: Some(rec),
                     ..CompactType::value()
-                }],
+                }]
+                .into_iter()
+                .collect(),
```
sortalongo: why not use `DomainSet::one`?

sortalongo (approved):

src/ccl/infer/solver/compact.rs
dpmills: Done — the helper was the one construction site not going through it. Now `DomainSet::one(...)` (`compact.rs`, `fun_over`).

src/ccl/infer/solver/simplify_type.rs
dpmills: Done

PR #121, opened 2026-08-21
Add a control port that replaces a running program with a new version of its source

dpmills:
Changing a running Cambra program's logic means restarting it, which discards every accumulator it holds and drops the sockets its clients are waiting on. This adds `--control` (default 8081), serving two endpoints against the running program: `/diff` reports how a new version of the source differs from it, at a pipeline phase the caller names, and changes nothing; `/reload` replaces the program in place — the listeners stay bound, every operator whose computation is unchanged keeps running with what it has accumulated, and every mutable variable the new version still declares is seeded with the value it was holding. So an edit to a guestbook's formatting leaves the entries already signed as they were and formats the next one the new way. The entry point is `LiveProgram::reload` (`src/live_program.rs`); the mechanism is [hot-reload.md](src/ccl/design/hot-reload.md) and what a reload means is [semantics.md, section 4](docs/operational-semantics/semantics.md#4-reload).

## What decides reuse

The compile that installs a version diffs its own planned tree against the running version's (`compile_replacement`) and records the `Content::Same` matches as a `Correspondence`. Conversion consults it at three places — every `Let` binding, every `Transact` store, and every writer's iteration input — and keeps the operator recorded at the corresponding node instead of building a second one. Those three are where a `FanOut` already sits, and a fan-out is the only thing that can carry a producer across a version, which is what bounds reuse to them ([What a reload does not do](src/ccl/design/hot-reload.md#what-a-reload-does-not-do)). Two further conditions: `reads_only_kept` declines a term with a free name among the bindings this compilation rebuilt ([Reuse is hereditary](src/ccl/design/hot-reload.md#reuse-is-hereditary)), and a site that needs the operator to still supply something declines one whose subscribers released it in full. Keeping a binding does not walk into it, so `keep_region` records what a kept subtree holds — without it a store under a kept binding leaves the handover on the first reload that keeps that binding.

## What carries state

State is keyed by `VarPath`: the author-written bindings enclosing the declaration, the variable's own spelling, then its index among the variables sharing both, assigned by one walk (`set_var_paths`) whose answers both the guard and conversion read. `live_state` reads each variable off its store's own cyclic fan-out at handover; a rebuilt store seeds from that and resumes one position past `FanOut::released_position` rather than at `0`. Value and position travel by different mechanisms: a value is an unsplittable tile that has to be carried, and a position is a guard on a tiling that can express one ([Rebuilding a store resumes it](src/ccl/design/hot-reload.md#rebuilding-a-store-resumes-it)).

## What is refused, and what is only reported

One guard, run on the planned tree before anything is torn down: a variable the new version no longer declares, declares at a different type, or would move between two declarations the source tells apart only by position ([The one guard](src/ccl/design/hot-reload.md#the-one-guard-a-version-must-be-able-to-take-over-the-state)). Nothing else — logic changes freely, endpoints come and go, and a variable may move between loops or into and out of a transaction. One outcome is silent in the same way and is reported rather than refused: a loop that begins above the beginning of what it reads, because the elements are gone and the source does not say whether the whole or "from here" was meant ([A variable that begins above its loop's input](src/ccl/design/hot-reload.md#a-variable-that-begins-above-its-loops-input)). `/diff` answers that report too, so an author reads it before the swap.

## Endpoints outlive a version

`SourceSinkRegistry` (`src/ccl/context.rs`) holds the open sources, each `http_serve` route, and the listener per bound port; every compilation seeds a `LoweringContext` from it, so a route already open is bound rather than reopened. A route the new version stops binding is retired — the address answers 404, requests already in flight are answered rather than left waiting (`answer_in_flight`), and the port is released once its last route goes. The `Planning` compile the guard runs on binds ports the version adds (`compile_to_opening`), so an unbindable address is a compile error raised while the program is whole; `/diff` runs `Endpoints::Inherited` and opens nothing.

## Making a graph swappable

Registrations that used to live for the process now live as long as their subscriber: fan-out slots, scheduler wake-ups and sink dispatch all hold a weak reference with the strong side in the producer that made it, and `FanOut::reopen` compacts dead slots and renumbers the survivors. `FanHold` splits a fan-out's readers into a downstream one that owns it and a recurrence sitting inside its own input chain that does not — the cycle that otherwise retains every retired version's subgraph. `ProducerReleases` names the per-producer release bookkeeping every source keeps, so a replacement's new producers start at the agreement rather than replaying the stream, and a retired producer's record is handed back on drop.

## One IR change to look at

A writer decision's `writes` is now a record keyed by the variable written (`writes.acc`) rather than a positional tuple, and reply taps are minted as `__to_<defer>_<n>` in the namespace user code cannot bind. Both are what let a second compilation match a store's slots back to declarations; the tap rename also closes a collision where a variable spelled like a tap had its read resolve to the tap. This touches every mutable-variable program, not only reloaded ones. A `stage` → `phase` rename accounts for nearly all of the change in `diff.rs`, much of it in `diffing.md`, and nineteen lines of `context.rs`, which defines `Phase`.

## Tests and docs

`tests/hot_reload.rs` is the suite: 69 cases over the induction, transactional and `stdin` domains, with the guard table and the fixture shapes in the module header. `tests/programs/hot_reload/` is the gallery entry (a guestbook and the version with its accumulating loop edited), and the HTTP glue moves from `tests/programs/common/mod.rs` to `tests/support/serving.rs` so both suites share it. New docs: [hot-reload.md](src/ccl/design/hot-reload.md) and semantics.md's Reload section; `README.md`, `docs/design.md` and `docs/demo-programs.md` pick up the feature.


src/ccl/design/live-update.md
```diff
@@ -0,0 +1,517 @@
...
+go on waking a consumer that no longer exists. So an operator is reusable only where a fan-out
+already sits, which is what decides the three places above: [What an update does not
+do](#what-an-update-does-not-do) says why they are not everywhere.
+
+Keeping an operator keeps the whole subgraph under it, including its stores and their accumulated
+values. Two conditions bound that: every binding the term reads must have been kept too ([Reuse is
+hereditary](#reuse-is-hereditary)), and the operator must not have been released in full by its
+subscribers, since it can then only answer empty (`OpConversionContext::keepable`). An iteration
+input is the exception to the second — released in full is what a finished collection looks like,
+and a fold that reached the end of its list is meant to read nothing further.
```
sortalongo: Hm... I don't see why iteration is an exception. If it's been released, it will still be empty, and we shouldn't be trying to read it. Whatever it would be driving should get its own iteration.

src/ccl/design/hot-reload.md
sortalongo, line 1: Something critically important is missing from this design: a tiling semantics. We need to be able to state unambiguously what it means to update a program, quantified by its current state. The current doc works in terms of operators, subscriptions, etc., which is useful and sufficient for the moment. But I think the next step is to specify the operational semantics of upgrades using tilings, along with integrity properties that must hold for the upgrade to be well defined.

src/ccl/design/hot-reload.md
```diff
@@ -0,0 +1,517 @@
...
+  shared-state ledger `./ci.sh shared_state` maintains: no value crosses between operators outside a
+  tile.
+- **The name.** State is keyed by a `VarPath`: the author-written bindings whose definitions enclose
+  the declaration, outermost first, then the variable's own spelling, then its index among the
+  variables sharing both. The spelling carries the meaning — a writer's write set is keyed by the
+  variable written, so the name survives from the source text to the store — and the chain is what
+  tells two declarations of one spelling apart. So a stateful loop added anywhere shifts nothing:
+  `a`.`n` is `a`.`n` however many other `n`s the version declares
+  (`a_loop_added_ahead_of_a_same_spelled_one_starts_at_its_init`,
+  `reordering_two_same_spelled_variables_keeps_their_state_apart`).
```
sortalongo, line 403: I expect we'll want to annotate variables that can cross versions like this. We don't want unintentional name collisions to lead to unexpected behavior. An early idea is something like

V0:
```
my_state := 0
transient := 0
for i in input():
  my_state += i
  transient += i
```

V1:
```
@Persisted
my_state
transient := 0
for i in input():
  my_state += i
  transient += i
```

Here, `transient` would restart its counter, and `my_state` would carry it over. The annotation makes people opt-in to that carry-over, which means they checked for name collisions. Otherwise, the variable starts fresh.

This also gives us a handle for dealing with functions that have internal mutable state, and are called from multiple places. They could take an identifier of some kind like:

```
def f(input_stream) requires CallerLabel:
  @Persisted(summon(CallerLabel))
  deferred my_stream

  for i in input_stream:
    my_stream << do_stuff(i)

  return my_stream
```

Then, an upgrade of `f` can correctly carry state across callers of `f`, even if those callers are re-arranged somehow.

sortalongo (approved):
Idea: we should call this feature "hot reload". People know what "hot reload" means, but it's associated with frontends and ephemeral state. In marketing materials, we can put "transactional" in front. "transactional hot reload" evokes a "ohh, that sounds cool!" reaction.

I think the rename is worth doing now before the name leaks into `main`.

src/ccl/design/hot-reload.md
dpmills: Added

src/ccl/design/hot-reload.md
dpmills: Let's discuss this more and do it as a follow-up based on what we decide.  Requiring annotations is pretty burdensome here

sortalongo:
*Written by Claude Code, from a review session with @sortalongo. The naming question is covered in his review above and is left out here. Reviewed against replacement-only scope — running two versions at once is out of scope by decision, so nothing below asks for it.*

[live-update.md](src/ccl/design/live-update.md)'s guard table is the most convincing part of this PR: measured rather than argued, and it answers the questions a reader actually arrives with. Two findings, one of which is invisible in the upgrade the design was tested against.

## A new store over a persistent source starts at the wrong place

The guard table says:

> A whole stateful loop is added under new names | Accepted; existing state untouched, and **the new store starts where its source has got to**

That is right for the storefront's V0→V1, and for a reason that upgrade happens to supply. V1 adds `is_promo_spent`, a view over the `orders` feed that persists across the swap, so the new store starts at the feed's current position and sees no prior orders — computing cumulative discount spend as zero. `v1.cambra` says why that is correct there: "Pre-sale orders carry discount=0, so no sale-start timestamp is needed — the ledger anchors itself." The sum over pre-`t_new` orders is zero either way.

That is a property of this upgrade rather than of the mechanism. V2 adds a day-keyed revenue mart over the same feed — a view whose value *does* depend on pre-upgrade history. Under this rule it starts at the current position and silently omits every order committed before the V1→V2 swap. No guard fires, nothing is refused, and the numbers are simply wrong.

So the rule as stated cannot distinguish a new store whose value is a function of history from one that only accumulates from here. Worth deciding which of these it is:

- A new store over a source with history replays that history, and "starts where its source has got to" applies only to a source with no retained history.
- The distinction is real but unbuilt, and the row should say a view over a persistent source is the case it does not handle, so a reader building one knows.

The second is a fine answer for this PR; what I would not want is the row reading as settled, because the shape it gets wrong is the next thing the demo does.

## The tests are 96% of one gallery entry

Test additions here:

| File | Lines |
| --- | --- |
| `tests/programs/live_update/mod.rs` | 3,309 |
| `tests/http_server.rs` | 56 |
| `tests/programs/common/mod.rs` | 30 |
| `tests/compilation_pipeline/feeds_cases.rs` | 23 |
| `program.cambra` + `updated.cambra` | 22 |
| `tests/compilation_pipeline/transactions.rs` | +2 / −8 |

`mod.rs` holds 58 tests, 77 helper functions, and 41 embedded programs across seven named bases — larger than every source file in the PR, and three times the biggest one.

The volume is largely justified. Each case needs a running program, a state mutation, a swap, and an assertion that state carried, and the guard table has around twenty rows to cover across the induction, transactional, and `stdin` domains. What does not follow is the location. `docs/demo-programs.md` defines a gallery entry as "a `program.cambra` source plus a `mod.rs` with one or more `#[test]` functions", and the module doc here agrees where the entry stops:

> The gallery entry is one program and the version that replaces it… That pair is what a reader should look at to see what an update *is*. Everything else the cases drive is scaffolding.

So the entry is the 22 lines of `.cambra`, and the other 3,309 are a feature suite. They are also not `compilation_pipeline` tests — they drive `LiveProgram::start`, `drive_main_to_terminal` and `launch_under_control` rather than compiling and evaluating — so the slot the layout already has for them is a top-level suite beside `tests/compilation_pipeline.rs` and `tests/http_server.rs`. Given the rename, `tests/hot_reload.rs` plus `tests/hot_reload/`.

Three smaller things inside it:

- **77 helpers against 58 tests** is more harness than cases. If the infrastructure is that substantial it wants to be named and shared rather than living in a gallery entry, and the `test-helpers` feature is where that goes.
- **The seven inline bases get none of the gallery's benefits.** A gallery program runs by hand, opens under `--inspect-only`, and is walked by ratchet 5 of `tests/inspector_goldens.rs` through the wire validator. An `indoc!` string gets none of that. For the *variants*, which differ from a base by the single edit their case is about, inline is clearly right and the doc's justification holds. For the bases, they are real programs and it is worth saying why they are not files.
- **No row in `docs/demo-programs.md`.** `mod live_update;` is registered in `tests/programs/main.rs`, so the entry exists without its table row, and the policy in that doc is explicit: "When you add a program under `tests/programs/`, add a row here." The row is also where a reader would learn the entry demonstrates this feature at all.

The `indoc!` discipline across all 41 embedded programs is the right call and easy to let slip at that volume.

## A generated name colliding with a user identifier, twice, fixed twice

The `feeds_cases.rs` addition pins a real bug, and a nasty one:

> a tap was named `to_<defer>_<n>`, which is a spelling user code can write, and the tap and the accumulator share one decision record — so the record's last writer won and the accumulator's read resolved to the reply's value (`300` here, in release; an internal panic in debug)

A wrong answer in release and a panic in debug, from a user writing an ordinary identifier. Good find, and the fix — minting taps into the `__` namespace user code cannot bind, via `Name::defer_tap_field` — is the right general answer.

It is the second bug of this exact class in the stack. #159 fixed a store key that was `Value::String(field_key)`, so a keyed map's data key could collide with a mutable variable's name — "the string `"pool"` would alias the scalar register spelled `pool`" — and fixed it with a tagged sum instead. Same class, two mechanisms, both failing silently or near-silently.

Two point fixes suggest nobody has swept the places a generated name is minted into a space user code can write. Worth doing once, since the failure mode is a wrong value rather than an error.

## Noted, not asked for

Concurrent load is out of scope by decision, so "Behaviour under concurrent load is not measured" needs nothing here.

With replacement as the decided scope, this closes the storefront's upgrade steps — which means `docs/demo-programs.md` and `tests/programs/storefront/mod.rs` still describe the version dimension as open design waiting on a `Versioned` surface, and that is now only true of shadowing. Someone owes those two a correction; flagging it rather than assuming it is yours.

## Scope

This pass covered [live-update.md](src/ccl/design/live-update.md), the guard and its table, the test layout, and the two `compilation_pipeline` changes. It did not cover `operator_conversion.rs` (1,105 added lines and the real change here), `fanout.rs`, `control_port.rs`, or `live_program.rs`.

src/ccl/design/live-update.md
dpmills: You're right that nothing about iteration is exceptional, and chasing it turned up a bug.

The flag is gone. `keepable(term, released_in_full_is_fine: bool)` was one function serving two callers, with the boolean constant per call site — which is exactly what made one of them read as an exception. There is now one lookup, `correspondent`, and each site asks its own question of it. A binding wants an operator its readers can pull, so a spent one is rebuilt. A recurrence continuing over an input wants the progress that operator recorded, and `FanOut::released_position` is a watermark that only rises — so reading it off a spent operator is reading a number, not reading a tile.

"Whatever it would be driving should get its own iteration" is now what happens, with one condition. A recurrence carrying no value starts at the value it declares, so it wants its input from the beginning, and it gets a fresh iteration wherever that input can be built again. Where the input reads a data source it cannot be — a source hands a new producer only what its retired producers had not released — so the recurrence takes the kept iteration and the reload reports that the variable begins above its input.

The decline you were suspicious of was doing damage, in a way neither of us named. A binding rebuilt because its operator was spent was recorded as *rebuilt*, which means "this version changed what it computes" and makes `reads_only_kept` refuse to keep any operator reading it. So every recurrence over that binding lost its own iteration, restarted at `0`, and folded its whole input on top of the value it was carrying. A reload that changed nothing about `n` turned `"abc"` into `"abcabc"` — silent, and pinned now by `a_loop_added_over_a_folded_collection_reads_it_whole`.

dpmills:
All three land. Taking them in order, with a correction to my own first answer.

## A new store over a persistent source

I took your second option, then found it did not go far enough — and that the way I first drew the distinction was wrong.

"A source with retained history" versus one without is not a distinction the mechanism makes. `StreamBuffer::first_index_for_a_new_producer` is `released_prefix(…) + 1`, and `FanOut::released_position() + 1` is the same number: both say the readers that finished with those elements released them. I had written the row as though a source's prefix were withdrawn where a collection's is released, and the release path is the same one in both cases. Worse, the accept line I built on it fell on whether a *correspondent* happened to exist at the iteration source — true only if the previous version had a store folding that term, which is bookkeeping rather than semantics.

What is there instead: every variable that begins above the beginning of its loop's input is **reported**, in the reload's report and in `/diff` before the swap. Not refused, because refusing would refuse an accumulator added to a live endpoint along with a view over a retained feed, and those are the same term — what separates them is what the author meant, which the source cannot currently say. The declaration is where that belongs, and it is filed alongside the `@Persisted` thread rather than being a second surface.

The row now says a stateful loop added under new names folds its input whole where this version can build that input again, and from where the input starts where it cannot.

**Your V2 mart is still not covered**, and I would rather say so than let the row imply otherwise. It reaches `orders` through a `__hist.k` dense read on a kept store rather than through an iteration input, so it is a different site from the one this fixes. The mechanism underneath is the same — `FanOut::subscribe` hands a new subscriber the release its predecessors made — but closing it is its own change, and V2 cannot be written today regardless.

Two things fell out of chasing this that were in neither finding:

- A binding rebuilt because its operator was spent was recorded as a *change* to what it computes, which makes `reads_only_kept` refuse to keep any operator reading it. So every recurrence over that binding lost its own iteration, restarted at `0`, and folded its whole input on top of the value it was carrying. A reload that changed nothing about a variable silently doubled it.
- The report was per-store when the situation is per-variable, and empty for commit stores. One loop drives one position sequence, so a variable added to a loop whose other variables carry begins where that loop resumes — the most ordinary shape this has, and it was reporting nothing. Both found by instrumenting `reload` and listing every case in the suite that reports, rather than by reasoning about it.

## The tests

Moved. `tests/hot_reload.rs` plus `tests/hot_reload/{harness,cases}.rs`; the gallery keeps the two `.cambra` files and the one case they demonstrate, and now has its row in `docs/demo-programs.md`.

The relocation forced a second split. What both binaries need went to `tests/support/serving.rs` — the sink compile, the scheduler pump, and the `TcpStream` client — because `common` compiled into the hot-reload binary reports its gallery-only helpers as dead. `exchange` was in the hot-reload module and is the collapsed form of a shape `http_greeter`, `http_counter` and `http_accumulator` each hand-roll; it moved there too and those three now use it.

Not done, each its own change: promoting the seven inline bases to files, and the harness behind `test-helpers`.

One thing your count understates: five gallery entries besides this one have no `demo-programs.md` row — `http_accumulator`, `http_counter`, `join_then_groupby`, `refinement`, `source_shared`. I added the row for this entry and left those, rather than inventing five descriptions inside this diff.

## The generated-name sweep

Agreed, and not done here. Two point fixes for one class is the signal, and the failure mode being a wrong value rather than an error is why it wants doing once rather than a third time. Filed.

## The versioning docs

Corrected in both `docs/demo-programs.md` and `tests/programs/storefront/mod.rs`: replacing one version with another is implemented, and what stays open is serving two at once.

## Scope

Worth re-flagging. `operator_conversion.rs` has moved substantially since you read it — the rebase onto #162, the `NodeId` work that came with it, and everything above. If there is a second pass, that file and `live_program.rs` are where it is.

PR #123, opened 2026-08-22
Carry a suffix morphism out of exponential eta as `map`

dpmills:
`[sum([s.amount for s in g]) for g in groupby(sales, \r -> r.region)]` did not compile. λ-elimination closes the inner lambda over both the group and the element and re-splits it with `curry`, and operator conversion has no `curry` arm. `simplify`'s exponential-eta rule discharged `curry(⟨.1, .0 ≫ 𝑓⟩ ≫ apply)` but not the projection sitting after `apply`. It now carries that suffix out as `map(𝑔)`, and `groupby_rollup` moves from `expect_compile_error` to `expect_scalar`.

## The rule

`curry(⟨.1, .0 ≫ 𝑓⟩ ≫ apply ≫ 𝑔) ⟹ 𝑓 ≫ map(𝑔)`. Both morphisms are optional — a bare `.0` arm is `𝑓 = id`, an empty suffix `𝑔 = id` — so the former rule and `curry(⟨.1, .0⟩ ≫ apply) ⟹ id` are its degenerate cases. Suffix and curry go in one rewrite, which keeps the standalone split `curry(𝑓 ≫ 𝑔) ⟹ curry(𝑓) ≫ map(𝑔)` out of the rule set: a residual `curry` in a zip arm is what `try_exponential_beta` matches on, and adding the split unguarded fails 20+ join and group-by cases with the same undischarged `curry`.

## Rebuilt function types carry their kind

`Type::fun` mints `FunKind::Compute` and no Pi binder, so rebuilding the curry's arrow from its parts reads a collection as a compute function and unbinds a dependent refinement's key. With 𝑓 absent the `map` node reuses the curry's arrow as it stands; with 𝑓 present it takes 𝑓's through `Type::fun_like_or_hole`.

`fun_ty_or_hole` named neither case and is retired. Fifteen of its sites build a built-in's function-slot type, where `Compute` is the answer rather than a default, and take `Type::compute_fun_or_hole`; the sixteenth rebuilds an eliminated lambda's own type and had been post-patching the kind and binder back onto the result, and now builds the arrow once.


dpmills:

- #123 ◀
    - #127

docs/demo-programs.md
```diff
@@ -83,7 +84,7 @@ plan and the full dependency map.
...
 | [generator_pipeline](../tests/programs/generator_pipeline/) | Compose two generators (filter then square), then `max` | `def` + `yield`, generator composition, aggregate | ✅ working | Demonstrates UDF call sites being inlined and fused through to operator conversion. |
-| [groupby_rollup](../tests/programs/groupby_rollup/) | Group sales records by region and total per region | `groupby` over records, `lambda`, inner projection, `sum` | 🚧 blocked | Operator conversion panics on the inner-projecting comprehension (`curry` combinator unsupported). See [known issues](#known-issues-surfaced-by-these-programs). |
+| [groupby_rollup](../tests/programs/groupby_rollup/) | Group sales records by region and total per region | `groupby` over records, `lambda`, inner projection, `sum` | ✅ working | Returns `Function [ "east" -> 200, "south" -> 75, "west" -> 300 ]`. The inner projection reaches operator conversion through the exponential-eta rule in `src/ccl/simplify.rs`: λ-elimination closes `s.amount` over both binders and re-splits it with `curry`, which operator conversion has no arm for. |
```
sortalongo: No need to document exactly why it works. Just leave it blank if there's nothing to say.

sortalongo (approved):

docs/demo-programs.md
dpmills: Done

PR #126, opened 2026-08-22
TEMPORARY: vendor #93 so the live-update work sits on its structures

dpmills:
**Drop this commit when rebasing after #93 merges.** It is a copy of that PR's single commit, carried here so the work above it can be written against the structures it introduces (`InductionDrive` / `TransactDrive` / `DriveWindow`) rather than against the `WriterBuffer` it deletes.

Nothing here is for review — review it as [#93](https://github.com/cambra-dev/cambra/pull/93).


dpmills:
Closing: this was created in error. The vendored #93 commit belongs inside #121, which is where it now sits.

PR #127, opened 2026-08-22
Filter a partition's inner collections with a `MapFilter` operator

dpmills:
A filter inside a per-group comprehension — `[sum([s.amount for s in g if s.qty > 2]) for g in groupby(sales, \r -> r.region)]`, the `WHERE … GROUP BY` shape — refines the *inner* collection's domain, so the surviving elements differ per key. `Filter` and `Restrict` each narrow one domain and reject anything but a `SealedFunction`, so nothing could express it and lambda elimination asserted the program away. `MapFilter` is that operator, and planning emits it.

## The operator

`MapFilter` takes a `CurriedFunction(K → I → V)` and a `Bool`-codomain predicate over the same keys and inner domain. The predicate's flattened codomain is the mask over the CSR rows directly, so `domain1` and the per-key structure are untouched and `Tile::retain` rebuilds `offsets`. Both operands read one upstream through a `FanOut`, and they release together, the same coupling `Filter` has.

## Getting the refinement there

Lambda elimination previously asserted on a cast-wrapped lambda whose value mentions the outer binder — the shape an inner comprehension over `g` produces. It now moves the refinement onto the inner lambda's binder, since `cast(λ 𝑦 : 𝐷 → body, {𝐷 | 𝑝} ⤇ 𝑉)` and `λ 𝑦 : {𝐷 | 𝑝} → body` are the same function, and hands the result to the nested-lambda rule. Re-targeting the cast onto the eliminated morphism instead does not typecheck: it narrows a data function's domain inside a codomain. The binder keeps the refinements it already carried — the partition's own domain — since the cast target names only the narrowing it imposes and a refinement set is a conjunction.

That makes an inner binder's type mention the outer binder for the first time, so the nested-lambda rule now substitutes into it when it builds the pair type. Without that the `Proj` rule compares a rewritten codomain against an unrewritten tuple component. The projection is then *dependent*, so `emit_proj` opens its codomain before comparing it against the domain it was read off — the reading `compose_chain` already takes on a chain adjacency. `DEEP_TYPECHECK=1 ./ci.sh test` is what checks any of this: it alone typechecks inside predicates, per operation.

## Planning

`insert_map_filters` matches a morphism `(𝑔: {𝐷 | 𝑟} ⤇ 𝑊) ⇒ ({𝐷 | 𝑟, 𝑝(𝑔)} ⤇ 𝑉)` and materializes the **difference** between the two refinement sets: a set has no outermost layer to read, and the site's domain already carries whatever upstream materialized (the group predicate, here). Each added predicate reads `__elem ▷ (𝑔 ≫ 𝑞)`, so filtering the collection 𝑔 denotes leaves 𝑞 as the value predicate, and a set narrows by the conjunction of its members, so a whole difference becomes one `map_filter(𝑞₁ and … and 𝑞ₙ)` rather than a chain. `wrap_with_iterate` reaches none of it: it materializes refinements on a node's own domain, and this one is a codomain in.

The rule and the coordinate it reads the binder in are in [`map_filter.rs`](src/ccl/planning/map_filter.rs). The pass runs after predicate compilation, which is what puts the predicates in point-free form.

## The conversion walk counts a rewrite it applied from its memo

`PiWalk` (`src/ccl/subst.rs`) applied a conversion on a memo hit without counting it, and an enclosing predicate reads that counter to decide whether the clone it just walked was touched. A predicate whose only conversions are nested therefore kept its unconverted term, and a type came back from `open_pi_binder` still spelling references as indices. Which of walking and hitting the memo converts a nested occurrence depends on the order the set holds its members, so `CAMBRA_REFINEMENT_ORDER=reverse` is the configuration that reaches it. `a_nested_conversion_applied_from_the_memo_is_kept` builds both orders.

`groupby_filtered_rollup` is the new gallery program.


dpmills:

- #123
    - #127 ◀

sortalongo (approved):

PR #129, opened 2026-08-25
Accept an `if`/`match` block as an assignment's value, and add the one-line `match`

dpmills:
CHL makes `if` and `match` value-yielding by position — a block's value is its last statement's — but gave that value no spelling where one was required. `x = if c: … else: …` did not parse, and a `match` inside a refinement predicate had no way to delimit its arms, because `{…}` suppresses the layout tokens the indented form depends on. Both are expressions now, through one AST node: `Expr::Block` wraps the `Stmt::If` or `Stmt::Match` already built. Nothing new reaches the IR — every spelling routes through `lower_final_stmt` to the same `Case` the ternary and the tail-position statement forms produce.

## The bracket closes the arm list

A one-line `match` is reachable only from `bracketed_expr`, so a `)`, `]` or `}` always terminates its arms. An arm body is a plain `expr`, which cannot derive a one-line `match`, so a nested one carries its own bracket and two arm lists never compete for the same `case`. Restricting the arm body instead does not hold — a lambda body, a ternary else-branch, an `<<` right operand and a `=>` codomain each take a whole expression. [The one-line form](docs/chl-spec.md#the-one-line-form) records the rule.

## Two productions in place of two would-be copies

`match_arms` is the arm grammar over a parser for how a body is spelled, and `assign_tail` is the six assignment operators over a parser for the right-hand side — which is why all six take a block rather than only `=`. Both are described in `src/chl_parser/design-chl-parser.md`.

## Scope on a block right-hand side

`lower_assigned_value` carries the enclosing scope in: a `for` inside one of the branches needs the names bound above the assignment to tell a loop-carried accumulator from a branch-local. Three positions take the scope-free `lower_expr` path instead — the one-line `match`, and a block right-hand side in a loop body or a `with begin():` block. None admits a `for` in a branch, so the dropped scope is unobservable.

`lower_aug_binop` and `lower_define` now take an already-lowered value, so each caller picks the path.

## Coverage

`tests/compilation_pipeline/` runs the block `if`, both `match` spellings, and a nested one-line `match` end to end; the parser and lowering tests pin the bracket rule and the six operators. A refinement predicate stops at lowering, because no refinement annotation discharges through the pipeline yet.

Spec: [§4.3](docs/chl-spec.md#43-assignment-forms) grammar and table, §4.5, §4.10, and the `[Open]` in [§6.4](docs/chl-spec.md#64-refinement-syntax), now closed.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -1313,11 +1313,11 @@ that block's value — its last statement's, by the rule in §4.5:
 
 ```python
 label = if score > 90:
-    "high"
-elif score > 50:
-    "mid"
-else:
-    "low"
+      "high"
+  elif score > 50:
+      "mid"
+  else:
+      "low"
 
 n = match msg:
     case `ping(seq):
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -258,6 +258,7 @@ Python's `ast.Module`; the name is historical.)
 
 ```ebnf
 statement       ::= simple_stmt NEWLINE
+                 |  block_assign_stmt
                  |  compound_stmt
 
 simple_stmt     ::= return_stmt
@@ -279,17 +280,32 @@ aug_assign_stmt ::= assign_target aug_op expression
 define_stmt     ::= assign_target "<<=" expression
 expr_stmt       ::= expression
 
+-- The same six assignment forms with a block statement on the right (§4.3).
+-- The block's DEDENT ends the statement, so no NEWLINE follows it.
+block_assign_stmt ::= assign_target "=" block_value
+                   |  assign_target ":" expression "=" block_value
+                   |  assign_target [ ":" expression ] ":=" block_value
+                   |  assign_target aug_op block_value
+                   |  assign_target "<<=" block_value
+
+block_value     ::= if_stmt | match_stmt
+
 aug_op          ::= "+=" | "-=" | "*=" | "//="
 
 assign_target   ::= ident
                  |  "(" assign_target ( "," assign_target )* [ "," ] ")"
                  |  assign_target ( "," assign_target )+ [ "," ]
 
-compound_stmt   ::= if_stmt | for_stmt | with_stmt | def_stmt
+compound_stmt   ::= if_stmt | match_stmt | for_stmt | with_stmt | def_stmt
 
 if_stmt         ::= "if" expression ":" block
                     ( "elif" expression ":" block )*
                     [ "else" ":" block ]
+
+match_stmt      ::= "match" expression ":" NEWLINE INDENT case_arm+ DEDENT
+case_arm        ::= "case" case_pattern ":" block
+case_pattern    ::= "`" ident [ "(" ( ident | "_" ) ")" ]
+                 |  "_"
 
 for_stmt        ::= "for" assign_target "in" expression ":" block
 
@@ -360,6 +376,15 @@ expression ::= lambda_expr | yield_expr | feed_expr
              | ternary | bool_or | bool_and | bool_not
              | comparison | log_or | log_xor | log_and | collection_union
              | sum_expr | product | unary | postfix | atom
+
+-- Every position a bracket encloses: a list or tuple element, a call argument,
+-- a subscript index, a record field, a brace item, a refinement predicate, a
+-- comprehension clause. The one-line `match` (§4.10) is legal only here, so the
+-- enclosing `)`, `]` or `}` always closes its arm list. The bracketed atom forms
+-- in §2.4 recurse through this production rather than through `expression`.
+bracketed_expression ::= oneline_match | expression
+
+oneline_match ::= "match" expression ":" ( "case" case_pattern ":" expression )+
 ```
 
 ### 2.4 Atoms
@@ -1282,6 +1307,30 @@ is supported at any nesting depth.
 (the mutation `:=` is a statement, §1.8). **No multi-target chained
 assignment** (`a = b = c` is not in the grammar).
 
+**A block statement may sit on the right.** Every form in the table above
+accepts an `if` chain or a `match` where it accepts an expression, and binds
+that block's value — its last statement's, by the rule in §4.5:
+
+```python
+label = if score > 90:
+    "high"
+elif score > 50:
+    "mid"
+else:
+    "low"
+
+n = match msg:
+    case `ping(seq):
+        seq
+    case `close:
+        0
+```
+
+The block's DEDENT ends the statement, so nothing follows it on the line. Every
+branch must produce a value, and an `if` with no `else` is rejected (§4.5); a
+one-line spelling of the same thing is the ternary (§3.6) or, for `match`, the
+bracketed form in §4.10.
+
 Annotated assignment **requires** a value (`x: T` alone is a parse error,
 unlike Python's bare type-only declarations).
 
@@ -1417,8 +1466,9 @@ else:
     block_else
 ```
 
-`if`/`elif`/`else` is a *statement*; for an expression form use the
-ternary (§3.6).
+`if`/`elif`/`else` is a statement. It is value-yielding by position: as the
+last statement of a block, and on the right of an assignment (§4.3). The
+one-line spelling of the same choice is the ternary (§3.6).
 
 The branches are tried in source-text priority: the value of the
 statement is the value of the block under the first guard that holds,
@@ -1432,9 +1482,9 @@ non-winning blocks contribute no effects.
 
 Each branch's block is itself a statement block. When the `if` chain
 occurs in a position that requires a value (function body, program
-value), every branch — including `else` — must end in a value-yielding
-statement, and the missing-`else` case is rejected as "if used as an
-expression, all branches must produce a value."
+value, an assignment's right-hand side), every branch — including `else` —
+must end in a value-yielding statement, and the missing-`else` case is
+rejected as "if used as an expression, all branches must produce a value."
 
 ### 4.6 `for` — iteration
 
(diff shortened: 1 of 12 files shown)


dpmills:

- #129 ◀

docs/chl-spec.md
```diff
@@ -1282,6 +1307,30 @@ is supported at any nesting depth.
...
+accepts an `if` chain or a `match` where it accepts an expression, and binds
+that block's value — its last statement's, by the rule in §4.5:
+
+```python
+label = if score > 90:
+    "high"
+elif score > 50:
+    "mid"
+else:
+    "low"
```
```suggestion
      "high"
  elif score > 50:
      "mid"
  else:
      "low"
```

sortalongo (approved):

src/ccl/lower/mod.rs
```diff
@@ -933,6 +933,23 @@ fn lower_expr_inner(
...
+        // `for`: a one-line arm body is a single expression, and a `for` nested
+        // in either enclosing block is rejected already. The statement path
+        // through `lower_assigned_value` carries the scope where a `for` can
+        // appear.
+        ChlExpr::Block(stmt) => {
+            debug_assert!(
+                matches!(stmt.node, ChlStmt::If { .. } | ChlStmt::Match { .. }),
+                "Expr::Block wraps an `if` or a `match`; the parser builds no other block value"
+            );
+            lower_final_stmt(stmt, &[], &HashSet::new(), ctx)
```
groundlar, line 951: **Critical — an accumulator write inside a block right-hand side is silently discarded.**

Verified on merged `main` (`86960fe`); every file this PR touched is byte-identical between the PR head and that commit, so this is live on main.

The `&[]` / `HashSet::new()` here is one half. The other half is that `collect_mutation_loop_vars`, `find_nested_mutation_var`, and `first_outer_plain_assign` (`src/ccl/lower/loops.rs`) each walk `Spanned<ChlStmt>` lists and descend into `ChlStmt::If` branches only. A `+=` inside `ChlExpr::Block(If)` is invisible to all three: the loop is never classified as a mutation loop, `acc_names` never gains the name, and the `MutWrite` that `lower_middle_stmt` still emits is claimed by no recurrence.

The two programs below differ only in where the `if` sits.

```python
acc := 0
for i in [1, 2, 3]:
    if i > 1:
        acc += i
    yield 0
acc          # 5 — correct
```

```python
acc := 0
for i in [1, 2, 3]:
    t = if i > 1:
        acc += i
        0
    else:
        0
    yield t
acc          # 0 — the update is silently discarded
```

The block `match` right-hand side does the same. No error, no warning. This is the failure mode `in_loop_aug_assign_error` exists to reject — "the update would be discarded at the iteration boundary, and `op=` reads the old value" — reached through a path that bypasses it.

Regression test for `tests/compilation_pipeline/mutability.rs`. The second case currently yields `Value::Int(0)`:

```rust
/// A `+=` inside a block right-hand side's branch is the same loop-carried
/// accumulator write the statement `if` spells, so the two agree. The
/// accumulator scan walks statements, so a write inside `ChlExpr::Block`
/// reaches it only if the walk descends into an assignment's value.
#[rstest]
#[timeout(Duration::from_secs(10))]
#[case(indoc! {"
    acc := 0
    for i in [1, 2, 3]:
        if i > 1:
            acc += i
        yield 0
    acc"})]
#[case(indoc! {"
    acc := 0
    for i in [1, 2, 3]:
        t = if i > 1:
            acc += i
            0
        else:
            0
        yield t
    acc"})]
fn test_block_right_hand_side_accumulates(#[case] code: &str) {
    check_scalar(code, Value::Int(5));
}
```

Minimal fix: descend into a `ChlExpr::Block` value in the three walkers. Structural fix: make `lower_assigned_value` the only entry point for a block value, thread the enclosing context into it, and make this arm an error rather than a silent fallback to the general lowerer.

Two neighbouring shapes reach internal assertions instead — the same program with a trailing read rather than a `yield` panics at `src/ccl/mut_elim.rs:1428`. The statement-`if` analogue panics identically on the parent commit, so that panic predates this PR; only the spellings reaching it are new.

src/ccl/design/lowering.md
```diff
@@ -140,6 +140,27 @@ The inference pass types `Defer`/`Feed`/`Define` directly — `Defer : Feed(Chan
...
+right-hand side. The names bound above the assignment are what a `for` inside
+one of the block's branches consults to tell a loop-carried accumulator from a
+branch-local (`find_mutation_loop_vars`), and `lower_expr` has no scope to pass.
+
+Three positions reach the block through `lower_expr` and so drop that scope: a
+one-line `match`, and a block right-hand side inside a loop body or a `with
+begin():` block, each of which lowers an assignment's value itself. None admits
+a `for` in the branch — a one-line arm body is a single expression, and a
+nested `for` is rejected in both enclosing blocks — so the dropped scope is
+unobservable.
```
groundlar, line 162: **Important — this claim does not hold. The dropped scope is observable, as a diagnostic that contradicts the source.**

The nested `for` is rejected, but the scope loss changes how. With `&[]` passed as `preceding`, the inner loop's `acc_names` is computed against an empty mutation scope, so `in_loop_aug_assign_error` fires:

```python
balance := 0
with begin():
    balance := if True:
        for i in [1, 2]:
            balance += i
        balance
    else:
        0
balance
```

> `` `balance` is not a mutable variable, so `balance op= …` inside a for-loop body cannot accumulate into it: declare it before the loop with `balance := …` ``

It is declared before the loop, on line 1. The loop-body position gives the same message for `acc`. Both reproduce on merged `main` (`86960fe`).

Either thread the scope through those two positions as well, or reject `ChlExpr::Block` there explicitly with a message that names the real restriction.

src/chl_parser/parser.rs
```diff
@@ -1438,6 +1365,196 @@ enum AssignTail {
 
 /// Convert an [`Expr`] parsed in target position into an [`AssignTarget`].
 ///
+/// What follows an assignment target, over parsers for the right-hand side and
```
groundlar, line 1368: **These functions were inserted into the middle of `expr_to_assign_target`'s doc comment.**

`assign_tail`'s rustdoc now opens with line 1366 — `/// Convert an [`Expr`] parsed in target position into an [`AssignTarget`].` — which is the summary line the rustdoc index shows. `expr_to_assign_target` lost it, so its docs now lead with "CHL binding patterns are bare names and…". Nothing lints this. Moving lines 1366–1367 back down to `expr_to_assign_target` fixes both.

src/chl_parser/parser.rs
```diff
@@ -1270,87 +1253,30 @@ where
...
+        // `Dedent` ends the statement, so this form takes no `stmt_terminator`
+        // and cannot ride `simple_stmt`. It is tried before `simple_stmt`, which
+        // reparses the target when the right-hand side turns out to be an
+        // ordinary expression.
+        let block_value =
+            choice((if_stmt.clone(), match_stmt.clone())).map(|stmt: Spanned<Stmt>| {
+                let span = stmt.span;
+                Spanned::new(span, Expr::Block(Box::new(stmt)))
             });
+        let block_assign = bare_tuple
```
groundlar, line 1277: **Nit — this re-parses `bare_tuple` for every statement in the module.**

`block_assign` is tried before `simple_stmt`, and `.clone()` on a chumsky combinator duplicates the parser without memoizing, so the LHS — a full `expr` chain — is parsed twice for every assignment and every expression statement whose right-hand side is not a block. Error quality is unaffected (the furthest-progress alternative wins). Worth a line in the comment above recording the cost, or a measurement, so the next reader doesn't have to rediscover it.

src/chl_parser/design-chl-parser.md
```diff
@@ -96,7 +96,16 @@ highest precedence:
 13. `*`, `//`
 14. unary `-`
 15. postfix: call `f(…)`, subscript `x[…]`, attribute `x.name` / `x.0`
-16. atom: literal, name, parenthesised, list, dict/record, comprehension
+16. atom: literal, name, parenthesised, list, record, brace type, comprehension
```
groundlar, line 99: **Nit — the rustdoc counterpart of this line still says "dict/record".**

`src/chl_parser/parser.rs:164` reads `/// 16. atom: literal, name, parenthesised, list, dict/record, comprehension`. `Expr::Dict` no longer exists, so dropping it here is right and the parallel line should follow. Two test comments (`parser.rs:1682` and `:1841`) also still mention a dict.

groundlar:
Post-merge review. The parser work here is solid — `assign_tail`, `match_arms`, and `annotation_mode` collapse shapes that were already duplicated, the "bracket closes the arm list" rule is a real invariant argued from the grammar, and the spec edits close the §6.4 `[Open]` rather than deferring it. The findings below are all on the lowering side.

Verified against merged `main` (`86960fe`). Every file this PR touched is byte-identical between the PR head and that commit, so all of these are live on main.

**Critical**

- [An accumulator write inside a block right-hand side is silently discarded](https://github.com/cambra-dev/cambra/pull/129#discussion_r3876646990) — `src/ccl/lower/mod.rs`. Two spellings this PR describes as equivalent produce `5` and `0`. `collect_mutation_loop_vars`, `find_nested_mutation_var`, and `first_outer_plain_assign` all walk statement lists and never look inside `ChlExpr::Block`. Includes a ready-to-paste regression test.

**Important**

- ["the dropped scope is unobservable" does not hold](https://github.com/cambra-dev/cambra/pull/129#discussion_r3876647079) — `src/ccl/design/lowering.md`. It is observable, as a diagnostic telling the author to declare a variable that line 1 already declares.
- [`assign_tail` carries `expr_to_assign_target`'s summary line](https://github.com/cambra-dev/cambra/pull/129#discussion_r3876647152) — `src/chl_parser/parser.rs`. The new functions were spliced into the middle of an existing doc comment.

**Nits**

- [`block_assign` re-parses the LHS for every statement](https://github.com/cambra-dev/cambra/pull/129#discussion_r3876647212) — `src/chl_parser/parser.rs`.
- [The rustdoc counterpart of the atom list still says "dict/record"](https://github.com/cambra-dev/cambra/pull/129#discussion_r3876647281) — `src/chl_parser/design-chl-parser.md`.

The underlying shape behind the Critical is that `ChlExpr::Block` lets statements appear in expression position, and `lower_expr`'s `Block` arm routes them to `lower_final_stmt` — the general, top-level statement lowerer — regardless of which of the three block-lowering contexts (`stmts.rs`, `loops.rs`, `transactions.rs`) called it. Each of those enforces its own statement grammar and its own notion of what a write means; the new node erases that context. `lower_assigned_value` is the right abstraction — the problem is that it is one of two entry points, and the other one substitutes top-level context silently.

PR #130, opened 2026-08-26
Declare a curried function's non-`Mut` parameter annotations instead of dropping them

dpmills:
A `def` with a `Mut(…)` parameter lowers to a curried chain of named lambdas, and that path bound every *other* parameter at `Hole` and discarded its annotation. `def f(a: Int, m: Mut(Int))` therefore accepted `f("hello", total)`, inferring `a` from the body alone — the defect the single-parameter arm's comment records having fixed there. Each non-history annotation is now a checking-mode declaration, as in the other two arms.

`Feed(T)` had to come with it: `lower_type_annotation` rejects `Feed(…)` as an unknown type application, so declaring one needs the interception `Mut(…)` already gets. `feed_param_history_type` recognizes it as the `Append` `Type::History` carrier — exact, one argument, for the reasons `Mut` is. Two existing feed cases relied on the drop.

Read the fold in `lower_function_body` first. The tests pin the enforcement, that a refinement naming a parameter to its left still resolves, and both malformed `Feed` spellings.

### DIFF
--- a/tests/compilation_pipeline/mutability.rs
+++ b/tests/compilation_pipeline/mutability.rs
@@ -1545,3 +1545,112 @@ fn a_mut_param_still_binds_a_mut_var() {
         cambra::interpreter::Value::Int(1),
     );
 }
+
+#[test]
+fn a_non_mut_parameter_annotation_is_enforced_in_a_curried_function() {
+    check_compile_error(
+        indoc! {r#"
+            def f(a: Int, m: Mut(Int)):
+                m := 1
+
+            total: Mut(Int) := 0
+            f("hello", total)
+            total
+        "#},
+        "Type mismatch for Apply: expected Int, found String",
+    );
+}
+
+#[test]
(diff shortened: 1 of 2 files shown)


tests/compilation_pipeline/mutability.rs
```diff
@@ -1545,3 +1545,126 @@ fn a_mut_param_still_binds_a_mut_var() {
         cambra::interpreter::Value::Int(1),
     );
 }
+
+/// A non-`Mut` parameter's annotation in a `Mut`-curried function is a
+/// **checking-mode** declaration, exactly as it is in the single-parameter arm.
+/// The curried path used to bind every non-`Mut` parameter at `Hole` and drop the
+/// annotation, so the call below type-checked and `a` was inferred from the body
+/// alone — the same defect the single-parameter arm's comment records.
```
sortalongo: Don't talk about previous versions of the code.

sortalongo (approved):

PR #131, opened 2026-08-26
Carry provenance panes down to `post-lambda-elim`, with the topology declared once

groundlar:
Carries provenance capture down to a fifth pane, `post-lambda-elim`, and replaces
the hand-spelled pane wiring with one table. Four commits, each green on its own
under `CAMBRA_PROVENANCE_GATE=1`.

## What lands

**`ccl(provenance): assert the copy channel's coverage, not just the mint channel's`**
A node reaches a tree by being minted or by being copied. `on_mint` asserted that
a mint under a row-writing phase scope has a recording to land in; `on_copy`
asserted nothing, so a rewrite that only duplicates produced `Unrecorded` at the
fold with no panic anywhere. Over the suite, copies outnumbered mints among the
uncaptured nodes 4548 to 988. The one gap it found in an already-instrumented
phase is inference freshening a refinement predicate per scheme instantiation,
which takes a `copy_frame` — a duplication with nothing being rewritten and so no
node to name.

**`ccl(provenance): the fold's id domain is the slot domain the passes rewrite`**
`collect_tree_ids` reached `expr.ty`, `user_annotation` and a `Cast`'s target, but
never a binder's declared type or annotation. `TypedExpr::walk_type_slots` reaches
all five, and it is what the rewriting passes walk — so the fold's explanation
domain was narrower than the slot domain being rewritten. A predicate riding only
a `let` binder's annotation was never enumerated, and a rebuild of it named an id
the fold had not heard of: `DanglingParent` against an id the input pane
demonstrably holds. `f: (Int => {Int where _ == 9}) = …` is the shape.

Nothing in tree exercised it, because no instrumented phase rebuilt a predicate
parked in a binder slot. `lambda_elim` is the first that does, which is why this
sits below the pane that needs it: with the pane pair in place and this fix
absent, `CAMBRA_PROVENANCE_GATE=1` fails 16 group-by and dependent-group-by
programs.

**`ccl(provenance): declare the pipeline's pane topology in one place`**
Three panes and two pairs were spelled across six consts, six named
`MaterializedPanes` fields and two methods, which does not survive going to six
panes: five pairs times three fold products is fifteen fields. `PANES` declares
the topology once — each pane's name, the phases that produced it from the pane
before it, and whether its pair is gated — and `materialize_panes` folds a sliding
window over it.

`Phase::AsOfRead` is new, declared where the rewrite *runs* rather than where its
code sits: reusing `Phase::Transact` would have put one phase on both sides of the
post-channelize pane, and a fold restricting by it would resolve straight past
that pane.

All of it moves to `src/ccl/panes.rs`, along with `gate_leaks`, the two
`CompiledProgram` methods that fold and enumerate the panes, and the pane tests.
`context.rs` drops from ~2900 to ~1980 lines and keeps what it is about: the
pipeline, the `Phase` axis, and the env switches. The inherent
`impl CompiledProgram` sits in `panes.rs` rather than beside the struct because
the methods are the pane layer's — which is also why no call site changed.

**`ccl(lambda_elim): the point-free traversal records what it rewrites`**
Two recordings, one per recursive entry point, plus two arm relabels, and the
phase scope that makes them reach a table. `elim_lambdas_impl` names the node it
was handed; `elim_lambda_impl` names the lambda **body**, because `elim_lambda`
recurses over the body's structure and naming the enclosing `Lambda` would
collapse a whole body's combinator tree onto one parent. The filter and
value-`Case` arms open a second recording on the same node — not a capture gap,
but a recording carries one label and one nature for its whole extent, and those
two arms expand a construct the user wrote rather than re-mint plumbing.

## Where the gate stands

Every pair from `pre-inference → post-inference` down to
`post-as-of-read → post-lambda-elim` is gated: zero `Unrecorded` and zero
`DanglingParent` over the whole test suite under `CAMBRA_PROVENANCE_GATE=1`, not
merely over the eleven programs `corpus()` lists. `pre-inference →
post-inference` is gated here for the first time, the first two commits having
closed both gaps that held it back. `post-lambda-elim → post-planning` stays
ungated until planning records.

`PaneSpec::gated` is one bit per pair and it says whether an assertion has been
turned on, not how clean a pair is allowed to be — a leak is a bug wherever it
appears. What an uninstrumented phase leaves behind is a count of how little it
records, a constant no correct recording elsewhere can drive down, so gating it
would pin churn rather than catch a defect.

## What this does not do

`lambda_elim` is not made id-preserving here: the catch-all arm's
`TODO(preserve)` carries why, `planning/groupby` relying on the re-mint to launder
the key it lifts out of a refinement predicate. The pair holds at zero regardless
— every node is explained — but almost nothing in it gets a self-edge, so a
pass-through node reads to a consumer as a rewrite of its predecessor rather than
as itself. 323 deaths across the new pair over the corpus are the expected
consequence.

## Verification

- `cargo test` and `CAMBRA_PROVENANCE_GATE=1 cargo test` clean at each of the four
  commits; `cargo clippy --all-targets` and `cargo fmt --check` clean.
- The `full` audit span moves to `post-inference..post-lambda-elim` and reports no
  leaks on any corpus program.
- `design/provenance.md` follows throughout. "Gate and audit coverage" ends by
  naming what retires once the last phase records — the audit, its two switches,
  and `PaneSpec::gated` — against a `TODO(provenance-audit-retire)` at both sites,
  so the scaffolding for this campaign does not outlive it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/ccl/design/provenance.md
+++ b/src/ccl/design/provenance.md
@@ -411,6 +411,13 @@ sink, mirroring `infer_var::ACTIVE_ARENA`:
   recording open around a rewrite *captures* the births and copies in its dynamic
   extent, innermost first. An empty stack means recording is off, which costs one
   emptiness check on the construction hot path.
+
+  **Both hooks assert coverage**, under a phase scope that writes rows: a node
+  reaches a tree by being minted or by being copied, so a check on one channel
+  alone leaves the other silent. The silence was not hypothetical — when the
+  copy-side check was added, copies outnumbered mints among the uncaptured nodes,
+  and the one gap it found in an instrumented phase was inference freshening a
+  refinement predicate per scheme instantiation.
 - `TableSession` installs the table for a **whole compile**; `PhaseScope` names
   the phase rows are tagged with, for **one phase**. The two nest that way because a
   row's key is a process-unique `NodeId` and needs no phase set to disambiguate
@@ -444,16 +451,20 @@ a loud failure rather than a silent misattribution):
   all in the mutability phases, so `parents` alone is what recovers a source
   location for almost every node.
 
-`enter` is the only constructor a phase uses. `copy_frame` is the one recording
-that names **no** node: uncurry's template-interior freshens and the compare-chain
-operand freshens duplicate nodes with nothing being rewritten, and each captured
-copy carries its own origin from the hook. A recording that names no node has
-nowhere to attach a mint or a consume, which `OpenRecording::assert_copy_only`
-enforces.
+`enter` is the constructor a phase uses to record a rewrite. `copy_frame` is the
+one recording that names **no** node, for a duplication with nothing being
+rewritten and so no id to name: lowering's uncurry template-interior and
+compare-chain operand freshens, and inference freshening a refinement predicate
+per scheme instantiation. Each captured copy carries its own origin from the
+hook, so the named node is never read. A recording that names no node has nowhere
+to attach a mint or a consume, which `OpenRecording::assert_copy_only` enforces —
+which is also what keeps the two constructors apart, since a mint captured in a
+copy-only recording is a site that should have named a node.
 
 **Where recordings are open today.** Lowering's leaf appends and copy sinks, plus
 recordings inside the two pane pairs: `infer/solve` (`mono.specialize`,
-`mono.coalesce_let`), `infer/emit` (`infer.lit_singleton`), `inline`
+`mono.coalesce_let`), `infer/emit` (`infer.lit_singleton`),
+`infer/solver/scheme` (`infer.freshen_predicate`), `inline`
 (`inline.alias`, `inline.udf`, `inline.beta`), `mut_elim` (`letrec.loop`,
 `letrec.bare_write`, `letrec.hoist_writer_body`, `letrec.terminalize_write`),
 `transact_phase` (strip, unwrap block, writer, commit record, history binding,
--- a/src/ccl/design/provenance.md
+++ b/src/ccl/design/provenance.md
@@ -15,12 +15,10 @@ features: the inspector's consumption of the AST snapshots (panes), and adoption
 further phases. A reader can tell the two apart by the marker alone; unmarked prose
 describes code you can go read.
 
-Three sites record without reaching any table in a normal build, because
-`compile_program` opens no `PhaseScope` around them: `simplify`,
-`planning/iterate`, and `transact_phase`'s as-of-read rewrite. `lambda_elim`
-records nothing at all, and operator conversion has no identity to record
-against; see
-[Known prerequisites for panes past `post-channelize`](#known-prerequisites-for-panes-past-post-channelize).
+Two sites record without reaching any table in a normal build, because
+`compile_program` opens no `PhaseScope` around them: `simplify` and
+`planning/iterate`. Operator conversion has no identity to record against; see
+[Known prerequisites for panes past `post-planning`](#known-prerequisites-for-panes-past-post-planning).
 
 ## Mechanism at a glance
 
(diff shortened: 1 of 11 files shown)


src/ccl/design/provenance.md
```diff
@@ -444,37 +451,51 @@ a loud failure rather than a silent misattribution):
...
-`simplify` (one combinator covering all thirteen `&mut` rule invocations, with no
-rule-body edits), `planning/iterate`, and `transact_phase`'s as-of-read rewrite.
-`compile_program` opens no `PhaseScope` around any of them, so their recordings
-are inert in a normal build and land only under an audit a caller opens.
-`CAMBRA_PROVENANCE_AUDIT=full` (`post-inference..post-as-of-read`) covers the
-as-of-read rewrite; reaching `simplify` and `planning/iterate` needs
-`CAMBRA_PROVENANCE_AUDIT=planning` (`recognized..join-planned`), which is what the
-coverage figures below were measured with. `lambda_elim` records nothing, which
-is what an audit's endpoint stops in front of.
+Two sites record where nothing reads their rows: `simplify` (one combinator
```
dpmills, line 484: This stuff is all temporary, right? It's confusing, but if it's all going away soon it doesn't matter.  It might be worth leaving a TODO to clean up these docs (and also code) once provenance flows end-to-end.  Seems like there is a ton of churn happening just to support and describe temporary asserts, and it will be easy for that churn to leave a bunch of residue at the end

src/ccl/design/provenance.md
```diff
@@ -828,9 +858,32 @@ that exist:
...
+  set is what makes the crossing read as a leak. It says nothing about the pane
+  pairs that are gated, whose phases rebuild predicates through `PredMemo` and
+  are recorded there.
+- **Operator conversion has no identity**, which is what stops a pane *after*
+  `post-planning`. `TileOperator` carries none and there is no `OperatorId`, so a
+  pane there has nothing to resolve against.
+
+### `lambda_elim` re-mints, so its pair relates panes by parentage alone
+
+This is not a blocker, and is recorded here because it used to be listed as one.
```
dpmills: Is this section important at all? The fact that lambda elim rewrites the entire tree isn't relevant from the provenance design perspective.

dpmills (approved):
I'm a bit suspicious that the PR split isn't right.  This PR is thousands of lines changed, but as far as I can tell all but ~100 lines of that is churn around temporary gating logic that only exists because the feature isn't complete yet

src/ccl/design/provenance.md
groundlar: Partly, and I've made the split explicit rather than leaving it implied. `Gate and audit coverage` now ends with two lists under a `TODO(provenance-audit-retire)`, tagged at `ProvenanceAudit` and at `PaneSpec::gated`.

**Retires**: `ProvenanceAudit` and `CAMBRA_PROVENANCE_AUDIT`, plus the audit clause in `provenance_capture_enabled` — capture becomes unconditional once nothing else contends for the phase scopes; `CAMBRA_PROVENANCE_PREDICATES`, whose only reader is the audit's live set; `PaneSpec::gated` and `gated_pane_pairs`; and the paragraph you're commenting on.

**Stays**: the recordings at each phase, `PANES`, the fold, `gate_leaks`, and the two switches that measure rather than scaffold — `CAMBRA_PROVENANCE=0` for capture cost, `CAMBRA_PROVENANCE_GATE=1` for the whole-suite corpus.

One qualifier on "temporary asserts": the asserts aren't the temporary part. `gate_leaks` is the permanent mechanism; what's temporary is the *ungated* state and the audit that measures a phase mid-instrumentation. The doc already carried the two rules for retiring that state piecewise — flip `gated` in the commit that instruments the pair's last phase, move the audit endpoint the same way — so the churn each commit does is bounded. What was missing is the terminal statement, which is exactly the residue you're pointing at.

Taking that seriously turned up one live instance: the `PANES` doc comment still claimed two ungated bottom pairs and described the fix as "the two commits below this one", both false now that `lambda_elim` records and `post-lambda-elim` is gated. Rewritten to state the facts rather than stack positions.

src/ccl/design/provenance.md
groundlar: Agreed — cut. It was an ex-blocker kept after it stopped being one, and it was the third copy of the argument: the `TODO(preserve)` at `lambda_elim.rs:1787` already carries the groupby-freshen reasoning in full, and the PR body repeated it (now trimmed to a sentence).

One design-relevant consequence survived, moved to `The edge labels` where self-edges are defined: a phase that re-mints its pass-through nodes has almost no self-edges, so its pair's relation comes from the parents walk alone and it cannot say that an untouched node was untouched. That bounds what a consumer can conclude from a pair, which is a fact about the model rather than about `lambda_elim`. Six lines there instead of twenty under a heading for prerequisites it explicitly wasn't one of.

No `TODO` reproduced in the doc, per your other thread — the code note is the single copy, and the doc just points at the arm that holds it.

PR #132, opened 2026-08-26
Gate the last pane pair by recording what planning rewrites

groundlar:
Planning was the last phase rewriting expression nodes with nothing recording, which held `post-lambda-elim → post-planning` ungated and capped the `full` audit span short of the pipeline's end. This opens one `Phase::Planning` scope over planning and a recording at each of its rewrite sites, so **every pane pair is gated** and capture is total from the lowering handoff to the end of the CCL pipeline. `DEEP_TYPECHECK=1 CAMBRA_PROVENANCE_GATE=1 ./ci.sh test` is green, so both leak classes are zero over every compile the suite performs rather than over the programs `corpus()` lists.

## What each recording names

The node its products replace, with two choices worth a pointer. Continuation reads take a recording *finer* than the recognition one enclosing them — without it every read descends from the `LetRec` and which read went where is lost. The group-by and hash-join recordings are coarse by construction, which is [Where to open a recording](src/ccl/design/provenance.md#where-to-open-a-recording)'s rule for a product spanning several nodes. `Nature` is stated at each site and ruled in [Choosing between `Expansion` and `Machinery`](src/ccl/design/provenance.md#choosing-between-expansion-and-machinery).

## The hash join blames the predicates it plans from

`planning.hash_join` named its iteration site and nothing else, so the refinement predicates its plan is read out of reached the map only where a clone happened to carry one. It now blames every refinement predicate the site's domain carries, through `ccl_utils::walk_refined_predicates`. Blame rather than a parent: the rewrite leaves those predicates alive — the original value-producer is spliced into the compose with its type intact — and `predicate.rebuild` consumes them later in the same phase, where the ancestry is already recorded.

The walk reaches what a return value cannot. `join_plan_to_expr` re-enters `convert_refinement_to_join` on an arm's own refinement to plan a nested join, three frames below the guard and with no access to it. Across the suite the walk names exactly the predicates the recogniser read: 40 of 42 hash-join sites carry one refinement, and the two that carry two are the nested cases, whose second entry is the id the inner read consumed. `a_nested_join_blames_both_join_conditions` pins that against a flat control.

## Raising needed no new channel

The three sites that lift a term out of a type keep the *predicate's* parentage, because a clone's copy names the node it was freshened from. The channel that looked missing was predicate ids sitting outside the enumerated domain, which `main`'s binder-slot widening already fixed. The staging around a lifted predicate does not: the chain-head `iterate` and each `restrict` descend from the site they materialise.

Two pre-existing sites minted into whichever recording was open, invisible until a scope existed: `simplify::take`'s placeholder literal, now `mem::take`, and `recognize_group`'s deep clone of every accumulator seed.

## One statement of a phase boundary's post-conditions

Every phase ended with the same three assertions written out by hand: `assert_unique_node_ids`, the debug dumps, and the stage's consistency check with its own panic string. `settle(ir, boundary, check)` is now the one place they live, so each phase's tail is one line and `pre_channelize_wall` is reached through it. `Check` names which check the stage answers to — the relaxed one for the mutability phases, the strict one from channelization on, and the relaxed one reporting a residual `Type::Infer` as an ambiguous program at the two boundaries that can still carry one.

## Review guide

**Start here:** `try_hash_join_rewrite` in `src/ccl/planning/join.rs`.

`tests/programs/join_then_groupby` is a new fixture putting both recognizers on one tree, and joins the provenance corpus. Two unit tests pin the planning labels' `Nature` and the single scope. `the_pane_folds_derive_a_non_vacuous_provenance_map` asks for one witness per gated pane pair rather than pinning which `program / pane pair` names rewrote: which programs supply the witness is a fact about the corpus, and pinning it made every corpus edit a list edit. `planning.map_filter` is instrumented alongside the rest — `main`'s per-group filter planning mints under the planning scope.

The `planning` audit span retires as redundant with the pair that now gates the same phase; `letrec` and `mutelim` stay, per [What gating every pair does not retire](src/ccl/design/provenance.md#what-gating-every-pair-does-not-retire). Comments and docs stop enumerating the corpus, the panes and the rule table — each count restated a population another file owns.

### DIFF
--- a/src/ccl/context.rs
+++ b/src/ccl/context.rs
@@ -563,7 +563,7 @@ pub struct CompiledProgram {
     /// `via` is what a fold between two panes restricts by. The phases that record are
     /// [`Phase::Infer`] (everything monomorphization mints inside `infer`, which
     /// bridges the pre-inference ⇄ post-inference panes) and [`Phase::Inline`],
-    /// [`Phase::Transact`], [`Phase::Letrec`], [`Phase::Channelize`] (the four phases
+    /// [`Phase::Transact`], [`Phase::Letrec`], [`Phase::Channelize`] (the phases
     /// between the post-inference and post-channelize snapshots) — see [`PANES`],
     /// which declares which phases each pane pair folds.
     ///
@@ -995,7 +995,7 @@ pub(crate) fn provenance_capture_enabled() -> bool {
     !std::env::var(PROVENANCE_ENV).is_ok_and(|v| v == "0") && provenance_audit_span().is_none()
 }
 
-/// Whether every compile folds both pane pairs and gates the leak classes,
+/// Whether every compile folds every pane pair and gates the leak classes,
 /// rather than only the programs a test asks about.
 ///
 /// **What this buys.** The always-on gate is
@@ -1006,7 +1006,7 @@ pub(crate) fn provenance_capture_enabled() -> bool {
 /// another loop's accumulator) and `flatten_spine`'s value-position writer hoist
 /// both sat outside it. With this on, the corpus is every program the caller
 /// compiles — point it at `tests/compilation_pipeline` and the gate covers the
-/// whole suite instead of eleven programs.
+/// whole suite instead of the programs `corpus()` lists.
 ///
 /// Off by default because it folds the table twice per compile, which is
 /// superlinear work the ordinary pipeline does not need; CI turns it on. It is
@@ -1040,15 +1040,16 @@ fn recorded<R>(capture: bool, phase: Phase, f: impl FnOnce() -> R) -> R {
 /// A driver-capture audit over one span of the pipeline: install a phase recorder at
 /// the input pane, fold at the output pane, and print what the capture explains.
 ///
-/// Opt-in via `CAMBRA_PROVENANCE_AUDIT=1` so the whole test suite can run either
+/// Opt-in by naming a span in `CAMBRA_PROVENANCE_AUDIT` — the spans opened below
+/// are `full`, `letrec` and `mutelim` — so the whole test suite can run either
 /// way. It is a **measurement**, not a gate: nothing declares a fate, so every
 /// genuinely-dead input id reaches the audit through the fold's death
 /// collection, which it counts apart from the [`Leak`]s that really are
 /// recording bugs.
 ///
 /// A span survives gating when its enclosing pair bundles more phases than the
-/// span does: the `letrec` and `mutelim` spans below isolate phases inside the
-/// four-phase `post-inference → post-channelize` pair. See
+/// span does: the `letrec` and `mutelim` spans below each isolate a phase inside
+/// the `post-inference → post-channelize` pair. See
 /// `src/ccl/design/provenance.md`, "What gating every pair does not retire".
 struct ProvenanceAudit {
     span: &'static str,
@@ -1407,8 +1408,8 @@ fn run_passes(
     lowering_projection: &SourceProjection,
     panes: &mut BTreeMap<Phase, Expr>,
 ) -> Result<Expr, Vec<CompileError>> {
-    // Phase recording for the rows the pane folds read. One scope per phase,
-    // opened and closed in place: a phase scope is per-thread and
+    // Phase recording for the rows the pane-pair folds read. One scope per
+    // phase, opened and closed in place: a phase scope is per-thread and
     // non-reentrant, so the scopes are sequential, never nested.
     let capture_provenance = record && provenance_capture_enabled();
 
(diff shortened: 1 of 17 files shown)


src/ccl/planning/join.rs
```diff
@@ -1081,6 +1081,44 @@ pub(super) fn try_hash_join_rewrite(expr: &mut Expr, domain_ty: &Type) -> bool {
         "Attempting hash-join rewrite at iteration site: {}",
         symbolic(expr),
     );
+    // The iteration site is what the recording names. Every product of the plan
```
dpmills, line 1084: This comment is huge and seems to be making an argument that doesn't really hold.  The key wrong point is "The arms the plan reads come from the domain **type**, and a type carries no identity to consume." The arms actually come from the nodes inside the refinement in the type, and those absolute do have node Ids, so this recording is throwing away important information

src/ccl/planning/iterate.rs
```diff
@@ -394,11 +394,10 @@ pub(super) fn wrap_with_iterate(expr: &mut Expr) {
     let Some(domain_ty) = expr.ty.domain() else {
         return;
     };
-    // The recording names the site being wrapped.
-    //
-    // These rows reach no table in a normal compile: `compile_program` calls
-    // `planning::run` outside every pass scope it opens, so they land only under
-    // `CAMBRA_PROVENANCE_AUDIT=planning` (`recognized..join-planned`).
+    // The recording names the site being wrapped. The predicate lifted out of
```
dpmills: I don't think this holds entirely; for example, the restrict node that applies the predicate will be parented off the expression represented the data instead of the filter

src/ccl/context.rs
```diff
@@ -1550,16 +1539,17 @@ pub fn compile_program(
     // Running post-elim is what keeps ONE letrec representation through
     // channelize and lambda_elim; the point-free guard matcher re-checks
     // causality at this wall. See the `mut_elim` recognition docs.
-    let recognized = planning::plan_loops(lambda_elim);
+    let recognized = recorded(capture_provenance, Phase::Planning, || {
```
dpmills: Doesn't need to happen in this PR, but this compilation script function is getting very repetitive and filling up with unnecessary comments.  Each phase does almost the same thing (record, run, dump ccl, typecheck, assert unique nodeids) and that could all be factored out.  We also wouldn't need the explicit audit start/finish if things were more structured. (PR #41, merged, already did some refactoring that will make this easier.  It could still be cleaned up a lot more though).

src/ccl/panes.rs
```diff
@@ -471,32 +471,46 @@ mod tests {
     /// difference visible, so it lists names and not a total.
     const EXERCISED_BOUNDARIES: &[&str] = &[
```
dpmills: Is this still useful? In general we expect most phases to do stuff, so why do we need to track which ones do?

dpmills (approved):

src/ccl/planning/join.rs
groundlar: Done. You were right that information was being thrown away: a `Refinement.predicate` is an `Rc<TypedExpr>` with a node id, and `collect_tree_ids` walks into it. The site now blames every refinement predicate its domain carries (`ccl_utils::walk_refined_predicates`), which reaches the arm's own refinement that `join_plan_to_expr` re-enters `convert_refinement_to_join` on for a nested join — three frames below the guard, so a return value could not carry it. Across the suite the walk names exactly what the recogniser read: 40 of 42 sites carry one refinement, and the two that carry two are the nested cases, whose second entry is the id the inner read consumed.

Blame rather than a parent: the rewrite leaves the predicates alive (the original value-producer is spliced into the compose with its type intact), and `predicate.rebuild` plus the simplifier consume them later in the same phase, where the ancestry is already recorded — measured on `[x + y for x in [2] for y in [a + b for a in [1, 2] for b in [1, 2, 3] if a == b] if x == y]`, both conditions now carry blame edges at `post-lambda-elim → post-planning`, pinned by `a_nested_join_blames_both_join_conditions` against a flat control. The false premise is gone and the comment is about half its length.

src/ccl/planning/iterate.rs
groundlar: Done. The claim only ever held for the predicate. The chain-head `iterate` and each `restrict` are minted under the site's recording and descend from it, not from the filter they apply; only the predicate they carry keeps its own parentage, as a `Copy` of the term it was freshened from. That narrower fact is all `both_raising_paths_attribute_the_predicate_to_the_predicate` pins, and the comment now says so.

src/ccl/context.rs
groundlar: Done. `settle(ir, boundary, check)` carries every boundary's post-conditions — unique ids, the debug dumps, the stage's consistency check — so each phase's tail is one line and `pre_channelize_wall` is reached through it. `Check` names which check the stage answers to. What is left between `recorded` and `settle` is the phase's own error channel, which differs at every site.

The audit spans stayed: a span covers a range of phases rather than one, so there is no per-phase structure to fold them into.

src/ccl/panes.rs
groundlar: Done — pinning which programs rewrote was the wrong question, as you say. The test now asserts each gated pane pair is exercised by at least one corpus program, which is the property the 42-entry list was standing in for, and a corpus edit is no longer a list edit. (The doc comment's arithmetic was stale too: 18 of 60 pane pairs are pure identity, not six of twenty-two.)

PR #135, opened 2026-08-27
Add refinement templates to trait instances, refined ^+ operator

octalsrc:
This PR adds an optional `RefinementTemplate` to each associated type of a `TraitInstance`. The `RefinementTemplate` is a function that produces a refinement for the associated type, when applied to the input operands of an operator.

This means that `TraitObligation` now carries the input expressions of the operator that produced it, so that those can eventually be passed into any relevant `RefinementTemplate` when a trait instance is finally chosen.

This PR also adds a `^+` operator, to be used temporarily to experiment with refined operators. `^+` mirrors `+` in every way, except it can only be used on `Int`, and the output type of `x ^+ y` is refined to `{Int where _ == x ^+ y}` using the trait refinement system described above.

### DIFF
--- a/src/ccl/infer/solver/traits.rs
+++ b/src/ccl/infer/solver/traits.rs
@@ -74,9 +74,13 @@ use std::fmt;
 use std::rc::Rc;
 use std::sync::atomic::{AtomicU32, Ordering};
 
-use crate::ccl::{BaseType, FieldKey, InferVar, InferVarId, Type};
+use crate::ccl::{
+    ArithmeticKind, BaseType, BinOpKind, CompareKind, FieldKey, InferVar, InferVarId, Name,
+    Refinement, RefinementSet, RefinementTemplate, Type, TypedExpr,
+};
 
 use super::constrain::{ConstrainCache, ConstrainError, constrain_subtype};
+use super::prim;
 
 /// A trait: a named requirement on types, together with any types it associates
 /// with them.
@@ -92,6 +96,11 @@ pub enum Trait {
     /// why surface `+` on strings types through arithmetic; `simplify` rewrites it to
     /// `Concat` later.
     Addable,
+    /// `^+` over `(𝐴, 𝐵)`, associating `Output`.
+    ///
+    /// This is an experimental version of Addable that gives the
+    /// output type an input-dependent refinement.
+    AddableRefined,
     /// `-` over `(𝐴, 𝐵)`, associating `Output`.
     Subtractable,
     /// `*` over `(𝐴, 𝐵)`, associating `Output`.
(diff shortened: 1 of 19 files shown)


src/ccl/infer/solver/traits.rs
```diff
@@ -150,27 +164,59 @@ impl TraitInstance {
...
+fn refinement_for_add(args: &[TypedExpr]) -> TypedExpr {
+    let [a1, a2] = args else {
+        panic!("AddableRefined is binary, so its instance's refinement receives two operands");
+    };
+    let int = prim(BaseType::Int);
+    TypedExpr::binop(
+        TypedExpr::var(Name::elem()).with_ty(int.clone()),
+        BinOpKind::Compare(CompareKind::Equals),
+        TypedExpr::binop(
+            a1.clone(),
```
dpmills, line 192: This means that any refinements on the args end up inside this new refinement so we still have all available information in one place, right? Would be good to have a test that confirms this

src/ccl/infer/solver/traits.rs
```diff
@@ -515,12 +575,20 @@ impl TraitObligation {
...
                 continue;
             };
             position.deposited.set(true);
             let target = position.ty.borrow().clone();
-            constrain_subtype(&Type::Base(settled), &target, cache)?;
+            let ty_base = Type::Base(settled);
+            let ty = match maybe_refinement {
+                Some(template) => Type::Refinement(
+                    Box::new(ty_base),
+                    RefinementSet::one(Refinement::born_from_template(template, &self.input_exprs)),
```
dpmills, line 598: You'll need provenance tracking here

dpmills (approved):

octalsrc:
Strange, `cargo test` and `./ci.sh` pass with no failures on my machine.

src/ccl/infer/solver/traits.rs
octalsrc: I've now added the operator's node ID to its resulting `TraitObligation`, which is then eventually used as provenance for the refinement that is generated.

src/ccl/infer/solver/traits.rs
octalsrc: It does, though I think we don't actually need that, since we will get those refinements more easily from the environment.

One issue I noticed is that refinements-on-exprs-in-refinement-bodies causes the complete refinement to blow up in size very fast, which at least makes debugging the result more difficult. Not sure if it's a performance issue, given that refinement bodies are shared through `Rc`s.

octalsrc: I'm going to defer whether we want to utilize or erase refinements-on-exprs-in-refinement-bodies for now.

PR #136, opened 2026-08-28
Carry a write inside a conditional's branch past the conditional, and indent a block right-hand side's chain one level in

dpmills:
A write inside a conditional's branch reached nothing after the conditional. A `MutWrite` is a shadowing advance over the rest of the scope it sits in, and a branch's scope ends at the branch, so the update was silently dropped everywhere except a for-loop body — where `transform_chain` merges a guard-`Case` into one writer decision over the whole source. Two rewrites put the write back on a statement spine by pushing what follows the `Case` into its branches, so every spelling of one conditional compiles to the same recurrence.

```python
acc := 0
for i in [1, 2, 3]:
    t = if i > 1:
            acc += i
            0
        else:
            0
    yield t
acc                     # was 0, now 5
```

The same held for a statement `if` outside a loop body, and for a block right-hand side in any position.

## The two rewrites

`push_binding_into_case` handles a `Case` a `Let` binds, `push_continuation_into_case` one in statement position. Both walk children first: pushing a nested conditional leaves a `Case` standing where a statement stood, and its writes are then on that `Case`'s branches rather than on the enclosing spine — so `spine_writes_mut` follows a `Case`'s branches as well as a spine, and a terminal `Case` that still writes is descended into rather than substituted whole. That is what carries a write out at any nesting depth.

[mutability.md](src/ccl/design/mutability.md), "A write inside a `Case` bound by a `Let`" carries the reasoning — why a branch's terminal is substituted rather than bound, and why the statement-position rewrite is suppressed inside a for-loop body, where that shape is the recurrence's. Both run ahead of `transact_phase::run`, which walks a `with begin():` block itself, and again at the letrec phase for the writes inlining buries between the two.

## The chain indents one level in

`elif` and `else` went back to the statement's own column. They now indent one level in and the branch bodies one further, and the old spelling is rejected. A block right-hand side opens a layout level of its own, floored at the column of the statement that opens it and fixed by the first line to dedent into it — [design-chl-parser.md](src/chl_parser/design-chl-parser.md), "A block right-hand side opens a level of its own". It closes with one `Dedent` more than the blocks inside it opened, which `block_value` consumes; that is what makes the two spellings different token streams. Every site in the tests and docs is reindented.

## Scope

`lower_block_value` is the sole entry point for a block right-hand side, and every position that assigns carries the enclosing scope to it — a statement block through `preceding` and `outer_bindings`, a for-loop body through `ForSite`, a `with begin():` block through `lower_tx_block`. That retires the "the dropped scope is unobservable" claim in [lowering.md](src/ccl/design/lowering.md): the loss was observable, as a diagnostic telling the author to declare a variable line 1 already declares.

**`match` is bounded below this layer.** An arm computing over a name bound outside it reaches operator conversion as a bare `BinOp`, and lifting a write onto an arm's spine leaves exactly that. The bound holds with no write involved and fails identically on `main`; `test_match_arm_computing_over_an_outer_name` records it, ignored.

Review nits from #129: `assign_tail` no longer carries `expr_to_assign_target`'s summary line, the atom list drops `dict`, and the comment above `block_assign` records the second parse of the target it costs.

### DIFF
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -1359,7 +1359,8 @@ acc                     # 5
 
 A branch's value may read what that branch wrote (`acc += i` then `acc` yields
 the updated value), and a write inside a `with begin():` block commits with the
-rest of its block.
+rest of its block. The statement `if` carries its writes the same way, in a
+for-loop body and outside one alike.
 
 Annotated assignment **requires** a value (`x: T` alone is a parse error,
 unlike Python's bare type-only declarations).
--- a/docs/chl-spec.md
+++ b/docs/chl-spec.md
@@ -1341,22 +1341,25 @@ branch must produce a value, and an `if` with no `else` is rejected (§4.5); a
 one-line spelling of the same thing is the ternary (§3.6) or, for `match`, the
 bracketed form in §4.10.
 
-A block on the right may not write a variable declared outside it. A write
-reaches the rest of the block it sits in, and the block ends at the value the
-assignment takes, so the update would reach nothing:
+A branch may write, and the write carries past the assignment. The two
+spellings below agree, so a conditional that both writes and yields a value
+needs no rewriting into a statement `if`:
 
 ```python
 acc := 0
-t = if c:
-        acc += 1        # rejected: the update is discarded
-        0
-    else:
-        0
+for i in [1, 2, 3]:
+    t = if i > 1:
+            acc += i
+            0
+        else:
+            0
+    yield t
+acc                     # 5
 ```
 
-Write `acc` from a statement `if` beside the assignment, or bind the block's
-value and write `acc` after it. A mutable variable the block itself declares is
-written normally — its declaration and its writes share one scope.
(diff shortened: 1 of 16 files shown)


docs/chl-spec.md
```diff
@@ -1326,11 +1326,42 @@ n = match msg:
...
+            0
+    yield t
+acc                     # 5
+```
+
+A branch's value may read what that branch wrote (`acc += i` then `acc` yields
+the updated value), and a write inside a `with begin():` block commits with the
+rest of its block. The statement `if` carries its writes the same way, in a
+for-loop body and outside one alike.
+
```
sortalongo: rm cruft. This is the only sane behavior, and we don't need to documnet it.

sortalongo (approved):

docs/chl-spec.md
dpmills: Deleted this plus other unnecessary docs

PR #137, opened 2026-08-28
Walk `elim_lambda`'s result at the `match`-arm rule

dpmills:
`elim_lambda` abstracts one binder and leaves the walk to its caller. `build_scrutinee_case_cform` passed its scrutinee, its guard-`Case` siblings and its default arm through `elim_lambdas` but not the arm morphism, so a `match` arm reaching a `const` rule kept whatever the rule lifted and operator conversion refused it. Every arm not mentioning its payload binder was affected, in all three `match` spellings.

```python
base = 3
t = match `some(3):
        case `some(n):
            base + n           # fine — the payload binder is free here
        case `none:
            base + 0           # refused at operator conversion
t
```

Adding the walk at the caller covers all three rules that lift a body under `const` — the plain constant rule, the Pi-const rule beside it, and the cast-const arm — where fixing one rule reaches one. It also leaves `elim_lambda` calling nothing in `elim_lambdas`, so the cast-const arm's own call goes: `Cast` has no arm in `elim_lambdas`, and the catch-all's `try_map_children` reaches its value. [optimization.md](src/ccl/design/optimization.md), "`elim_lambda`'s caller walks its result" records the contract.

It unblocks a write in a `match` arm: the mutability phases in #136 lift the write onto the arm's spine and leave the arm computing over the accumulator, which is that shape, so `match` joins `if` in carrying writes past a conditional. `test_match_arm_write_carries` covers those and replaces the ignored test that recorded the bound; `test_const_lifted_arm_is_point_free` covers the write-free shapes, including a `match` nested inside a binder-free arm.

### DIFF
--- a/src/ccl/design/optimization.md
+++ b/src/ccl/design/optimization.md
@@ -124,6 +124,19 @@ where `src_refined` is `src_elim` with its domain wrapped in `Type::Refinement` 
 
 The filter check happens at the `Compose` level (rather than inside the `Lambda` arm) because the refinement must be attached to the source, which is only visible alongside the lambda at the compose level.
 
+### The constant rule reaches through its body
+
+`λ 𝑥 → 𝑒` with `𝑥 ∉ fv(𝑒)` becomes `𝑒 ▷ const`, and `𝑒` is eliminated first. `const` lifts a
+*value*, and the value operator conversion accepts is point-free — a `BinOp` or a nested `Lambda`
+left inside a lifted body reaches conversion as written and is refused there.
+
+The `Lambda` arm of `elim_lambdas` re-enters on the rule's result, so it covers for a body the rule
+leaves alone. The `match`-arm rule calls `elim_lambda` directly and does not, which is the caller
+the descent is load-bearing for: an arm reaches the constant rule whenever it does not mention its
+payload binder, so an arm computing over a name bound outside the `match` needs it. So does every
+write in a `match` arm — the mutability phases lift the write onto the arm's spine and leave the arm
+computing ([mutability.md](mutability.md), "A write inside a `Case` bound by a `Let`").
+
 ### `Let` nodes after rule 7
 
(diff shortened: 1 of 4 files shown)


src/ccl/design/mutability.md
```diff
@@ -672,10 +672,11 @@ write inside the continuation's own expression. What the two rules together guar
 path carries its writes ahead of the continuation, at any nesting depth
 (`test_nested_conditional_write_carries_out_of_a_block_right_hand_side`).
 
-A `match` reaches neither: an arm computing over a name bound outside it is not point-free by
-operator conversion, and lifting a write onto an arm's spine leaves exactly that. The bound is below
-this layer and holds with no write involved — see
-`tests/compilation_pipeline/mutability.rs`, `test_match_arm_computing_over_an_outer_name`.
+A `match` reaches both, and by the same route: an arm is a branch, so a write in one lifts onto the
```
sortalongo: This whole subsection seems wayyy to detailed. I can't make heads or tails of it; I don't understand what it's trying to document; and the fact that it actively documented obviously-buggy behavior suggests it didn't help claude. I recommend deleting it.

src/ccl/design/optimization.md
```diff
@@ -124,6 +124,25 @@ where `src_refined` is `src_elim` with its domain wrapped in `Type::Refinement`
 
 The filter check happens at the `Compose` level (rather than inside the `Lambda` arm) because the refinement must be attached to the source, which is only visible alongside the lambda at the compose level.
 
+### `elim_lambda`'s caller walks its result
```
sortalongo: This also seems worthless. It's an extremely verbose way of restating the contract of `elim_lambda`, then explaining the purpose of `elim_lambdas`. There was a bug that violated the already-documented contracts of those functions. Over-explaining those contracts wouldn't have prevented it, and it doesn't aid human understanding either.

sortalongo (approved):
Reading this PR, I think we need to be much more aggressive about policing AI's documentation additions. This is a bug with a one-line fix, and it came with a 30 line docs change that has negative value.

src/ccl/design/optimization.md
dpmills: Deleted

src/ccl/design/mutability.md
dpmills: Deleted (downstack)

PR #138, opened 2026-08-28
Iterate a `mut` accumulator over the positions its filtered source delivers

dpmills:
A `mut` accumulator over a filtered collection — `kept = [l for l in stdin() if l != "skip"]` then `for l in kept: n := n + 1` — panicked in the post-planning `typecheck`, with no test covering the shape. Two defects stacked, so fixing the first alone turned the panic into a silent `0`. The shape now compiles and computes the right value, batched or streaming.

## The sequencing domain is a type slot

`Transact.domain` was excluded from `TypedExpr::walk_type_slots{,_mut}` by a documented decision that predicted this failure: `planning::compile_refinement_predicates` rewrites predicates through that walk, so the domain kept its bare predicate while every sibling slot took the compiled one, and the post-planning `typecheck` compares refinements structurally.

## Positions, not a step counter

`restrict` keeps an element at its original position, so a filtered source's domain is a subset of its extent's. `InductionDriver` conflated the store's tick cursor with its item cursor and read the absent position 0 as end-of-source; it now keeps `emitted_through` and takes the smallest delivered position above it, while `InductionStore` consumes decisions in ascending position order rather than at consecutive ones. Ticks still follow positions (`p` decides tick `p + 1`), so `StoreDenseRead` needs no change: an excluded position folds to the last write below it.

## Two consequences

Commit 2 puts the last two hand-rolled copies of the type-slot set on the shared walk. `predicate_id_collisions` was blind to a predicate riding only a binder annotation — the live shape at the `post-lowering` boundary, where that tripwire runs.

Commit 3 rejects a `mut` loop over a **product** domain at op-conversion. Positions are `UInt` throughout the recurrence, and a comprehension across two sources is keyed by `(i, j)`; the join spelling of it reaches that check only because commit 1 lets it past the `typecheck` it used to fail at. `decode_source_positioned` drops a non-`UInt` key rather than failing on it, so without the check a product domain is a loop that runs zero times.

Reasoning is in [`type-inference.md`](src/ccl/design/type-inference.md), [`mutability.md`](src/ccl/design/mutability.md) and [`design-operators.md`](src/interpreter/design-operators.md). Five pipeline cases cover the filtered sources (let-bound, inline, interior gap, empty filter, feed) and four the product rejection.

### DIFF
--- a/src/ccl/design/mutability.md
+++ b/src/ccl/design/mutability.md
@@ -66,22 +66,13 @@ carry-forward, so channels never close a causal cycle — see [merge laws](#the-
 | `with begin():` transaction | `Txn`, an anonymous total order issued by the runtime | `get_prev_txn` |
 | `while` loop *(future)* | a prefix of `Nat` bounded by the running condition | `get_prev_seq` over a self-ceiling domain |
 
-A **filtered** loop source is the same row read literally: its domain is the refined extent
-`{[0, 𝑁] | 𝑝}`, so the recurrence runs at the positions that survived the filter and at no
-others. `restrict` keeps an element at its original position, so those positions are a subset
-of the extent's rather than a renumbering of them, and the induction store's ticks follow them
-(position `𝑝` decides tick `𝑝 + 1`, and an excluded position's tick is never occupied). A read
-of the accumulator's history still spans the whole extent: at an excluded position it folds to
-the latest write below it, the value the accumulator held there. The alternative — iterating
-the extent densely and gating the write, which is how an in-body `if` compiles — would put the
-store at odds with the domain the type carries.
-
-A **product** loop source is rejected at op-conversion. A comprehension across two sources
+A **product** loop source is unsupported. A comprehension across two sources
 (`[e for x in xs for y in ys]`, join or not) is keyed by `(𝑖, 𝑗)`, and the induction recurrence is
 sequenced by `UInt` position throughout: the driver pairs items with `UInt` domain keys, the commit
-engine's ticks are positions, and the dense read folds tick `𝑝 + 1` at each. The rejection is at the
-boundary rather than deeper because the driver's decode drops a non-`UInt` key instead of failing on
-it, so an unguarded product domain would be a loop that runs zero times.
+engine's ticks are positions, and the dense read folds tick `𝑝 + 1` at each. Supporting one means
+giving the recurrence a sequencing order over `(𝑖, 𝑗)`. Op-conversion rejects such a source until
+it has one, because the driver's decode drops a non-`UInt` key instead of failing on it, so an
+unguarded product domain is a loop that runs zero times.
 
 A mutable variable's domain is the domain of the context that **writes** it (`Txn` excepted — never
 inferred, always spelled at the introduction). What the **introduction** site fixes is not that
--- a/src/ccl/design/mutability.md
+++ b/src/ccl/design/mutability.md
@@ -76,6 +76,13 @@ the latest write below it, the value the accumulator held there. The alternative
 the extent densely and gating the write, which is how an in-body `if` compiles — would put the
 store at odds with the domain the type carries.
 
+A **product** loop source is rejected at op-conversion. A comprehension across two sources
+(`[e for x in xs for y in ys]`, join or not) is keyed by `(𝑖, 𝑗)`, and the induction recurrence is
+sequenced by `UInt` position throughout: the driver pairs items with `UInt` domain keys, the commit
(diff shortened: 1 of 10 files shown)


src/ccl/design/mutability.md
```diff
@@ -66,6 +66,23 @@ carry-forward, so channels never close a causal cycle — see [merge laws](#the-
...
+of the extent's rather than a renumbering of them, and the induction store's ticks follow them
+(position `𝑝` decides tick `𝑝 + 1`, and an excluded position's tick is never occupied). A read
+of the accumulator's history still spans the whole extent: at an excluded position it folds to
+the latest write below it, the value the accumulator held there. The alternative — iterating
+the extent densely and gating the write, which is how an in-body `if` compiles — would put the
+store at odds with the domain the type carries.
+
+A **product** loop source is rejected at op-conversion. A comprehension across two sources
+(`[e for x in xs for y in ys]`, join or not) is keyed by `(𝑖, 𝑗)`, and the induction recurrence is
+sequenced by `UInt` position throughout: the driver pairs items with `UInt` domain keys, the commit
```
sortalongo: This is an undesirable limitation, but it reads as a decision.

src/ccl/design/mutability.md
```diff
@@ -66,6 +66,23 @@ carry-forward, so channels never close a causal cycle — see [merge laws](#the-
...
 
+A **filtered** loop source is the same row read literally: its domain is the refined extent
+`{[0, 𝑁] | 𝑝}`, so the recurrence runs at the positions that survived the filter and at no
+others. `restrict` keeps an element at its original position, so those positions are a subset
+of the extent's rather than a renumbering of them, and the induction store's ticks follow them
+(position `𝑝` decides tick `𝑝 + 1`, and an excluded position's tick is never occupied). A read
+of the accumulator's history still spans the whole extent: at an excluded position it folds to
+the latest write below it, the value the accumulator held there. The alternative — iterating
+the extent densely and gating the write, which is how an in-body `if` compiles — would put the
+store at odds with the domain the type carries.
```
sortalongo: rm cruft. Has nothing to do with the semantics being documented here.

src/ccl/design/type-inference.md
```diff
@@ -615,15 +615,15 @@ in the `provenance-design` note under projects/program-inspector in the internal
...
-rewrite.
+`expr.ty`: the node's own type, its `user_annotation`, a `Cast`'s `target`, a
+`Transact`'s `domain`, and — per binder — both the binder's declared type *and* its
+annotation. Each holds an independent predicate `Rc`. `Expr::walk_type_slots{,_mut}`
+is the single source of truth for that set, precisely because hand-rolling it per
+pass is how a pass silently acquires a blind spot — `Cast.target`, where a
+comprehension filter's predicate actually lives, is the one that costs most.
+`count_free`/`is_free` are on it too, and that is not cosmetic: several passes *skip
+work* when `is_free` says no, so a slot the free-variable walk cannot see is a slot
+those passes decline to rewrite.
```
sortalongo: more unnecessary docs changes. The list being updated here should not be taken as exhaustive. It's just examples.

src/ccl/design/type-inference.md
```diff
@@ -635,13 +635,18 @@ A walk over the slot set must still cover it, because the passes that use
...
 slot the walk misses is a slot the check cannot report on.
 
+A `Transact`'s **sequencing domain** is on the walk for the same reason. The node is
+born by `plan_loops` and its domain is the extent of the source it iterates,
+refinements and all, so a `mut` accumulator over a filtered collection carries that
+filter's predicate there as well as on every other slot the same extent reaches.
+Planning's predicate compilation runs through this walk and the post-planning
+`typecheck` compares refinements structurally, so a domain the walk skipped would
+hold the bare predicate while its siblings held the compiled one, and the carrier
+would contradict its own type.
```
sortalongo: more useless content

sortalongo (approved):

src/ccl/design/mutability.md
dpmills: Fixed

src/ccl/design/mutability.md
dpmills: Done

src/ccl/design/type-inference.md
dpmills: Reverted

src/ccl/design/type-inference.md
dpmills: Deleted

PR #139, opened 2026-08-28
Compile a `match` in a loop body and a transaction block, and tag a reply tap with its firing

dpmills:
A `for`-loop body rejected `match` and a `with begin():` block admitted no tag dispatch, so per-element dispatch went through a `def` call — which computes a value per element but cannot write the loop's accumulators or feed from an arm. Both statement contexts now take a `match`: a statement-position tag-`Case` is rewritten into the boolean-guard `Case` its consumers already compile, and a reply tap is retyped `` {`fired{𝑉} | `idle} `` so a fed value may read the arm it sits in. An arm becomes a control-flow path like an `if` arm's, and `` out << v `` inside `` case `data(v): `` compiles beside an accumulator the other arms write.

## Tag dispatch as a boolean guard

`ccl_utils::tag_case_to_guard_case` turns each arm's `Pattern` into a ``variant_is(`tag)`` guard over the scrutinee and substitutes ``variant_project(`tag)`` for the payload binder rather than binding it; a `match` with no `case _:` gains the trailing `true → unit` carry. `Builtin::VariantIs` is the new total tag test — its doc comment carries why a decision's guard position cannot hold a restricting projection — with a tile operator beside `VariantProject`. `mut_elim` and `channelize` convert at their own consumption points; `transact_phase::strip` converts a whole block once, since a footprint scan and a decision walk both read it.

## A tap's tag replaces its gate

`F_FIRE_SUFFIX` and its companion `to_<defer>__fire : Bool` field are gone: a tap holds `` {`fired{𝑉} | `idle} `` (`V_FIRED`/`V_IDLE`), and a channel read composes ``variant_project(`fired)`` onto the field. A gate beside a same-position value obliged the value to answer wherever the record commits, which an arm's payload projection cannot. [mutability.md](src/ccl/design/mutability.md#general-in-transaction-conditionals-and-conditional-writes) and [design-operators.md](src/interpreter/design-operators.md#the-decision-record) hold the encoding, and `body_decision_at` reads the tag where it read the gate.

## One `match` production, three contexts

`lower_match_over` generalizes `lower_match` over how an arm's body lowers, so the tag rules and the arm's payload scope are shared by statement lowering, a loop body and a transaction block. The loop's yield, feed and mutation-variable walks recurse into arms, and the guarded-induction-write diagnostic names a `match` arm beside an `if`.

## Tests and spec

`tests/compilation_pipeline/variants.rs` runs from a bare two-arm `match` to a tap value reading its payload and the accumulator; `transactions.rs` adds in-block dispatch, an arm-fed reply and the rejection. `variant_project_forwards_a_partial_domain_release_to_its_scrutinee` replaces a `should_panic` case. The spec updates [`match`](docs/chl-spec.md#410-match--tag-dispatch) and [transactions](docs/chl-spec.md#82-transactions-with-begin).

### DIFF
--- a/tests/compilation_pipeline/variants.rs
+++ b/tests/compilation_pipeline/variants.rs
@@ -1441,3 +1441,105 @@ p"#;
 fn test_const_lifted_arm_is_point_free(#[case] code: &str, #[case] expected: Value) {
     check_scalar(code, expected);
 }
+
+// A `match` in a for-loop body dispatches on `variant_is` guards, so its arms
+// are the boolean-guard `Case` an `if` chain builds and the writer decision
+// merges them the same way (`src/ccl/design/mutability.md`, "A `match` in a
+// for-loop body").
+//
+// Selecting an arm by *reading* it — the fan-out of `variant_project`s a `match`
+// compiles to everywhere else — cannot drive a writer body: the decision selects
+// on a predicate over its whole driver domain, and a projection restricts that
+// domain rather than answering over it.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[case(indoc! {"
+    acc := 0
+    for m in [`a, `b, `a]:
+        match m:
+            case `a:
+                acc += 1
+            case `b:
+                acc += 10
+    acc"}, Value::Int(12))]
+// `case _:` is already the fallback the trailing `true` guard names.
+#[case(indoc! {"
+    acc := 0
+    for m in [`a, `b, `a]:
+        match m:
+            case `a:
+                acc += 1
+            case _:
+                acc += 10
+    acc"}, Value::Int(12))]
+// An arm that writes nothing is the carry position, as an `if` without an
+// `else` is.
+#[case(indoc! {"
+    acc := 0
+    for m in [`a, `b, `a]:
+        match m:
+            case `a:
+                acc += 1
+            case `b:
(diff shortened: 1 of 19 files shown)


tests/compilation_pipeline/variants.rs
```diff
@@ -1441,3 +1441,111 @@ p"#;
...
+fn test_match_arm_reading_its_payload_in_a_loop(#[case] code: &str, #[case] expected: Value) {
+    check_scalar(code, expected);
+}
+
+/// A `yield` or `<<` inside a `match` arm in a for-loop body is rejected: a
+/// conditional feed rides its path's predicate as a fire gate, and a `match`
+/// arm is selected by its tag rather than by a predicate.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+fn test_feed_inside_a_match_arm_in_a_loop_is_rejected() {
```
sortalongo: This justification sounds contrived. Is it just a temporary limitation? If so, document it as such.

sortalongo (approved):

tests/compilation_pipeline/variants.rs
dpmills: There's been enough change in the codebase over the past two weeks that this isn't actually too hard to just fix now, so I've done that instead of pinning the error

PR #141, opened 2026-08-28
inspector_model: build the inspector's read-only payload from the retained IR panes

groundlar:
The pane folds on `CompiledProgram` produce a source projection per retained pane and a node→node map per adjacent pair, and outside the leak gate nothing consumes them. This adds `src/inspector_model/`, which turns a `CompiledProgram` into one `InspectorPayload` — the source text, a node table and a span table per pane, the links between adjacent panes, source-level name resolution, and diagnostics — so `cambra-inspector` can compile a program once and serve one document its frontend answers every interaction from. The layer's reference is [`src/inspector_model/design.md`](src/inspector_model/design.md); the module sits in the core crate because it walks `ccl` internals, and its wire types derive `Serialize` only under the default-off `serde` feature, so a plain build compiles none of it.

## Every static fact ships; nothing answers a position

The module offers no positional lookup. `SpanIndex` and `NameBinderIndex` build their tables and enumerate them onto the wire; "which node is at this position" is the consumer's question, answered over the shipped `(span, node)` rows and the shipped tree, whose depth breaks a tie between byte-identical spans — see [The usage model](src/inspector_model/design.md#the-usage-model). The entry surface is `InspectedProgram::new` and `build_payload`.

## The pane set is the compiler's

`panes[]` carries one entry per `PANES` entry in pipeline order and `paneLinks[]` one per adjacent pair, each pane's `id`, `label` and `kind` read off what that pane declares. Adding a pane is an edit to `PANES` and nothing here. `ANCHOR_PANE` names `post-inference` rather than indexing it, and `InspectedProgram::new` panics if `PANES` does not declare it, so a pane inserted ahead of it cannot silently move the anchor.

## A node table, not a tree

A pane ships each of its nodes exactly once in first-visit pre-order, children named by id. A refinement predicate is an expression tree with `NodeId`s of its own and those ids are pane-link endpoints, so every walk here descends into predicates, and a predicate shared across type slots is one entry that several `where.N` edges name ([Predicates are nodes](src/inspector_model/design.md#predicates-are-nodes)). A node's type ships as its `Display` rendering plus `typeKind` and `predicateRefs` — the two facts a consumer cannot recover from the string ([Types on the wire](src/inspector_model/design.md#types-on-the-wire)). Pane links ship dense, self-edges kept, each edge carrying its `descends`/`relates` label set ([Pane links are dense and labelled](src/inspector_model/design.md#pane-links-are-dense-and-labelled)).

## A binder's type is recovered, not read

A binder is not a node, so nothing attributes it. `InspectedProgram::binder_type` searches the anchor pane among the nodes covering the binding site — by binder name, then by exact span equality for a `def` whose binder monomorphization replaced — and answers `None` for a substituted multi-param parameter. Either of two channels would replace the search with a read; both are follow-up work, carried as `TODO(binder-site)` and stated in [A binder's type is the binder's](src/inspector_model/design.md#a-binders-type-is-the-binders).

## One error model, two renderers

`Diagnostic::from_compile_error` renders a `CompileError` as `{severity, stage, message, span, labels}`, reusing each variant's `Display` so the JSON and the terminal's ariadne label carry the same line. That needs `Display` on `LoweringError` and `LexError` and a `span()` accessor on `LexError` and `ParseError`; `ParseError`'s `Display` renders a nested lex error through that `Display` rather than `Debug`. A failed compile ships `InspectorPayload::degraded` — the same payload type, empty panes, `meta.payloadKind: "failed"` ([Diagnostics and the degraded payload](src/inspector_model/design.md#diagnostics-and-the-degraded-payload)).

Supporting `ccl` changes: `NodeId::as_u64`, for projecting an id onto the wire outside a `Serialize` context; `pane_trees` visible to the crate; and explicit lifetimes on `walk_binders`, `walk_type_slots` and `walk_children` so a caller can retain what the closure hands it. `src/ccl/design/provenance.md`'s inspector-consumers section hands the payload shape off to the new reference and keeps only what this layer reads out of the provenance layer.

## Tests

31 unit tests, each in the module it exercises. `wire.rs`'s run over a three-program corpus — a comprehension, a monomorphized definition, a mutable loop — and pin what the consumer's validator re-checks: one pane per declared pane and one link per pair, every link endpoint and child id live in the table it points into, one table entry per node, no repeated span row, every node attributed.


src/inspector_model/design.md
```diff
@@ -0,0 +1,407 @@
...
+
+What the rendering still forecloses: a consumer cannot diff a node's type across panes except by
+string comparison, and cannot elide a long type by structure.
+
+A node's `where.N` children cover every type slot it carries — its own type, an annotation, a `Cast`
+target, each binder's — so they cannot say which predicate refines the node's *own* type.
+`predicateRefs` is that subset, and it is the leading part of those children, since the walk reaches
+the node's own type first.
+
+### A binder's type is the binder's
```
dpmills: All of this binder stuff seems confusing.  Do we need it? If binders aren't nodes yet, they could just be not independently clickable (so clicking on them would highlight the whole let).  When we build one of the real designs later, we could add the missing functionality

Claude found:

`scopes` is consumed by no production frontend code on any branch through `skylar/provenance/11-pane-filter`. It appears in `web/src/types.ts` (declaration) and `web/src/wireValidate.ts` (validation) and nowhere else. `predicateRefs` has the same status.

What ships in `arithmetic.snapshot.json`: 10 scope rows, 24 bindings, 2.3 KB, carrying 4 distinct facts — two of them `None`.

| binder | `defSpan` text | `type` |
|---|---|---|
| `g` | `g` | `Int@10` |
| `p` | `p` | `None` |
| `q` | `q` | `None` |
| `f` | `def f(p, q):\n  p + q + g\n` | `((Int@1, Int@2) ⇒ Int)` |

And the scenario this section is written to fix is not fixed by the shipped stack. A source-side hover goes through `typesAt` in `web/src/indices.ts`, which is span containment over `spanIndex`, so clicking `g` in `g = "abc"` still answers with the covering `Let`'s type — the wrong answer this section opens with. The machinery is built, shipped and unread.

So "not independently clickable" costs nothing anyone currently gets. Deleting the path removes `binder_type` + `binder_at_site` (101 lines) and their three tests (116 lines); `ANCHOR_PANE` and `anchor()`, whose only non-doc caller is `binder_type`; `ScopeEntry`/`ScopeBindingEntry` and their emission; and `NameBinderIndex::scopes` + `visible_bindings` + `region_key` + `Event::Scope` (~70 lines plus 4 emission sites).

src/inspector_model/name_binder.rs
```diff
@@ -0,0 +1,860 @@
+//! The source-level name-resolution index over the parsed CHL AST.
+//!
+//! This is the [`NameBinderIndex`] — the source-level answer to lexical name
```
dpmills: Having to build all this machinery just because our multi-arg function calls are weird feels bad, especially since the current function call mechanism doesn't handle refined types well anyways and probably need to change.  Maybe there's some smaller bit of metadata we could capture instead that would allow us to keep all of the scoping logic in one place in CCL?

Claude found:

CCL already has the use→binder relation this module reimplements. From `src/ccl/names.rs`: `Name::Unique` carries a `Uid`, "identity is the `uid`", "copies preserve the `uid`", and "after uniquification every binder is `Unique`" is debug-checked by `assert_all_binders_minted` (`src/ccl/uniquify.rs`). Uid equality *is* lexical resolution, post-uniquify, capture-free, by construction.

The only missing fact is the source span of a binder's name site. `TypedBinding` is `{name, ty, user_annotation}` with no span; `Param` already carries `name_span` (`src/chl_parser/ast.rs`). That is the smaller bit of metadata — one `Option<Span>`, after which both `definitions` and `binder_type` become reads rather than searches.

Uid is not injective (lowering copies pre-minted subtrees, so a uid can sit at several binding sites), but those sites are copies of one source binder and share its name span, so `uid → name_span` is a function even where `uid → site` is not.

The two places uid resolution genuinely fails are the two places a pass swaps a source binder for a compiler one, and both already know what they replaced:

- **Monomorphization** — `SyntheticKind::Mono(Box<Name>)` already carries the source binding's `Name`. With a site on the binder it resolves transitively, no new plumbing.
- **`uncurry_params`** — `fresh_tuple_arg` returns a plain `String` (`src/ccl/lower/mod.rs`), so `__arg_tuple_N` lowers as a `Name::Raw` and uniquify mints it to a `Name::Unique` whose `base` is `"__arg_tuple_0"`. No `SyntheticKind` covers it. That contradicts what `names.rs` says `Unique` means — "`base` is always a real identifier (trustworthy in errors)", "every binder traces to source". The pass holds `params[i].name_span` at the substitution site and drops it.

So this module is 860 lines paying for one missing `SyntheticKind` variant. Closing that gap where the information is — a kind for the uncurried tuple binder carrying its per-position name spans, plus the site on `TypedBinding` — lets CCL's own uid relation answer goto-definition in about 25 lines. Same fix as `TODO(binder-site)`, placed at the pass that broke the invariant rather than bolted onto the consumer.

Also worth stating in the PR either way: this is the fourth place CHL scoping is written, after `src/ccl/scope.rs`, uniquify's walk, and lowering's `let`-nesting.

src/inspector_model/span_index.rs
```diff
@@ -0,0 +1,166 @@
+//! The span → node index over one pane's IR tree.
+//!
+//! A pane's [`SourceProjection`](crate::ccl::provenance::SourceProjection) maps
```
dpmills: Why do we need this and SourceProjection? The comments here describe this as a one-directional map, but the actual data structures are non-directional entries, so consumers of this can presumably look things up in either direction

Claude found:

Confirmed, and the table carries no information the node table does not. Measured across all seven fixtures, every pane:

| fixture | nodes | `spanIndex` rows | nodes with no span |
|---|---|---|---|
| arithmetic | 157 | 157 | 0 |
| udf_fanout | 162 | 162 | 0 |
| polymorphic | 189 | 189 | 0 |
| mutation_loop | 98 | 98 | 0 |
| list_min | 128 | 128 | 0 |
| defer_lift | 200 | 200 | 0 |
| txn_multi_read | 157 | 157 | 0 |

Rows equal nodes everywhere, so every node has exactly one span, so `spanIndex` is literally `(node.span, node.nodeId)` — 5–10 KB per payload of duplicate. There is also no lookup in either direction: `entries()` only enumerates, and `build_pane_nodes_and_index` walks the pane twice to build two views of one `(tree, projection)` pair.

Multi-span attributions are real in the model — `fold` unions blame spans, pinned by the `attr.spans == vec![s1, s2, s3]` case in `src/ccl/provenance.rs` — so the equality above is a corpus coverage gap rather than an impossibility. That argues for `spans: Span[]` on the node entry, not a parallel table: the node-table walk already visits every node and already reads `attr.spans` to pick the narrowest. `span_index.rs`, its round-trip test and `SpanEntry` all go, and the frontend's `tightestNodeAt`/`typesAt` iterate `pane.nodes`.

src/inspector_model/walk.rs
```diff
@@ -0,0 +1,98 @@
...
+/// The label is for display; the edge's `predicate` flag is what a consumer
+/// branches on to tell "this subtree lives inside a type" from "this subtree is
+/// an operand". Order is [`TypedExpr::walk_type_slots`] order, which is stable,
+/// so the labels are stable too, and a consumer can compare one node's predicate
+/// edges across panes.
+///
+/// Mirrors `collect_tree_ids`' type-slot descent, and must: a predicate that walk
+/// enumerates and this one does not is a node the fold explains and the table
+/// omits.
+pub(super) fn predicate_children(expr: &Expr) -> Vec<(String, &Expr)> {
```
dpmills: There's probably other code that needs to walk all Exprs including inside refinements.  Also, this is going to output duplicate node IDs when the same refinement shows up in multiple places; is that ok? This also discards type structure.  If you have a top level refinement or a refinement on a single field inside a tuple, those will both show up the same way here

Claude found:

There are three copies of this descent, and one of them documents the coupling instead of removing it:

| site | what it collects |
|---|---|
| `collect_tree_ids::from_ty`, `src/ccl/context.rs` | the id domain the fold must explain |
| `predicate_children::from_ty`, here | the `where.N` child edges |
| `own_type_predicates::walk`, `src/inspector_model/wire.rs` | `predicateRefs` |

This function's doc says "Mirrors `collect_tree_ids`' type-slot descent, and must". The invariant is real, so it wants to be one function in `ccl` yielding `(position, &Expr)` pairs that all three callers project from.

Both of the other concerns reproduce in the shipped `arithmetic` fixture, in `post-inference`:

```
Lambda(__arg_tuple_0) : ((Int@1, Int@2) ⇒ Int)
  predicateRefs: [47, 50]
  where.0 → 47   where.1 → 50   where.2 → 47   where.3 → 50

Proj(Index(0)) : ((Int@1, Int@2) ⇒ Int@1)
  predicateRefs: [47, 50, 47]
```

Four predicate children naming two ids: `walk_type_slots` yields the lambda's own `ty` and its binder's `ty`, which for a lambda are the same type, so `where.2`/`where.3` repeat `where.0`/`where.1` — duplication from the slot walk rather than from genuine sharing. Three nodes in this pane alone ship duplicate child ids.

`predicateRefs: [47, 50, 47]` carries one id twice in one list: `Int@1`'s predicate sits in the domain tuple and again in the codomain. Structurally distinct positions, identical wire representation. `where.N` being a flat counter over `walk_type_slots × from_ty` order is what discards the structure; a path (which slot, then the position inside that type) keeps it.

`predicateRefs` has no consumer — like `scopes`, it is declared in `web/src/types.ts`, checked in `web/src/wireValidate.ts`, and read nowhere. So the cheap resolution is to drop it, keep the `predicate` flag (the frontend does use that, for the depth tie-break in `indices.ts`), and make the edge label a path if a consumer ever needs the positions apart.

src/inspector_model/wire.rs
```diff
@@ -0,0 +1,1771 @@
...
+pub const SCHEMA_VERSION: u32 = 5;
+
+/// The `GET /api/snapshot` bulk payload — the whole static read-only model.
+///
+/// Field order/names mirror the schema. `outline` is intentionally absent (see
+/// the module note); `diagnostics` is always empty on the success path.
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct InspectorPayload {
```
dpmills, line 71: Five of this payload's field families are written and never read. Grepping every production `.ts` on the follow-up branches through `skylar/provenance/11-pane-filter` (excluding `types.ts` declarations and `wireValidate.ts` checks):

| field | producer | read by |
|---|---|---|
| `scopes[]` + `bindings[].type` | `NameBinderIndex::scopes`, `binder_type` | nothing |
| `nodes[].predicateRefs` | `own_type_predicates` | nothing |
| `nodes[].typeKind` | `type_kind` | nothing |
| `paneLinks[].edges[][2]` | `wire_labels` | nothing |
| `meta.tick` | one field | nothing |

Measured on `arithmetic.snapshot.json`:

```
total payload           63736
scopes                   2292
typeKind                 2986
predicateRefs            3042
edge label sets          1807
--- unconsumed          10127  (15%)
spanIndex (redundant)    8199  (12%)
```

27% of the payload is fields nothing reads plus fields the node table already carries.

Two of these have rationales that do not survive the measurement. `typeKind` exists because "a rendered type gives a consumer nothing to branch on" — and no consumer branches on `typeKind` either; same for `predicateRefs`, so the "two narrow facts ride beside the string" pair is 6 KB per program of unread payload.

The edge label set is the one to be careful with: it is dead *on the wire* but live in `ccl`. `links.ts` declares `PaneEdge = [number, number, EdgeLabel[]]` and its resolver never reads element `[2]` — resolution is bidirectional and transitive, so it treats `descends` and `relates` identically. Upstream the distinction is load-bearing (`has_ancestry` drives leak accounting in `src/ccl/provenance.rs`), so the suggestion is to stop projecting it, not to stop tracking it.

Taken together these read as one thing rather than five: the wire is specified against an anticipated consumer rather than the built one, and most of the machinery under review serves the specification.

src/inspector_model/wire.rs
```diff
@@ -0,0 +1,1771 @@
...
+    pub span: Option<Span>,
+    /// Labelled spans (one per pointed-at range). Empty when no span is known.
+    pub labels: Vec<DiagnosticLabel>,
+}
+
+/// A `{ span, message }` label inside a [`Diagnostic`].
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct DiagnosticLabel {
```
dpmills: `labels` is `span` and `message` copied. `from_compile_error` builds it as `span.map(|span| vec![DiagnosticLabel { span, message: message.clone() }])`, so every diagnostic carries zero or one label and the label duplicates two fields already on the parent — it can hold nothing else by construction.

Nothing reads it: `diagnosticLines` in `web/src/main.ts` renders `severity`, `stage`, `message` and `span`. `DiagnosticLabel` and the field can both go, and `from_compile_error` loses the `message.clone()`.

If the intent is that a future diagnostic points at several ranges with distinct texts, that is a different type from this one — it would need the labels to carry their own messages, which means the producer has to have them, which is the part that does not exist yet.

src/inspector_model/wire.rs
```diff
@@ -0,0 +1,1771 @@
...
+use crate::chl_parser::ast::Span;
+
+use super::name_binder::{Definition, ScopeRegion};
+use super::program::{InspectedProgram, PaneProjection};
+use super::walk::{node_label, predicate_children};
+
+/// The current `/api/snapshot` wire-format version, emitted as `meta.schema` on
+/// both the success and degraded payloads. See [`Meta::schema`] for the
+/// versioning contract and the current version's field set.
+pub const SCHEMA_VERSION: u32 = 5;
```
dpmills: Versions 2 and 3 existed only on unmerged commits of this stack (`b602524b9` "schema 2", `1bcbcdd9d` "schema 3"). Nothing outside the stack has ever read 1–4, so a wire whose stated purpose is "a client reads this to detect an incompatible payload" lands by encoding this PR's own iteration history into a public contract. It should be 1.

Related, on `Meta::payload_kind` below: the two values are `"post-inference"` (compiled) and `"failed"`, but `"post-inference"` is simultaneously a pane id — it is in `PANES`, it is `ANCHOR_PANE`, and it appears in `panes[].id` and `paneLinks[].to`. One spelling, two unrelated meanings on one wire. `main.ts` renders it as a header badge, so a successful compile displays the text "post-inference" to the user.

`inspector_model: a diagnostic's stage is a compiler stage, not a pane` fixes exactly this confusion for `Diagnostic::stage`. The same fix wants applying here — the success value is a payload kind (`"program"`, `"ok"`), never a pane name.

src/inspector_model/mod.rs
```diff
@@ -0,0 +1,83 @@
+//! The program inspector's read-only model over the retained IR panes.
+//!
+//! `src/inspector_model/design.md` is the reference for this layer: what the
+//! payload carries, when it is built, and what the wire promises. See "The usage
+//! model" there before adding an entry point, and "Decided, not yet built" for
```
dpmills: The citation two lines up is broken and CI cannot catch it. There is no `Decided, not yet built` heading in `design.md`.

`./ci.sh doc_refs` passes green — I ran it — because a quoted title is only checked when it is *directly adjacent* to the doc path, and here the path is followed by "is the reference for this layer:", whose first ordinary word ends the title run (`.github/scripts/doc-refs/README.md`, "The quoted-title citation convention"). So both quoted titles in this sentence are unchecked; one happens to be a real heading and one never was. Making the citation adjacent puts it back under the checker and would have caught this.

src/inspector_model/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+//! consumer over the shipped `(span, node)` rows and the shipped node table,
+//! which is the copy that runs. See `src/inspector_model/design.md`, "The usage model".
+
+mod name_binder;
+mod program;
+mod span_index;
+mod walk;
+mod wire;
+
+pub use name_binder::{Binding, Definition, NameBinderIndex, ScopeRegion};
```
dpmills: `cambra-inspector` — this module's only declared consumer — names exactly five items: `InspectedProgram`, `InspectorPayload`, `Diagnostic`, `diagnostics_from_compile_errors`, `SCHEMA_VERSION`. This list exports 19.

The wire *field* types need `pub` to be nameable through `InspectorPayload`'s public fields, so most of the `wire` re-exports are unavoidable. These are not field types and have no caller outside the module: `SpanIndex`, `NameBinderIndex`, `Binding`, `Definition`, `ScopeRegion`, `dense_edges`. They are builders, and the module docs say the payload is the thing that ships.

src/inspector_model/program.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    /// Build from a single materialized pane, named by `pane`, which becomes the
+    /// anchor. Equivalent to [`new`](Self::new) for one pane: there is no
+    /// adjacent pair, so [`build_payload`](Self::build_payload) ships a single
+    /// pane with no pane links.
+    ///
+    /// `pane` must be one of [`PANES`]' names — a pane's wire id is the pane's
+    /// declared name, so a model built over the post-channelize tree must say
+    /// so rather than inherit the anchor's label.
+    pub fn from_parts(
```
dpmills: `from_parts` is `pub` with one caller — a test in `wire.rs` — and it produces a *second payload shape*: one pane, no pane links.

That contradicts the module docs, which say `InspectedProgram::new` and `build_payload` are "the module's whole entry surface"; `design.md`'s API-shape table lists `from_parts` as an entry surface alongside them. The two docs disagree, and the test-only shape is why.

If it exists to let a test build a model over a single pane, it wants to be `#[cfg(test)]` or `pub(super)` and out of the API table. If a real caller is coming, the docs should say which of the two claims is true.

src/inspector_model/design.md
```diff
@@ -0,0 +1,407 @@
...
+the **payload** is what ships.
+
+Unit tests live in the module they exercise. `wire.rs`'s run over a three-program `corpus()`
+chosen for the shapes that reach the payload's moving parts: a comprehension for refinement
+predicates, a monomorphized definition for one source span over several nodes, and a mutable loop
+for a recurrence.
+
+## Design considerations
+
+### The wire belongs to this module
```
dpmills, line 314: This section is codebase history rather than design. "Before, the payload rode `pretty_tree::InspectNode` — a rendering type holding the tile-producer fields `annotations` and `tiling`, which the payload never set and no consumer read, plus a hand-written `Serialize` spelling this module's schema. A wire change was then a diff in a renderer." And "There were two, the second on `SourceAttribution`, each with its own copy of the null-compression rule; it went with the query results that were its only reader."

None of that is checkable against the code as it stands, which is the test CLAUDE.md sets for a comment and the doc of record inherits. The invariant is one sentence — `pretty_tree` carries no serde and no `ccl` type, and one encoder writes a `RewriteTag` to the wire — and the before/after belongs in the PR description. §Cost has the same shape: the `2,443 node positions for 465 nodes` measurement justifies a change already made.

Six lead-ins in this document are argument rather than record, against "a design doc records a decision rather than arguing for it":

- `**Status.**` (13) — a status callout
- `Two rules follow.` (36) — a sentence announcing another sentence
- `The choice is provisional.` (175)
- `What the rendering still forecloses:` (187)
- `**Why the payload cannot simply say so.**` (227) — answers an objection nobody raised
- `**A better design exists and is not built.**` (251) — an apology

The last two are the clearest. 227 wants to be the invariant it is dancing around ("no channel names a binder site", which §What the model cannot say already states), and 251 wants to be the `TODO(binder-site)` and nothing else.

src/inspector_model/design.md
groundlar: Done, and then some — the machinery is gone rather than reworked.

Deleted: `binder_type`, `binder_at_site`, `ANCHOR_PANE`/`anchor()`, `ScopeEntry`/`ScopeBindingEntry`, `NameBinderIndex::scopes`/`visible_bindings`/`region_key`, and the `scopes[]` wire field.

Your suggested fallback is what ships. Clicking a binder resolves the node whose span contains it — the enclosing `Let` — and a definition's `defSpan` is now that binding's span too, so a use of `g` in `g = 10` jumps to the statement rather than to the name. `design.md`, "What the model cannot say" states the gap.

src/inspector_model/design.md
groundlar: Done. All six lead-ins are gone (`**Status.**`, `Two rules follow.`, `The choice is provisional.`, `What the rendering still forecloses:`, `**Why the payload cannot simply say so.**`, `**A better design exists and is not built.**`).

The `pretty_tree::InspectNode` before/after and the `There were two…` paragraph are replaced by the two invariants they were narrating, and §Cost's `2,443 node positions` measurement — which justified a change already made — is gone.

src/inspector_model/mod.rs
groundlar: Done, and thank you for the mechanism rather than just the instance. The citation is adjacent to the doc path so `doc_refs` checks it, and it names a real heading.

The adjacency rule went on to catch two further dangling citations in this work — when headings moved, and when `name_binder.rs` was deleted. Neither would have been caught otherwise.

src/inspector_model/mod.rs
groundlar: Done. The list is 12 items, all wire field types reachable through `InspectorPayload`'s public fields.

Every builder you named is private or gone: `SpanIndex`, `NameBinderIndex`, `Binding`, `Definition` and `ScopeRegion` no longer exist; `dense_edges` is `pub(super)`.

src/inspector_model/name_binder.rs
groundlar: Done, and this one drove most of the rest.

You were right that CCL already has the relation: uid equality is lexical resolution after uniquification. `name_binder.rs` is deleted — 664 lines out, 135 in (`definitions.rs`), resolving uses to binders by uid over the pre-inference pane. Pre-inference because monomorphization copies a `def` body per specialization, so a later pane would report each use once per copy.

One correction to my first reply: it was the **third** implementation of CHL scoping, not the fourth. Lowering's `let`-nesting encodes the rule but resolves nothing.

The smaller metadatum you predicted turned out not to be needed at all. I did build it — a source span on `TypedBinding` — and then removed it, because a definition can take the span of the node that binds it. That was #144, now closed; the only piece kept is three lines tagging a substituted parameter's projection with `Param::name_span`, a field that already existed.

The one place the surface walk did better: a defer handle used at a `<<` feed site. `Feed`, `Define` and `MutWrite` name a binder in a *field* rather than a node, so the only span available is the whole node's, which would contain and shadow the narrower uses inside it. Those two rows (of 31 across the corpus) are dropped rather than emitted wrong.

src/inspector_model/program.rs
groundlar: Done. `#[cfg(test)] pub(super)`, and out of the API-shape table. The docs now agree that `new` and `build_payload` are the entry surface.

src/inspector_model/span_index.rs
groundlar: Done. `span_index.rs`, `SpanEntry` and `PaneEntry::span_index` are deleted; a node carries `spans: Span[]`, narrowest first, each once.

Your reading was right, and the multi-span case is exactly why it is a list rather than one span. `a_nodes_spans_are_its_attributions_narrowest_first` and `no_node_repeats_a_span` pin it.

src/inspector_model/walk.rs
groundlar: Done on all three.

**Duplicate ids.** Predicates are deduplicated by node id, pinned by `no_node_repeats_a_predicate_edge`. It was 40 nodes across two fixtures, not the three I first reported: `walk_type_slots` yields a `Lambda`'s own type and its binder's type, which for a lambda are the same `Type`.

**Three copies.** `Type::walk_refinements` is now the one descent, and the callers project from it — this module's `where.N` edges, `collect_tree_ids`' id domain, and `predicate_id_collisions`' per-term `Rc` keys. Dropping `predicateRefs` removed a fourth. `walk.rs` no longer says it *mirrors* `collect_tree_ids`' descent; it shares it, which is what made the coupling safe to assert.

**Discarded structure.** `where.N` stays a flat counter rather than a path, deliberately: nothing on the wire needs the slots told apart now that `predicateRefs` is gone. A path is what a consumer would need to tell two slots apart, and `walk.rs` says so.

src/inspector_model/wire.rs
groundlar: Done both. `SCHEMA_VERSION` is 1.

The success `payloadKind` is `"program"`, and `a_payload_kind_is_never_a_pane_id` checks both values against `PANES` so the collision cannot return.

src/inspector_model/wire.rs
groundlar: Done. `scopes[]`, `bindings[].type`, `predicateRefs`, `typeKind`, `meta.tick` and `paneLinks[].edges[][2]` are off the wire.

`EdgeLabels` stays in `ccl`, where `has_ancestry` drives leak accounting — only the projection went, as you advised.

src/inspector_model/wire.rs
groundlar: Done. `labels` and `DiagnosticLabel` are deleted and `from_compile_error` loses the `message.clone()`. The doc now states why a multi-range diagnostic would be a different type.

groundlar:
## Two findings from acting on this review

Neither was raised in review; both came out of the work and are worth separating from the thread replies.

### A defect in refinement scoping, reachable on `main`

A `let` in a function body whose binder reaches the result refinement produces an ill-formed function type. No multi-argument function is involved:

```chl
def g(n: Int):
    m = n ^+ 1
    m ^+ 1
```
```
Scope violation (compiler bug, design §6.2):
  type (Int ⇒ {Int | __elem == m ^+ 1})
  at λ n : Int → let m : {Int | __elem == n ^+ 1} = n ^+ 1 in …
  references out-of-scope binder(s) ["m"]
```

`^+` is `AddableRefined`, so its result type records the sum and the predicate holds the operand terms themselves. Inference derives that refinement from the body and lands it on the lambda's own type, where a `let` inside the body has not opened, so the predicate names a binder that is not in scope at the position the type sits. `ScopeViolation` is labelled a compiler bug, so an ordinary program reports an internal invariant failure.

The contrast isolates it: `def h(n: Int): n ^+ 1` compiles, because its refinement names the parameter and the function type binds that — the type comes out as `((n: Int) ⇒ {Int | __elem == n ^+ 1})` with the reference closed into a `PiBound` index.

Pinned by `let_binder_escaping_into_a_result_refinement_is_ill_scoped` with a `TODO(refinement-let-scope)`, alongside `a_result_refinement_over_the_parameter_is_well_scoped` as the contrast. **The first test asserts the bug and is meant to start failing when it is fixed.**

The fix is to eliminate the binder as the refinement leaves its scope, by substituting the `let`'s definition into the predicate: `m` defined as `n ^+ 1` makes the refinement `{Int | __elem == (n ^+ 1) ^+ 1}`, whose only free name is the parameter. Two things to settle first — nested `let`s compose, so the predicate grows multiplicatively and wants a size bound; and a binder that cannot be eliminated should widen the type to the unrefined base rather than error.

This is also what blocks binding a multi-argument function's parameters rather than substituting them away, which would give each parameter a real binder and make a refinement over one render as `{Int | __elem >= a}` rather than `{Int | __elem >= __arg_tuple_0.0}`. Substituting the projection is currently what guarantees a tupled lambda's body mentions only the tuple binder, which is the one name the function type can bind. So the two are one fix, and it sits in the area the trait-and-refinements work is already touching.

### `CompiledProgram::source_ast` is removed

It retained the parsed surface AST on every compiled program so the inspector could resolve names over it, and its doc called that AST "the source-of-truth for source-level (lexical) inspector queries" — a claim that stopped holding when resolution moved onto the IR. Nothing else read it, so it was a whole `Module` kept alive per compile to answer a question nobody asks. `Frontend::module`, which carried it out of the frontend, is gone with it.

src/inspector_model/design.md
```diff
@@ -0,0 +1,343 @@
...
+`Infer`'s is resolved at the `compile_program` boundary; `ChannelizeDefers`, `LambdaElim`,
+`Conversion` and `Unsupported` carry none, so a consumer has nothing to underline for them.
+
+`meta` carries `payloadKind` — `"program"` for a compiled program, `"failed"` for a degraded
+payload — and `schema`. Neither kind is a pane id, so one word never names both a pipeline position
+and a document kind; `a_payload_kind_is_never_a_pane_id` (`wire.rs`) pins that. The schema's
+`outline` field and a live-value `tick` are both omitted rather than stubbed, until something reads
+them.
+
+### The schema version
```
dpmills, line 256: Probably worth a note here that we don't need to bump this during development; we have no durable users at this point, so it can stay at 1 for now regardless of what changes

src/inspector_model/design.md
```diff
@@ -0,0 +1,343 @@
...
+
+### Diagnostics and the degraded payload
+
+`diagnostics` is empty on a successful payload: it describes a program that compiled, and there are
+no warnings. A failed compile ships `InspectorPayload::degraded` instead — same type, so the two
+shapes cannot drift — carrying the source text, the diagnostics, `meta.payloadKind: "failed"`, and
+empty `panes`, `paneLinks` and `definitions`.
+
+A failed compile ships no panes even where the pipeline reached some: inference can fail after
+channelization succeeded, and that channelized IR is displayable. Shipping what a failed compile did
```
dpmills: This doesn't make sense.   Inference runs first, it can't fail after a later phase.  The justification for not shipping panes on a failed compile doesn't hold up.  
It's fine if we don't want to for now, but we shouldn't document that as a forced choice.

src/inspector_model/design.md
```diff
@@ -0,0 +1,343 @@
...
+`Rc` keys. Written out per caller it was the same match-and-recurse three times, and a predicate this
+module enumerated but `collect_tree_ids` did not would be a node the pane fold never explains.
+
+### Types on the wire
+
+A node's `type` is rendered by `Display for Type`, and `Type`'s `Serialize` impl delegates to that
+same `Display`, so every type on the wire is one rendered string. Every node has a type, so the
+field is never absent.
+
+A structural type would carry a refinement's predicate, which is an expression tree with node ids,
```
dpmills: Doesn't Display also carry the refinement? I think I see what this is trying to say, but it's probably unnecessary, and unlikely to be the final design since we'll almost certainly want richer type info later.  I think we can cut everything except the first paragraph of this section

src/inspector_model/walk.rs
dpmills: I looked deeper into the where.N labels, and I think they are unnecessary.  Given that they are also confusing (they lose positional info, they show up from places in the node that aren't the node's type), I think we should just cut for now until we have a clearer picture of how we want to represent refinements in the inspector.
We don't lose anything by doing this; the entire `edge` field carries 0 new information beyond the other fields.  (the boolean predicate field does the same work already since everything is already order-preserving)

After looking at this, I'm quite convinced that the current way refinements are being shipped is wrong, even after we do this, but it's good enough for now.  We should clearly mark it as provisional in the docs/code so that future work in the area knows to discard instead of building on top.  For simple cases, this structure blind version happens to do more or less the right thing, but for anything more complicated and if we want to add any other functionality in the inspector around refinements, it won't work

src/inspector_model/design.md
```diff
@@ -0,0 +1,343 @@
...
+A `ProvenanceMap` is the compiler's node→node relation between two adjacent panes: what each node of
+the upstream pane became, and what each node of the downstream pane came from. It ships verbatim,
+self-edges included, so a node preserved across the phase appears as its own `[id, id]` edge and the
+consumer may follow edges with no identity special case.
+
+An edge is `[upstream, downstream]` and carries no label. `EdgeLabels` distinguishes `descends` (the
+downstream node was made from the upstream one) from `relates` (it is about the upstream node but
+was not made from it), and that distinction is load-bearing in `ccl`, where `has_ancestry` drives
+leak accounting — see [provenance.md](../ccl/design/provenance.md#the-edge-labels). It stops at the
+wire: link resolution is bidirectional and transitive, so a consumer following edges treats the two
```
dpmills: There are a bunch of statements like this that read as real decisions but are actually just temporary.  It's not at all true that we never want to distinguish between parent and blame, we just don't happen to yet so the wire format doesn't surface it.

dpmills (approved):

src/inspector_model/design.md
groundlar: Reframed. The `descends`/`relates` description and the `has_ancestry` pointer stay; the "It stops at the wire" claim is now a **Provisional** paragraph — today's consumer resolves links bidirectionally and transitively so it treats the two alike, a consumer that separates descent from mention wants the label, and carrying it is an additive field. Same edit on `dense_edges`.

Swept the rest of the doc for the shape. The other two instances were the `where.N` and structural-type passages, both now cut outright. `outline`/`tick` ("omitted rather than stubbed, until something reads them") and the unresolved-name gap already read as temporary, so I left them.

src/inspector_model/design.md
groundlar: Cut to the first paragraph, plus one sentence marking the rendered string as provisional rather than a settled omission.

`Display` does carry it — `{T | p}` renders the predicate through `symbolic` — so the second paragraph's premise was false as written.

src/inspector_model/design.md
groundlar: Fixed. `src/ccl/context.rs` runs `Infer` before `Inline`/`Transact`/`Letrec`/`Channelize`, so the example was backwards.

It now states what happens: a program that fails inference has its pre-inference pane already built, and one that fails a later phase has that pane and every pane before the failure; all of them are dropped. The forced-choice framing is gone. The `TODO(degraded-panes)` comment on `InspectorPayload::degraded` carried the same backwards claim and got the same fix.

src/inspector_model/design.md
groundlar: Added, and mirrored on `Meta::schema`.

Also fixed `cambra-inspector/CLAUDE.md`, which instructed the next person to bump on any schema change — it now says the version stays at 1 and points here for what a bump means once a durable consumer exists.

Taken at its word in this PR: dropping the `edge` field is a breaking wire change and `SCHEMA_VERSION` is untouched.

src/inspector_model/walk.rs
groundlar: Cut — the whole `edge` field, not just `where.N`. `IrChild` is `{ id, predicate }`.

Nothing is lost on the consumer side, because the label is derived rather than dropped. Children ship in order (value children, then predicates), so `treeView` renders a value child's index in `children` as its prefix and gives a predicate none. Value rows are byte-identical to before — `0: Operand : Int #2`. A predicate row now reads `Pred : Bool #3` instead of `where.0: Pred : Bool #3`, in the pane and in the copied text alike.

Marked provisional in three places, each saying the shape gets replaced rather than extended: `predicate_children`, the `inspector_model` module doc, and design.md "Predicates are nodes".

The fallout is folded into the commits that own each piece rather than landing here: the Rust validator's `edge` assert (06), `types.ts` / `wireValidate.ts` / `indices.test.ts` (07), `serializeTree` and its tests (10), plus the fixture re-bless and bundle rebuild wherever the corpus moved. No `SCHEMA_VERSION` bump, per your note on the schema version.

PR #144, opened 2026-08-31
ccl: carry a binder's source site

groundlar:
A binder is not an expression, so nothing attributes one and no pane's node → span map can say which node binds the name written at a given source position. A consumer asking "where is this name defined" therefore had to re-resolve CHL's scoping over the parsed surface AST — a third implementation after `ccl/scope.rs` and `uniquify`'s minting walk, and the only one nothing checked. This adds the channel that makes the question two reads, so that implementation can go.

Uid equality is lexical resolution. After uniquification every binder carries a globally-fresh `Uid`, copies preserve it, and a bound occurrence names its binder — capture-free by construction. The one missing fact was where the binder was written.

## `TypedBinding::name_span`

The source span of the name at a binding site, `None` for a compiler-minted binder, which was written nowhere. Lowering sets it, `uniquify` rewrites `name` and leaves it alone, no later pass reads it.

The field sits **outside `PartialEq`**, and that is load-bearing. Structural equality on terms coincides with α-equivalence after uniquification, and every equality-mediated decision depends on it. Passes rebuild a binder from its name and type — `mut_elim`'s and `transact_phase`'s `binding` helpers do — so a rebuilt binder carries no span, and comparing spans would make a term unequal to its own image. `Name`'s derived equality is safe for the opposite reason: its `base` is minted with the `uid` and copies preserve both, so the two can never disagree.

## Every source binder form carries its site

Assignment targets (`=`, `x: T = e`, `:=`), a `def`'s own name, `def`/`lambda` parameters single and curried, `for` targets in both loop encodings, comprehension targets, and a match arm's named payload binder. Two functions set the field, split by what the lowering arm holds: `lower::with_binder_site` takes an already-built node, `TypedBinding::at_name_span` a binding still under construction. Neither the field nor the assignment appears at a lowering arm, so a binding form added later cannot ship `None` silently. `with_binder_site` debug-asserts that the node it is handed introduces a binder, and a caller pointing it at the wrong node fails on the next test run rather than in some program — the node kind is spelled out at the call site, never decided by the program being lowered. A binder with no source name goes through neither, which is how lowering's own plumbing keeps `None` by construction.

Two AST nodes gained a span, both mirroring `Param::name_span`, which already existed: `Stmt::FunctionDef::name_span` and `PayloadPattern::Named::name_span`. Those are the two source binders the parser was not recording a position for.

## A substituted parameter's site rides its projection

A multi-argument function's parameters bind nothing: `uncurry_params` substitutes `__arg_tuple_N ▷ .i` into every occurrence, so no `TypedBinding` corresponds to `p`. The projection is what became of the parameter, so it is now tagged with the parameter's declaration rather than the whole function's span — it was tagged with `fn_span` because the template is manufactured once per parameter before any occurrence exists. A substituted parameter therefore has two spans that together answer goto-definition: the projection root carries the occurrence (substitution is root-carry), and the projection interior carries the declaration (every freshened copy inherits the template's tag).

This uses the existing node → span channel and nothing else — no `SyntheticKind` carrying spans, no second channel for the pane folds.

The pair is readable only while no node carries both halves. A `SourceAttribution`'s `spans` is a set and the fold unions attributions along its edges, so this is a per-pane property rather than a property of the tree. The halves stay on distinct nodes through `pre-inference`, `post-inference`, `post-channelize` and `post-as-of-read`; `lambda_elim` rewrites the body point-free and carries neither span to its pane, so a consumer needing a parameter's site reads a pane above it.

**Binding the parameters instead does not work**, and was tried. `let p = __arg_tuple_0.0 in …` gives each parameter a real binder and reads better, but inference derives refinements from the body and lands them on the lambda's own type, where a body-internal `let` has not opened: `def add(a, b): a + b` infers `((Int, Int) ⇒ {Int | __elem == a ^+ b})` and the scope check rejects it, correctly. Substituting the projection is what keeps a tupled lambda's body mentioning only the tuple binder, which is in scope where the type sits.

## Review guide

`src/ccl/design/ir.md` gains two sections, "A binder carries its source site" and "A substituted parameter's site rides its projection", which record the mechanism, the equality exclusion, and that rejection. The rest is the field, its population, and the four tests: `a_use_resolves_to_its_binders_source_site_by_uid` and `every_source_binder_form_carries_its_site` (`uniquify.rs`) read the source back at each span a binder reports; `a_substituted_parameter_carries_occurrence_and_declaration_spans` (`lower/functions.rs`) pins both projection spans as lowering emits them; `a_substituted_parameters_two_spans_stay_on_distinct_nodes` (`panes.rs`) pins that folding four pane pairs onto them keeps the halves apart, and pins where the answer stops.

The payoff lands in the PR stacked on this one, which deletes the surface-AST scope walk.


src/ccl/design/ir.md
```diff
@@ -75,6 +75,82 @@ A scope is a type, `Binders`, not a collected list of binder references: it borr
...
+record, the reserved spelling standing in for a payload an arm declined to name.
+
+Two AST nodes gained a span to make this possible, both mirroring [`Param::name_span`]:
+`Stmt::FunctionDef::name_span` and `PayloadPattern::Named`'s second field.
+
+A multi-argument function's parameters are the exception, and
+[a substituted parameter's site rides its projection](#a-substituted-parameters-site-rides-its-projection)
+says how they are answered instead.
+
+### A substituted parameter's site rides its projection
```
dpmills, line 119: This section is confusing and seems like it doesn't actually do enough to recover the information we want.  Will anything actually map back to a span corresponding to just `p`? In my comments on the other PR (which led to this one) I proposed reworking how we handle multi-arg functions; might be able to use that sort of idea to fix this issue too.

The only test for the multi-arg case here just asserts that the function name has the right span; nothing is checking the behavior of spans for arguments.

src/ccl/design/ir.md
```diff
@@ -75,6 +75,82 @@ A scope is a type, `Binders`, not a collected list of binder references: it borr
 
 Consumers may rely on three properties of the walk, each asserted in `scope.rs`'s tests: the `Child` items come in `walk_children` order (so the `&mut` walk can pair scopes with children positionally), the binders it declares are exactly `walk_binders`', and a scope's children are consecutive with a binder-introducing scope entered only once. The last is what lets a consumer build per-scope state once per scope rather than once per child, and the `&mut` walk hands it the change point directly, as a `Scope` item ahead of the run of children it covers. Substitution needs both halves of that: restricting the substitution across the scope (`Subst::shadow_all`, which borrows instead of cloning whenever the scope names nothing the substitution acts on) and the Barendregt no-capture check, which is release-active and walks every replacement term — so a `LetRec` group of *n* binders owes *n* checks, and paying per child instead would cost *n*(*n*+1).
 
+### A binder carries its source site
```
dpmills, line 78: Why Span instead of NodeId? We have the ability to map NodeIds to spans, so that would let us handle the case of compiler-derived binders too; they can point back to whatever caused them to be created so we'll always have provenance information back to source.  The prose later claims "a binder is not an expression", but is there a fundamental reason we can't treat it like one here?

groundlar:
Closing this — the binder-site channel is abandoned to reduce scope, and the one piece worth keeping has moved into #141.

Your question on `ir.md:119` was the right one. Investigating it turned up two things that took the ground out from under this PR:

**The channel had no consumer that needed it.** `definitions` powers Ctrl/Cmd-click goto-definition and the tooltip's jump button, both on *use* sites. `definitionAt` is keyed on use spans only, so a click on a binder is inert — your original observation, reproduced. And for the definition side, span containment already resolves a binder's position to the enclosing `Let`, which is the "just span the whole `let`" behaviour you suggested. So `TypedBinding::name_span` was buying an exact def-span where the coarse one was already acceptable.

`definitions` now takes the span of the node that binds the name. A use of `g` in `g = 10` lands on the statement; a call lands on the whole `def`. Row counts across the fixture corpus are unchanged — nothing dropped, nothing added — and every `defSpan` simply widens to its binding.

**What survives is three lines.** A multi-argument function's parameters bind nothing, so they have no binder to widen to; they resolve through the projection that replaced them. That needed only tagging the projection template with `Param::name_span`, a field that already existed on `main`. It is now the first commit of #141 rather than a PR of its own.

Everything else is gone: the `TypedBinding` field, its `PartialEq` exclusion, `lower::with_binder_site`, the two AST spans this PR added to `Stmt::FunctionDef` and `PayloadPattern::Named`, the ~12 construction sites, and the `uniquify.rs` tests.

**On your `ir.md:78` question — `NodeId` instead of `Span`** — worth recording since it does not survive here either. There is no fundamental reason a binder cannot be treated as a node. But an id answers only through a `SourceProjection`, and only the lowering projection is always built, so an in-compiler consumer past inference loses the answer a field read gives; and `SourceAttribution::spans` is a set, where a scalar was wanted. I measured the motivating case: for a for-loop accumulator, all five binders `mut_elim` mints attribute to one span — the whole `for` statement — and carry the same rewrite tag, so an id would report exactly what the enclosing node already reports. `SyntheticKind::Mono` already carries the source binder's `Name` for the one case where a finer cause exists. Your instinct was right where there is no binder to hold a field: substituted parameters are answered exactly that way, by node ids and the projection.

**One finding worth keeping regardless**, because it is a defect in reachable code rather than an artifact of this PR. Binding a multi-argument function's parameters rather than substituting them — the rework you proposed — is blocked by refinement scoping, and not because of anything multi-arg. A single-parameter function with an ordinary `let` fails today on `main`:

```chl
def g(n: Int):
    m = n ^+ 1
    m ^+ 1
```
```
Scope violation (compiler bug, design §6.2):
  type (Int ⇒ {Int | __elem == m ^+ 1})
  references out-of-scope binder(s) ["m"]
```

Inference derives the refinement from the body and lands it on the lambda's own type, where a `let` inside the body has not opened. `def h(n: Int): n ^+ 1` compiles fine, because its refinement names the parameter and the function type binds that. Substituting the projection is what currently guarantees a tupled lambda's body mentions only the tuple binder. This is what you meant by the call mechanism not handling refined types well, and it is sharper than I credited at first — I will raise it separately rather than leave it buried in a closed PR.

PR #145, opened 2026-09-01
Serve the program inspector from the `cambra` binary and pin its wire with golden fixtures

groundlar:
`src/inspector_model` builds a read-only payload that nothing serialized, nothing served, and no test read end-to-end; its docs pointed at a `cambra-inspector` crate that does not exist. This PR adds `src/inspector_server` — the JSON transport, a loopback `tiny_http` server, and a public wire validator — reaches it from two new CLI modes (`--inspect-only[=PORT]` and `--dump-snapshot`), and pins the resulting bytes with a committed golden-fixture corpus plus a `ci.sh` drift gate. A `.cambra` file now has a path to bytes a frontend can fetch, and a change to that byte format fails CI instead of landing unnoticed. The frontend itself lands separately; `cambra-inspector/` holds only its fixtures, its regen script, and [its README](cambra-inspector/README.md).

## The transport edge

`snapshot_json` / `snapshot_json_pretty` render an `InspectorPayload`; `serve.rs` compiles the program once at startup and replays two pre-built bodies (`/api/snapshot`, `/api/diagnostics`) so the diagnostics the two routes report cannot disagree, degrading rather than erroring when the compile fails. `wire_check.rs` is `pub`, not `#[cfg(test)]`, because `tests/inspector_goldens.rs` links the library as a consumer does — the module docs give the argument. Nothing here decides what the payload says.

## Two CLI modes, exclusive by construction

`main.rs` gains a `Mode` enum instead of a check over three flags: `Run`, `InspectOnly`, `DumpSnapshot`. `--dump-snapshot` prints the pretty payload and exits, which is what makes the corpus regenerable — the server never exits.

## The `serde` feature is retired

`serde` and `serde_json` become unconditional dependencies, so every `#[cfg_attr(feature = "serde", …)]` on a wire type collapses to a plain derive. `ci_clippy_serde` and its CI step go with it: the gate is now three clippy configurations, not four. The measured tradeoff is recorded in `Cargo.toml`. Two consequences worth a look: `Cargo.toml` gains a one-package `[workspace]` pinned to `resolver = "3"` (an unset workspace resolver silently downgrades an edition-2024 package's default), `ci.sh` scopes its cargo invocations with `-p cambra`, and one assertion in `src/ccl/planning/join.rs` needs an explicit element type now that `serde_json`'s `PartialEq` impls are always in scope.

## The corpus and what guards it

`cambra-inspector/scripts/fixtures.manifest` is the single corpus list, read by both `regen-fixtures.sh` and the tests, so the regen path and the assertions cannot disagree about membership. `./ci.sh fixtures` regenerates through that same script into a temp dir, byte-diffs against the six committed fixtures, and flags an orphan fixture with no manifest row. `tests/inspector_goldens.rs` adds six ratchets over the real CLI, spawning a fresh process per dump because `NodeId`s come from a process-global counter; ratchet 5 walks every gallery source through the validator at no re-bless cost, which is what keeps the pinned set at six. `ci_all` now names the gates that failed.

## Eleven new gallery programs

Each carries a `mod.rs` asserting its own value and a row in [docs/demo-programs.md](docs/demo-programs.md#the-inspector-reads-this-gallery), so a program that pins an IR shape also runs. `txn_multi_read` and `mutation_loop` are corpus members with no committed fixture — five-figure payloads nobody could review — and are asserted structurally instead. `type_error` is the only source of the degraded payload; `defer_generators` is a known compile failure and the sole entry in `DUMP_PANICS`. `polymorphic` returns a tuple, so `tile_to_canonical` grows a `Tile::Record` arm.

Roughly 21,800 of the ~24,400 changed lines are those six fixture documents. Read the manifest and the gate; the JSON is generated output.


cambra-inspector/examples/accumulator.chl
dpmills, line 1: Any reason these need to be separate examples from the existing test program examples?

cambra-inspector/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+edition = "2024"
+license = "Apache-2.0"
+
+# The user-facing program-inspector: HTTP server (tiny_http) + embedded
+# CodeMirror frontend, serving the read-only post-inference snapshot from
+# `cambra::inspector_model`. This crate owns serde + the JS toolchain so the
+# `cambra` core stays serde-free.
+#
+# W0 scaffold only: depends on `cambra` with the `serde` feature on (this is
```
dpmills: Why is it important to keep serde out of the regular build? Seems like that requirement is adding unnecessary build complexity

src/inspector_server/serve.rs
```diff
@@ -0,0 +1,272 @@
+//! The user-facing inspector HTTP server.
```
dpmills: This doesn't seem like it's actually the inspector server? Isn't this the component that lives inside an actual cambra program that can serve the data the inspector needs? If so, it should be in Cambra/src not Cambra/cambra-inspector

cambra-inspector/scripts/fixtures.manifest
```diff
@@ -0,0 +1,45 @@
...
+# Slimming trims what the byte-exact corpus pins on the big, volatile programs:
+# `paneLinks` is ~4 lines/edge and grows super-linearly in the collapsed
+# channelize window, and the panes are the bulk of a large fixture's bytes. It
+# never changes the LIVE wire — `--dump-snapshot`
+# without flags, and the served payload, stay full. goldens.rs applies the SAME
+# per-row flags to its fresh dumps (so committed == fresh stays exact) and adds
+# structural link/pane assertions over a FULL dump to keep the bug-catching that
+# the pinned bytes used to provide.
+#
+# KEEP FULL (do not slim): list_min and arithmetic are the end-to-end canaries
```
dpmills: Why are there more examples than fixtures?

cambra-inspector/README.md
```diff
@@ -0,0 +1,139 @@
+# cambra-inspector
+
+The user-facing **Cambra program inspector**: a small, read-only web app that
+compiles a Cambra (CHL) program once and serves an interactive view of it —
+the source alongside one IR tree pane per compiler stage, cross-linked by true
+node→node provenance.
```
dpmills: Why is this the architecture that we want? It seems like an actual cambra program is the thing that should be serving the inspector info.  If you want a way to get the data without actually running the program, that's just an --inspect-only or --dry-run etc flag

cambra-inspector/tests/goldens.rs
```diff
@@ -0,0 +1,390 @@
...
+}
+
+/// RT-1a — the golden ratchet: every committed fixture equals a fresh dump of
+/// its example program. Compared at the JSON-content level (both sides parsed)
+/// so the comparison is independent of the pretty-printer `regen-fixtures.sh`
+/// pipes through; the ci.sh `ci_fixtures` gate covers the literal file bytes by
+/// regenerating through the same script and diffing. On failure the message
+/// names the first drifted JSON path and the fix path is its one command.
+#[test]
+fn committed_fixtures_match_fresh_dumps() {
```
dpmills: How painful has this been during rebasing? Naively, it seems like this is testing way more than it needs to.  Is there any sort of property-based testing we can do instead of verbatim dumps?

cambra-inspector/Cargo.toml
groundlar: Compile time was the reason, and it doesn't survive measurement. Same machine, cold builds in separate target dirs:

| | serde-free | serde + `serde_json` |
|---|---|---|
| `cargo build -p cambra` | 22.77 s | 23.72 s |
| `cargo build --release -p cambra` | 35.49 s | 36.49 s |
| cold at `-j1` (serial CPU work) | 52.02 s | 46.28 s |

Run-to-run spread on the release build alone was 35.5–41.2 s, so a ~1 s delta is at or under noise — the `-j1` number came out *faster* after, which is the same noise from the other side. In isolation the serde tree is 4.80 s of serial work, 7.06 s with `serde_json`, and those units sit off `cambra`'s critical path behind `chumsky`/`ariadne`.

What the split actually cost was standing, not seconds: `ci_clippy_serde` was a fourth clippy configuration whose only job was to compile the gated impls, every gate had to name `-p cambra` to keep the feature-off build alive, and 21 `cfg_attr` sites carried the gating. All of that is gone — serde and `serde_json` are ordinary dependencies and CI lints three configurations instead of four.

One real consequence worth knowing about: `serde_json`'s `PartialEq<Value> for usize` made an empty `vec![vec![]]` ambiguous in `planning/join.rs`, whose element type is now spelled out.

cambra-inspector/examples/accumulator.chl
groundlar: No good reason. They're now gallery programs under `tests/programs/`, with the IR shape each was chosen to pin recorded in its `mod.rs` and asserted by its own value — so a program that pins a pane shape is also a program that runs and checks its result.

Three were already in the gallery under another name and are deleted: `accumulator` is `for_accumulator`, `comprehension` is a smaller `filter_and_aggregate`, `generator_min` a smaller `generator_pipeline`.

Giving them tests for the first time turned up two that were dead:

- **this one.** `accumulator.chl` hasn't compiled since `:=` became required for a loop-carried binding — it still used `=`. Nothing ran it, so nothing said so.
- `defer_pipeline.chl` panics `post-lambda-elim`: the `zip` of two generators types as a compute function (`⇒`) where the letrec binder wants a data collection (`⤇`). It survives as `defer_generators`, a documented blocked program expecting `16`.

And a third that runs and is wrong: `txn_multi_read` returns `Function [ 100 ]` where 100 − 10 − 20 − 30 is `Function [ 40 ]`. The write sits inside an `if` inside the transaction and doesn't survive the conditional — which looks like what #136 is about. Pinned with `expect_scalar_currently_buggy` so it goes red when that lands.

src/inspector_server/serve.rs
groundlar: Agreed, and moved. The Rust half of the inspector is now `src/inspector_server/` in the core crate, next to the `inspector_model` it renders: `mod.rs` is the JSON edge, `serve.rs` this server, `wire_check.rs` the structural validator. `cambra-inspector/` keeps only `web/` and the fixture corpus and is no longer a Cargo package, so the workspace is one member again.

The binary reaches it through `cambra --inspect-only [=PORT]`, with `--dump-snapshot` as the one-shot form. A `Mode` enum makes those exclusive with `--inspect` by construction rather than by a check over two flags that each mean something.

One thing left open, flagged rather than fixed: `--inspect` and `--inspect-only` still share port 8080 and both answer `GET /` and `GET /api/snapshot`. Being exclusive is what makes that a later change rather than a collision now, and the convergence is real — `web_inspector`'s deprecation note is down to one blocker (no live per-tick view) once the operator pane lands upstack.

cambra-inspector/scripts/fixtures.manifest
groundlar: The gap was real, and part of it was rot. Six of the fourteen examples had no fixture, and two of those had stopped working with nothing to notice: `accumulator.chl` hadn't compiled since `:=` became required, and `defer_pipeline.chl` panics `post-lambda-elim`.

The corpus is now the demo gallery — the inspector has no example directory of its own — and the manifest selects the few programs whose *bytes* a test reads. It stays two columns so a fixture name can survive a program rename without moving the frontend's imports, which is why the `arithmetic` row points at `udf_closure`.

What closes the gap properly is that a program no longer needs a fixture to be covered. `every_gallery_program_produces_a_valid_payload` in `tests/inspector_goldens.rs` walks every `*.cambra` under `tests/programs/` through `wire_check`, dispatching on `meta.payloadKind`: **30 payloads validated against 6 pinned**, at no re-bless cost. So a program added for pane coverage gets a directory and no manifest row.

cambra-inspector/README.md
groundlar: You're right, and that's what it is now: `cambra --inspect-only <program>` compiles the program, serves its panes, and never runs it. `--inspect` keeps its current meaning — run the program, serve the live runtime dashboard — and the two are exclusive by construction rather than by a flag check.

The serve-a-snapshot-and-don't-recompile part stays, because it's what makes every interaction a client-side lookup over one payload rather than a round trip. What moved is where it lives: `src/inspector_server/` in the core crate, no second binary, no second Cargo package.

The remaining duplication is with `src/web_inspector.rs`, which serves the runtime dashboard on the same default port and the same two routes. Its deprecation note now names one blocker rather than two — the operator graph it lacked is the terminal pane upstack, so the live per-tick view is all that's left. Collapsing the two servers is the change after that, and I've kept it out of this PR rather than pulling four frontend commits into it.

cambra-inspector/tests/goldens.rs
groundlar: Not painful, and I went looking for numbers rather than trusting the impression. Across the 19 non-empty commits of this stack there was **one** re-bless for a wire change and **zero** id-only re-blesses. So the cost you'd expect from `NodeId` churn is real in principle — about 15% of every fixture's lines are id-bearing, and one process-global counter means a pass that mints a different number of ids renumbers ~4,500 lines corpus-wide — but it has been paid once.

You're right that it tests more than it needs to. What I don't think it can give up is being the frontend's only realistic payload input: `committed_fixtures_match_fresh_dumps` is what stops the consumer becoming the source of truth for the producer's shape, and a builder can't do that. `minimalSuccess()` in `wireValidate.test.ts` is the demonstration — one node per pane, empty edges, can't exercise cross-pane resolution at all.

So the answer is property tests *above* the dumps rather than instead of them. Two new ratchets in `tests/inspector_goldens.rs`:

- **Every gallery program through the validator.** Walks each `*.cambra` under `tests/programs/`, dispatching on `meta.payloadKind`: 30 payloads validated against 6 pinned, zero re-bless cost. Programs whose dump panics are a pinned list, not a silent skip, so one that starts working has to leave it.
- **Ids are the only volatile field.** Compiles a program twice in one process and asserts the payloads agree once ids are canonicalized. This is the premise the whole golden strategy rests on and nothing stated it. It does *not* subsume the cross-process determinism check — both compiles share one process's hash seeds, so only a second process catches hash-order dependence.

The second one earned itself on first run: node ids are **not** the only volatile number on the wire. `Display for Type` spells an unresolved `Infer` as its variable id, and `mutation_loop`'s post-inference pane ships `Mut(Int, ?N)` from a second process-global counter. No committed fixture carries a `?N` today, so nothing is churning — but a fixture over such a program would churn arbitrarily. Worth a separate look at why the mutability domain is still open after inference.

Rejected, for the record: a CHL program generator (none exists — `tests/type_gen/` generates `Type`s only, and `formal/design.md` lists one as future work), `insta` inline snapshots, and `schemars`. If renumbering churn ever does bite, the fix is a `--canonical-ids` dump flag, not a generator.

Also relevant to this thread: #152 is folded into this PR, which deletes the fixture-slimming grammar and 5,526 pinned lines outright.

src/inspector_server/serve.rs
dpmills: Comment is wrong here; this serves the raw data and is not user-facing

cambra-inspector/examples/accumulator.chl
dpmills: I think we could deduplicate a bit more aggressively; this is still a lot of new examples.  We don't need to match 1-1 with what cambra-inspector/examples was doing like the three that deduped, anything that has the same coverage of CHL/CCL shapes should be good enough

dpmills (approved):

cambra-inspector/examples/accumulator.chl
groundlar: Done — eleven new programs down to six, and the fixture corpus from six documents to five (21,774 lines to 19,788).

The criterion I used is the one you named: a program earns its place only if it covers a CHL/CCL shape, or a `rewritten.via` set, that no gallery program already covers. Measuring that turned up five that didn't.

Two merged into the program that subsumes them:

- **`udf_fanout` → `polymorphic`.** Same CHL shape (a lambda applied at N sites) and an identical via set. `polymorphic` now calls `dup` at three `Int` sites and one `Bool` site, so one program carries both duplication mechanisms — monomorph clones from inference, and `inline`'s freshened `Replicated` copies within the `Int` specialization. `udf_fanout`'s window was strictly the poorer of the two anyway (6 non-identity edges, max fan-out 1, against 13 and 2).
- **`defer_fanin` → `defer_lift`.** `defer_fanin`'s via set was a strict subset of `defer_lift`'s. `defer_lift` now feeds the returned channel again from a loop, so its panes carry both `channelize.defer_lift` (6 nodes) and `channelize.cluster` (12) — the lift and the two-site union in one program.

Three deleted outright:

- **`scopes`** — same shape and same via set as `udf_closure`, one extra parameter.
- **`mutation_loop`** — same via set as `for_accumulator`, which also produces an equivalent dense channelize window (95 non-identity edges and max fan-out 43, against 99 and 45). The golden tests that named it now name `for_accumulator`. That also fixed a claim that was already false: `the_recurrence_phases_reach_the_wire` asserted `mutation_loop` was the only program tagging nodes `Letrec`, and `for_accumulator` always did too.
- **`udf_fanout`'s fixture**, with it.

The six that stay each cover something nothing else does: `list_min` (smallest payload), `udf_closure` (the byte-for-byte canary), `polymorphic` (both fan-outs), `defer_lift` (`Channelize`), `txn_multi_read` (the only `Transact`/`AsOfRead` program), `type_error` (the only degraded payload).

One I kept against the criterion, so you can overrule it: `defer_generators` contributes zero pane coverage — it doesn't compile, and it's the sole `DUMP_PANICS` entry, so deleting it would remove that carve-out and leave ratchet 5 with no exception list at all. It's the gallery's only record of the incomparable-kinds failure (`zip` of two generators typed `⇒` where the letrec binder wants `⤇`), and deleting a bug pin to reduce a count seemed like the wrong trade. Say the word and it goes.

src/inspector_server/serve.rs
groundlar: Done. The module doc opened with "The user-facing inspector HTTP server", which was wrong on both halves — it answers `GET /api/snapshot` and `GET /api/diagnostics` with JSON and 404s everything else, so nothing it returns is a page.

It now leads with what it serves and names the frontend as the thing a user looks at:

```
//! The inspector's data endpoints: a read-only `tiny_http` server that answers
//! with JSON and never with a page. What a user looks at is the frontend, which
//! fetches these two routes.
```

groundlar:
Correction to my dedup reply above, fixed in #154.

I said `polymorphic` gained a second `Int` call site so that one specialization would inline twice and carry the same-type replication `udf_fanout` had. That premise is wrong: a literal's type is its singleton refinement, so `dup(1)` and `dup(2)` are `Int@1` and `Int@2` and monomorphization makes a third clone rather than a second call site on the first.

The site was never needed. At its original two call sites `polymorphic` already dominated `udf_fanout` on every measure the wire carries — `inline.beta` and `inline.udf` provenance labels against `inline.udf` alone, 25 paneLinks edges against 20, 10 non-identity against 6, a max fan-out of 2 against 1. So deleting `udf_fanout` stands; the extra call site was more of the same shape, which is what the deduplication was for.

#154 drops it and corrects the claim in `polymorphic/mod.rs` and `docs/demo-programs.md`. The fixture loses another 1,421 lines, so the corpus is 18,369 lines against the 21,774 it started at.

PR #146, opened 2026-09-01
Serve the inspector frontend from the `cambra` binary and document fixture re-blessing

groundlar:
`cambra-inspector/web/dist/index.html` is committed and gated by `ci_web`, but nothing served it: `cambra --inspect-only` offered `/api/*` only, so the frontend had to be run separately out of `web/`. `serve.rs` now `include_str!`s the bundle and answers `GET /` and `GET /index.html` with it, borrowing the static string per request as the snapshot body already does — a `cargo build` with no Node toolchain produces a binary that serves the whole inspector.

The rest is the maintenance surface around the golden corpus: a new `cambra-inspector/CLAUDE.md` holding the re-bless procedure, a `fixture()` wrapper in `web/src/__fixtures__/helpers.ts` that turns a stale-corpus `WireError` into a pointer at `regen-fixtures.sh` (pinned by a new `helpers.test.ts`; the four test files that loaded fixtures through `validateSnapshot` now go through it), and comments marking `web_inspector` and `InspectNode::to_json` removable once the program inspector gains an operator-graph pane and a live view.

### DIFF
--- a/cambra-inspector/web/src/__fixtures__/README.md
+++ b/cambra-inspector/web/src/__fixtures__/README.md
@@ -73,3 +73,9 @@ cargo golden test compares each committed fixture against a fresh dump.
 `store.test.ts` keys its assertions on node **labels/types**, not raw NodeIds, so
 ordinary id churn does not require touching the tests — but a shape change will,
 by design (and `wireValidate.ts` enforces the shape).
+
+Re-blessing pins more than it looks like. Every node's `type` is `Display for
+Type`'s rendering (`src/ccl/ty.rs` hands `collect_str` that string), so the
+corpus pins the type notation on every node in every pane — a literal's
+singleton spelling included. Changing that notation is a corpus-wide re-bless,
+not a formatting tweak.
--- a/cambra-inspector/web/src/__fixtures__/README.md
+++ b/cambra-inspector/web/src/__fixtures__/README.md
@@ -1,19 +1,24 @@
 # Golden snapshot fixtures
 
-Real `/api/snapshot` payloads emitted by the
-`cambra-inspector` backend, used by `store.test.ts` to test the frontend's
-derived logic (the B5 hole→type stitch, cross-pane resolution) **and** to guard
-the wire-shape contract: if the Rust payload drifts from the TypeScript wire
-types in `../types.ts`, these tests break. The corpus (fixture ↔ program
-mapping) lives in `../../../scripts/fixtures.manifest`, shared by
-`regen-fixtures.sh`, the ci.sh drift gate, and the cargo golden tests
-(`tests/inspector_goldens.rs`).
+Real `/api/snapshot` payloads emitted by the `cambra` backend, used by
+`store.test.ts` to test the frontend's derived logic (the B5 hole→type stitch,
+cross-pane resolution) **and** to guard the wire-shape contract: if the Rust
+payload drifts from the TypeScript wire types in `../types.ts`, these tests
+break. The corpus lives in `../../../scripts/fixtures.manifest`: which program each
+fixture is dumped from, and a comment per row saying what that row pins. It is
+shared by `regen-fixtures.sh`, the ci.sh drift gate, and the cargo golden tests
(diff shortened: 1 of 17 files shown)


cambra-inspector/web/src/__fixtures__/README.md
```diff
@@ -28,6 +28,41 @@ a `where.N`-labelled child edge instead of a positional index. `predicate: true`
...
+- `links=none` **omits** `paneLinks` entirely (the field is dropped, not shipped
+  as an empty array — an empty array would falsely claim the wire has no edges).
+
+The fixture loader (`fixture()` in `helpers.ts`) and `wireValidate.test.ts`
+validate slimmed fixtures under `validateSnapshot(json, { fixtureSlimming: true })`
+(the TS twin of Rust's `FixtureContext::Fixture`), which accepts a pane subset
+and an absent `paneLinks`. Production callers pass no options and get the strict
+live contract, so a live payload that drops a pane or its links still fails.
+
+**Full (all panes + links):** `arithmetic`, `polymorphic`, `list_min`,
```
dpmills: This is all unnecessary detail.  If there's anything important, it should be comments in the manifest so it doesn't rot

cambra-inspector/web/src/__fixtures__/helpers.ts
```diff
@@ -3,6 +3,28 @@
...
+ * error must not tell users to regenerate test fixtures.
+ */
+export function fixture(json: unknown): Snapshot {
+  try {
+    return validateSnapshot(json);
+  } catch (e) {
+    throw new Error(
+      `${String(e)}\n` +
+        "  This is a golden fixture. If the wire shape changed on purpose, regenerate via\n" +
+        "  cambra-inspector/scripts/regen-fixtures.sh, updating SCHEMA_VERSION (wireValidate.ts)\n" +
```
dpmills: We're not updating SCHEMA_VERSION yet, cut

dpmills (approved):

cambra-inspector/web/src/__fixtures__/README.md
groundlar: Done — `## The corpus` is gone from the README, and each manifest row carries a comment saying what that row pins.

Your prediction was already true when you wrote it. The `polymorphic` entry in that list described the program as "applied at four sites, three at `Int` and one at `Bool`", which #145 made true and #150 made false again within the day; `main` carries the wrong description right now. A per-fixture list sitting three directories away from the manifest has nothing holding it to the corpus.

What moved into `fixtures.manifest`, one comment per row: the shape each fixture pins, and for the two that a test names by hand, which test walks them. What was already there — the fixture↔program mapping, why the two columns can differ, that a program needing only backend coverage takes no row — stays and is no longer restated in the README.

Two things from that section are elsewhere rather than in the manifest, because they are not per-row facts:

- Re-bless discipline (classify every diff; the corpus pins `Display for Type` and `Display for Tiling` notation on every node) belongs to `cambra-inspector/CLAUDE.md`, "The golden fixtures", which this PR adds. The README now links there instead of restating it — that was a third copy.
- The wire-shape description (pane order, node table, predicate edges) stays in the README, since it describes the payload rather than the corpus.

This README goes from 75 lines on `main` to 54 here, and the `source_shared` and operator-pane entries that would have been appended upstack are manifest comments instead.

cambra-inspector/web/src/__fixtures__/helpers.ts
groundlar: Cut. The message now reads:

```
  This is a golden fixture. If the wire shape changed on purpose, regenerate via
  cambra-inspector/scripts/regen-fixtures.sh, updating both validators and any
  hand-written expectations in the same change — re-bless discipline in
  web/src/__fixtures__/README.md.
```

It was not just premature, it contradicted the policy this PR writes down: `cambra-inspector/CLAUDE.md` says `SCHEMA_VERSION` stays at 1 because nothing durable consumes the payload. An error message telling the next person to bump it would have been the only instruction to do so.

`helpers.test.ts` matches on `/regen-fixtures\.sh/`, so the assertion is unaffected.

PR #150, opened 2026-09-01
Add the inspector's CodeMirror frontend and a `ci.sh web` gate

groundlar:
The inspector backend already serves `/api/snapshot` and a committed golden-fixture corpus, but nothing renders them — `cambra-inspector/README.md` described a frontend that did not exist. This adds it: `cambra-inspector/web/`, a framework-free TypeScript / CodeMirror 6 single-page app that fetches one snapshot and computes every interaction locally, plus a `ci.sh web` gate (typecheck, vitest, bundle freshness) wired into `ci_all` and the workflow. The inspector is now usable end to end by serving the frontend from `web/`; the `GET /` route that embeds the bundle in the binary lands separately, so there are no Rust changes here.

Read `src/types.ts` first — it mirrors the shapes `cambra::inspector_model` serializes ([`src/inspector_model/design.md`](src/inspector_model/design.md#a-pane-on-the-wire)) and names every term the rest of the diff uses.

## One snapshot, one shared selection

`store.ts` owns the payload, one `Indices` per pane, the `LinkGraph`, and a single `selection`. A click anywhere — a source position or a tree row — resolves through that graph into a per-pane highlight set plus source spans, and each view renders its own slice, so the views never talk to each other. It also precomputes, per pre-inference node, the type its downstream counterpart resolved to, which is what the hole tooltips show.

Three pure modules carry the correctness and are where review effort pays:

- `offsets.ts` — Cambra spans are UTF-8 byte offsets and CodeMirror positions are UTF-16 code units, so every crossing routes through `OffsetMap` (built once per source, binary-searched).
- `indices.ts` — per-pane `nodeById` / `depthById` / `predicateIds` plus `tightestNodeAt`, `typesAt`, `definitionAt`. `tightestNodeAt` is the only implementation of the D4 tightest-enclosing policy anywhere; a Rust port is a followup, and `indices.test.ts` is that port's spec.
- `links.ts` — `buildLinkGraph` / `resolveLinks` walk the dense `paneLinks` bidirectionally and transitively. Self-edges carry identity, so there is no identity special case and a fan-out is just the `u !== d` edges.

The invariant to check hunk by hunk: refinement predicates are nodes on the wire but are not the program's value tree, so every positional query filters on `IrChild.predicate`.

## The wire shape is validated, not asserted by cast

`wireValidate.ts` replaces the unchecked `as Snapshot`: `validateSnapshot` walks the payload and throws naming the failing path. It pins `PANE_IDS`, the per-pane kinds, and the adjacent link windows as constants instead of reading them off the payload, so a pane added upstream fails the validator rather than passing silently. It is the TypeScript twin of the Rust `assert_snapshot_shape` and the file says so — extend the two together.

## The committed bundle and its gate

`vite-plugin-singlefile` inlines all JS and CSS into one `dist/index.html`, committed on purpose so the server can `include_str!` it without putting Node in the `cargo build` path ([`cambra-inspector/web/README.md`](cambra-inspector/web/README.md#committed-bundle-r7)). `ci_web` runs `npm ci`, typecheck and vitest, then rebuilds and `cmp`s the bundle against the committed copy, and skips (rather than fails) when `npm` is absent. Two things there look odd and are commented in `ci.sh`: every step is `|| exit 1` because `ci_all` calls the gate on the left of a `||`, which disables errexit for its whole extent; and freshness is a file comparison rather than a `git diff`, which false-fails in a colocated jj repo. The workflow gains `Setup Node` (v4, Node 20) and a `Web` step — the `TODO` asking for SHA pinning is stale, the action is already pinned.

## Tests and docs

94 vitest cases over nine files, listed by module in the new [`web/README.md`](cambra-inspector/web/README.md#tests). `views.dom.test.ts` is the one worth reading: it mounts the real `SourceView` and `TreeView`s under jsdom over `list_min.snapshot.json` and drives the selection both ways, covering the store→view subscriber path where one throwing subscriber starves the rest. `cambra-inspector/README.md` drops its "frontend pending" note, moves from "stage" to "pane" throughout, and gains an "Editing the frontend" section pointing at the re-vendor step.

## Two review fixes folded in

`web/README.md`'s "Running the inspector" links to [`cambra-inspector/README.md`](cambra-inspector/README.md#running-it) instead of restating it. The restatement had gone stale as well as duplicative — it named `cargo run -p cambra-inspector`, a `--port` flag, and `cambra-inspector/examples/`, none of which survived #145.

`polymorphic` loses the `dup(2)` call site #145 added for the inspector corpus. A literal's type is its singleton refinement, so `dup(1)` and `dup(2)` are `Int@1` and `Int@2` and monomorphization makes a third clone rather than a second call site on the first — the site added a specialization, not the same-type inline fan-out it was there for. At two call sites `polymorphic` already dominates what it replaced on every measure the wire carries: `inline.beta` and `inline.udf` provenance labels against `inline.udf` alone, 25 paneLinks edges against 20, 10 non-identity against 6, a max fan-out of 2 against 1. The fixture drops 1,421 lines, so the committed corpus is 18,369 lines against the 21,774 it held before #145.


cambra-inspector/web/README.md
```diff
@@ -0,0 +1,219 @@
...
+> cheap, immutable, and computed once at compile time. The post-M1 live surface
+> **cannot** ship everything: runtime values are `O(N·K·T)` and recomputed on
+> demand, so they move to per-query/streaming over the CQRS transport. The
+> `Store` here builds all indices eagerly in its constructor from one payload;
+> adding live values is **not** a matter of adding a field — it will require
+> reworking the store into an incremental/streaming model. That is an expected,
+> additive migration (the static read model is a true subset of the live one),
+> but flagged here so the eager-load architecture is understood as scoped to M1.
+
+## Running the inspector
```
dpmills, line 28: This section is duplicative; just link to the right other docs

dpmills (approved):

cambra-inspector/web/README.md
groundlar: Done. The section is now two lines pointing at [Running it](../README.md#running-it), keeping only what is specific to this directory (the `npm run dev` proxy).

It was worse than duplicative — it was wrong. It documented `cargo run -p cambra-inspector -- program.chl --port 9000` and an examples directory, and #145 removed the separate crate, the `--port` flag, and `cambra-inspector/examples/` alike. Linking rather than restating is what would have kept it honest through that change, which is your point.

Same edit fixes the intro paragraph, which called the server `cambra-inspector`; it is `src/inspector_server/` reached through `cambra --inspect-only`.

PR #152, opened 2026-09-01
Drop inspector fixture slimming for whole-payload goldens and structural assertions

groundlar:
The inspector's golden corpus admitted two payload shapes: the live wire, and a *slimmed* fixture holding a pane subset with `paneLinks` omitted, so the two largest corpus programs could commit smaller files. That second shape cost a manifest retention grammar parsed twice (once in shell, once in Rust), two `--dump-snapshot` flags, a `serde_json::Value` pruning path pinned byte-identical to the direct one, and a validator parameterized by which contract it was checking. This change deletes all of it: every committed fixture is the whole wire, `txn_multi_read` and `mutation_loop` carry no fixture at all, and `tests/goldens.rs` asserts what they were in the corpus for over a fresh dump instead.

## One payload shape

`FixtureRetention`, `render_pretty`, `prune_for_fixture`, and the `FixtureContext::{Live,Fixture}` split are gone; `snapshot_json_pretty` serializes the typed payload and nothing else. A manifest row is now two words, so `regen-fixtures.sh` and `goldens.rs` have no grammar left to disagree about, and each loses its retention branch. `serde_json` loses `preserve_order`, which existed only to make the `Value` round-trip byte-safe.

## What stands in for 5,526 pinned lines

`txn_multi_read` and `mutation_loop` keep their dense-channelize-window assertions and gain `the_recurrence_phases_reach_the_wire`: their dumps must still tag nodes `Transact` and `Letrec`. The `rewritten.via` allowlists in both validators reject a new tag but cannot notice a pinned one disappearing, which is the gap that test closes. `snapshot_dumps_are_cross_process_deterministic` now names both programs explicitly, since the manifest loop no longer reaches them.

## The validator becomes public

`test_support` becomes `pub mod wire_check`; `tests/goldens.rs` links the library as a consumer does, so a `#[cfg(test)]` validator is invisible to it, and its doc comment records why a feature gate would be worse. The new fourth ratchet, `committed_fixtures_are_structurally_valid`, reads each committed file rather than comparing it — a wrongly built dump equals a fresh copy of itself, so byte equality alone lets it reach the frontend unchallenged.

## Server and gate fixes

`serve` binds `127.0.0.1` rather than `0.0.0.0`: the payload is the user's source and the IR for it, from a local development tool. Responses borrow the pre-rendered bodies instead of cloning megabytes per request, and a failed `respond` logs rather than vanishing. `diagnose_json` and `snapshot_body` are removed as unused — `build_bodies` already renders both routes from a single compile.

`ci.sh fast` narrows its clippy pass to `-p cambra`, matching every other core gate; the workspace-wide form unified the inspector's `serde` feature into the library. `CLAUDE.md` records that, and that `--fix` also runs the inspector and fixtures gates. The lib's embedded CHL test programs move to `indoc!`.


dpmills:
This whole PR looks like a bugfix for #145.  Consider merging it down to make review simpler

PR #153, opened 2026-09-01
Meet both sides when reading a negative position

dpmills:
A domain variable is read from its polarity-correct side alone, so the value flowing in, which the one-way `arg <: domain` edge deposits as a lower bound, is dropped once the uses have settled a shape. The two readings of one invariant data domain then differ: `groupby([1], \x -> x)`'s key variable carries `Int@1` below and `Int` above, which the post-pass tree check rejects. `compact_go` now merges the opposite side into a negative position's reading at the position's own polarity — the uses state what they demand, the lower bounds what arrives — so group-by, `set`, and `map` over a literal whose elements share one singleton type compile and run.

## Only a projection needs a rebuild

`specialize_lambda_domain` is deleted: a lambda's binder is its domain variable, so the meet at the position is the whole reading. `specialize_projection_domain` stays, because a projection is polymorphic in its input's width — its untouched positions are variables no edge reaches, and op-conversion needs the one width the use site has. `recovered_input` loses its position argument and always derefs a mutable-variable handle; [The collapse happens at the position](src/ccl/design/type-inference.md#the-collapse-happens-at-the-position) carries the rule.

## Identifying two spellings of one domain

The merge settles one position; equating two spellings of one domain is separate. `bind_annotation` draws the second half of a `SharedHole`'s equation, since the annotation edge alone leaves the id below every domain naming it, and comprehension lowering writes the id into a `groupby` source annotation that states only a kind. Recorded in [A shared hole naming a domain states an equation](src/ccl/design/type-inference.md#a-shared-hole-naming-a-domain-states-an-equation).

Two `#[should_panic]` tests become value assertions; `a_negative_position_meets_both_sides` pins the merge; `a_consumed_boxed_map_annotation_escapes_its_scope` now names the two colliding domain spellings rather than an out-of-scope binder.

### DIFF
--- a/src/ccl/infer/context.rs
+++ b/src/ccl/infer/context.rs
@@ -611,6 +611,41 @@ impl Typing for InferCtx {
                 inferred: inferred_ty,
             })
         })?;
+        // **A `SharedHole` naming a domain is an equation, not an ordering.** The edge
+        // above is contravariant in the domain, so it leaves the shared variable *below*
+        // every domain annotated with it: a common lower bound, which orders each domain
+        // under it and says nothing between them. Lowering writes the id to claim that two
+        // positions are one domain, and a data domain is invariant, so the claim is an
+        // equation and this draws its other half.
+        //
+        // Only where the id names the domain. A domain variable reached any other way may
+        // receive several domains deliberately — a conditional collection's arms, a
+        // domain-generic consumer's parameter — and equating those is what `constrain_go`'s
+        // invariant-domain arm declines to do for a variable-sided edge.
+        //
+        // Never against a **bound witness**. A sum's domain is its binder's reference, and
+        // entering a sum is a term (`src/ccl/design/type-inference.md`, "Only a term builds
+        // a sum"), so an equation between that reference and a free variable is not a claim
+        // about two positions but an escape of the binder — the coercion the one-way rule
+        // above exists to leave to the Σ rule.
+        if let Type::Fun {
+            domain: claimed, ..
+        } = ann_to_normalize.peel_refinements()
+            && matches!(**claimed, Type::SharedHole(_))
+            && let Type::Fun { domain: shared, .. } = ann_simple.peel_refinements()
+            && let Type::Fun {
+                domain: inferred_dom,
+                ..
+            } = inferred.peel_refinements()
+            && !matches!(inferred_dom.peel_refinements(), Type::WitnessRef(_))
+        {
+            constrain_subtype(inferred_dom, shared, &mut self.cache).map_err(|_| {
+                self.raise(InferError::AnnotationMismatch {
+                    annotation: ann.clone(),
+                    inferred: coalesce_for_error(inferred),
+                })
+            })?;
+        }
         Ok(ann_simple)
     }
 
--- a/src/ccl/infer/context.rs
+++ b/src/ccl/infer/context.rs
@@ -601,14 +601,14 @@ impl Typing for InferCtx {
             _ => ann,
         };
         let ann_simple = self.normalize_annotation(ann_to_normalize);
-        // Snapshot the inferred type before the annotation bounds are added so
-        // the error shows what was actually inferred, not the partially
-        // modified state after a failed constrain_subtype.
+        // Snapshot the inferred type before any annotation bound is added. Both reports
+        // below read this one snapshot, so each shows what was inferred rather than the
+        // partially modified state a failed `constrain_subtype` leaves behind.
         let inferred_ty = coalesce_for_error(inferred);
         constrain_subtype(inferred, &ann_simple, &mut self.cache).map_err(|_| {
             self.raise(InferError::AnnotationMismatch {
                 annotation: ann.clone(),
-                inferred: inferred_ty,
+                inferred: inferred_ty.clone(),
             })
         })?;
         // **A `SharedHole` naming a domain is an equation, not an ordering.** The edge
@@ -642,7 +642,7 @@ impl Typing for InferCtx {
             constrain_subtype(inferred_dom, shared, &mut self.cache).map_err(|_| {
                 self.raise(InferError::AnnotationMismatch {
                     annotation: ann.clone(),
-                    inferred: coalesce_for_error(inferred),
+                    inferred: inferred_ty,
                 })
             })?;
         }
--- a/src/ccl/infer/solver/compact.rs
+++ b/src/ccl/infer/solver/compact.rs
@@ -2055,10 +2055,10 @@ fn compact_go(
             // This fallback handles a *bare* under-determined domain var: it
             // materializes the type locally at coalesce from the lower-bound
             // side. The other half of the contravariant-domain story — a
-            // *structured* domain (a tuple/record with `Infer`s inside, which
-            // this per-var read cannot reassemble) — is recovered separately
-            // by `coalesce_node`'s `specialize_projection_domain` /
-            // `specialize_lambda_domain`. Both Apply edges are one-way (no
+            // *structured* domain whose untouched positions are variables no
+            // edge ever reaches, which is a `Proj`'s dense-prefix requirement
+            // and nothing else — is recovered separately by `coalesce_node`'s
+            // `specialize_projection_domain`. Both Apply edges are one-way (no
             // emit-time reverse whose eager cross-component propagation would
             // cover these halves); see `design/type-inference.md` ("Apply is
             // one-way" and "Closing the single-sided blind spots (no separate
@@ -2087,17 +2087,16 @@ fn compact_go(
                 let bc = compact_go(&b.ty, pol, &inner_acc, Some(&new_parents), st);
                 bound = CompactType::merge(pol, bound, bc);
             }
-            // Opposite-polarity fallback: walk the other side too if the
-            // primary walk did not produce any concrete (atom / shape)
-            // contribution. Without this, a variable whose only concrete
-            // information lives on the opposite polarity coalesces to
-            // `Type::Infer(?N)` instead of its real type — most commonly
-            // a fresh lambda param whose Apply-site bound flows in at the
-            // opposite polarity from where the lambda is coalesced. This is the
-            // coalesce-time read of monomorphization; it is sound because every
-            // var reaching coalesce is monomorphically determined (one type or
-            // an `IncompatibleBounds` error). See the rationale above, and
-            // [`fallback_allowed`] for why it happens only at a position.
+            // Whether the *shape* collapse fires: the primary walk produced no
+            // concrete (atom / shape) contribution. Without it a variable whose only
+            // concrete information lives on the opposite polarity coalesces to
+            // `Type::Infer(?N)` instead of its real type — most commonly a fresh
+            // lambda param whose Apply-site bound flows in at the opposite polarity
+            // from where the lambda is coalesced. This is the coalesce-time read of
+            // monomorphization; it is sound because every var reaching coalesce is
+            // monomorphically determined (one type or an `IncompatibleBounds`
+            // error). See the rationale above, and [`fallback_allowed`] for why it
+            // happens only at a position.
             let no_concrete = {
                 let CompactType {
                     atoms,
@@ -2131,7 +2130,13 @@ fn compact_go(
                     && history_slot.is_none()
                     && !var_is_shape
             };
-            if no_concrete && allow_fallback {
+            // A negative position reads the opposite side whether or not the shape
+            // needed recovering, because both sides narrow it: the requirement side
+            // records what the uses demand, the value side what actually arrives (see
+            // `src/ccl/design/type-inference.md`, "The collapse happens at the
+            // position").
+            let read_opposite = allow_fallback && (no_concrete || !pol);
+            let recovered = read_opposite.then(|| {
                 let mut recovered: Option<CompactType> = None;
                 for b in opposite_bounds.iter() {
                     let inner_acc = Subst::then(&b.render_subst(), subst_acc);
@@ -2141,17 +2146,23 @@ fn compact_go(
                         Some(acc) => CompactType::merge(!pol, acc, bc),
                     });
                 }
-                if let Some(mut recovered) = recovered {
+                recovered
+            });
+            match (recovered.flatten(), no_concrete) {
+                (Some(mut recovered), true) => {
                     // Carry what the polarity-correct walk *did* find — variable
                     // identities and refinement demands — across without letting
                     // it into the structural fold.
                     //
-                    // Replacing rather than merging is what the *negative* case
-                    // needs: there the primary result may hold a variant shape,
-                    // and it is the arms the body can handle rather than anything
-                    // that flowed in, so merging would union those tags into the
-                    // domain. (At a positive position `no_concrete` now implies
-                    // there is no shape at all to lose.)
+                    // Replacing rather than meeting is what an *undetermined*
+                    // position needs, and the difference is that the collapse here is
+                    // a choice rather than a narrowing: there is no settled structure
+                    // for the two sides to narrow jointly. At a negative position the
+                    // primary result may still hold a variant shape, and it is the
+                    // arms the body can handle rather than anything that flowed in, so
+                    // meeting would intersect those tags into the domain. (At a
+                    // positive position `no_concrete` implies there is no shape at all
+                    // to lose.)
                     //
                     // Refinements union instead of intersecting: a demanded
                     // predicate is checked against the value the fallback found,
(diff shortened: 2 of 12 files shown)


src/ccl/infer/solver/compact.rs
```diff
@@ -2131,7 +2130,13 @@ fn compact_go(
                     && history_slot.is_none()
                     && !var_is_shape
             };
-            if no_concrete && allow_fallback {
+            // A negative position reads the opposite side whether or not the shape
+            // needed recovering, because both sides narrow it: the requirement side
+            // records what the uses demand, the value side what actually arrives (see
+            // `src/ccl/design/type-inference.md`, "The collapse happens at the
+            // position").
+            let read_opposite = allow_fallback && (no_concrete || !pol);
```
groundlar, line 2144: **The shipped version pays the doubling too, and that is not recorded anywhere.**

`compact_go` resets `parents` to `None` at every structural child and has no per-position result memo, so this gate re-opens one nesting level down and the walk doubles per contravariant flip — in the shipped version, not only in the rejected ungated one.

Measured here (debug, 3 runs each, same test counts):

| | `main` | this branch |
|---|---|---|
| `tests/type_check.rs` | 0.57s | 0.85s |
| `test_groupby_key_relation_is_per_occurrence` | 0.36s | **0.71s** |
| `tests/compilation_pipeline` | 3.06s | 4.87s |

That middle row is `def by_key(c, f): groupby(c, f)` at two types — the doc's own example. The design doc and both commit messages record the doubling only as the cost of the rejected variant ("1.4s to 10s"), which reads as though this one is free.

The paper is explicit that the cache is what keeps this from going exponential (§3.2.2 on `constrain`'s cache: it "avoids repeating identical work … which is important in order to avoid making the algorithm exponential in complexity"). Its `coalesceType` never resets `inProcess` across structural children; ours passes `None`. The applicable memo is §4.3.2 hash consing — and note a `(uid, pol)` key would be **wrong**, since the result here also depends on `subst_acc` and on `st.scope`. Hash-consing the rendered contribution sidesteps that.

Ask: record the shipped number beside the rejected one and name the absent memo. Fixing it can be a follow-up.

src/ccl/infer/solver/compact.rs
```diff
@@ -2165,6 +2176,15 @@ fn compact_go(
                     }
                     bound = recovered;
                 }
+                // Both sides are narrowings of a structure the primary walk already
+                // settled, so the reading is their meet — which is what merging at the
+                // position's own polarity computes. One rule covers every slot: a
+                // refinement set narrows a position exactly as a record's fields do,
+                // and the negative merge unions both.
+                (Some(recovered), false) => {
```
groundlar, line 2196: **This branch performs the operation the comment directly above gives as the reason not to merge** — "meeting would intersect those tags into the domain". At negative polarity `merge_variants` intersects, so whenever the demand side produced a shape, tags are intersected.

The intersection is the correct meet and is sound, so I think the comment needs rewording rather than the code needing an exclusion — its argument applies to the *undetermined* case only.

What does change materially is **openness**, not the tags: a closed demand meeting a closed value gives the same tags either way, but `CompactVariant::meet_openness` returns `Closed` unless both sides are open, and its own doc says the one-side-open case is an approximation that "**No program reaches it**". This makes a closed value-side variant meet an open demand-side one at every settled negative position. `CompactType`'s field doc calls compaction "the only place an `Open` demand can be dropped silently (a lost marker turns a `case _:`'s scrutinee demand back into an exact sum)".

Probed at `compact_type`: value `` {`some{Int}} `` against open demand `` {`some, `none} `` yields a domain naming `` `some `` alone. Nothing in the suite pins it. Worth a pin on the **openness marker**, and a reworded comment.

src/ccl/infer/solver/compact.rs
```diff
@@ -2165,6 +2176,15 @@ fn compact_go(
...
+                    bound = CompactType::merge(pol, bound, recovered);
```
groundlar, line 2197: Minor, and latent rather than live: this unions the value side's `vars` into a negative position's reading, and `simplify_type` sits between `compact_type` and `coalesce_compact` in the production path (`solve.rs:1068`) reading exactly those sets. Its polar-only elimination, co-occurrence merging and atom-sandwich absorption are all defined on occurrences *in the type* with polarity (the paper's §4.3.1), and they now see occurrences the coalesced type does not contain.

Not a defect today — that file's own doc says the pass "is currently cosmetic (all types are monomorphic) but becomes load-bearing once let-polymorphism introduces genuine polar asymmetry." But its stated refinement-safety argument ("co-occurring variables sit in the same position and therefore carry the same refinements") is weakened once a position's refinements come from two sides. One line in the design doc would keep this from being rediscovered later.

Style, separately: `read_opposite.then(|| …)` yielding `Option<Option<_>>` to immediately `.flatten()` reads less directly than an `if` producing `Option<CompactType>`.

src/ccl/infer/context.rs
```diff
@@ -611,6 +611,41 @@ impl Typing for InferCtx {
...
+        if let Type::Fun {
+            domain: claimed, ..
+        } = ann_to_normalize.peel_refinements()
+            && matches!(**claimed, Type::SharedHole(_))
+            && let Type::Fun { domain: shared, .. } = ann_simple.peel_refinements()
+            && let Type::Fun {
+                domain: inferred_dom,
+                ..
+            } = inferred.peel_refinements()
+            && !matches!(inferred_dom.peel_refinements(), Type::WitnessRef(_))
```
groundlar, line 640: This is the better half of the PR: an identity claim recorded where the claim still exists, so it survives instantiation (the paper copies both bound lists symmetrically in `freshenAbove`/`extrude`, Fig. 7) instead of being re-decided per route. The two qualifications — only where the id *names* the domain, never against a bound witness — are things the coalesce-time merge cannot express, which is why that one needed a gate.

One asymmetry worth resolving: this excludes `WitnessRef` deliberately, and the compact merge has no counterpart. `AtomKey::Witness` meeting a concrete atom is two shapes, so the two halves of this PR disagree about whether witnesses need excluding. Is the merge safe there because the eager closure already rejects the pair, as it does for the data-domain case?

groundlar (commented):
Reviewed both commits. `./ci.sh` is clean here (all three clippy passes, doc, doc-refs, formal, golden fixtures) and `cargo test` is green. The direction is right: this removes a special case rather than adding one. Comments below are one that should block and a set of smaller ones.

## The merge is sound, and costs principality

Checked against *The Simple Essence of Algebraic Subtyping* (Parreaux, ICFP 2020). Fig. 11 gives a variable's two readings, and the soundness argument in §5.3.1 needs only that `E⁺(vs) ≤ E⁻(vs)` whenever the bounds are consistent. This computes `E⁻_new = E⁻_old ⊓ E⁺`, which under consistent bounds is `E⁺` — it collapses the interval to its bottom — so the property still holds and Fig. 12's unifying instantiation still exists. `constrain_go` maintains the consistency premise eagerly, replaying Fig. 9's two `Variable` arms including the closure over existing opposite bounds. A negative domain therefore reads as a *subtype* of the principal one and the enclosing function type as a *supertype*, still derivable by T-SUB (Fig. 3). Loss of principality, not of soundness.

The caveat that applies is §4.1: the paper's lattice identifies `{tag:0} ⊔ {tag:1}` with the componentwise merge, and it says plainly that this assumption "may not hold in languages with more advanced features", naming flow typing, "where the former holds more information". Cambra has singleton refinements and an invariant data domain, so it is the language that caveat is about. Every residual risk here is one instance of `CompactType::merge` not being the lattice meet it stands in for; the slot-level comments below are those instances.

On principality: `transactions.rs`'s "the arity is the workaround" (from #159) is the symptom already written down. This PR narrows it rather than causing it, and `a_one_entry_seed_does_not_reach_a_mutable_map` is an honest record of the residue.

## Verified

- Reverting `read_opposite` to `allow_fallback && no_concrete` fails **34 tests across 5 binaries**, including the new unit pin. Coverage is real.
- Deleting `specialize_lambda_domain` is more than subsumed. On `def by_key(c, f): groupby(c, f)` at two types, `main` leaves the key lambda narrowed *inside refinement predicates* (`λ r : {b: String} → r.b`) while specializing only the outer occurrence. The meet fixes both: `λ r : {a: Int, b: String} → r.b`.
- The golden wire corpus does not move, which is good evidence the merge is a no-op wherever the two sides already agree.

## Could not reproduce, recorded so it is not re-raised

I expected `CompactFun::merge` to hand one position's two readings to `data_domains_disagree` and manufacture a `DomainJoinConflict` on a program with one domain. It does not: the eager closure rejects the pair at `constrain` before the merge sees it. Likewise the atoms-union hazard — atom subtyping is reflexive-only (`constrain.rs:878-886`), so two differing primitive atoms are already a constrain error. `AtomKey::Witness` is the exception, which is the `context.rs` comment below.

## Freshness

The branch is 2 commits behind `origin/main` (#121, #175), and those touch `src/ccl/lower/mod.rs` (where `fresh_shared_hole` lives) and `src/ccl/infer/emit.rs`. Worth a rebase and a re-run of `./ci.sh` before merge.

src/ccl/infer/context.rs
```diff
@@ -611,6 +611,41 @@ impl Typing for InferCtx {
...
+        {
+            constrain_subtype(inferred_dom, shared, &mut self.cache).map_err(|_| {
+                self.raise(InferError::AnnotationMismatch {
+                    annotation: ann.clone(),
+                    inferred: coalesce_for_error(inferred),
```
groundlar: `coalesce_for_error(inferred)` runs *after* both `constrain_subtype` calls, so it reports the partially-modified state that the `inferred_ty` snapshot 30 lines up exists to avoid — "so the error shows what was actually inferred, not the partially modified state after a failed `constrain_subtype`".

`inferred_ty` is moved into the first closure, so this needs a clone of the snapshot rather than a second coalesce.

src/ccl/lower/comprehension.rs
```diff
@@ -269,7 +269,30 @@ pub(super) fn lower_list_comp(
         let source = match source.user_annotation.as_ref().and_then(Type::fun_kind) {
             Some(annotated) => {
                 result_kv.contributes_first(annotated.clone());
-                source
+                // An annotation states a kind, a domain, or both, and only the domain
```
groundlar, line 272: The comment names `groupby`, but the arm fires for three source shapes: `groupby`, the `set`/`map` re-keying constructors (via `lower_rekeyed`'s `data_fun(Hole, Hole)` in `exprs.rs`), and a multi-generator nested comprehension (which leaves `domain: Hole` further down this file).

Verified — `[x for x in set([1])]` and `[v for v in map([(1,10)])]` both gain the singleton under this change. The equation holds for all three, since an unfiltered single-generator comprehension's domain is its source's domain whatever shape that is, so this is comment accuracy rather than a bug. Naming one instance as if it were the rule is how a comment rots.

The enclosing comment above ("Only a source that did not already name its domain is stamped") now describes both branches and reads as contradicted until you reach this one.

src/ccl/infer/solver/compact.rs
```diff
@@ -2189,8 +2209,67 @@ fn compact_go(
...
+            Type::Record(vec![field("a"), field("b")]),
+            Type::Record(vec![field("a")]),
+        );
+        let fields = wide.rec.expect("a record shape");
+        assert!(
+            fields.contains_key(&FieldKey::Name(SmolStr::from("a")))
+                && fields.contains_key(&FieldKey::Name(SmolStr::from("b"))),
+            "both the demanded field and the one only the value carries: {fields:?}"
+        );
+    }
```
groundlar, line 2361: Good pin — it is what fails first when the gate is reverted.

Missing case: nothing pins **"a positive position is unaffected"**. The design doc states it and it is load-bearing, since it is what keeps the `var_is_shape` asymmetry at positive polarity intact. One assertion on a positive position would close it.

groundlar (approved):

src/ccl/infer/context.rs
dpmills: Fixed in 3a33dcfbb: both reports read the one snapshot, the first cloning it.

src/ccl/infer/context.rs
dpmills: Yes, and the reason makes it not an asymmetry: `bind_annotation` excludes a bound witness because it *draws an edge*, and the merge draws none.

A reference meeting a concrete type is settled where the edge is recorded. `constrain_go`'s `(WitnessRef(w), other)` arm distributes the demand over the kind's candidates — one invariant edge each — and reports `Mismatch` where the kind names none. A variable is not a concrete type there, so it falls through to the variable arms and the reference lands in the bounds like any leaf, meeting the concrete side when the closure propagates. Compaction reads back only what those edges admitted.

Where both do land at one position it is still not equated. `CompactType::shapes` counts witness atoms as one shape *between them* but not with a non-witness atom, so a reference beside `Int` is two shapes and `coalesce_compact` reports `IncompatibleBounds`.

Stated in the design doc in c1cc56354.

src/ccl/infer/solver/compact.rs
dpmills: Recorded in 81c7b888f, beside the rejected number.

Two of the three rows reproduce here (debug, median of 3, this branch rebased onto `ca4068412`):

| | `main` | this branch |
|---|---|---|
| `tests/type_check.rs` (314 tests both sides) | 0.83s | 1.29s |
| `test_groupby_key_relation_is_per_occurrence` | 0.56s | 0.94s |
| `tests/compilation_pipeline` | 2.72s (923) | 2.83s (927) |

The pipeline suite does not show it on this machine — its time is execution rather than inference. The other two match your ratio, and the doc carries them.

It names the absent memo too, minus the section numbers, since I did not verify those against the paper: `in_process` prunes cycles and is removed again as each walk returns, so a position reached twice is compacted twice; `(uid, pol)` would be a wrong key, the reading depending also on `subst_acc` and on the refinement scope; hash consing the rendered contribution keys on what was produced instead. Left as a follow-up per your note.

src/ccl/infer/solver/compact.rs
dpmills: Reworded in 4b289d182, and the openness finding is real — but one level in from where the probe lands.

That probe reads the same on `main`. I ran value `` {`some{Int}} `` against open demand `` {`some, `none} `` on both and got `` {`some} `` `Closed` either way. At that shape the demand is the whole primary contribution, so `no_concrete` is true at negative polarity (`var_is_shape` is `pol && var.is_some()`) and the *replace* arm fires — the path that was there before this PR.

What this branch changes is a variant under a settled shape, which is what reaches the merge arm at all:

| domain reading | `main` | this branch |
|---|---|---|
| value `` {`some} `` vs open demand `` {`some, `none} `` | `` {`some} `` Closed | same |
| the same pair under a record field | `` {`some, `none} `` **Open** | `` {`some} `` **Closed** |

And the tags matter more than the marker. Value `` {`some, `other} `` against open demand `` {`some} `` reads `` {`some} `` Closed: the value's `` `other `` is intersected away, and the marker then claims an exhaustiveness the position does not have. That is the one-side-open case `meet_openness` said no program reaches; its doc now names the caller.

Pinned rather than fixed, since the exact rule is a change to the variant meet itself: `a_settled_negative_position_closes_an_open_child_demand`. No program observes the narrowed reading — `test_default_arm_under_a_record_field` (new) runs `def pick(r): match r.v: case `a(n): … case _: …` against `pick((v=`b(5)))`, and the default still fires, the arm being compiled from the `Case`. Say the word if you want the exact meet in this PR.

The comments: the undetermined arm's tag argument is now scoped to it ("a negative position reaches this arm holding at most a variant shape"), and the merge arm states what it does to a child variant.

src/ccl/infer/solver/compact.rs
dpmills: Both in a835febbb.

The design doc records it where `simplify_type` is described: all three rules read the variable sets `compact_type` deposits, a negative position's set holds both sides' variables, and the occurrences the analysis sees there are not the ones the coalesced type materializes. The refinement-safety argument survives as stated — refinements sit on the position and the rules rewrite variable ids — but it rested on co-occurring variables sharing a position, and a position's contents now come from two sides.

Style: `read_opposite.then(…)` is a plain `if` filling the `Option<CompactType>`.

src/ccl/infer/solver/compact.rs
dpmills: Added in a835febbb. `a_negative_position_meets_both_sides` reads its two shapes at a codomain as well, and each asserts what a positive merge would have destroyed: the singleton the value establishes, and the field the demand omits.

Verified discriminating rather than vacuous — ungating `read_opposite` to `allow_fallback` fails the refinement assertion, `Some(0)` against `Some(1)`.

src/ccl/lower/comprehension.rs
dpmills: Reworded in cecb52c1f.

The inner comment states the test as the annotation's shape — a kind with `domain: Hole` — rather than `groupby`, and names what a comprehension can iterate carrying it: a `groupby`, a `set` / `map` re-keying (`lower_rekeyed`), or a conditional comprehension's arm. Confirmed `[v + 1 for v in map([(1, 10)])]` takes the path; `set` reaches it too and then fails on its `Unit` element rather than in lowering.

The enclosing comment no longer reads as contradicted: it says what separates the branches — where the kind comes from, and whether the annotation already names the domain — instead of claiming only an un-named domain is stamped.

PR #155, opened 2026-09-03
Record a mutation loop's products against the body statements that produce them

groundlar:
`mut_elim` opened one `letrec.loop` recording over everything a `for` statement becomes, so the loop statement claimed every product and the write and feed statements in its body reached nothing — the inspector had no answer for the lines that do the mutating. Each feed statement, and each accumulating variable, now carries the `StmtSite` of its statement (its `ExprStmt` plus its effect marker), and the products of its tap or its recurrence slot are recorded under a nested `letrec.feed` or `letrec.accumulator`. A recurrence slot belongs to the variable rather than to any one write, so a variable the body writes twice is named on its first write and blames the rest: every write answers with the slot's products, and only the first claims ancestry. On a probe loop this moves 9 of the loop statement's 46 operators to the write (6) and the feed (3). Node ids and the golden fixtures are unchanged — the change moves which recording adopts a mint rather than minting anything.

### Reading order

- `StmtSite` and `AccumulatorVariable` in `src/ccl/mut_elim.rs`. `collect_writes_in` threads the enclosing `ExprStmt` the walk arrived through, and `collect_feed_only` does the same for feeds. `AccumulatorVariable` is per variable — its position in `InductionFold::accs` is the variable's slot index — and holds one `StmtSite` per write; `enter` splits those into the named first and the blamed rest, with a `TODO` for the per-product attribution it does not attempt.
- The three places that open the nested recording: `fold_induction_loop`'s slot machinery, `InductionFold::acc_view`, and `transform_feed_only_loop`. `acc_view` opens its own because its only caller, `transact_phase::fold_cross_domain_loops`, holds a recording on the loop statement and cannot see `AccumulatorVariable`'s writes.
- The two bounds on the split. An accumulator-free loop is its feeds, so the whole expansion moves to them and the `for` rides their blame column (`StmtSite::blamed_in`) rather than an ancestry edge it did not earn. A conditional feed rides one lambda over the whole body, so no per-feed recording has anything to adopt and the products stay on `letrec.loop`.

### Tests and docs

`panes.rs` splits each statement's image into what descends from it and what is only blamed on it, and pins three programs' images non-empty and pairwise disjoint. Two unit tests read the parent of a mint, because the products share the shape the fold builds for the trailing final read and a pane pair cannot tell them apart: `acc_view`'s nodes resolve to the write statement, and a two-write body's slot blames both writes. [`provenance.md`](src/ccl/design/provenance.md#where-to-open-a-recording) records the per-construct rule as a fourth refinement, notes that one construct can be several statements, and adds the induction fold to the shared helpers.

### DIFF
--- a/src/ccl/mut_elim.rs
+++ b/src/ccl/mut_elim.rs
@@ -1146,8 +1146,18 @@ impl InductionFold {
     /// A per-position cross-domain read of that accumulator (`acc(pos)`) is this
     /// applied at `pos`; the transaction phase uses it to resolve a `commits(r)`
     /// decision that reads an induction accumulator at its request position.
+    ///
+    /// The view is the accumulator's product, so the recording is the
+    /// accumulator's: one slot's stream, named on the statement that writes it.
+    /// Opening it here rather than at the call site is what puts it on the write
+    /// statement at all — the caller is `transact_phase`, which holds the
+    /// enclosing `transact.cross_domain_fold` recording on the *loop* statement
+    /// and has no access to [`Accumulator`]'s site.
     pub(crate) fn acc_view(&self, i: usize) -> Expr {
-        let vty = &self.accs[i].ty;
+        let acc = &self.accs[i];
+        let _g = acc
+            .site
+            .enter(ACCUMULATOR_LABEL, provenance::Nature::Expansion);
         writes_index_view(
             &self.hist,
             &self.hist_ty,
@@ -1155,7 +1165,7 @@ impl InductionFold {
             &self.writes_ty,
             &self.decision_ty,
             i,
-            vty,
+            &acc.ty,
         )
     }
 }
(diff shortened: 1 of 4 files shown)


src/ccl/mut_elim.rs
```diff
@@ -802,6 +843,20 @@ fn transform_loop(target: TypedBinding, iter: Expr, loop_body: Expr, cont: Expr)
...
 
+/// One accumulator of an induction loop: the mutable variable, the value type its
+/// recurrence slot carries, and the write statement that made it an accumulator.
+///
+/// The site is what the slot's products are recorded against — the snapshot
+/// projection, the guard's seed, the trailing final read — so each accumulator
+/// of a two-accumulator loop resolves to the line that writes it rather than
+/// both resolving to the loop.
+#[derive(Clone)]
+pub(crate) struct Accumulator {
```
dpmills: There can be multiple writes to a single variable (either across loops or in a single loop).  Is this class tracking the variables or the writes? It either needs a list of sites or needs a different name for the struct.  Also, do writes always occur inside ExprStmts? I don't remember offhand.  For constructs where returning unit is fine, we don't fundamenally need an ExprStmt

src/ccl/mut_elim.rs
```diff
@@ -59,6 +59,41 @@ use crate::ccl::{
...
+///
+/// A recording names a single node, and a statement's expansion is about both
+/// of these. [`StmtSite::enter`] names the `ExprStmt` — the outermost node the
+/// expansion replaces — and blames the marker, so the attribution carries the
+/// construct's own span as well as the statement's and either node reaches the
+/// products.
+#[derive(Clone, Copy)]
+pub(crate) struct StmtSite {
+    stmt: NodeId,
+    effect: NodeId,
```
dpmills, line 89: "stmt" and "effect" are confusing names here.  At least in my head, what is labeled "effect" here is actually the statement, and the exprstmt is a container that holds the statement and whatever else follows.  Maybe stmt and expr_stmt?

dpmills (approved):

src/ccl/mut_elim.rs
groundlar: Renamed the container to `expr_stmt`, kept `effect` for the marker — agreed the old `stmt` was the misleading half. Went with `effect` rather than your `stmt` only because 27 destructures across `mut_elim`, `transact_phase`, `channelize`, `lower/stmts` and `panes` already bind the `ExprStmt`'s child that way, so `stmt` here would be a second name for it. Both fields now carry a doc line saying which is which, including that they are equal for a marker reached off any statement chain.

src/ccl/mut_elim.rs
groundlar: Variables, and you were right that it needed both changes. It is now `AccumulatorVariable` holding `writes: Vec<StmtSite>` — the struct always tracked the variable (its position in `InductionFold::accs` is the slot index, so a variable written five times has always had one entry), but a single `site` field claimed otherwise.

The behaviour change: `collect_writes` no longer skips a repeat write, and `AccumulatorVariable::enter` names the first write and blames the rest, so every write to the variable answers with the slot's products. Only the first claims ancestry — a later write mints nothing, since `transform_chain` inlines its value into the read-your-writes environment and its marker and `ExprStmt` die. This mattered most for `if p: total += x` / `else: total += y`, where the second arm previously reached nothing at all. New test: `a_repeat_write_blames_the_slot_it_shares`. No fixture movement — no corpus program writes one accumulator twice.

`enter` also carries a TODO for the precision this does not have: blaming every write says which writes a product is about, not which one it came from. Two of the five products (the guard's seed and the trailing final read's default) are the variable's pre-loop binding, so they really belong to its `MutDecl` — which needs its `NodeId` threaded down from `rewrite`'s `MutDecl` arm, an ancestor of the loop statement and outside the body `collect_writes` walks. Happy to do that as a follow-up if you want the finer attribution.

On `ExprStmt`: not assumed. `collect_writes_in` threads `stmt: Option<NodeId>` and falls back to the marker for both ids, and `StmtSite::enter` skips the blame edge when they are equal, so a construct that drops the `ExprStmt` degenerates to a one-node site with no code change — it just loses the statement-span half of the attribution. Post-`flatten_spine` every write *on a statement chain* is a direct `ExprStmt` effect, which is what the loop paths rely on; that is now stated on the field rather than only at the collector.

PR #156, opened 2026-09-04
Type a group-by at its present-key domain, and lower the `Map`/`Set`/`FullMap` annotations

dpmills:
`groupby(c, key)` typed as a total function from the bare key type, so a lookup at any key was accepted and an absent key answered with the empty group; `Map(K, V)` and `Set(K)` annotations were a lowering error, leaving no surface spelling for a keyed collection. This gives a group-by the type it has — a `FullMap` whose domain is the **present-key domain**, the keys its key morphism produces — and lowers `Map`, `Set`, and `FullMap` onto `Type::map_of`, `Type::set_of`, and `Type::full_map_of`. A group-by is now consumable at an annotation (`FullMap(_, _)`, the key elided, being the only form its domain satisfies), and a bare-key lookup is rejected rather than silently total.

## The present-key domain

`Builtin::CollectionContains` is new: the characteristic predicate of a collection's outputs, `∀ι κ. (ι ⤇ κ) ⇒ (κ ⇒ Bool)`, carried as an `OperatorSchemes` scheme because `κ` pins the key type from inside the refinement. `present_key_domain` in `src/ccl/lower/exprs.rs` builds `{K | __elem ▷ ((c ≫ key) ▷ collection_contains)}` and stamps it on `groupby`'s key binder; the `SharedHole` pinning the key type moves into that refinement. Nothing executes it — planning still rebuilds the site as `converse(c ≫ key) ≫ map(c)`, so the predicate rides on types. Why the morphism is named, and what retires that: [The key domain is the key morphism's image](src/ccl/design/collections.md#the-key-domain-is-the-key-morphisms-image).

## What stops type-checking

`g(k)` and a lookup keyed from the source, `g[x]`, both fail naming `collection_contains`: no rule discharges membership yet ([Lookup: membership discharge](src/ccl/design/collections.md#lookup-membership-discharge)). Their runtime cases move to `tests/type_check.rs` as pinned rejections; the two dependent-application discharge tests move to a filtered comprehension, dependent without needing a membership proof. Two defects are pinned alongside: a consumed boxed `Map(_, _)` annotation escapes the comprehension's binder, and a consumed `g <: FullMap(_, _)` records an open bound.

## Binders and scopes

`Type::map_of` and `Type::full_map_of` mint a Pi binder, since a map's value may depend on its key. `Subst::aligned` names the binder alignment `constrain_go` already drew and `complete_annotation` now needs when filling an exact annotation's codomain. `uniquify` walks a lambda's own arrow at the two scopes its halves belong to; `Typing::type_annotation_predicates` types a refinement riding `param.ty` in the enclosing scope (Emit only — Check trusts a resolved predicate); `strip_iterate_markers` rewrites a `Compose` in place so its kind stamp survives.

## The group-by recognizer splits in two

`partition_key_of` asks whether a refinement partitions by a key, independent of spelling; `match_pointful_site` asks whether the term is the shape `lambda_elim` emits. Declining is safe — the site falls back to the generic iterate/restrict lowering.

Docs: [chl-spec, Direction: collection types](docs/chl-spec.md#63-direction-collection-types-decided) lists `FullMap` among [six collection types](src/ccl/design/collections.md#the-six-collection-types), and `collections.md` gains the group-by's exact type, lookup discharge, and keyed-entry sections.


dpmills:

- #10
    - #11
        - #45
            - #156 ◀
                - #157
                    - #158
                        - #159

sortalongo (approved):
Approving. The present-key domain is the right shape and the argument for naming the morphism rather than the collection is the correct one for the lattice we have today. `FullMap` as the unboxed `(𝑘: 𝐾) ⤇ 𝑉` is a real simplification over trying to make `Map` describe a group-by, and the four new rejection pins (`a_key_from_the_source_does_not_yet_carry_its_key_domain`, `a_groupby_is_a_full_map_but_its_key_type_is_not_yet_writable`, `full_map_lookup_needs_no_presence_proof`, `full_map_annotations_are_satisfiable`) are exactly how a partly-built rule should be recorded — each one goes red the day its gap closes.

Six things, the first of which I'd want before merge.

## The seventeen `#[ignore]`s are the wrong mechanism, on either reading

The description argues both sides in one paragraph: "The rejection is by design and permanent" and "Those cases come back restated as a checked lookup." Both readings say `#[ignore]` is wrong.

If the rejection is permanent, the assertion (`sum(g(2)) == 2`) is behavior we have decided against, and an ignored test asserting rejected behavior is dead code. The right encoding is twenty lines away in the same file: `a_key_from_the_source_does_not_yet_carry_its_key_domain` calls `infer_program_err` and asserts the error names `collection_contains`. That test runs and it goes red when the discharge lands.

If they come back restated, they are not permanent — and what comes back is a different program, `g[k]?` with `Option` handling rather than `g(k)`. Keeping `sum(g(2)) == 2` on ice preserves nothing anyone would write. #158 confirms the drift: two return un-ignored and the rest are *retitled*, so some are on their third state and still inert.

The cost is measurable. Nothing runs ignored tests — no `--include-ignored` in `ci.sh`, in `.github/workflows/`, or in `scripts/` — so each one reports the same green whether the feature landed, regressed, or was deleted. This branch takes `tests/` from 8 `#[ignore]` attributes to 17.

One of them is pure collateral. `test_grouping_built_once_through_a_lookup` tests the *sharing* property — one `converse` per program however many uses — which has nothing to do with membership proofs. It was deferred because its cases happened to spell three uses as `g(1) + g(2) + g(3)`. `test_grouping_built_once` is now down to a single comprehension-only case, so the invariant most likely to break has the least coverage. Re-express it with a spelling that compiles.

Per function, one of three: delete it, convert it to a live rejection pin, or re-express it where the property under test is not about lookups. None of the eight needs `#[ignore]`.

## The transparency argument is contingent, and reads as settled

"Representation: the key domain is the key morphism's image" justifies naming the morphism this way:

> refinements relate by structural predicate equality rather than implication, so there is no entailment step for a proof to land in instead.

That clause is load-bearing and temporary. [type-inference.md](src/ccl/design/type-inference.md) already lists **SMT-backed refinements** as planned work — "reasoned about — implication, not just equality — by an external SMT solver such as Z3" — and says implication is what the lattice "deliberately does not have" today. Once it exists, an opaque named key-set predicate is dischargeable from an axiom, and naming the *collection* beats naming the *morphism* on size, spelling-independence, and not carrying a producer inside a type. Transparency is precisely what entailment retires.

Not asking for a redesign. Please cite the SMT entry from that section and say the morphism-naming is what the absence of implication forces, so the opaque form is its successor. As written the section reads as a decision rather than as the current answer, which is the `[Decided]`/`[Open]` line.

The timing matters for the same reason as the `Map <: Collection` pin: this branch and the three above it put `SKU`-shaped types into program signatures, and migrating the representation afterwards changes types programs are written against.

## Domain identity is term identity — where does canonicalization happen?

The section concedes that "distinct spellings of one collection (a `let`-bound source and its inlined literal) are distinct domains." Implication does not fix that: the obligation needs the definitional fact `c = [1,1,2]` rather than an entailment between predicates over values, so it is canonicalization work, not solver work.

The sharper version is the point-free split. The new expectation in `joins_aggregates_groupby.rs` renders a key domain as `{Int | __elem ▷ (([1, 2, 3, 4] ≫ (λ y : Int → y // 2)) ▷ collection_contains)}`, and its comment notes the morphism "stays **pointful** — planning point-frees the predicates it reifies into a `Restrict`, and this one it never reaches." So point-freed and pointful predicates coexist. Is a shape reachable where one collection's key domain is point-freed at one position and not at another? If so the two are structurally unequal and the membership fact does not transfer; if not, it is worth saying why in the section.

Two consequences either way. A key domain embeds its whole producer, so the type grows with the producer and appears wherever the key type does. And identity of a *type* now depends on a rewrite that runs in planning, which is the same identity-by-shape hazard the Σ rule's ρ substitution had.

## `FullMap` is an unenforced assertion

"Totality is claimed, not checked," and "Nothing distinguishes a `FullMap` from any other dependent data function, so every data function satisfies a `FullMap` annotation." Both are stated plainly, which is right, but together they mean the annotation currently carries no property. Worth a `[Tentative]`-style marker on the type in the six-types list, and a pointer to what would check it, since `FullMap` is the type the total-lookup story rests on.

## Vocabulary this stack retired keeps coming back

Four terms settled on #45 at my request re-appear here:

- `Any` as the kind name, twice — the `Collection(𝑇)` line in the six-types list and `test_keyed_collection_widens_to_collection_like_any_other`'s comment. It is `TypeKind::Type` in the code.
- `width-to-top`, once, in the same test comment.
- "realized", promoted into a section heading — `Operations: how the trait layer is realized`. The #45 answer for keeping the word was that it names `TypedExprNode::Realize`; a heading about the trait layer is not that.
- `arrow`, seventeen times, including doc prose ("operations dispatch on the kind rather than on the arrow"). The #45 answer was that a vocabulary pass settled on "function" throughout.

One sweep of the stack rather than four spot fixes, and worth a check on the branches above this one.

## Two smaller things

`test_keyed_collection_widens_to_collection_like_any_other` pins `Map(Int, Int)` into a `Collection(Int)` parameter as a property, with a doc comment arguing there is nothing to withhold. That question is open in `collections.md`; please restate it and `test_map_consumption_is_kind_blind_interim` as interim pins naming the open question. Detail on #45.

The branch's commit message says "This adds `TypeKind::Keyed(K)` as a Σ witness kind." `TypeKind` at this head has four variants — `Enumerated`, `UIntRanges`, `SubtypesOf`, `Type` — and zero occurrences of `Keyed`. The PR body does not make the claim, so a squash-merge from the body drops it, but `jj log` readers see a variant that was never built.

## Scope

I read the design sections, the type-level tests, and the regression handling. I have not reviewed `planning/groupby.rs`'s `partition_key_of` / `match_pointful_site` split, `Subst::aligned` and the non-termination it fixes, or `Typing::type_annotation_predicates` — the places a correctness bug would sit rather than a design question. Happy to take those in a second pass if you would rather not carry them into #157.

dpmills:
All six done. `tests/` is back to the 8 `#[ignore]`s it inherits, from 17; `a_bare_key_lookup_on_a_groupby_is_refused` carries every deferred program and asserts the rejection names `collection_contains`. The two discharge tests are re-expressed over a filtered comprehension rather than deleted — that was #158's idea, moved down here.

Canonicalization is answered and measured: no split is reachable, because `fn_of_bare_predicate` returns `f` verbatim from a bare `__elem ▷ f`, which is the form a key domain is born in — compilation is the identity on it. Implication wouldn't relate a `let`-bound source to its inlined literal either; that needs the definition.

`FullMap`'s claim is marked where it's empty: `FullMap(_, 𝑉)` asks only for a data function, and a written key is an ordinary invariant-domain obligation.

The transparency section cites the SMT roadmap entry and names the opaque form its successor. Vocabulary: see the thread on #45 — `Any` and `width-to-top` are absent at every head. `TypeKind::Keyed(K)` is gone from the commit message.

PR #157, opened 2026-09-04
Add the `set` and `map` constructors, lowering both as a `groupby` with a collapse aggregate

dpmills:
The nominal `Set` and `Map` types exist with no term that builds a value at one; `set(xs)` and `map(kvs)` are that surface. Each lowers to a `groupby` on a key morphism with every group collapsed by an aggregate, so the value is a data function at this site's present-key domain and injects into the nominal Σ through a `box`. A repeated key is decided by the collapse aggregate — `set` absorbs it, `map` rejects it.

## One shape, two collapses

`lower_groupby` is extracted from the `groupby` arm and now returns the key domain it stamps on its binder; [`lower_rekeyed`](src/ccl/design/collections.md#constructor-lowering-runtime-groupby-now-constant-folding-later) is the constructor shape over it. `set`'s key morphism is the identity and its collapse the total `Drain`, folding a group of any size to the one `unit` of `Set(K) = Map(K, unit)`. `map`'s morphism is `.0` and its collapse `Sole ▷ .1`, where `Sole` is partial — a group of two entries has no value — and that partiality is the duplicate-key rejection. `lower_rekeyed` carries the three load-bearing details: the shared key domain, the η-expanded shape, two stamps on the iteration lambda.

## Partiality becomes a stated aggregate law

`AggregateKind::is_partial` is a fourth per-variant law beside output extent, identity and fold ([ir.md](src/ccl/design/ir.md#aggregate--first-class-aggregation-node)), and `initial_accumulator` asserts what it decides: a seed column is empty exactly for a partial aggregate. `Builtin::as_aggregate` replaces `operator_conversion.rs`'s hand-maintained `builtin_to_aggregate` and supplies `iterates_arg`'s aggregate half, which a second list had let `Sole` fall out of.

## Deferred, and recorded

Compile-time construction and the spec's compile-time duplicate-key error both wait on constant folding, and a duplicate today faults the process rather than the query ([collections.md](src/ccl/design/collections.md#a-duplicate-key-is-a-process-fault-today)). Two known failures are tests — one `#[should_panic]`, one `#[ignore]`d: a re-keying over a literal whose elements share one singleton type, which a plain `groupby` also fails, and a `set(…)` inlined as a comprehension source, which planning leaves without an iteration site. `tests/type_check.rs` pins the inferred key domain and the nominal injection.


dpmills:

- #10
    - #11
        - #45
            - #156
                - #157 ◀
                    - #158
                        - #159

sortalongo (approved):
Approving, contingent on the items below. The one-shape-two-collapses lowering is the right factoring — `set` and `map` differing only in a key morphism and a collapse aggregate is a genuine collapse of a duplicated shape rather than two constructors sharing a helper — and minting the `KeyDomain` once per site so the group-by's key binder and the constructor's iteration binder relate is the detail that would have been easy to get wrong. Enforcing the key type through the nominal annotation (`x <: Set(String) = box(set([1,2,3]))` rejected) is the property I most wanted to see pinned.

## `test_rekeying_over_a_singleton_literal` is ignored for a bug #153 fixes

The description calls this pre-existing: "the element type staying `Int@1` where the key domain's base widens to `Int`." That is #153's bug, in the same words it uses — "the comprehension over it resolved its own domain to `Int` while every occurrence of that domain's binder resolved to `Int@1`, and a data function's domain is invariant, so the post-lambda-elim check rejected the tree with `DataDomainMismatch`". Same singleton refinement, same widened base, same invariant-domain rejection.

#153 is not an ancestor of this stack. So this test starts passing the moment the stack rebases onto a main carrying it, and because it is `#[ignore]`d nothing will report that. This is the `#[ignore]` argument from #156 demonstrated rather than asserted: an inert test cannot tell you its gap closed. It also sits on the path to #159, since the description notes it blocks the single-entry seed a mutable map wants.

Rebase and drop the ignore, or at minimum stop calling it pre-existing and name #153 as the fix.

## `Sole`'s duplicate-key check is a process fault on user data

```rust
(AggregateKind::Sole, acc, vs) => {
    let incoming = end - start;
    assert!(
        acc.len() + incoming <= 1,
        "sole: {} elements under one key; a map literal's keys are distinct",
```

The comment says what makes this different from everything around it: "enforced here rather than at compile time because only the key *values* decide it."

`interpreter/tile_operators` holds 169 `panic!` and 55 `assert!`, and every one I sampled is about internal tile structure — "Accumulator must be Aggregation tile", "not a function tile", "codomain should be UInts", "Can only converse functions". Those say the compiler handed the operator a shape no pass should produce. This is the first whose trigger is a comparison of user data values, and there is no runtime-fault channel for it to use instead, so a duplicate key takes the process down. In a program serving four endpoints, that fails all of them rather than the one request that carried the duplicate.

`assert!` rather than `debug_assert!` is the right choice — the alternative is silent corruption — so this is not a request to weaken the check. Two things: say in `collections.md` that a `Sole` violation is a process fault today and what it should become, alongside the constant-folding note that already covers the timing; and confirm whether the map comprehension `[𝑘 ↦ 𝑣 for …]` will lower through `map`/`Sole`, since that decides whether a map built from request data inherits this. The doc mentions a map comprehension only as a shape whose domain is unknowable until coalesce, so the lowering is not stated anywhere yet.

Related and worth one sentence in the doc: `set([1,2,2,3])` absorbs the duplicate through `Drain` while `map([(1,10),(1,20)])` faults through `Sole`. Same `lower_rekeyed`, opposite treatment of the same input condition. Both are right — a set absorbs duplicates by definition — but the only thing distinguishing them is which collapse aggregate the constructor picked, and the doc should say why that is the whole difference.

## `Sole` makes "aggregate" stop meaning "total fold"

Every other `AggregateKind` is a total fold with an in-band identity. `Sole` is partial, and its identity is out-of-band — `none` as an empty column, presence carried by column length, because "there is no in-band identity to seed with as `Sum` seeds with `0` and `Max` with `MIN`".

That is a new shape in the table, and the partiality currently lives implicitly in one variant's `accumulate` arm: nothing lets a consumer ask whether an aggregate can fault. If partiality is a real property of an aggregate — and `Sole` says it is — the table should carry it rather than leaving it to be rediscovered at the next partial law.

## Vocabulary, and one sweep that is already working

"realization" / "realized" now heads three `collections.md` sections — `Operations: how the trait layer is realized`, `Lowering realization: the key binder states its domain`, and `Realization instantiates the witness inside the predicate`. The #45 resolution was that the word stays only where `TypedExprNode::Realize` is discussed.

Credit where it is due: a commit on this branch drops "keyed collection" for exactly the right reason — "keyed-ness is not a primitive, so 'a keyed collection' named a category the design doc denies" — and renames `test_undriven_keyed_collection` to say what the gap is instead of what kind of value hits it. That is the sweep doing its job. It needs to run down the stack rather than stopping here: #156 is titled "Type a keyed collection's keys as its key morphism's image" and carries `test_keyed_collection_widens_to_collection_like_any_other`.

## Scope

I read `aggregate.rs`, the design sections, the deferred tests, and the type-level cases. I did not review `lower/exprs.rs` — the `lower_rekeyed` / `lower_groupby` split, and the largest change here — or `infer/schemes.rs`. Neither is where a design question sits, but both are where a lowering bug would.

dpmills:
`test_rekeying_over_a_singleton_literal`: #153 is still open, so the rebase option wasn't available. It's a live `#[should_panic]` pin instead of an ignore, so it reports the day a base carries the fix, and the comment names the mechanism rather than calling it pre-existing.

`Sole`: the doc states the process fault, what it should become, and why the assertion stays until that channel exists. The map comprehension does inherit it — the spec reads `[k -> v for …]` as a `Map` exactly as a literal does, and `lower_rekeyed` is the one shape both re-keyings take — and that lowering is undecided, which is where it gets settled. The `Drain`/`Sole` asymmetry is stated as the collapse aggregate and nothing else.

Partiality is `AggregateKind::is_partial`, and load-bearing: `initial_accumulator` asserts a seed is empty exactly for a partial aggregate, so a new partial one seeding a value is caught rather than making its empty group indistinguishable from a group of one.

PR #158, opened 2026-09-04
Add the checked lookup `c[k]?` as its own typing rule and tile operator

dpmills:
CHL's only subscript is the proven `c[k]`, an application demanding an index whose type proves it lies in the collection's domain — a proof only the collection hands out, so a lookup at a key from anywhere else has no spelling. This adds the checked form `c[k]?`, answering `Option(V)` for any key of the collection's key type: a `?` suffix in the parser, `Builtin::LookupChecked` through lowering and inference, and a `CheckedLookup` tile operator deciding presence against the collection's domain. A `Map` and a `Set` answer it today, at a point and over a key stream.

## The rule is not an application

`emit_apply` intercepts the builtin instead of a scheme, which would have to name a key type only a `SubtypesOf(K)` kind states. The key owes the collection's key base, membership refinement peeled off, so presence stops being a typing question; a dependent codomain discharges its key binder to the key term, so a group-by's `g[k]?` answers the group refined at `k`. Emission computes that discharge and stamps the operator; a check reads it back through the new `Typing::keyed_value_at` rather than re-substituting into a predicate planning may already have compiled point-free. The rule: [collections.md, "The checked lookup"](src/ccl/design/collections.md#the-checked-lookup-𝑐𝑘).

## From a pair to an operator

Lowering emits the tupled `(c, k) ▷ lookup?`, so the operator never carries a function value. Where the collection does not vary, `simplify`'s `try_partial_lookup` rewrites `⟨const(c), g⟩ ≫ lookup?` to `g ≫ (c ▷ curry(lookup?))` — required, since a streamed collection cannot be broadcast into every row. Op-conversion compiles that partial application and the syntactic pair alike to `CheckedLookup`, which searches the collection's domain column and withholds `` `none `` until that domain is terminal, an empty tile being no evidence of absence.

## Where a lookup is refused

Typing rejects a range domain (`xs[0]?` — no membership refinement to relax), a tuple, a `<:`-bounded parameter target, and a key of the wrong base type. Op-conversion rejects a group-valued answer, which nothing materializes. The proven form is unchanged; a mismatch against a present-key domain now carries a hint naming `c[k]?`.

## Adjacent runtime fixes

`VariantProject` passes a domain release through to a streamed scrutinee, which a `match` inside a `with begin():` block needs (`a_variant_match_inside_a_block_releases_per_commit`); seven multi-input operators drop hand-rolled consumer sharing for `shared_consumer`/`forwarding_consumer`.

Typing is pinned in `tests/type_check.rs` (`a_checked_lookup_is_not_an_application`, `checked_lookup_boundaries`), execution in `checked_lookup_answers_presence` and `checked_lookup_over_a_key_stream`, whose filtered cases cover row alignment. [chl-spec.md, "3.9 Subscript and attribute access"](docs/chl-spec.md#39-subscript-and-attribute-access) records `Set`'s new subscript and the open question of pinning a live domain at a cut instead of waiting for termination.


dpmills:

- #10
    - #11
        - #45
            - #156
                - #157
                    - #158 ◀
                        - #159

sortalongo (approved):
Approving. `c[k]?` as its own total operation rather than an application is the right call and the argument for it is the one I wanted: an application would first have to relax `𝑐`'s domain, and a collection type with its domain relaxed is a type no value has. That closes the question I was going to raise about intercepting the apply rule. The withholding design is right too — terminality as the *readiness* condition, absence decided rather than read off an empty tile — and the operator's three cases are the minimal form of it.

Two questions below worth taking as follow-ups rather than resolving here, one thing to fix before merge, and the vocabulary drift measured across the stack.

## Before merge: the description states the inverse of the rule

The body says:

> What has no answer is rejected by name rather than as a shape mismatch: a key-dependent codomain — a group-by's group, the deferred `m[𝑘] : 𝑉(𝑘)` — and a target unresolved where the rule runs

and cites `a_key_dependent_collection_has_no_checked_lookup`. That test is not at this head. What is there is `a_key_dependent_lookup_discharges_the_key_binder`, and `collections.md` states the opposite rule:

> **A key-dependent codomain answers at the key.** A group-by's group is refined by the key binder, and `𝑔[𝑘]` reads the group refined at `𝑘`, the binder discharged to the key term… That is the deferred `m[𝑘] : 𝑉(𝑘)` this document names under `groupby`'s exact type, now paid.

The behavior is better than advertised, which is the good version of this problem — but we squash-merge, so the body becomes the commit message on main, and it currently describes a rule the branch does not implement and names a test that does not exist. Please re-read it against the head.

This is the third one in the stack: `box` is idempotent on #45, `TypeKind::Keyed(K)` on #156, and a reversed rule here. The common cause looks like descriptions drafted against an earlier design and not re-read, which is worth a habit change rather than three fixes.

## Follow-up 1: the condition is a cut, not termination

`CheckedLookup` withholds until the domain is terminal, so on a source that never terminates the lookup answers `` `some `` or never answers. The doc notes "a live source is the ordinary case here" without following that through, and a live source is what #159 puts behind this lookup.

Terminality is standing in for "the domain has a definite value", and we have a better notion of that already: a cut. `/stats` reads `orders.filter(\o -> o.time < txn.current_time())` — a pinned view of a feed that never ends, and it is decided. A `Mut(…, Txn)` store read inside `with begin():` is pinned by the transaction the same way. So the condition for `c[k]?` should be that the domain is pinned at a cut, which admits both of those and rejects only the bare live feed.

Rejecting on *unboundedness* instead would be too strong — it rejects `txn_kv`, which is a north-star program doing an `Option` lookup on a live transactional store — so the predicate has to be the pin rather than the bound.

Two alternatives, both declined: provisional `` `none `` with a later correction is not a separate design, it is a cut with better ergonomics, and it lands with IVM; a timeout or watermark makes the answer a function of wall-clock, which is the same nondeterminism this design already refuses when it declines to read absence off an empty tile.

Worth noting that the precedent is yours and already in this branch. `reject_unanswerable_lookup_collection` rejects a non-scalar codomain because "the operator answers `None` for 'the collection has not decided this key yet', and a permanent shape mismatch reported that way is a lookup that spins instead of erroring." An unpinned live domain is the same category through the same channel, one case wider.

Please write the cut formulation up as a spec Direction note rather than leaving it in a review thread, and state the consequence in the doc here — that a lookup on an unpinned live domain never decides absence.

## Follow-up 2: the operator pins both legs and rescans everything

Separate from the semantics, and worth fixing independently so it does not park behind that decision.

`get_impl` calls `self.keys.get(self.keys.tiling().universal_guard())` and the same for the collection on every poll, and for a `SealedFunction` that guard is `FunctionGuard::Domain(Predicate::True)` — the whole domain, unrestricted. A consumer holding a universal guard pins its upstream at full extent, so neither leg can release history; the `compact()` calls act on the fetched copy, not on the producer. On the streaming-key shape this branch tests, retained memory then grows without bound whether or not the collection terminates. I have not traced the release path end to end, so tell me if `ProducerReleases` already covers this — but the universal guard on both legs is what makes me doubt it.

Second, `answer_for` does a linear `(0..domain.len()).find(…)` per key and `get_impl` re-scans every key on every poll, so a poll costs O(keys × collection) with both terms growing. `/stats` would pay the whole order history against the whole key set on each request. Answering the new suffix, and indexing the domain column, are the two obvious halves.

Please file both of these as issues; no need to carry the fixes in this PR.

## Good catch

`a_checked_lookup_is_not_an_application` pins that relating the key and the collection's keys to a *common* type — the literal reading of `SubtypesOf` — is satisfied by any join, so a `String` key against an `Int`-keyed map widens the key type instead of failing. That is the same class as the Σ rule's kind premise being satisfiable by widening, and finding it unprompted is the right instinct.

One residual question, small: if `c[k]?` is categorically not an application, why is the rule reached by intercepting `emit_apply` rather than through its own emission path? The stated category and the code path disagree, which is where the next reader loses the thread.

## Vocabulary across the stack

Occurrences in `collections.md` at each branch head:

| | collections-design | groupby-keyed-collection | set-constructor | keyed-lookup |
|---|---|---|---|---|
| "keyed collection" | 0 | 3 | 7 | 9 |
| "realiz*" | 1 | 7 | 10 | 11 |

Both grow monotonically, and this branch adds ten more "keyed collection" occurrences in its own added lines — the term a commit on #157 drops as "a category the design doc denies", having only brought its own count from 3 to 7. Per-branch cleanups are adding more than they remove, so this wants one sweep at the top of the stack.

## Credit

This is the first branch in the stack to shrink the inert test set — eight `#[ignore]` attributes removed against six added. That is the direction I asked for on #156.

## Scope

I read the design sections, the emission rule, the operator, and the type-level tests. I did not review `simplify.rs`'s `partial_lookup` rewrite or the `Curry` shape it introduces in op-conversion.

dpmills:
Description rewritten; you were right that it stated the inverse.

The cut is written up as a Direction in `chl-spec` §3.9 — the condition should be the pin, not a bound, since unboundedness rejects `txn_kv` — with both alternatives recorded as declined. `collections.md` states the consequence: a lookup on an unpinned live domain never decides absence.

Both perf items are filed. `ProducerReleases` does not cover the first: the keys leg releases, but `release_impl` releases the collection leg only universally, by design, since every key is answered against the whole of it. It's the unindexed search rather than a wrong guard.

On `emit_apply` — the category is a claim about the rule, not the term. Lowering emits an ordinary builtin application, so that is where the node arrives; a separate emission path means a separate `TypedExprNode` for no gain. Said so in the doc.

One to flag: restacking this branch dropped five of its own tests (`checked_lookup_answers_presence`, `a_group_valued_lookup_is_rejected_by_name`, `checked_lookup_over_a_key_stream`, and the two discharge tests) and broke the build at three heads. Restored, and every head builds and passes.

PR #159, opened 2026-09-04
Add the keyed write `m[k] := v` as a whole-collection `insert`

dpmills:
CHL could declare a mutable `Map` and replace it wholesale, but not write one key: `m[k] := v` had no assignment target, no typing rule, and no runtime. This adds it on both sequencing domains — a loop's induction store and a `with begin():` transaction — by rewriting the write to what it denotes, `m := insert(m, k, v)`, before any phase that builds a recurrence. A mutable collection is now written and read back in one program, and the whole collection is the unit of commit: two transactions touching one map conflict whatever keys they touch.

## The write

`AssignTarget::Subscript` holds the collection and the key as expressions, and only `:=` takes one: every other statement form rejects it at lowering, and `m[k]? := v` is refused in target position. Typing is one rule with the checked lookup, so `m[k]` names the same value type read or written — the key against the key type, the value against the codomain, a key-dependent codomain rejected.

`MutWrite` gains `key: Option<Box<TypedExpr>>`, and its lifetime is one window: lowering mints it, `mut_elim::desugar_keyed_writes` erases it at the head of the transact phase, so every phase below sees one write shape. Within that window `walk_children`/`walk_children_mut` reach the key, carrying every pass built on them; the walks that hand-roll their own match are extended alongside, and `scope.rs`'s corpus instance stays `key: None`, so its child-agreement assertion never sees a keyed write.

## The abstract map type

A `Mut(Map(K, V), _)` binder holds a sum, so a concrete seed reaches it through a `box` introduction, not a subtyping edge ([type-inference.md](src/ccl/design/type-inference.md#only-a-term-builds-a-sum)); op-conversion compiles that box to the identity. `transact_phase` now reads a mutable variable's value type off that binder and `recognize_txn_group` off the history binding's codomain, in place of deriving it from the seed. A seed is one contribution to that type: a map seeded at two keys and written at a third holds more. Reading the binder is what lets `mut_var_value_ty` stop stripping refinements.

## Reading it back

A mutable variable holds its collection as one materialized map value, so `CheckedLookup` answers from that value's bindings rather than from a domain column, and gains a second operand shape for the `(collection, key)` rows a transaction's snapshot tuple produces. A store seeds itself from a collection init. Store keys become a tagged sum, one arm per mutable variable and per reply tap, replacing a bare `String`; every arm is `Unit` today, and the tag is what will keep a keyed variable's keys from aliasing another variable's name.

## Tests and docs

Nine pipeline tests cover both domains, a write through a `Mut` parameter, a read-back through `m[k]?`, and a computed key with a computed value; `type_check.rs` pins the key and value obligations separately. `keyed_writes_at_distinct_keys_compose_across_transactions` is what the whole-collection footprint buys, and has to survive [narrowing it](src/ccl/design/mutability.md#future-work). The type itself is stated in [mutability.md](src/ccl/design/mutability.md#a-mutable-collections-key-set-varies-with-the-position).

### DIFF
--- a/src/ccl/channelize.rs
+++ b/src/ccl/channelize.rs
@@ -1031,8 +1031,9 @@ fn drop_expr_stmts(expr: Expr) -> Expr {
             iter: Box::new(drop_expr_stmts(*iter)),
             body: Box::new(drop_expr_stmts(*body)),
         },
-        TypedExprNode::MutWrite { name, value } => TypedExprNode::MutWrite {
+        TypedExprNode::MutWrite { name, key, value } => TypedExprNode::MutWrite {
             name,
+            key: key.map(|k| Box::new(drop_expr_stmts(*k))),
             value: Box::new(drop_expr_stmts(*value)),
         },
         TypedExprNode::Begin { body } => TypedExprNode::Begin {
@@ -1152,7 +1153,12 @@ fn assert_no_defer_residue(expr: &Expr) -> Result<(), DeferError> {
             assert_no_defer_residue(iter)?;
             assert_no_defer_residue(body)
         }
-        TypedExprNode::MutWrite { value, .. } => assert_no_defer_residue(value),
+        TypedExprNode::MutWrite { key, value, .. } => {
+            if let Some(key) = key {
+                assert_no_defer_residue(key)?;
+            }
+            assert_no_defer_residue(value)
+        }
         TypedExprNode::Lit(_)
         | TypedExprNode::Var(_)
         | TypedExprNode::Builtin(_)
@@ -1882,7 +1888,12 @@ fn collect_feed_target_names(expr: &Expr) -> Vec<Name> {
                 rec(body, bound, out);
                 bound.pop();
             }
-            TypedExprNode::MutWrite { value, .. } => rec(value, bound, out),
+            TypedExprNode::MutWrite { key, value, .. } => {
+                if let Some(key) = key {
+                    rec(key, bound, out);
+                }
+                rec(value, bound, out);
+            }
             TypedExprNode::Lit(_)
             | TypedExprNode::Var(_)
             | TypedExprNode::Builtin(_)
--- a/src/ccl/context.rs
+++ b/src/ccl/context.rs
@@ -1560,6 +1560,12 @@ fn run_passes(
     // commit. Reject it before the phase strips the sites.
     check_transact_rejections(&expr, &txn_mut_vars)?;
 
+    // A register's seed enters the value type its binder declares — a concrete collection
+    // reaches an abstract one only through an introduction.
+    mut_elim::view_seeds_at_value_type(&mut expr);
+    // `m[k] := v` becomes the whole-value write it denotes, `m := insert(m, k, v)`, so
+    // every phase below sees one kind of `MutWrite` and none of them needs a key.
+    mut_elim::desugar_keyed_writes(&mut expr);
     expr = recorded(capture_provenance, Phase::Transact, || {
         transact_phase::run(expr, &txn_mut_vars)
     })
--- a/src/ccl/context.rs
+++ b/src/ccl/context.rs
@@ -1422,7 +1422,7 @@ fn run_passes(
         return Ok(expr);
     }
 
-    // Register every source (pre-registered + discovered during lowering) with
+    // Mutable variable every source (pre-registered + discovered during lowering) with
     // inference and operator-conversion now that the full source set is known.
     for (_name, source) in ctx.lowering_ctx().take_sources() {
         let name = source.borrow().get_id().to_string();
@@ -1566,7 +1566,7 @@ fn run_passes(
     // reachable.
     expr = recorded(capture_provenance, Phase::Transact, || {
         let mut expr = expr;
-        // A register's seed enters the value type its binder declares — a concrete
+        // A mutable variable's seed enters the value type its binder declares — a concrete
         // collection reaches an abstract one only through an introduction.
         mut_elim::view_seeds_at_value_type(&mut expr);
         // `m[k] := v` becomes the whole-value write it denotes, `m := insert(m, k, v)`, so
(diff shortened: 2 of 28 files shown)


dpmills:

- #10
    - #11
        - #45
            - #156
                - #157
                    - #158
                        - #159 ◀

sortalongo (approved):
Approving. `m[k] := v` denoting `m := insert(m, k, v)` is the right reduction — the history stays an ordinary overwrite, and every phase below `mut_elim` sees one kind of `MutWrite` and needs no notion of a key. Reading the value type off the binder rather than the seed is the correct fix for the third-key rejection, and the `store_key` change is the best catch in the branch.

One thing to add before merge, one stack-wide pass to run, and two things done well.

## Before merge: every keyed write in the tests is a constant

Collecting every keyed write across the branch's test programs:

| Write | Key | Value | Domain |
|---|---|---|---|
| `m["c"] := 3` (transactional store) | literal | literal | `Txn` |
| `m["c"] := 3` (round trip) | literal | literal | `Txn` |
| `m["c"] := 3` (type_check) | literal | literal | `Txn` |
| `m[x] := 99` (induction store) | loop binder | literal | induction |
| `m[x] := 99` (type_check) | loop binder | literal | induction |
| `m[7] := 3`, `m["c"] := "nope"` | literal | literal | rejection pins |

Three gaps follow, and they compound.

**No test writes a variable value.** Every right-hand side is a literal.

**A variable key is exercised only on the induction domain.** Every `Txn`-domain write uses the literal `"c"` — and the `Txn` path is where this branch's own fix lives. `transact_phase` and `recognize_txn_group` previously read the value type off the seed, "so a write at a third key was rejected"; both now read the binder. That fix is on the transactional path, and the transactional path is only ever tested with a constant key.

`a_keyed_write_reaches_an_induction_store`'s own comment says which case is load-bearing: "a variable key's type is reachable from no other slot, since a write is `Unit`, so a pass that skips the key slot leaves it an unresolved inference variable and nothing downstream notices." That is exactly right, and it is why the case wants covering on both domains rather than one.

**Even the variable key has a statically known extent.** `m[x] := 99` binds `x` over the literal `[3, 4]`. A key off a request stream — the shape the storefront writes, `inventory[req.body.sku] := stock - qty` — has no such extent.

So the shape a real program writes is a variable key with a computed value on the transactional domain, and no test combines even two of those three.

Worth one test rather than six: a pipeline case on `Txn` with a variable key and a computed value — `m[k] := n + 1` with `k` off the loop and `n` another mutable variable — which is both the demo's shape and the one the comment above calls dangerous.

There is a second reason to want it. Because every seed, key, and value here is a literal, a constant-folding pass can settle these writes at compile time and the runtime path stops being exercised while the tests stay green. That pass is already scoped — #157 says the constructor lowering's timing "rides on constant folding" — so these are the tests most likely to be hollowed out by planned work, and a variable-keyed case is what survives it.

## Please run one vocabulary pass over the whole stack

"register" is back. Occurrences in the docs:

| | main | this head |
|---|---|---|
| `mutability.md` | 0 | 3 |
| `collections.md` | 0 | 3 |
| `chl-spec.md` | 5 | 7 |

Plus 52 in this branch's added lines and two test names (`a_map_valued_register_reads_inside_a_block`, `a_checked_lookup_on_a_register_answers_none_for_an_absent_key`). There is no `Register` type, field, or variant anywhere, so this is prose, comments, and those two names — not a refactor. The name the code already gives it is **mutable variable**, which `MutDecl` and `Mut(𝑉, 𝐷)` carry and which this very doc uses in the same sentence ("like any other mutable variable").

That makes five terms across four PRs — `Any`, `width-to-top`, `realization`, `arrow`, `register` — and the per-branch cleanups have been adding more than they remove: this branch also carries four more "keyed collection" and four more "realiz*" in its added lines, and #157's own de-coining commit took its branch from 3 to 7. So please run this once across the stack against a named target list rather than per branch.

## Two things done well

**The `store_key` fix.** A store key was `Value::String(field_key)`, so a keyed collection's *data* key could collide with a mutable variable's *name* — "the string `"pool"` would alias the scalar register spelled `pool`". That is silent cross-variable corruption, and a tagged sum over variables and reply taps is the right shape for it.

**The key's blast radius is bounded, and the description undersells it.** Listing seven folds that gained key walks reads like seven manual additions waiting to rot. The actual containment is structural: `key` becomes an ordinary child in `expr.rs`'s `walk_children` / `walk_children_mut`, so `check_fully_typed` reaches it through `collect_expr_errors` and a pass that forgets it fails loudly rather than silently. Combined with the key existing only between parse and `mut_elim`, that is a bounded window with a generic guard — worth describing that way, since it is the reason the seven walks are safe.

Also clean on `#[ignore]`: this branch adds none.

## Conflict granularity

Noted and not a concern for now — one store key per collection means two transactions touching one map conflict whatever keys they touch, and the Future work entry scopes the narrowing with its mechanism already identified (`` (.i, key, value) ▷ zip ≫ insert ``). Fine to defer; it is a scalability question rather than a correctness one.

## Scope

I read `mut_elim`, the store-key change, the `expr.rs` node changes, the tests, and the design sections. I did not review `transact_phase`'s `MutVarDecl` value-type change or `plan_loops::recognize_txn_group` beyond what the description says about them.

dpmills:
`a_keyed_write_commits_a_computed_value_at_a_computed_key`: `Txn` domain, key off the loop, value reading another mutable variable written in the same transaction.

One correction to the table — `keyed_writes_at_distinct_keys_compose_across_transactions` already wrote a variable key on `Txn`, so the real gap was narrower than three: no computed *value* on either domain. The new case closes that, and it is the one that survives constant folding.

`register` → mutable variable across 53 sites and the two test names. Three senses, not one: the verb is untouched (registering a source, a consumer, a route), and `KIND_VARS` was called a register too, so that one names what it reads. Chasing the last site turned up `kind_var_owners` having stolen `dump_kind_vars`'s doc block.

PR #161, opened 2026-09-05
Delete the per-operation `deep-typecheck` probe and enforce the closure invariant unconditionally

dpmills:
The `deep-typecheck` feature and the per-operation `debug_typecheck` probe it gated are deleted, along with its call sites in `lambda_elim` and `simplify` and the `Derivation::SubTree` excuse the probe alone required. The probe types a sub-tree mid-rewrite, where a node's kind can disagree with a stamp a later step settles, so it reports a pass's transient states as failures. CI coverage is unchanged — the workflow stopped setting `DEEP_TYPECHECK=1` in `769e24621`, before this change — so what goes is a local opt-in bisect aid that was already reporting those false failures. `settle`'s per-boundary walls, `check_pre_channelize` and `typecheck`, remain the always-on correctness net.

## The closure invariant loses its exemption

`observe_bound_scope` becomes `enforce_bound_scope`: no derivation is excused, so a recorded bound whose free reference escapes the holder's telescope is an internal error in debug builds. The `CAMBRA_TELESCOPE_LOG`/`CAMBRA_TELESCOPE_BT` observation log goes with it, since it existed to tell excused gaps from real ones. [`type-inference.md`](src/ccl/design/type-inference.md#the-invariant) drops the excuse in both places it stated it.

`Derivation` is now two variants controlling one behavior: `opens_unconditionally`, read at the `Fun`/`Fun` codomain edge and nowhere else.

## An invariant left unguarded

`realize_and_unbox`'s `in_predicate` early return rests on the per-leg discharge replacing the whole predicate, so nothing in the copy it skips reaches compilation. Nothing checks that: the pass-boundary typecheck does not descend into predicates, and the probe was what made a violation visible. Pre-existing, surfaced rather than created here.

## Tests

`a_rejection_holds_at_every_derivation` swept a variant that no longer exists; its assertions hold in `kind_compute_where_data_is_rejected` and `conditional_is_subtype_by_candidate_subset`, and its `Map`/`Collection` pair moves into `candidates_lie_below_a_kind_that_admits_them`. The concrete-pair opening test now asserts both answers of `opens_unconditionally` from one pair of types.


dpmills:

- #161 ◀

.github/workflows/ci.yml
```diff
@@ -198,14 +198,6 @@ jobs:
         # Read more about the "hidden ifs" and general inanity here
         # https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions
         if: steps.filter.outputs.code == 'true' && (success() || failure())
-        # The opt-in per-operation typecheck (the `deep-typecheck` feature) is **not**
```
groundlar, line 201: **The commit message's CI claim contradicts the comment it deletes.** The body says the probe "ran only in CI" and that "the workflow's test step drops `DEEP_TYPECHECK=1`". At `HEAD~1` this step ran `CAMBRA_PROVENANCE_GATE=1 ./ci.sh test` with no `DEEP_TYPECHECK`, and these very lines say so: "is **not** enabled here, pending its removal". The ci.yml half of this diff is a comment deletion and nothing else.

On a squash-merge repo the PR body becomes the commit message, so this matters past pedantry. The true story is a strictly stronger argument for the change: *the probe has been dead in every configuration since it started reporting a pass's transient mid-rewrite states as failures; this deletes dead code, and the coverage delta is exactly zero.* That is also what someone bisecting a miscompile later needs, so they do not go looking for coverage they think this removed.

Related: `ci.sh`'s deleted comment ("The GitHub workflow sets it so automated runs keep exercising that check") was itself already stale. Worth saying it was stale rather than implying it was live.

src/ccl/infer/solver/constrain.rs
```diff
@@ -3552,11 +3461,7 @@ mod tests {
     /// cache and used one.
     #[test]
     fn a_rejection_holds_at_every_derivation() {
-        for derivation in [
-            Derivation::LiveSolve,
-            Derivation::PostPass,
-            Derivation::SubTree,
-        ] {
+        for derivation in [Derivation::LiveSolve, Derivation::PostPass] {
```
groundlar: **This loop is now a no-op, not a narrowed one.** Dropping `SubTree` from the array is forced by the enum shrinking, but what the test became is worth a decision.

Evidence — patch `Derivation::opens_unconditionally` to `panic!("reached")` and run this test alone:

```
test ccl::infer::solver::constrain::tests::a_rejection_holds_at_every_derivation ... ok
```

It passes, so `cache.derivation` is never read on any of the four paths. The loop runs identical code twice.

That was not true before. `SubTree` weakened the `WitnessRef` arm, so iterating it was a real guard that the abstention did not leak into a rejection. With two variants differing in exactly one predicate — `opens_unconditionally`, read at one site (the named-Pi codomain guard around line 1080), which none of these types reach, since they are unnamed `Fun`s and Σs whose codomains do not reference the enclosing function — the doc's "One rule, every derivation" overclaims.

Two honest resolutions:

1. Delete the test. The risk it guarded left with `SubTree`.
2. Retarget it at the edge that actually differs: a named Pi whose codomain references the enclosing function, constrained against a concrete counterpart with no `Infer` on either side — the one shape where `LiveSolve` declines to open and `PostPass` opens.

I'd take (2). That guard is now the entire behavioral content of `Derivation`, and nothing tests it.

src/ccl/infer/api.rs
```diff
@@ -1656,21 +1639,18 @@ fn check_pre_channelize_as(expr: &Expr, derivation: Derivation) -> Result<(), Ve
...
-/// can hand such an alias to [`debug_typecheck`] as a standalone subtree;
-/// keying only on defer nodes would wrongly check it strictly and reject the
-/// legitimate channel type. `Mut` is analogous: a mutable reference carries a
-/// `Mut` type whose `Infer` domain the pre-channelize relaxation must tolerate
-/// until the unified phase resolves it.
+/// The type check matters because a defer-read alias (`Var(x) : feed(_)`)
+/// carries a `Feed` type with no defer node; keying only on defer nodes would
+/// check such a tree strictly and reject the legitimate channel type. `Mut` is
+/// analogous: a mutable reference carries a `Mut` type whose `Infer` domain the
+/// pre-channelize relaxation must tolerate until the unified phase resolves it.
```
groundlar: **This rationale names a scenario no remaining caller produces.** The premise of "a defer-read alias (`Var(x) : feed(_)`) carries a `Feed` type with no defer node" was the probe — `substitute` handing a standalone alias subtree. Every remaining caller passes a whole tree, where a `Feed`-typed `Var` implies the `Defer`/`Feed`/`Define` node is in the same tree and the node-kind clause fires first.

Evidence — split the predicate into its three clauses and panic when the node-type clause decides alone:

```rust
let full = has_pre_channelize_artifacts_go(expr);
if full && !node_kind_only(expr) && !binder_only(expr) {
    panic!("node-type clause decided alone");
}
full
```

Full suite, zero hits.

The paragraph below this one *is* load-bearing — `let cnt: Mut(Int, _) = 0 in (cnt := 1)` carries no `Defer`/`Feed`/`Define` node at all — and is what should carry the justification. Keep the clause (it is cheap, and the walk happens regardless); either attribute it to the `Mut` case or name a whole-tree witness for the `Feed` one. A rationale nothing witnesses is the rot CLAUDE.md's "every claim is checkable" rule targets.

src/ccl/infer/solver/constrain.rs
```diff
@@ -1269,15 +1237,7 @@ fn constrain_go_impl(
...
-            //
-            // A **sub-tree** probe is the one abstention: a binder its refinements reference
-            // may be held by the context the cut removed, so Γ classifies neither reference
-            // and the probe has no verdict to give ([`Derivation::sees_every_binder`]).
-            let same = a == b
-                || (!cache.derivation.sees_every_binder()
-                    && (cache.lctx.type_kind_of(a).is_none()
-                        || cache.rctx.type_kind_of(b).is_none()));
-            if same {
+            if a == b {
```
groundlar: The reduction is exact — `&&` binds tighter than `||`, so the guarded disjunct was statically false on both surviving derivations and `same` already meant `a == b`. Two things the comment should now say:

- **This arm no longer consults Γ at all.** `lctx`/`rctx` survive only for the concrete-meets-witness arm below (~line 1264). An unclassified `WitnessRef` compared against itself still passes here silently, as it did before — nothing in this arm is the free-witness check. That guard is `enforce_bound_scope` at the `Infer` arms plus `debug_assert_no_free_witness`. A reader arriving at an arm whose comment mentions Γ twice will assume otherwise.
- **"at every derivation" is vestigial.** It was written to contrast with the one derivation that answered otherwise; with the abstention gone it reads as if some derivation might still disagree.

src/ccl/infer/solver/constrain.rs
```diff
@@ -218,41 +218,9 @@ pub enum Derivation {
     /// passes spelled in different coordinates, but every binder the tree's
     /// refinements reference is a binder the walk itself can enter.
     PostPass,
```
groundlar, line 228: Nit, and a flag rather than a request: `Derivation` is now a two-variant enum whose entire behavioral content is one bool read at one call site. Keep it — it names the concept, which a bare bool would not — but per CLAUDE.md's "recurring shapes" guidance this is the kind of thing worth noticing out loud in the PR body, so the next person does not have to rediscover it.

src/ccl/infer/check.rs
```diff
@@ -259,10 +255,11 @@ impl Typing for CheckCtx {
...
         //
-        // The cache serves this walk's derivation: a whole tree enforces the
-        // closure invariant like the live solve, a sub-tree probe cannot (the
-        // binders its refinements reference are held by the context it was cut from).
-        let mut cache = ConstrainCache::for_derivation(self.derivation);
+        // Check re-derives at a pass boundary, reconciling types two passes
+        // spelled in different forms, so its cache takes `Derivation::PostPass`.
+        // That is the one respect in which Check's constraint solving differs
+        // from the live solve's (see [`Derivation`]).
+        let mut cache = ConstrainCache::for_derivation(Derivation::PostPass);
```
groundlar: Nit: `Derivation::PostPass` is now hardcoded here and again at ~line 400, each followed by the same `seed_context(&self.witness_ctx)` call. "Check is always `PostPass`" is one fact stated twice. A `CheckCtx::cache(&self)` helper returning the seeded cache holds it once and collapses the duplicated seeding.

src/ccl/infer/check.rs
```diff
@@ -794,21 +791,20 @@ fn check_predicates(
...
-/// is compiled out of release builds; the remaining callers (`typecheck`,
-/// post-planning validation in `context.rs`) run once per pipeline stage.
+/// Cost note: the full-tree clone makes each call O(tree). Every caller
+/// (`typecheck`, `check_pre_channelize`, post-planning validation in
+/// `context.rs`) runs once per pipeline stage.
 ///
-/// **TODO(scratch-copy): ripe for refactoring — this is quadratic.** One tree
-/// copy per fired rule is O(tree x rules) over a debug compile, for a value that
-/// is read and dropped. The clone exists only because the shared per-node rules
+/// **TODO(scratch-copy): ripe for refactoring.** The clone is a whole tree per
```
groundlar: Nit: correctly downgraded from "quadratic", but the TODO still reads as urgent for what is now O(passes × tree) beside passes that are themselves O(tree). Either state the new bound or drop the TODO — with the hot caller gone, the clone is the same order as the work it validates.

src/ccl/planning/conditionals.rs
```diff
@@ -159,9 +159,8 @@ fn realize_and_unbox(
     // predicate with one reading that leg's arm ([`read_the_arm_instead`]), so nothing
     // inside this copy survives to be compiled. Descending would only *break* it, and does:
     // erasing the `box` off each arm leaves a `Case` joining collections over distinct
-    // domains, which is exactly the type error `box` exists to prevent. Nothing on the
-    // normal path notices — the pass-boundary typecheck does not descend into predicates —
-    // so it shows up only under the `deep-typecheck` feature.
+    // domains, which is exactly the type error `box` exists to prevent. Nothing reports it:
+    // the pass-boundary typecheck does not descend into predicates.
```
groundlar, line 163: This conversion is more accurate than what it replaced — the probe never ran anywhere, so it showed up nowhere. It also now documents a fully unguarded invariant: `realize_and_unbox` must not descend into a predicate's conditional, and nothing checks it. Pre-existing and surfaced rather than created by this PR, but worth a line in the body so it is not lost.

src/ccl/design/type-inference.md
```diff
@@ -1353,19 +1353,15 @@ of the bound's type is in the telescope or in the edge's substitution domain. Th
 the bound is recorded, and it is a lookup, since uniquify gives every binding site one uid. A
 violation names the variable and the reference and fails.
 
-Enforcement covers every derivation over a self-contained tree: the live solve, meaning emission and
-its specialization pins, and the pass-boundary re-derivations that check what a pass produced. A
+Enforcement covers every derivation: the live solve, meaning emission and its specialization pins,
+and the pass-boundary re-derivations that check what a pass produced. A
```
groundlar: Nit: the edit shortened the sentence without reflowing the paragraph — line 1357 lands at 71 columns mid-paragraph while its neighbours are at 98–99.

groundlar (approved):
LGTM after comments

The deletion is correct and complete: zero leftover references to the feature or to `Derivation::SubTree`, and zero behavior change for either surviving derivation. `sees_every_binder()` was `self != SubTree`, so on `LiveSolve`/`PostPass` the guarded disjunct in the `WitnessRef` arm was statically false and the condition already reduced to `a == b`.

Verified locally at `b258e30`:

| Check | Result |
|---|---|
| `cargo clippy -p cambra --release --all-targets -- -D warnings` | clean |
| `cargo test -p cambra -q --no-fail-fast` | green |
| grep for `deep-typecheck`/`debug_typecheck`/`dbg_typecheck_mv`/`DEEP_TYPECHECK`/`CAMBRA_TELESCOPE*`/`SubTree`/`sees_every_binder`/`enforces_closure`/`observe_bound_scope`/`check_pre_channelize_sub_tree` | zero hits |
| doc anchors `#the-invariant`, `#where-the-conversions-run` | both resolve |

Not run: full `./ci.sh` (formal/Lean, doc, doc_refs, fixtures).

Three comments below want a response before merge — the commit message's CI claim, `a_rejection_holds_at_every_derivation` having become a no-op loop, and the `has_pre_channelize_artifacts` rationale. The rest are nits.

.github/workflows/ci.yml
dpmills: Right, and the correction is the stronger claim. `769e24621` removed `DEEP_TYPECHECK=1` from the workflow, so the probe ran in no configuration — the body was claiming coverage that was already gone. Title and body rewritten: the delta is zero, and what goes is a local opt-in bisect aid that was already reporting transient mid-rewrite states as failures. `ci.sh`'s deleted comment was stale for the same reason, not live.

src/ccl/design/type-inference.md
dpmills: Reflowed.

src/ccl/infer/check.rs
dpmills: Done — `CheckCtx::cache()` returns the seeded `PostPass` cache and both sites call it. The fact is stated once, on the helper.

src/ccl/infer/check.rs
dpmills: TODO dropped. The note now states the bound — O(tree), the order of the walk it feeds — and keeps what would remove the clone rather than what would make it cheaper.

src/ccl/infer/api.rs
dpmills: Re-ran your experiment — three clauses split, panic when the node-type one decides alone — and got the same zero hits across the suite. Rationale moved onto the `Mut` binder-slot case, which is the clause that does decide alone. The node-type clause now states why it never does over a whole tree: a defer-read alias carries a `Feed` type with no defer node of its own, but the `Feed`/`Define` node that introduced the channel it reads is in the same tree, where the node-kind clause answers first. It stays because it names what the relaxation is keyed on — the transient type the check tolerates, not the node that minted it.

src/ccl/infer/solver/constrain.rs
dpmills: Both taken. The comment now says neither Γ is read here — `lctx`/`rctx` serve the concrete-meets-witness arm below, where a reference has to be classified before it can be constrained — that a reference neither Γ classifies compares equal to itself and passes, and that the free-reference guard is `enforce_bound_scope` at the `Infer` arms plus `debug_assert_no_free_witness`. "At every derivation" is gone.

src/ccl/infer/solver/constrain.rs
dpmills: Flagged in the body: two variants controlling one behavior, `opens_unconditionally`, read at the `Fun`/`Fun` codomain edge and nowhere else. Related — the comment at that call site still claimed `Derivation` answered "both this and the record sites' question", which stopped being true when `enforce_bound_scope` lost its parameter. Corrected in the same commit.

src/ccl/planning/conditionals.rs
dpmills: Noted in the body under its own heading, as pre-existing and surfaced rather than created here.

src/ccl/infer/solver/constrain.rs
dpmills: Took (2), and the fixture already existed: `a_concrete_pair_does_not_open_at_a_shared_spelling` builds exactly that shape — `(x: Int) ⇒ {Int | __elem == #0}` against `Int ⇒ {Int | __elem == x}` — and asserted only the `LiveSolve` half. Measured `PostPass` on the same pair: `Ok(())`. So it is now `the_derivation_decides_whether_a_concrete_pair_opens`, pinning both answers from one pair of types.

The loop is deleted rather than narrowed, and I checked what that costs before deleting. Two of its four assertions already exist as `kind_compute_where_data_is_rejected` and `conditional_is_subtype_by_candidate_subset`; the `Map`/`Collection` pair was the only unique coverage, and it moved into `candidates_lie_below_a_kind_that_admits_them`, where kind containment is already stated — a key bound joins the `⊤` loop and the narrower-than-`⊤` rejections.

PR #162, opened 2026-09-08
Give every operator a `NodeId` and ship the dataflow graph as the seventh inspector pane

groundlar:
The inspector's panes stopped at `post-planning` because operator conversion had nothing to record against: a `TileOperator` carried no identity, so a pane past the last expression tree had no id to resolve. Every operator now carries an `OperatorBase` holding a `NodeId` drawn from the same counter as an expression node's, conversion becomes an ordinary recorded phase (`Phase::Convert`), and the graph it builds becomes the `post-conversion` pane — on the wire, rendered in the web view, and joined to `post-planning` by a gated pane pair. Every operator therefore resolves to a post-planning node and transitively to a source span.

## Reading order

Thirteen commits, and two of them revise shapes the earlier ones introduce. Read the commits in order for the mechanism, but take the wire shape from the net diff: `EdgeKind::Feedback`, `roots` and `unowned` appear mid-stack and none of them survives.

Load-bearing, in this order:

- **`an OperatorBase for every operator`** — identity plus output tiling in one field, collapsing a `tiling` field and its trivial accessor on 35 operators. Mostly mechanical; the substance is `OperatorBase`, `impl_operator_base!` and `short_type_name` in `src/interpreter/tile_operators/mod.rs`.
- **`conversion records what it builds, and the operator graph is a pane`** — the core. `provenance::converting` opens one recording per conversion frame, `OperatorBase::new` rows the operator and the inputs it holds, and `src/interpreter/operator_graph.rs` accumulates them into a thread-local session `compile_program` installs. Recorded rather than walked because `subscribe` empties every `CycleSlot` before a `CompiledProgram` exists. `PaneSpec` gains `content: PaneKind`, and the fold reads `pane_ids` rather than trees, which is what lets a pane hold something that is not a tree.
- **`the operator graph has both program boundaries`** — a `Sink` per compiled field, and one `Source` per registered source rather than per read site. A source node is minted after the walk, since its row names every site that read it, and the reader edge is `Share` because a shared target has no single owner.
- **`the cycle set is the late-wired edges, and a label is its own type`** — retires `EdgeKind::Feedback`, which was stamped on every branch of a cyclic fan rather than on the edges that close a cycle. A cycle is now a `Value` edge with `deferred` set, which over the corpus is a minimum feedback arc set. `OperatorBase<T>` also gains a `PhantomData<fn() -> T>`, so a label that disagrees with its holder is a type error.
- **`derive the operator pane's walk starts`** — the pane shipped `unowned`; a node's owner is the one `value` edge naming it, so the table already answers it and the field goes, with its four consumers deriving it. That strengthens the closure check, which now fails on a node reachable only along a `share` edge and on a cycle among the `value` edges.

Then the two presentation commits: `the operator pane on the wire` (`PaneNodes` untagged with `kind` as the discriminant, `root` optional, both validators dispatching on pane kind, `ALLOWED_VIA` gains `Convert`) and `render the operator pane` (`operatorView.ts` — value edges are the child relation, share edges are leaf reference rows, so a shared subtree is drawn once and every cycle is a reference).

Mechanical: the rename commit, the two docs commits, and the fixtures README. The docs commit also carries a `cargo clippy --all-targets` fix that was red from the recording commit upward.

## Tests and fixtures

`source_shared` is a new gallery program and the only fixture with a `Source` node. `a_source_read_twice_is_one_node_attributed_to_both_reads` pins the span count, which is the only assertion that catches minting a source at its first read. `assert_store_edge_shapes` covers `deferred` and store-keyed edges structurally over a fresh dump of `txn_multi_read` and `for_accumulator` — the only corpus programs with a store, and whose bytes would re-bless on every change to how a store is built. `every_operator_graph_cycle_runs_through_a_deferred_edge` runs Kahn over the corpus twice, once whole and once with the deferred edges removed.

The re-bless of the four existing fixtures is purely additive: a `post-conversion` pane and a sixth `paneLinks` entry, zero removed lines, no id renumbering. `SCHEMA_VERSION` does not move, and `helpers.ts` no longer tells a re-blesser that it should.

## Docs

[provenance.md](src/ccl/design/provenance.md#panes-past-post-planning) replaces the prerequisites section with the four properties that carry a pane past the last expression tree; [inspector_model/design.md](src/inspector_model/design.md#an-operator-node-on-the-wire) specifies the operator node and its input edges.


src/ccl/design/provenance.md
```diff
@@ -916,24 +916,32 @@ bit is what says so until every phase in the pair records.
   catch a defect.
 
   Where it stands: **every pair is gated**, from `pre-inference → post-inference`
-  down to `post-lambda-elim → post-planning`. Capture is total across the whole
+  down to `post-planning → post-conversion`. Capture is total across the whole
   pipeline.
 
-## Known prerequisites for panes past `post-planning`
+## Panes past `post-planning`
```
dpmills: This title is obscure; is it just "Operator Conversion"?

src/ccl/panes.rs
```diff
@@ -87,16 +90,54 @@ impl CompiledProgram {
...
+            })
+            .collect()
+    }
+}
+
+/// What a pane holds.
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub(crate) enum PaneKind {
+    /// An expression tree.
+    Ir,
```
dpmills, line 119: Why IR instead of CCL or AST? Don't think we need a new term.  This has also spread into other constants/identifiers

---

Per-site, from `main` rather than this head: `IrNode`/`IrChild` are on `main` in the wire layer, `pre_inference_ir`/`post_*_ir` are on `main` as `CompiledProgram` fields, and `src/ccl/design/ir.md` exists on `main`. So the term is already established in two places; what this PR adds is `PaneKind::Ir` and `IR_PANE_COUNT` here in `panes.rs` — where the neighbours are `Expr`, `pane_trees` and `collect_tree_ids`, so the local name is `Tree`/`Expr` — plus `IrPane`/`isIrPane`/`irPaneById` on the frontend.

The spread is wider than the name. The IR-vs-operators distinction now has five representations: `PaneKind` here, `PaneContent<'a>` in `inspector_model/program.rs`, `PaneNodes` in `wire.rs`, the wire string in `pane_kind()`, and `IR_PANE_COUNT`. Each is individually justified (a const tag, a borrow, a table, a wire discriminant, an arity), so this is a note rather than an ask — but a third pane kind touches all five with nothing enforcing agreement.

src/interpreter/tile_operators/mod.rs
```diff
@@ -125,6 +146,101 @@ pub trait TileOperator {
...
+/// label names. Declaring it in the field type (`base: OperatorBase<Filter>`)
+/// makes a label that disagrees with its holder a type error rather than a call
+/// site's convention.
+pub(crate) struct OperatorBase<T: ?Sized> {
+    /// Identity, minted at construction. See the type's own docs for why here.
+    pub(crate) id: NodeId,
+    /// Output tiling for this operator.
+    pub(crate) tiling: Tiling,
+    /// `fn() -> T` rather than `T`: covariant, and neutral for auto traits.
+    _holder: std::marker::PhantomData<fn() -> T>,
```
dpmills: Do we actually need this? If so, the comment should say why instead of listing mechanical details which are already clear from the definition

src/interpreter/operator_graph.rs
```diff
@@ -0,0 +1,610 @@
...
+//! the operators rather than choices:
+//!
+//! * **An edge's kind is not recoverable from the operator.** [`FanOut::branch`]
+//!   returns a `Box<dyn TileOperator>` and `FanOutBranch` is private, so a shared
+//!   input's field type is identical to an owned one. Whether a fan is cyclic
+//!   lives in `FanOutShared`, which a branch exposes no accessor for.
+//! * **A `CycleSlot`'s only accessor consumes it.** `subscribe` calls `take`, and
+//!   `compile_program` subscribes before it returns, so a walk of a finished
+//!   `CompiledProgram` has already lost the commit store's writers and the
+//!   induction store's body.
```
dpmills: Neither of these seems like a good reason to create an entirely new structure that will need parallel maintenance over time.  If we need to make Operators store slightly more metadata, we can do that.  That will also work better with dynamic operator graphs once we have update/branching.

Once we have sorted this out, we should also have at least a small design section in interpreter/design-operators.md about how this works; currently none of the changes to operators in this PR are documented in design docs.

---

Two costs the construction-time capture has already imposed, as evidence:

- `CycleSlot::setter` grew `owner: Option<NodeId>`, `role: EdgeRole` and a `T: TileOperator` bound purely to feed the recorder. Both call sites are in `commit_operator.rs` (`:826`, `:1275`) and both already hold `self.base.id`, so a wrapper closure at each would have done the same job without touching the primitive. The doc's stated reason — "the owner is already gone by the time the setter runs" — argues for capturing at `setter()` time, which a wrapper also does, so it does not distinguish the two designs.
- The session is installed unconditionally on every compile. `record_operator` calls `tiling.to_string()` per operator, `push_edge` and `drop_operator` are linear scans over `nodes`, and `drop_operator` runs once per aligned `Var` use — so node removal is O(nodes x aligned uses) on a path with no inspector attached.

src/interpreter/tile_operators/aggregate.rs
```diff
@@ -61,7 +64,10 @@ impl TileOperator for Aggregate {
         let input_producer =
             self.input
                 .subscribe(self.input.tiling().universal_guard(), consumer, scheduler);
-        Box::new(AggregateProducer::new(self.tiling.clone(), input_producer))
+        Box::new(AggregateProducer::new(
+            self.base.tiling.clone(),
```
dpmills: Having to go through base to get the tiling is awkward.  It looks like the machinery to make self.tiling() already is built; just use that

---

Verified and it is total: I rewrote all 34 `self.base.tiling` sites outside the macro body to `self.tiling()`, and `cargo build` plus `cargo clippy --all-targets -- -D warnings` are both clean. No site needs the disjoint-field borrow that `&self.base.tiling` gets and `self.tiling()` does not.

src/interpreter/tile_operators/aggregate.rs
```diff
@@ -161,7 +167,8 @@ impl TileProducer for AggregateProducer {
 
 pub struct ExtractAggregate {
     input: Box<dyn TileOperator>,
-    tiling: Tiling,
+    /// Identity and the output tiling.
```
dpmills: We don't need 50+ identical "Identity and the output tiling." comments in every operator.  Delete the zero-info ones and revert the changes to the ones that are listing the tilings

src/interpreter/operator_graph.rs
```diff
@@ -0,0 +1,610 @@
...
+//!
+//! So the recording happens at construction, through [`OperatorBase::new`], and
+//! accumulates into a thread-local [`GraphSession`] that `compile_program`
+//! installs around conversion. Outside a session every record is a no-op, which
+//! is what keeps the operator constructions in engine tests from accumulating.
+//!
+//! What the edges mean: they are **construction** edges — which operator holds
+//! which, and how. Runtime dataflow follows `get` and `notify`, which is a
+//! different relation, and nothing here asserts the two coincide.
```
dpmills: This seems like a big downside to this approach

---

There is a second site, and it says the opposite: `wire.rs` documents `OperatorEdge` as "A subscription: the consumer holds this operator and calls `get` on it", which is exactly the reading this disclaimer denies. The field name (`subscribed`), `walk_starts`' doc ("the subscription forest") and `operatorView.ts` all follow the wire; the defining module doc stands alone. So the codebase already disagrees with the disclaimer in three places.

src/ccl/design/provenance.md
```diff
@@ -17,8 +17,8 @@ read.
 
 Every phase that rewrites expression nodes records, and every recording reaches
 a table: `compile_program` opens a `PhaseScope` around each. Operator conversion
-has no identity to record against, which is what stops the panes there; see
-[Known prerequisites for panes past `post-planning`](#known-prerequisites-for-panes-past-post-planning).
+records too, against the `NodeId` each operator carries; see
```
dpmills, line 20: Anchoring here because the lines are outside the diff: the Terms table at line 65 still defines a pane as "A retained AST snapshot the inspector displays, materialized after a set of phases. Each one costs a retained full-tree clone." Both halves are false for `post-conversion`, and this is the file that owns the term. `panes.rs`'s module doc and `inspector_model/design.md` were updated for exactly this; the definition at line 47 and the intro at line 8 say the same thing and were not.

The Terms table is also where the new vocabulary should land — `walk start` and `boundary node` are used across five files and defined nowhere.

src/web_inspector.rs
```diff
@@ -2,11 +2,11 @@
...
 //! [`inspector_server`](crate::inspector_server)) is the intended replacement,
-//! but cannot fully supplant this yet: it models only the static IR passes (one
-//! per declared pane) and has no operator/dataflow graph pane and no live
-//! per-tick view. Once the operator graph is instrumented (the
-//! `NodeId → OperatorId` provenance edge) and a live/tick layer lands, this
-//! module — and the `InspectNode::to_json` wire it depends on — can be
+//! and one thing still stands in the way: it has no live per-tick view. The
+//! operator graph it also lacked is now its terminal pane (`post-conversion`),
+//! reached by operators carrying a `NodeId` recorded at conversion, so this
+//! dashboard's remaining sole claim is the run. Once a live/tick layer lands,
```
dpmills, line 8: This makes `docs/design.md:206` stale, and the PR does not touch it. That line still reads: "the web inspector (`--inspect`) serves the CHL AST, the lowered CCL, the operator graph, and live per-producer runtime state for any running program; the program inspector (`--inspect-only`, …) serves one IR pane per compiler stage for a program it compiles and does not run." The operator graph is attributed to the web inspector there, which is the distinction this comment says is gone. CLAUDE.md asks for `docs/design.md` in the same change.

src/inspector_model/design.md
```diff
@@ -15,7 +15,7 @@ compile reads the lowering projection alone.
 | | |
 |---|---|
 | Input | a `CompiledProgram`: the pane trees, the provenance table, the lowering projection, the parsed surface AST, the source text |
```
dpmills: The Output row below was updated for the operator pane; this Input row still reads "the pane trees" and omits `operator_graph`, which is a new input to the same function.

cambra-inspector/web/src/__fixtures__/README.md
```diff
@@ -4,34 +4,57 @@ Real `/api/snapshot` payloads emitted by the `cambra` backend, used by
...
+order, and names no walk start. A walk begins at the nodes no `value` edge
+subscribes — a sink per compiled output, a fan input per share point, and a
+source per registered data source — which a consumer derives from `inputs`.
+Every node is reachable from there following `value` edges alone, which is the
+relation a consumer walks.
+
+A node carries `role` (`operator`, `source`, `sink`), a `tiling` for an operator
+and none for a boundary, and `inputs`: `{ role, kind, deferred, subscribed }`. An
+edge's `kind` is `value` for an exclusively owned input, `share` for one several
+consumers may reach, `feedback` for a share that closes a cycle. These are
```
dpmills: This documents an edge kind the PR deletes: "`feedback` for a share that closes a cycle". `ALLOWED_EDGE_KIND` is `["value", "share"]` in both validators.

The same sentence is also the place `deferred` — which the field list on the line above names — should be explained, and isn't.

src/interpreter/operator_conversion.rs
```diff
@@ -105,6 +113,94 @@ pub type RecordFieldOperators = Vec<(String, Box<dyn TileOperator>)>;
 /// branches off the same memoised upstream operators rather than each sink
 /// re-compiling the shared prefix into a fresh, independent subgraph.
 ///
+/// What a `Let` binding compiles to, for the `Var` uses that read it.
+///
+/// Both variants carry the input the body is converted with, because whether the
+/// enclosing input is fanned depends on which variant this is: a transaction
+/// store consumes no input and hands the body the original.
+enum LetBinding {
```
dpmills, line 112: `enum LetBinding` was inserted inside `convert_record_fields_to_operators`'s doc comment, so the doc lands on the wrong item. `LetBinding` now carries that function's summary line and its `FanOut`/`Memo` sharing rationale; the function is left with only its trailing "The expression must consist of zero or more `Let` bindings…" paragraph and no summary.

Same defect at `wire.rs:1308` and `wireValidate.ts:192`.

src/inspector_model/wire.rs
```diff
@@ -958,21 +1298,62 @@ mod tests {
...
     /// from that one attribution, so a repeat would mean the attribution itself
     /// carries a duplicate.
+    /// **An operator node's own fields**: `role` says which of the three node
+    /// kinds it is, `tiling` is present for an operator and absent for a
+    /// boundary, and `spans` is narrowest-first like an expression node's.
+    ///
+    /// The shape checks elsewhere range over ids and attributions, which both
+    /// node kinds share. These are the fields only an operator node has.
+    #[test]
+    fn an_operator_node_carries_its_role_tiling_and_ordered_spans() {
```
dpmills, line 1315: This test was inserted inside `no_node_repeats_a_span`'s doc comment. It now carries "No node claims one span twice…", the two comments are concatenated with no blank line between them, and `no_node_repeats_a_span` at line 1345 has no doc at all.

Same defect at `operator_conversion.rs:121` and `wireValidate.ts:192`.

cambra-inspector/web/src/wireValidate.ts
```diff
@@ -138,32 +171,81 @@ function validateNode(v: unknown, path: string): IrNode {
...
+
+// One entry of the operator pane's node table: an operator, or one of the two
+// program boundaries. It shares an id, a span table and a rewrite tag with a
+// tree node and nothing else, so the two fields that carry a tree node's
+// content are asserted absent here exactly as `tiling` is asserted absent
+// there.
+// Every span a node's attribution records, narrowest first and each once — the
+// table the spatial queries scan. Both node shapes carry the channel and both
+// carry the same contract, so both read it here.
+function validateSpans(v: unknown, path: string): void {
```
dpmills, line 188: `validateSpans` was inserted inside `validateOperatorNode`'s comment. It now carries "One entry of the operator pane's node table…" with its own comment concatenated below it and no blank line, and `validateOperatorNode` at line 210 has none.

Same defect at `operator_conversion.rs:121` and `wire.rs:1308`.

cambra-inspector/web/src/operatorView.ts
```diff
@@ -0,0 +1,269 @@
...
+// - **Value edges are the child relation.** They are exclusive ownership, so
+//   they form a forest — that is asserted at the producer, in
+//   `operator_graph::assert_graph_invariants`. Following them needs no cycle
+//   guard.
+// - **Share edges are reference rows.** A leaf naming its target, clickable, so
+//   a reader follows sharing without the subtree being drawn twice.
+//
+// Every cycle in the graph runs through a value edge wired late — the `late`
+// marker — so following the value edges from the walk starts terminates without a
+// cycle guard only because those late edges each land on a node already drawn
```
dpmills: This gives the wrong reason and contradicts lines 6-9 above it. Termination does not depend on "those late edges each land on a node already drawn above them" — it follows from the value edges being acyclic, which `assert_graph_invariants` does pin: single value parent plus reachability-from-the-starts implies it, because a value cycle's members each have their only value parent inside the cycle, so none is a start and nothing reaches in.

As written a reader will think `renderNode` and `serializeOperatorGraph` are order-dependent. Neither has a `seen` set, and neither needs one.

src/inspector_model/design.md
```diff
@@ -214,7 +251,7 @@ name is a field rather than a node, so the only span available is the whole node
 covering a statement would contain the narrower uses inside it, and a consumer takes the first
 containing row, so a broad row would shadow them. Those uses contribute none: `out` in
 `out << value` does not resolve to its declaration. Closing it needs a span on the name field, which
-is a `ccl` change. Over the fixture corpus this is 2 rows of 31.
+is a `ccl` change. Over the fixture corpus a small minority of rows.
```
dpmills: "Over the fixture corpus a small minority of rows." is a sentence fragment, and it replaces the measured "2 rows of 31" with a vague quantifier — an unwritten invariant, by CLAUDE.md's rule. The corpus grew, so give the new bound.

Same swap in `provenance.rs:1517` and `:3092`, where "318 nodes would be clobbered" became "clobbers nodes across the pipeline corpus".

src/ccl/context.rs
```diff
@@ -1722,8 +1745,8 @@ pub fn compile_program(
     main_consumer: Box<dyn Consumer>,
 ) -> Result<CompiledProgram, Vec<CompileError>> {
     // The frontend is [], shared with []: parse through
```
dpmills: This comment is rewritten here but keeps its empty markdown links: "The frontend is [], shared with []", "Phase-internal consistency checks (, )", and "keep their  inside the frontend". Pre-existing, but worth closing while the sentence is being edited anyway.

src/inspector_server/wire_check.rs
```diff
@@ -328,43 +361,163 @@ fn assert_node_table(pane: &Value, at: &str) {
             "{at}.nodes[{i}] repeats node id {id} — the table holds each node once"
         );
     }
-    let root = pane["root"]
-        .as_u64()
-        .unwrap_or_else(|| panic!("{at}.root is a number"));
+    if kind == OPERATOR_PANE_KIND {
+        // Superseded by the derivation below: a shipped start set could only
+        // repeat, or contradict, what the edges say.
+        for retired in ["root", "unowned"] {
```
dpmills: `unowned` was never on the wire — it existed only between two commits of this stack, so no payload a stale producer could ship has it. Every neighbouring absence check names a genuinely retired field (`ir`, `spanIndex`, `typeKind`, `predicateRefs`), which is what makes those worth their line.

The loop variable is also called `retired` while one of its two values, `root`, is not retired — only absent on this pane kind. Same in `wireValidate.ts:291`.

src/inspector_model/design.md
```diff
@@ -78,7 +79,10 @@ node is that pane's own answer.
 
 ### A node on the wire
 
-A pane ships `nodes`, every node of that pane exactly once, and `root`, the id its walk starts from.
+A tree pane ships `nodes`, every node of that pane exactly once, and `root`, the id its walk starts
```
dpmills: `root` is documented here as "the id its walk starts from", and the graph's equivalent is `walk_starts`. That is one concept with two names. Deriving rather than shipping is right; the naming is still a decision to make rather than inherit.

cambra-inspector/web/src/wireValidate.ts
```diff
@@ -175,33 +257,118 @@ function validateChild(v: unknown, path: string): IrChild {
...
+        }
+      });
+    });
+    // The `value` edges reach every node from the ones no `value` edge
+    // subscribes. That is the relation the renderer follows — value edges are the
+    // child relation, share edges are reference leaves — so a node outside this
+    // closure is one no view draws and no selection can reach. A `value` cycle
+    // fails it too: its members are all subscribed, so no walk enters them.
+    const inputsById = new Map(nodes.map((n) => [n.nodeId, n.inputs]));
+    const subscribed = new Set(
```
dpmills: This re-derives the walk-start set that `operatorView.ts:53` already exports as `walkStarts`, in the same directory. Removing the shipped field so no second channel could disagree, then implementing the derivation twice on the same side of the wire, gives back the exposure it removed.

tests/inspector_goldens.rs
```diff
@@ -371,6 +371,174 @@ fn for_accumulator_produces_dense_channelize_window() {
...
+            if let Some(kind) = edge["kind"].as_str() {
+                kinds.insert(kind.to_string());
+            }
+            if edge["deferred"] == Value::Bool(true) {
+                deferred += 1;
+            }
+            // A store key is rendered as its `Value`, which quotes a string —
+            // the one edge role that is neither a field name nor a position.
+            if let Some(role) = edge["role"].as_str()
+                && role.starts_with('"')
```
dpmills: `role.starts_with('"')` recovers the `EdgeRole::StoreKey` variant by depending on `serde_json::Value`'s `Display` quoting a string. `wire_edge` renders all three variants as one bare string, so the distinction existed as a type, the wire discarded it, and this test needs it back — which is the argument for keeping the discriminant on the wire.

src/inspector_model/wire.rs
```diff
@@ -102,17 +103,119 @@ pub struct PaneEntry {
...
+    /// Every node's id, in table order.
+    pub fn ids(&self) -> Vec<u64> {
+        match self {
+            PaneNodes::Ir(nodes) => nodes.iter().map(|n| n.node_id).collect(),
+            PaneNodes::Operators(nodes) => nodes.iter().map(|n| n.node_id).collect(),
+        }
+    }
+
+    /// How many nodes the pane holds.
+    pub fn len(&self) -> usize {
```
dpmills: `len` has no callers at all, and `is_empty` below it exists only to satisfy clippy's `len_without_is_empty`. `ids` (line 138) and `ir` (line 159) are called only from `mod tests`. Four public accessors on a wire type with no production consumer.

src/ccl/panes.rs
```diff
@@ -127,9 +168,15 @@ pub(crate) struct PaneSpec {
     pub(crate) gated: bool,
 }
 
+/// How many panes hold an expression tree.
+///
+/// [`CompiledProgram::pane_trees`]' arity. The [`PaneKind::Ir`] entries of
+/// [`PANES`] match that array element for element, in order, and nothing
+/// enforces it.
+pub(crate) const IR_PANE_COUNT: usize = 6;
```
dpmills, line 175: The doc says the invariant is unenforced in as many words: "The `PaneKind::Ir` entries of `PANES` match that array element for element, in order, **and nothing enforces it**." Counting the `Ir` entries of `PANES` is O(1) and free in a `const { assert!(…) }`.

`pane_ids` catches too-few trees with `unreachable!`; a surplus tree is dropped silently.

src/ccl/panes.rs
```diff
@@ -1102,6 +1158,203 @@ mod tests {
...
+            let deferred = graph
+                .edges()
+                .filter(|(_, e)| matches!(e.kind, EdgeKind::Value { deferred: true }))
+                .count();
+
+            if !subscription_relation_is_acyclic(graph, false) {
+                cyclic_programs += 1;
+                assert!(
+                    deferred > 0,
+                    "{name}: the graph has a cycle and no deferred edge, so a construct closed a                      cycle without a `CycleSlot` and the cycle set no longer names it",
```
dpmills: Two assertion messages in this test print about 20 consecutive spaces, from a source line wrapped inside the string literal rather than with a `\` continuation like the neighbouring messages: "…so a construct closed a⟨20 spaces⟩cycle without…", and again at line 1309. `cargo fmt` does not touch string contents.

dpmills (commented):
Mostly minor comments except for a request to change how we read the operator graph.  The construction-time graph extraction seems brittle and unnecessary

cambra-inspector/web/src/__fixtures__/README.md
groundlar: Fixed. The sentence now names the two kinds that exist and explains `deferred` as a `value` edge wired through a `CycleSlot` after its consumer was built — when it is held, not how.

cambra-inspector/web/src/operatorView.ts
groundlar: Right, and the old reason is worse than weak — it describes graphs the invariants forbid, since a late value edge lands either on an ancestor (a value cycle) or on a node with a second value parent. Rewritten to your argument. `assert_graph_invariants`' own doc gave the same unchecked reason and is fixed with it.

cambra-inspector/web/src/wireValidate.ts
groundlar: Fixed at all three. It's a recurring defect rather than three slips — thirteen across the repo's history — so #166 adds a CI check for the diff shape and repairs the six that were still visible.

cambra-inspector/web/src/wireValidate.ts
groundlar: Agreed. `walkStarts` and `isChildEdge` moved to `types.ts` beside `isIrPane`, so the derivation the validator pins is the one the renderer walks — the validator importing from the view would have been the wrong direction. One correction: `walkStarts` wasn't exported, so sharing it meant exporting it and widening the parameter to the node array.

src/ccl/context.rs
groundlar: The links are empty on `main` too, so there was nothing to restore. Rewritten to name `run_frontend` and `compile_to` in prose instead.

src/ccl/design/provenance.md
groundlar: Fixed in all three places — the Terms table, the intro, and the definition at line 47 — and both terms are now defined. The spread is wider than five files: `walk start` reaches eleven, `boundary node` twelve.

src/ccl/design/provenance.md
groundlar: Renamed to `## Operator conversion`, with its four inbound references updated.

src/ccl/panes.rs
groundlar: I agree the term is redundant — an IR here is either the AST or CCL — but it's `main`'s name, not this PR's: `IrNode`/`IrChild` on the wire, five `*_ir` fields on `CompiledProgram`, and `ir.md`, titled "CCL IR — the typed AST". Renaming only the new sites gives `TreePane { nodes: IrNode[] }` and `PaneNodes::Tree(Vec<IrNode>)`, so I'd rather drop it everywhere as its own change than split it — say the word and I'll queue that. On the five representations: four are exhaustive matches, so only `IR_PANE_COUNT` was unenforced, which is the next thread.

src/ccl/panes.rs
groundlar: Worse than unenforced — this PR removed the enforcement. `pane_trees` returned `[&Expr; PANES.len()]` on `main`, and the doc claiming the two cannot disagree was left in place while a doc ninety-odd lines further down says nothing does. A `const` assertion now pins that the `Ir` entries are exactly the leading `IR_PANE_COUNT`, which covers the position `pane_ids` zips against as well as the count; setting it to 5 fails the build twice. That makes the surplus case unreachable, so I left `pane_ids` alone rather than adding a check restating it.

src/ccl/panes.rs
groundlar: Fixed, with `\` continuations like the neighbours.

src/inspector_model/design.md
groundlar: Fixed — the row names the operator graph.

src/inspector_model/design.md
groundlar: They're less alike than the shared wording made them look: `root` is shipped because it *is* the pane's own `Expr`, handed over by the compiler, while `walk_starts` is derived from the edges. Both kept, and the prose now says which is which rather than calling `root` a walk start; `wireValidate.ts`'s local is renamed `rootId`.

src/inspector_model/design.md
groundlar: Deleted here — `definitions.rs` already carries the same passage without that sentence. The `318` is re-measured and pinned instead: it is 30 now, because the binder-slot widening brought the predicate ids into the enumerated domain, and `the_predicate_sweep_skips_this_many_nodes_over_the_corpus` holds it so it can't rot again.

src/inspector_model/wire.rs
groundlar: Confirmed across every commit above this one in the stack, not just here. `len` and `is_empty` are deleted; `ids` and `ir` are `#[cfg(test)]`.

src/inspector_model/wire.rs
groundlar: Fixed — see the `wireValidate.ts` thread for the check that catches this class now.

src/inspector_server/wire_check.rs
groundlar: Both right; `unowned` is gone from both validators and the loop with it, so `root` is a single absent-on-this-kind assertion. `roots` — the operator pane's key four commits earlier — was the one in-stack field genuinely unchecked, which decided it: I dropped the in-stack-staleness category rather than adding a line for it, since nothing outside this unmerged stack can produce either.

src/interpreter/operator_conversion.rs
groundlar: Fixed — the doc is back on the function and `LetBinding` carries its own.

src/interpreter/operator_graph.rs
groundlar: Agreed, and rewritten: the graph is walked from the program's outputs, and the recorder, `drop_operator` and both linear scans are gone. `TileOperator::visit_inputs` is now the single statement of an operator's inputs, which collapses the 28 `add_inspect_children` impls that were the second and disagreeing statement of the same thing — `inspect` derives its children from `visit_inputs` now. `CycleSlot::setter` loses all three additions, `OperatorBase` loses the `T` it carried for the recorded label, and a node holds its `Tiling` rather than a rendered string, so the per-operator `to_string` is paid once at the wire instead of on every compile; the walk itself stays unconditional because the graph is a pane's content. `design-operators.md` has the section you asked for, "Operator identity and the graph the inspector reads".

src/interpreter/operator_graph.rs
groundlar: Five, counting `panes.rs`'s `subscription_relation_is_acyclic` and the runtime, where the consumer holds its input and calls `get` on it. The disclaimer is gone; the module doc says an edge is a subscription and names `notify` as the relation running the other way. The two edges held without being subscribed — a source read, and a fan branch that lost `should_subscribe` — are stated as the exceptions they are.

src/interpreter/tile_operators/aggregate.rs
groundlar: Done, all 34 sites outside the macro body.

src/interpreter/tile_operators/aggregate.rs
groundlar: Twenty-eight, not fifty. The ones on fields `main` documented are reverted to its exact text; the ones on fields `main` left bare are deleted, except two in `fanout.rs` where the line says where the tiling comes from rather than restating the type. No `Identity and …` doc survives.

src/interpreter/tile_operators/mod.rs
groundlar: Gone — `OperatorBase` lost its `T` and the `PhantomData` with it, since the label is `kind()` read at walk time rather than a recorded one. Nothing is recorded at construction any more, so the label never passes through the base: it comes from `Self` in the trait impl, which is the holder by construction.

src/web_inspector.rs
groundlar: Fixed — `docs/design.md` now attributes the operator graph to the program inspector and leaves the deprecated dashboard the live per-tick view it is still the only source of.

tests/inspector_goldens.rs
groundlar: Agreed, and the discriminant is back: `role` ships as `{kind, name|index|key}`, so the test reads `role["kind"] == "storeKey"`. The real loss was upstream of the test — `Named("0")` and `Positional(0)` were indistinguishable on the wire — and both validators now check the shape.

dpmills (approved):

PR #166, opened 2026-09-10
Gate against an item inserted into another item's doc block

groundlar:
An item added at the bottom of an existing doc block adopts that doc and leaves the item below it undocumented. Nothing catches this today: the finished file reads as documented, and only the diff shows what happened. This adds `.github/scripts/doc-adoption.py`, a per-commit diff scan wired into `ci.sh` as `ci_doc_adoption`, and repairs the six blocks already merged that way.

Start with the script's module docstring for the diff signature it matches and the `doc-adoption-ok` suppression marker; `.github/scripts/test_doc_adoption.py` pins `scan` against nine diffs, including the appended-item and new-file cases that must stay quiet. The `ci.sh` comment covers the `jj` tip lookup — under `jj` the git `HEAD` is the working copy's parent, so `main..HEAD` would check nothing. The doc repairs span `src/ccl/` and `tests/`; only `src/ccl/transact_phase.rs` needed a move, where the `StoreWriters` alias had landed inside `run`'s doc block.

### DIFF
--- a/.github/scripts/doc-adoption.py
+++ b/.github/scripts/doc-adoption.py
@@ -1,0 +1,118 @@
+#!/usr/bin/env python3
+"""Fail on an item inserted at the bottom of another item's doc block.
+
+The inserted item adopts the doc above it and the item it displaced is left
+undocumented. Both halves are silent: the new item reads as documented, and the
+displaced one is only missing a comment.
+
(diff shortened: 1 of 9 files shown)


dpmills (approved):
Nice, thanks!

PR #168, opened 2026-09-10
Compile `pass` by dropping it at every block entry

dpmills:
`pass` parses into `ChlStmt::Pass`, which lowered nowhere: no block walker had an arm for it, so every position rejected it as unsupported, though the spec documents it. A `match` arm that does nothing then had no spelling — an `if` arm can be left out, while a `match`'s arms partition the scrutinee's tags. All five block walkers now drop `pass` on entry through one filter, `contributing_stmts`, so no statement grammar needs an arm.

## What a block that contributes nothing is

Each walker's own shape answers it. A value block rejects it (`block_contributes_nothing`); a `for`-loop body is `unit`, so the loop still runs once per element of its source; a mirror loop body and a `with begin():` block are the manufactured `Unit` their chain already starts from, which leaves `lower_tx_block`'s existing must-do-something rule to report a transaction with no footprint. The drop runs before a walker splits its last statement off, so a trailing `pass` is an identity.

## Docs and tests

[chl-spec.md](docs/chl-spec.md#47-pass) is rewritten around the `match` arm, and [lowering.md](src/ccl/design/lowering.md#pass-statements--dropped-at-every-block-entry) records the drop. The cases in `tests/compilation_pipeline/misc.rs` cover both loop-body grammars, `with begin():`, and the three rejections.


src/ccl/lower/loops.rs
```diff
@@ -620,6 +623,10 @@ fn lower_for_body_terminal(
             "a `with begin():` transaction inside a generator/nested for-loop body \
              is not yet supported",
         )),
+        // `pass` contributes no statement, and a loop body expects no value, so a body
+        // that ends in one ends in `unit` — the same terminal a guard with no `else`
+        // falls through to (`docs/chl-spec.md`, "4.7 `pass`").
+        ChlStmt::Pass => Ok(ctx.tag_image(Expr::lit(Lit::Unit), stmt.span)),
```
groundlar: A trailing `pass` breaks a generator/feed loop body. `lower_for_body_stmts_scoped` splits the terminal off at line 359, before `pass` is considered, so this arm claims the terminal and the feed, yield or nested `for` that held it becomes a non-terminal statement. The `_` arm at line 447 then rejects it with "only assignments and function definitions are supported as non-terminal statements in for-loop bodies", pointing at the feed.

This program fails to compile on the PR head, and evaluates to 3 with the `pass` removed:

```python
g = defer()
for i in [1, 2]:
    g << i
    pass
sum(g)
```

The same failure reaches `yield` and an `if` or `match` arm inside a generator body. The last is this PR's own case at `tests/compilation_pipeline/misc.rs:139` with a `pass` added after `good << n` on line 145.

The direct-mirror grammar (line 1181) and the `with begin():` grammar both accept a trailing `pass`, so this grammar is the outlier.

Fix: drop trailing `pass` statements before the `split_last` at line 359, and return `unit` when none remain. That change compiles all three shapes above and leaves the full suite green.

docs/chl-spec.md
```diff
@@ -1715,16 +1715,32 @@ begin():` block, and `while` loops.
...
+
+A `match` arm is where one is needed. An `if` arm can be left out — a guard
+with no `else` does nothing at the positions it rejects — while a `match`'s
+arms partition the scrutinee's tags, so an arm that does nothing is still
+written.
+
+The blocks that expect no value are a `for`-loop body, a `with begin():`
+block, and the `if` / `match` arms inside one. Everywhere else a block's value
+is its last statement, so `pass` stands before that statement and not as it:
+`def todo(x): pass` is rejected, because the function body must yield a value.
```
groundlar: This states a rule the code does not honor. A `for`-loop body rejects a `pass` standing last beside other statements, at the terminal arm in `src/ccl/lower/loops.rs` line 629, so "the blocks that expect no value are a `for`-loop body" holds for a leading `pass` and fails for a trailing one.

Land the `loops.rs` fix in this PR so the sentence holds.

docs/chl-spec.md
```diff
@@ -1715,16 +1715,32 @@ begin():` block, and `while` loops.
...
-
-A `pass` body is only valid where a block of statements is required and
-no value is expected. A function body of `pass` is rejected because the
-function body must yield a value.
+for m in entries:
+    match m:
+        case `debit(d):
+            total += d
+        case `credit(c):
+            pass
```
groundlar: This example does not compile as printed. `entries` and `total` are undeclared, and `total += d` requires `total` to be a mutable variable declared before the loop. The runnable form is the `acc := 0` case at `tests/compilation_pipeline/misc.rs:154`.

Let's move that here so doc examples are runnable.

src/ccl/lower/transactions.rs
```diff
@@ -314,6 +314,10 @@ fn lower_tx_block_scoped(
                     "nested `with begin():` transactions are not supported",
                 ));
             }
+            // `pass` contributes no statement, so the chain passes through
+            // (`docs/chl-spec.md`, "4.7 `pass`"). A block of nothing but `pass` has no
+            // footprint, which the caller's must-do-something check reports.
+            ChlStmt::Pass => chain,
```
groundlar: This is the sixth `ChlStmt::Pass` arm, after `loops.rs:446`, `loops.rs:629`, `loops.rs:1181`, `stmts.rs:441` and `stmts.rs:940`. One concept, spelled six ways.

Drop `pass` once, at each block-lowering entry, and let each grammar answer the one question left: what is an empty block? A value block errors, a loop body is `unit`, and a `with begin():` block takes the empty-chain path `lower_tx_block_inner` already has for `fallback_span`. That names a concept the code reasons about in four places without naming it, the three `assert!(!stmts.is_empty())` parser-invariant asserts and that `fallback_span`.

The same filter subsumes the trailing-`pass` bug on `loops.rs:629`: with `pass` gone before the split, no terminal arm is reachable by one.

As written, a statement-dispatch site added later that omits the arm fails at runtime with a generic "unsupported statement" rather than at compile time.

src/ccl/lower/loops.rs
```diff
@@ -620,6 +623,10 @@ fn lower_for_body_terminal(
             "a `with begin():` transaction inside a generator/nested for-loop body \
              is not yet supported",
         )),
+        // `pass` contributes no statement, and a loop body expects no value, so a body
+        // that ends in one ends in `unit` — the same terminal a guard with no `else`
+        // falls through to (`docs/chl-spec.md`, "4.7 `pass`").
```
groundlar: Nit: this comment and the tagging disagree. The guard fallthrough it points at is `tag_machinery(.., "lower.guard_fallthrough")` on line 562, while the `unit` on the next line is `tag_image`. `src/ccl/design/provenance.md`, "The seam" gives `label` no cross-site guarantee, so either tag is admissible, but a comment calling them the same terminal should not tag them differently. Name the reason `pass` takes an image tag, that it has a source statement and a missing `else` does not, or use `tag_machinery`.

docs/chl-spec.md
```diff
@@ -1715,16 +1715,32 @@ begin():` block, and `while` loops.
...
+with no `else` does nothing at the positions it rejects — while a `match`'s
+arms partition the scrutinee's tags, so an arm that does nothing is still
+written.
+
+The blocks that expect no value are a `for`-loop body, a `with begin():`
+block, and the `if` / `match` arms inside one. Everywhere else a block's value
+is its last statement, so `pass` stands before that statement and not as it:
+`def todo(x): pass` is rejected, because the function body must yield a value.
+
+A loop body of nothing but `pass` is a loop that does nothing, and compiles as
```
groundlar: Nit: the loop is not dropped. `for i in [1, 2]:` with a `pass` body, as a program's final statement, evaluates to `SealedFunction { domain: UInts([0, 1]), codomain: Scalar(Units(2)) }`, one `unit` per source element. The comment at `src/ccl/lower/loops.rs:1159` claims the phase "drops the loop as a no-op". Say here that the loop has no effect, or drop the `loops.rs` claim.

tests/compilation_pipeline/misc.rs
```diff
@@ -117,3 +118,86 @@ fn test_multi_arg_param_in_filter_predicate() {
...
+    Value::Int(2)
+)]
+// As a loop body's whole content: a loop that does nothing.
+#[case("for i in [1, 2]:\n    pass\nsum([1, 2])", Value::Int(3))]
+// Before an accumulator write, so the drop leaves the write as the body.
+#[case(
+    "acc := 0\nfor i in [1, 2]:\n    pass\n    acc += i\nacc",
+    Value::Int(3)
+)]
+fn pass_contributes_no_statement(#[case] code: &str, #[case] expected: Value) {
```
groundlar, line 243: Three cases are missing.

A trailing `pass` after an effect, which is the case that fails: append a `pass` after `good << n` on line 145 and that program stops compiling.

A `pass` inside a `with begin():` block's `if` or `match` arm, the position `docs/chl-spec.md` line 1735 names. Both work today, a `` case `b(k): pass `` arm inside a transaction gives 2 and an `if i > 1: pass` beside a spine `b += i` commits unconditionally, so these pin behavior rather than report a bug.

A standalone `with begin(): pass` outside a loop. It is rejected with "must do something", matching `docs/chl-spec.md` line 1741.

groundlar (approved):
One real finding, LGTM after that's addressed

src/ccl/lower/transactions.rs
dpmills: Done, in `4a994b9`. `contributing_stmts` (`src/ccl/lower/stmts.rs`) drops `pass` once, at the entry the five block walkers already share, and all six arms are gone. Each walker answers only what a block that contributes nothing is: `block_contributes_nothing` for a value block, `unit` for a `for`-loop body, and the manufactured `Unit` the chain already starts from for a mirror loop body and a `with begin():` block — which leaves the transaction with no footprint for `lower_tx_block` to report.

The trailing-`pass` bug is subsumed as you said: the drop runs before the split, so no terminal arm is reachable by one. `src/ccl/design/lowering.md`, "`pass` statements — dropped at every block entry" records the rule beside the five walkers the type-alias section already names.

src/ccl/lower/loops.rs
dpmills: Fixed by the entry-level drop rather than by a filter before `split_last`, so the same change covers the value block and the two others. Your program answers 3 now, and a trailing `pass` after `good << n` is `#[case]` 3 of `pass_contributes_no_statement`.

src/ccl/lower/loops.rs
dpmills: Gone with the arm. A body that contributes nothing has no source statement to be the image of, so its `unit` is `tag_machinery(…, "lower.loop_unit")` — the same label `lower_loop_body_chain_scoped`'s own chain seed carries.

docs/chl-spec.md
dpmills: Done. The section now carries the runnable program, which answers 3 (`total := 0` over ``[`debit(3), `credit(4)]``, read back as `total`). I checked it by extracting the fenced block verbatim and running it — `#` rather than `--` for the trailing comment, since §1.2 makes `#` the comment marker.

docs/chl-spec.md
dpmills: The `loops.rs` fix is in this PR, as the entry-level drop. The sentence holds at every position now, and says so: "`pass` contributes nothing at every position, last included, so a block reads as though it were not written."

docs/chl-spec.md
dpmills: Taken on the spec side: "A loop body of nothing but `pass` is a loop whose body has no effect; the loop still runs, once per element of its source." I left the `loops.rs` comment alone — it is about a bare expression statement that turns out pure after inlining, an arm `pass` never reached and now cannot, so your `Units(2)` evidence does not bear on it.

tests/compilation_pipeline/misc.rs
dpmills: All three added, each checked against the built binary first: a trailing `pass` after `good << n` and after a `<<` feed, a `pass` arm inside a `with begin():` `match` (2) and inside its `if` beside a spine write (3), and the standalone `with begin(): pass` as a second case of `a_transaction_of_nothing_but_pass_has_no_footprint`.

One expectation moved. `x = 1` followed by `pass` no longer reports the `pass`: the drop makes `x = 1` the block's value, and it earns the rejection an assignment earns at that position. `a_trailing_pass_leaves_the_statement_above_it_as_the_value` pins that, and `a_block_of_nothing_but_pass_has_no_value` keeps the `pass`-only cases on the message they had.

PR #169, opened 2026-09-10
Admit a feed above a for-loop body's last statement

dpmills:
A for-loop body could feed once, as its last statement: a `<<` or `yield` above it was a lowering error, because the prefix loop admitted only bindings and only the terminal lowerer recognized a feed. Every position now lowers through that one function, renamed from `lower_for_body_terminal` to `lower_for_body_stmt`, and the prefix carries a `PrefixStmt`: a `Bind` folds a `Let` around the rest of the body, an `Effect` sequences a feed before it with `ExprStmt`. Two feeds into one deferred collection, or into two, now stand in a body with no accumulator.

## What a non-final position still keeps out

An `if`, a `match` and a nested `for` are the body's value and stay last, because `channelize`'s fan-out dispatches on a bare `Case` at the lambda body; `lower_for_body_stmt`'s doc comment carries the rest. A mutable-variable introduction (`t := e`) there now reports its own rejection rather than the generic one.

## Tests

`interleaved_feeds_in_a_loop_body` covers one deferred collection and two, in the `<<` and `yield` spellings; `a_non_final_loop_body_statement_is_a_binding_or_a_feed` pins what the position rejects. The `#[ignore]`d interleaved case in `feeds_cases.rs` runs, with its expected values corrected: `x = i` then `x = x + i` rebind immutably per iteration, so the feeds see `i` and `2i`.


src/ccl/lower/loops.rs
```diff
@@ -444,44 +461,78 @@ fn lower_for_body_stmts_scoped(
...
 
-/// Lower the terminal (last) statement of a for-loop body.
+/// Lower one effect statement of a for-loop body, at any position in it.
 ///
-/// Valid terminals:
 /// - `yield e` — `Feed(defer_name, lower(e))` (requires `defer_name` to be set)
 /// - `r << e` — `Feed("r", lower(e))`
 /// - `if cond: body` (no else) — `Case` with `Unit` fallthrough
+/// - `match m: case …` — the same `Case`, dispatching on a tag
 /// - `for j in ys: body` — nested `Compose([ys, Lambda(j, body)])`
```
groundlar, line 563: Three of these six bullets are false at a non-final position, and two of them crash the compiler.

**`if` — panics.**

```python
o = defer()
for i in [1, 2, 3]:
    if i > 1:
        o << i
    o << 100
sum(o)
```

Debug: `src/ccl/infer_var.rs:512` — `open bound recorded on ?334: 'i' is free in the lower bound ({Unit | i > 1} ⇤ Int) ... the bound left its binder's scope without a mediating discharge`. Release: `src/ccl/context.rs:1501` — `post-channelize produced an invalid tree: [Expected function type at type of [1,2,3] ≫ (λ i : Int → λ __unused : {Unit | i > 1} → i) ⊎ [1,2,3] ≫ (λ i : Int → 100), found Int]`. The generator spelling (`yield x`, then `if x > 1: yield 100`) panics the same way, as does `if`/`else` with both arms feeding.

**Nested `for` — panics in both profiles.**

```python
o = defer()
for i in [1, 2]:
    for j in [10, 20]:
        o << i * j
    o << i
sum(o)
```

`post-channelize produced an invalid tree: [Expected function type at type of [1,2] ≫ (λ i : Int → [10,20] ≫ (λ j : Int → i * j)) ⊎ [1,2] ≫ (λ i : Int → i), found Int]`.

**`match` — a wrong-cause error.** A `match` whose arms all feed evaluates at the last position and fails one statement earlier with `deferred collection: deferred binding 'o' is fed from some but not all branches of a multi-way conditional; this shape is not supported yet`.

The base rejects all three with a lowering error, so each is a move from a diagnostic to a crash or to an error naming the wrong cause.

The cause is in `channelize`. Its arm fan-out dispatches on the lambda body directly — `try_extract_fanout_feed(&body, defer_name)` at `src/ccl/channelize.rs:2441` — and that function matches `body.node` for a bare `Case` (`src/ccl/channelize.rs:244-249`). This lowering produces `λ i → ExprStmt(Case{...}, Feed(...))`, so the match fails, the fan-out never runs, and the generic path emits the `λ __unused : {Unit | i > 1} → i` companion channel the post-channelize wall rejects. A terminal `Case` is bare and fans out correctly: `for i in [1,2,3]: if i > 1: o << i` returns 5.

The same body with an accumulator works, through `lower_loop_body_chain`: `acc := 0`, `if i > 1: o << i`, `acc += 1` returns 5. Extending the no-accumulator path to `if`/`match`/`for` requires `try_extract_fanout_feed` and its `Compose`-site dispatch to walk the `ExprStmt` spine, which is a separate change with its own tests.

src/ccl/lower/loops.rs
```diff
@@ -444,44 +461,78 @@ fn lower_for_body_stmts_scoped(
...
-                    stmt.span,
-                    "only assignments and function definitions are supported as \
-                     non-terminal statements in for-loop bodies",
-                ));
+                let effect =
+                    lower_for_body_stmt(stmt, defer_name, mutation_scope, &frame_introduced, ctx)?;
+                prefix.push(PrefixStmt::Effect {
+                    effect,
+                    span: stmt.span,
+                });
```
groundlar: Narrow this arm to the forms `channelize` supports, and the three crashes become lowering errors again.

```rust
ChlStmt::Expr(v) if matches!(&v.node, ChlExpr::Yield(_) | ChlExpr::Feed { .. }) => {
    let effect =
        lower_for_body_stmt(stmt, defer_name, mutation_scope, &frame_introduced, ctx)?;
    prefix.push(PrefixStmt::Effect { effect, span: stmt.span });
}
_ => {
    return Err(LoweringError::unsupported(
        stmt.span,
        "only assignments, function definitions, `<<` feeds and `yield` are \
         supported as non-terminal statements in for-loop bodies",
    ));
}
```

Under this arm `two_feeds_into_one_defer_in_a_loop_body` stays green and the full test suite stays green. `PrefixStmt`, the `ExprStmt` sequencing, and the rename all survive; the doc bullets at lines 524-526 come out with it. `docs/chl-spec.md`, "4. Statement semantics" licenses exactly this set: a non-terminal statement "either introduces a binding visible to the remainder of the block, or performs an effect (a feed into a deferred output)".

src/ccl/lower/loops.rs
```diff
@@ -444,44 +461,78 @@ fn lower_for_body_stmts_scoped(
...
 /// - `yield e` — `Feed(defer_name, lower(e))` (requires `defer_name` to be set)
 /// - `r << e` — `Feed("r", lower(e))`
 /// - `if cond: body` (no else) — `Case` with `Unit` fallthrough
+/// - `match m: case …` — the same `Case`, dispatching on a tag
 /// - `for j in ys: body` — nested `Compose([ys, Lambda(j, body)])`
+/// - `pass` — `unit`
+///
+/// The body's **last** statement is its value; [`lower_for_body_stmts`] sequences the
+/// rest before it. Both positions lower through here, so a body's statements have one
+/// grammar rather than one per position.
```
groundlar: This sentence obliges `lower_for_body_stmt`'s two rejection messages to stop naming a position, and they still name one. `src/ccl/lower/loops.rs:565` and `src/ccl/lower/loops.rs:683` both read "for-loop body must end in a yield, `<<` feed, nested for, if-guard, or match", and a non-final statement now reaches them:

```
 3 │     foo(i)
   │        ╰──── for-loop body must end in a yield, `<<` feed, nested for, if-guard, or match
```

The base says "only assignments and function definitions are supported as non-terminal statements in for-loop bodies" there. `return i` and `t := i * 2` at a non-final position take the same regression.

`lower_loop_body_chain` already carries position-neutral wording at `src/ccl/lower/loops.rs:1236-1239`: "only assignments (`x = ...`, `x op= ...`), `<<` feeds, `yield`, `if` guards, `match` dispatch, `with begin():` transactions, and bare side-effect calls are supported inside a for-loop body". Reuse that shape.

src/ccl/lower/loops.rs
```diff
@@ -444,44 +461,78 @@ fn lower_for_body_stmts_scoped(
             // `pass` contributes no statement, so it binds nothing here
             // (`docs/chl-spec.md`, "4.7 `pass`").
             ChlStmt::Pass => {}
+            // Everything else is an effect, lowered by the same rule that lowers it
+            // as the body's last statement and sequenced before what follows. The
+            // grammar of a body's statements is therefore one grammar, at whichever
+            // position the statement sits — which is what lets two feeds into one
+            // deferred collection stand in a body with no accumulator, the shape
+            // [`lower_loop_body_chain`] already admits when one is present.
```
groundlar: `x := e` is not an effect, and this arm calls it one. `ChlStmt::MutAssign` has no arm above, so both `:=` spellings fall through here onto `lower_for_body_stmt`'s terminal-position message:

| non-final statement | message |
|---|---|
| `t: Mut(Int) = i * 2` | `` `t` is a mutable variable introduced inside a for-loop body ... `` |
| `t := i * 2` | `for-loop body must end in a yield, ...` |
| `t: Mut(Int) := i * 2` | `for-loop body must end in a yield, ...` |

`in_loop_mut_var_error`'s doc at `src/ccl/lower/loops.rs:275-288` says the error is "Uniform across spellings on purpose ... at any spelling: `x := e`, `x: T := e`, or `x: Mut(V) = e`", and names annotation-gating as the split it closes. The split is open again on this path. Add a `ChlStmt::MutAssign` arm calling `in_loop_mut_var_error` after the outer-scope check.

src/ccl/lower/loops.rs
```diff
@@ -444,44 +461,78 @@ fn lower_for_body_stmts_scoped(
             // `pass` contributes no statement, so it binds nothing here
             // (`docs/chl-spec.md`, "4.7 `pass`").
             ChlStmt::Pass => {}
```
groundlar: Under #168's decision — `pass` evaluates to `Unit` and is an identity for statement sequencing at any position, including last — this arm is that identity law rather than a position special case. Keep it.

Deleting it is behaviour-preserving: the `_` arm below lowers `pass` to `unit` and wraps it as `ExprStmt(unit, body)`, which `drop_expr_stmts` collapses to `body` because `unit` carries no phase marker (`src/ccl/channelize.rs:945-951`). The full test suite passes with the arm removed. It earns its place by stating the identity where the grammar is, and by not minting two nodes per `pass` that move the inspector goldens.

The comment above it states a different rule. "`pass` contributes no statement, so it binds nothing here" is about binding; what holds is that `pass` lowers to `unit` at every position and `ExprStmt(unit, b)` is `b`. Saying that makes this arm an instance of the one-grammar claim at line 530 instead of an exception to it.

tests/compilation_pipeline/feeds_cases.rs
```diff
@@ -416,6 +417,39 @@ o"#;
...
+            o << i
+            o << i * 10
+            acc += 1
+        sum(o)
+    "#},
+    Value::Int(33)
+)]
+fn two_feeds_into_one_defer_in_a_loop_body(#[case] code: &str, #[case] expected: Value) {
+    check_scalar(code, expected);
+}
```
groundlar: `case_25` at `tests/compilation_pipeline/feeds_cases.rs:285-304` already pins the shape this PR unlocks, under an `#[ignore]` whose TODO this PR resolves ("this should work, but our filters on supported exprs in loops are too restrictive"). Drop the `#[ignore]` rather than adding a new program.

Un-ignoring it needs one correction. The pinned codomain is `Ints([0, 1, 3, 10, 30, 60])` and the program produces `Ints([1, 2, 3, 20, 40, 60])`; `domain`, `domain_predicate` and `deleted` match unchanged. `x = i` and `x = x + i` are per-iteration immutable rebinds, so for `i` in `{1, 2, 3}` the feeds are `x` in `{1, 2, 3}` and `x * 10` where `x = 2i`. The pinned values are the exclusive and inclusive cumulative sums of `[1, 2, 3]`, which is the loop-carried-accumulator reading of `x = x + i` that `=` no longer means.

With that correction `case_25` passes at this head in debug and release, and fails at `dmills/pass-statement` with `only assignments and function definitions are supported as non-terminal statements in for-loop bodies`. That makes it this PR's regression test.

The two cases here pin the adjacent-feed shape. What this PR newly admits is an interleaved feed inside a loop body with no accumulator — a feed, an ordinary statement, a second feed — across one defer and two, in the `<<` spelling and the `yield` spelling:

```python
o = defer()
for i in [1, 2]:
    o << i
    t = i * 10
    o << t
sum(o)                       # 33
```

Every program in that class is a lowering error on the base and the right answer here, in debug and release. A double `yield` in one iteration is admitted here too and has no test.

`#[case]` 2 compiles on the base already, through `lower_loop_body_chain`, so it pins routing rather than this change.

Add `check_compile_error` cases pinning the rejection wording for each form the narrowed grammar keeps out.

groundlar (approved):
LGTM once comments are addressed

src/ccl/lower/loops.rs
dpmills: The arm is gone rather than kept, because #168 took the other branch of its own review: `contributing_stmts` drops `pass` at each of the five block-walker entries, so it reaches no statement grammar at any position and there is no arm here to state the identity in. The identity is stated once, at the drop and in `src/ccl/design/lowering.md`, "`pass` statements — dropped at every block entry". Your point about the comment was right either way — it was about binding, and the rule is that `pass` lowers away wherever it stands.

src/ccl/lower/loops.rs
dpmills: Done, in `33cac4ac`, with your guard. All three programs are lowering errors again, and I reproduced each panic first to be sure the arm was the cause. The doc bullets came out with it, and `lower_for_body_stmts`' own doc now says which forms reach `lower_for_body_stmt` only last and why — `channelize`'s fan-out dispatches on a bare `Case`, and one under an `ExprStmt` is not one.

`a_non_final_loop_body_statement_is_a_binding_or_a_feed` pins the rejection for each kept-out form: `if`, `match`, nested `for`, a bare call, `return`, and the three mutable-variable spellings.

src/ccl/lower/loops.rs
dpmills: Added, guarded on a name target so `m[k] := v` stays a keyed write rather than an introduction. All three spellings now give `in_loop_mut_var_error`, checked against the binary, and all three are cases of `a_non_final_loop_body_statement_is_a_binding_or_a_feed`.

src/ccl/lower/loops.rs
dpmills: The `pass` bullet is gone (dropped at the block entry in #168), and the list is now split by what reaches `lower_for_body_stmt` at which position: a feed at either, and `if` / `match` / nested `for` only last. The doc names the reason — `try_extract_fanout_feed` dispatches on a bare `Case` at the lambda body, so one sequenced under an `ExprStmt` is never fanned out and the generic path builds a companion channel the post-channelize typecheck rejects.

src/ccl/lower/loops.rs
dpmills: The narrowing resolves this rather than a rewording. With the prefix arm restricted to `yield` / `<<`, `lower_for_body_stmt` is reached from a non-final position only by those two, which always match an arm above; its two `_` arms are terminal-only again, so "must end in a yield, `<<` feed, nested for, if-guard, or match" is accurate where it fires.

Checked both ends: `foo(i)` and `return i` at a non-final position get "only assignments, function definitions, `<<` feeds and `yield` are supported as non-terminal statements in for-loop bodies", and `return i` last gets the terminal message. `a_loop_body_ends_in_a_value` pins the latter.

tests/compilation_pipeline/feeds_cases.rs
dpmills: Done — `#[ignore]` dropped and the codomain corrected to `[1, 2, 3, 20, 40, 60]`, your reading of `x = i` / `x = x + i` as per-iteration rebinds. Confirmed it is this PR's regression test: `dmills/pass-statement` rejects the program with "only assignments and function definitions are supported as non-terminal statements", and it answers 126 here.

The adjacent-feed case is replaced by `interleaved_feeds_in_a_loop_body`: feed / statement / feed across one deferred collection (33) and two with the reads weighted apart (330), the `yield` spelling of the first, and a double `yield` in one iteration. The accumulator case is dropped, since it pins routing rather than this change, and `a_non_final_loop_body_statement_is_a_binding_or_a_feed` pins the rejection wording for each form the narrowed grammar keeps out.

PR #170, opened 2026-09-10
Fan a loop's conditional out once per deferred collection it feeds

dpmills:
A conditional inside a loop could feed only one deferred collection. `try_extract_fanout_feed` read an arm feeding a second defer as unfannable, sending the whole `Case` to the source-less path — `PartialFeedCaseUnsupported` for a `match`, and for an `if` a gate over the loop binder that lambda elimination rejects — and the extraction behind it collapsed the lambda body to `Unit`, which would drop the sibling's feeds. Such an arm is now a non-feeding arm, and the fan-out runs once per defer, each pass taking its own arms and leaving the rest standing, so `if c: good << i else: bad << i` is two channels over one conditional.

## The residual

`residual_after_fanout` replaces this defer's feed arms with `unit` and leaves every other arm alone, rebuilding from the original body rather than the guard form the fan-out matched on, so a `match` keeps its scrutinee and its patterns for the next pass. Fan-out and residual select arms by different traversals, and an assertion holds their counts equal.

## Tests

`feeds_cases` adds both spellings of complementary arms, a `case _:` default arm feeding its own defer, three defers off one `elif` chain, non-contiguous arms feeding one defer, and two defers in separate clusters. [`mutability.md`](src/ccl/design/mutability.md#value-selecting-case-and-conditional-induction-writes-partially-implemented) and the note in `conditionals.rs` record the in-loop form next to the source-less one.


tests/compilation_pipeline/feeds_cases.rs
```diff
@@ -417,6 +417,55 @@ o"#;
...
+        good = defer()
+        bad = defer()
+        for i in [1, 2]:
+            if i > 1:
+                good << i
+            else:
+                bad << i
+        sum(good) + sum(bad)
+    "#},
+    Value::Int(3)
```
groundlar: `sum(good) + sum(bad)` is symmetric in the two channels, so this test cannot fail for the reason it exists. Swap the arm bodies — ``case `a(n): bad << n`` / ``case `b(k): good << k``, and `bad << i` / `good << i` — and both cases stay green against the same `5` and `3`. A fan-out that routes an arm's values into the wrong defer ships.

Weight the two reads apart: `sum(good) * 10 + sum(bad)` gives `23` and `21`. `check_tile` is stronger and pins each channel's domain, which is what `if_else_both_feed_fan_out` (`:392`) and `multi_arm_if_elif_feeds_fan_out` (`:354`) already do for the single-defer fan-out.

src/ccl/channelize.rs
```diff
@@ -213,6 +213,53 @@ impl fmt::Display for DeferError {
...
+            scrutinee,
+            branches,
+        },
+        ty,
+        user_annotation,
+        node_id,
+    } = body.clone()
+    else {
+        return Expr::lit(Lit::Unit);
+    };
```
groundlar: Take `body` by value, and make the non-`Case` arm `unreachable!`.

The caller owns `body: Box<Expr>` and drops it at the `continue` on `:2578`, so `body.clone()` copies the whole loop body for nothing. `impl Clone for TypedExpr` (`src/ccl/expr.rs:845`) mints a fresh id per node through `provenance::copy_id`, so each pass re-identifies every node in the residual: O(defers x |body|) minted ids, and the surviving `Case` is a copy rather than the node the provenance fold would account for.

The `else` arm is unreachable. The only caller guards on `try_extract_fanout_feed` returning `Some`, which requires destructuring a `Case`. Returning `Unit` there reintroduces this PR's own bug — a sibling defer's feeds dropped, reported as `NoFeedOrDefine`. Name the invariant instead:

```rust
fn residual_after_fanout(body: Expr, defer_name: &Name) -> Expr {
    let TypedExpr {
        node: TypedExprNode::Case { scrutinee, branches },
        ty,
        user_annotation,
        node_id,
    } = body
    else {
        unreachable!("`residual_after_fanout` requires the `try_extract_fanout_feed` Case shape")
    };
```

with `residual_after_fanout(*body, defer_name)` at `:2570`. That signature keeps the suite and `cargo clippy --all-targets -- -D warnings` green.

src/ccl/channelize.rs
```diff
@@ -213,6 +213,53 @@ impl fmt::Display for DeferError {
...
+    let branches = branches
+        .into_iter()
+        .map(|b| Branch {
+            body: match &b.body.node {
```
groundlar, line 258: Two traversals decide which arms are this defer's feeds, and nothing checks they agree. `try_extract_fanout_feed` decides on the converted guard form, where `tag_case_to_guard_case` has substituted pattern payloads. This `.map` erases on the original form by name-matching `TypedExprNode::Feed`. If that substitution ever rewrote a `Feed` into another node, a feed is either extracted and not erased — counted into the channel and left standing for the next pass — or erased and not extracted, and lost. Neither failure has a check.

Pin it at the call site, before the `fanout_arms` loop at `:2534`, where both halves are in scope:

```rust
debug_assert_eq!(
    fanout_arms.iter().filter(|(_, v)| v.is_some()).count(),
    match &body.node {
        TypedExprNode::Case { branches, .. } => branches
            .iter()
            .filter(|b| matches!(&b.body.node,
                TypedExprNode::Feed { name, .. } if name == defer_name))
            .count(),
        _ => 0,
    },
    "the fan-out extracts and the residual erases the same arms"
);
```

The counts agree across the whole suite today.

src/ccl/channelize.rs
```diff
@@ -213,6 +213,53 @@ impl fmt::Display for DeferError {
...
+                TypedExprNode::Feed { name, .. } if name == defer_name => Expr::lit(Lit::Unit),
```
groundlar: `Expr::lit(Lit::Unit)` leaves `ty: Type::Hole` (`Expr::new`, `src/ccl/expr.rs:877-885`). Use `ccl_utils::unit_expr()`, which stamps `Base(Unit)` and is what `tag_case_to_guard_case` uses for the same job.

The `Hole` stays invisible today because `drop_expr_stmts` (`:662`) deletes the residual before `assert_no_type_residue` (`:677`) and the post-channelize `typecheck` see it. It still falsifies the module doc's "Type-preserving by construction" (`:29`) for a node this pass builds and hands to the next defer's pass.

Carrying the `Case`'s `ty` unchanged across the swap is sound because a `Feed` types as `Unit` (`src/ccl/infer/emit.rs:1443-1445`). That fact is load-bearing and unstated; one line in the doc comment covers it.

src/ccl/channelize.rs
```diff
@@ -224,13 +271,21 @@ impl fmt::Display for DeferError {
...
 /// (its first-match predicate is `¬⋁ⱼ gⱼ`).
 ///
+/// **An arm feeding a *different* deferred collection is a non-feeding arm here.**
+/// One `Case` fans out once per defer it feeds, each pass extracting its own arms
+/// and leaving the others standing ([`residual_after_fanout`]), so `if c: good << i
+/// else: bad << i` becomes one refined-source channel per defer. Reading such an arm
+/// as unfannable instead sent the whole `Case` to the generic handling, which reports
+/// [`DeferError::PartialFeedCaseUnsupported`] for a `match` and builds a source-less
+/// gate for an `if` — a gate whose predicate references the loop binder, which
+/// lambda elimination then rejects.
```
groundlar: Three points on this block.

`*different*` is a stress italic, which `CLAUDE.md`'s prose rules replace with the contrast in words: "An arm feeding another deferred collection is a non-feeding arm here."

The module doc at `:46-52` is the file's stated entry point and still reads "one refined-source channel per feeding arm", with no mention of the per-defer pass. Add the per-defer clause there. A reader who starts at the module doc learns the new rule only on reaching this function. (It also still says "multi-arm (`if`/`elif`)" where a `match` fans out too; that half predates this PR.)

The past tense here and at `:221-223` — "sent", "dropped" — reads as changelog. Present tense keeps it as the rejected alternative the style guide permits: "Reading such an arm as unfannable sends the whole `Case` to the generic handling.".

src/ccl/design/mutability.md
```diff
@@ -1206,8 +1206,14 @@ The value-`Case` positions ride the same union-of-restricts:
...
-  scrutinee / pattern feed stays rejected; a no-else partial feed is still blocked earlier at
-  lowering (bare `if` as a value expression).
+  scrutinee / pattern feed stays rejected there; a no-else partial feed is still blocked earlier
+  at lowering (bare `if` as a value expression).
+
+  A conditional feed **inside a loop** takes the per-arm refined-source channels instead, and a
+  pattern arm is admitted there: the fan-out converts the tag test to a guard at its own
+  consumption point. One `Case` fans out once per deferred collection it feeds, each pass
+  leaving the arms it did not take (`channelize`'s `residual_after_fanout`), so two defers fed
+  from complementary arms are two channels over one conditional.
```
groundlar: "The fan-out converts the tag test to a guard" contradicts `residual_after_fanout`, which this PR adds. The conversion runs on a throwaway `body.clone()` inside `try_extract_fanout_feed` (`channelize.rs:291-303`), and the `Case` left in the tree keeps its scrutinee and its arm patterns — which is what makes the second defer's pass possible at all. What survives the conversion is `variant_is` in the arm's refinement predicate and `variant_project` in its channel value:

```
letrec good : ({[0, 1] | __elem ▷ [`a(2), `b(3)] ▷ (λ m → m ▷ variant_is(`a))} ⤇ Int@2)
         = cast([`a(2), `b(3)]) ≫ (λ m → m ▷ variant_project(`a));
       bad  : ({[0, 1] | __elem ▷ [`a(2), `b(3)] ▷ (λ m → m ▷ variant_is(`b) and not m ▷ variant_is(`a))} ⤇ Int@3)
         = cast([`a(2), `b(3)]) ≫ (λ m → m ▷ variant_project(`b))
in (Sum(good)) + (Sum(bad))
```

A reader following the sentence as written expects the tree's `Case` to be guard-form after the first pass, which leaves the second pass unexplainable. Suggested: "the fan-out reads a pattern arm through the tag-to-guard rewrite, so `variant_is` becomes the arm's refinement predicate and `variant_project` its channel value, while the arm stays in tag form for the next defer's pass."

Two smaller points in the same sentence. "a pattern arm is admitted there" holds on `main` as well — a single-defer in-loop `match` feed compiles and answers correctly there — so only the multi-defer pattern case is this change's news, and a reader diffing the doc takes the whole clause as new. "consumption point" arrives from `ccl_utils.rs:325-332`, where it names a contrast between three consumers: the induction writer converting as it walks, `channelize`'s fan-out, and a transaction block's two walks. Without that contrast, "its own consumption point" has no antecedent here.

tests/compilation_pipeline/feeds_cases.rs
```diff
@@ -417,6 +417,55 @@ o"#;
...
+)]
+fn two_defers_fed_from_complementary_arms(#[case] code: &str, #[case] expected: Value) {
```
groundlar: Add the `case _:` default-arm program. It is a behavior change this PR makes and no case covers it:

```python
a = defer()
b = defer()
for m in [`x(1), `y(2), `z(4)]:
    match m:
        case `x(n):
            a << n
        case _:
            b << 7
sum(a) * 100 + sum(b)
```

This answers `114` on the branch and fails on `main` with `PartialFeedCaseUnsupported`. It reaches `tag_case_to_guard_case`'s `has_default` path (`ccl_utils.rs:132-140`) together with the per-defer residual, which neither added case does.

Three further shapes work and are untested: three or more defers off one conditional, the same defer fed from non-contiguous arms with a sibling interleaved, and two defers in separate clusters split by an intervening `let`.

groundlar (commented):
LGTM % comments

src/ccl/channelize.rs
dpmills: Done, in `5a8529e1`, with your signature: `residual_after_fanout(body: Expr, …)`, `residual_after_fanout(*body, defer_name)` at the call site, and `unreachable!` naming the `try_extract_fanout_feed` `Case` shape the only caller reaches it through.

src/ccl/channelize.rs
dpmills: Added at the call site, before the `fanout_arms` loop, with the `Case` destructured through an `unreachable!` rather than a `_ => 0` arm — `try_extract_fanout_feed` returning `Some` is what makes the body a `Case`, and the fallback would have read as a real case.

`assert_eq!` rather than `debug_assert_eq!`: the check is two walks of a handful of arms, and the repo's rule is that a check this cheap is an assert. The counts agree across the suite.

src/ccl/channelize.rs
dpmills: Switched to `ccl_utils::unit_expr()`, and the doc comment now states the fact that makes carrying the `Case`'s `ty` sound: a `Feed` types as `Unit`, so replacing one with `unit` leaves every arm at the type the `Case` already has.

src/ccl/channelize.rs
dpmills: All three taken. The stress italic is now "An arm feeding another deferred collection is a non-feeding arm here."; the past tense at both sites is present ("sends", "drops", "reports"); and the module doc carries the per-defer clause plus the `match` correction — it now reads "an `if`/`elif` chain or a `match`", and says one `Case` fans out once per deferred collection it feeds.

src/ccl/design/mutability.md
dpmills: Rewritten along your lines. The paragraph now says the fan-out reads a pattern arm through the tag-to-guard rewrite, which puts `variant_is` in the arm's refinement predicate and `variant_project` in its channel value, and that the rewrite runs on a copy so the arm in the tree keeps its scrutinee and its pattern for the next defer's pass.

The "a pattern arm is admitted there" clause is gone, since it holds on `main` too — the sentence now leads with the per-defer rule, which is the news. "consumption point" is gone with it rather than given an antecedent here; the contrast it names lives at `ccl_utils.rs` where the three consumers are in view. I made the same correction to the `conditionals.rs` note, which claimed the in-loop pattern feed as this change's.

tests/compilation_pipeline/feeds_cases.rs
dpmills: Taken, and taken the stronger way: each channel is read on its own and pinned with `check_tile`, so the domain is the source restriction its arm earned — `good` is `UInts([0]) ⤇ Ints([2])` and `bad` is `UInts([1]) ⤇ Ints([3])` in the `match`, and the reverse indices in the `if`. Swapping the arm bodies now fails on the domain, not only on the sum.

tests/compilation_pipeline/feeds_cases.rs
dpmills: Added as `a_default_arm_feeds_its_own_defer`, pinned per channel rather than on the combined `114`: `a` is `UInts([0]) ⤇ Ints([1])` and the default arm's `b` is `UInts([1, 2]) ⤇ Ints([7, 7])`, the complement `tag_case_to_guard_case`'s trailing `true` guard gives it. Confirmed it fails on `dmills/loop-body-one-grammar` with `PartialFeedCaseUnsupported`.

The three further shapes are covered too, in `more_defers_off_one_conditional`: three defers off one conditional (12300), one defer fed from non-contiguous arms with the sibling between them (42), and two defers split into separate clusters by an intervening read (21).

src/ccl/channelize.rs
```diff
@@ -224,13 +285,21 @@ impl fmt::Display for DeferError {
...
+///
 /// Returns each arm's `(guard, feed_value?)` in source order — `feed_value` is
 /// `None` for a non-feeding arm, whose guard still participates in later arms'
 /// predicate synthesis ([`synthesize_arm_predicate`]). Returns `None` (not
-/// fannable) if the trailing guard is not `true` or an arm body is neither this
-/// defer's feed nor `Unit` — so a `Case` mixing this defer's feeds with other
-/// effects falls through to the generic handling rather than being silently
-/// collapsed.
+/// fannable) if the trailing guard is not `true` or an arm body is neither a feed
+/// nor `Unit` — so a `Case` mixing feeds with other effects falls through to the
```
sortalongo: This presents falling through as the safe outcome, but for a loop-sourced `if` the generic path does not reject cleanly. It panics in inference with an open-bound error naming an inference variable and no source span. The line above already says lambda elimination rejects the gate, which understates the same thing.

This is pre-existing and reproduces on main, so it is not yours to fix here. Correct the claim in this PR and file the follow-up.

src/ccl/channelize.rs
```diff
@@ -2464,6 +2537,38 @@ fn extract_for_defer_impl(
...
+                                original_branches
+                                    .iter()
+                                    .filter(|b| matches!(
+                                        &b.body.node,
+                                        TypedExprNode::Feed { name, .. } if name == defer_name
+                                    ))
+                                    .count(),
+                                "the fan-out extracts and the residual erases the same arms"
+                            );
                             let source_prefix = if new_elts.len() == 1 {
```
sortalongo: Each feeding arm clones the whole source prefix, so N defers off one conditional compile to N independent restricted scans of the same source and N evaluations of the guard. With a data source and a UDF condition the UDF runs once per defer per element, and there is no common-subexpression pass downstream to collapse it.

State here that the per-defer scan is the accepted cost.

src/ccl/channelize.rs
```diff
@@ -224,13 +285,21 @@ impl fmt::Display for DeferError {
 /// ordinary arms here, so an `else` that feeds fans out just like a guard arm
 /// (its first-match predicate is `¬⋁ⱼ gⱼ`).
 ///
+/// **An arm feeding another deferred collection is a non-feeding arm here.** One
+/// `Case` fans out once per defer it feeds, each pass extracting its own arms and
```
sortalongo: The fan-out-once-per-defer sentence appears here, on `residual_after_fanout`, in the module doc, and in `mutability.md`. Drop it from both function docs and let each say what its own function does.

tests/compilation_pipeline/feeds_cases.rs
```diff
@@ -480,6 +480,192 @@ fn a_comprehension_reads_a_feed_channel_fed_outside_a_transaction() {
...
+/// (`residual_after_fanout` in `channelize`), so an arm feeding the other defer is a
+/// non-feeding arm for this one.
+///
+/// Each channel is read on its own, so the pinned domain is the source restriction the
+/// arm earned — `good` takes index 0 and `bad` index 1 in the `match`, and the reverse
+/// in the `if`. A fan-out routing an arm's values into the sibling's channel fails here,
+/// which a symmetric read of the two would not.
+///
+/// Both spellings are listed because the failure differed by spelling and the cause did
+/// not. A `match`'s arms carry patterns, so the whole `Case` reached the source-less
```
sortalongo: Past-tense narration of what the compiler used to do. Delete the paragraph. The failure is already stated in the PR body and in `try_extract_fanout_feed`'s doc.

tests/compilation_pipeline/feeds_cases.rs
```diff
@@ -480,6 +480,192 @@ fn a_comprehension_reads_a_feed_channel_fed_outside_a_transaction() {
...
+/// path and was reported as `PartialFeedCaseUnsupported`; an `if`'s arms are guards, so
+/// the same path built a `Unit` gate whose predicate references the loop binder, which
+/// lambda elimination rejected as a non-dependent arrow over a dependent codomain.
+///
+/// The loop with an accumulator compiles either way — the tap on the writer decision is
```
sortalongo: This is the only place recording that the repo now holds two representations of a conditional feed in a loop, the per-arm channel and the tap on the writer decision. A test doc comment is not where the next person extending either one will look.

Move the sentence into `mutability.md` beside the new paragraph.

src/ccl/design/mutability.md
```diff
@@ -1207,8 +1207,16 @@ The value-`Case` positions ride the same union-of-restricts:
 - **Source-less conditional feeds** (`if c: o << 1 else: o << 2` outside any loop) — *implemented*
   (`channelize`): each feeding arm becomes a gated one-shot lift `λ __unused : {Unit | π̂ᵢ} → 𝑣ᵢ`,
   one channel per arm — replacing `PartialFeedCaseUnsupported` for guard-only `Case`s. A
-  scrutinee / pattern feed stays rejected; a no-else partial feed is still blocked earlier at
-  lowering (bare `if` as a value expression).
+  scrutinee / pattern feed stays rejected there; a no-else partial feed is still blocked earlier
+  at lowering (bare `if` as a value expression).
+
+  A conditional feed **inside a loop** takes the per-arm refined-source channels instead. One
```
sortalongo: The in-loop rule is a sub-paragraph of the bullet headed for feeds outside any loop, so the taxonomy no longer lists the in-loop case where a reader would look for it. Promote it to its own bullet.

sortalongo (approved):
Approving. The fan-out is a real generalization rather than a special case bolted on, the residual being rebuilt from the original body is the right call, and the count assertion at the seam is the correct defensive check. Exercised against programs beyond the ones here, it fixes more shapes than the description claims: five defers off one chain, non-consecutive and reversed declaration order, a UDF condition, a computed scrutinee, a loop with an accumulator plus two defers, two defers in complementary arms inside a transaction block, and a generator yield mixed with a defer feed.

All seven of Skylar's threads are resolved with code changes. None of the notes below blocks the merge.

The first note is about a pre-existing crash, not a regression. A loop-sourced `Case` that the fan-out declines still panics the compiler rather than reporting a clean error, and it panics the same way on main. Only the doc comment needs correcting here; the fix itself is a follow-up.

Two minor things not worth their own threads: the rustdoc link to `Clone for TypedExpr` will not resolve and renders as literal brackets, and `residues` is used as a verb where `leaves` would do.

src/ccl/channelize.rs
dpmills: Dropped from both function docs. `try_extract_fanout_feed` now states only its own rule — an arm feeding another defer is a non-feeding arm here, which is what leaves `if c: good << i else: bad << i` fannable for either defer. `residual_after_fanout` states only what collapsing the whole body to `Unit` would lose. The module doc and `mutability.md` keep the once-per-defer sentence.

src/ccl/channelize.rs
dpmills: Confirmed, and the claim is corrected. I measured three shapes on `main` at `357b8615`:

- an arm body that is itself a conditional — `if i > 1: (if i > 2: o << i else: o << 0) else: o << 100` — panics in inference: `open bound recorded on ?168: ... is free in the lower bound ({Unit | i > 1} ⤇ ({Unit | i > 2} ⤇ Int))`, at `src/ccl/infer_var.rs:482`, no source span. Same panic on this branch, which does not touch that path.
- the two-defer `if` this PR fixes, on `main`: lambda elimination, on `λ i : Int → λ __unused : {Unit | true and not i > 1} → i`.
- the two-defer `match`, on `main`: `PartialFeedCaseUnsupported`.

So "lambda elimination rejects the gate" was accurate for the shape this PR fixes, and the inference abort is a second shape reaching the same generic handling one phase earlier. The doc now names all three and drops the safe-outcome framing: reaching the generic handling from a loop is a defect wherever it happens, not a fallback.

Filed as `vault/issues/defer-loop-sourced-case-fallthrough-panics.md` — refuse at the point `try_extract_fanout_feed` returns `None` with a `DeferError` carrying the `Case`'s span, rather than building a gate a later phase cannot type.

src/ccl/channelize.rs
dpmills: Stated at the `source_prefix` binding:

```
// Every feeding arm composes its own restriction onto a
// clone of this prefix, so N defers off one conditional
// compile to N restricted scans of the same source and N
// evaluations of the guard — with a data source and a UDF
// condition, one UDF call per defer per element. No
// downstream pass collapses the shared prefix. That is the
// accepted cost of giving each defer a channel whose domain
// is its own arms' restriction: the restrictions are
// disjoint by first-match, so there is no one domain the
// arms could share a scan over.
```

src/ccl/design/mutability.md
dpmills: Promoted to **Loop-sourced conditional feeds**, its own bullet in the taxonomy, alongside the source-less one rather than inside it.

tests/compilation_pipeline/feeds_cases.rs
dpmills: Deleted. What replaces it says why both spellings are listed without the history: they reach the fan-out by different routes, a `match`'s arms carrying patterns and an `if`'s carrying guards.

tests/compilation_pipeline/feeds_cases.rs
dpmills: Moved into `mutability.md`, under the new **Loop-sourced conditional feeds** bullet, with the mechanism named rather than alluded to:

> A conditional feed in a loop has **two representations**, chosen by whether the loop carries a mutable variable. A loop that does reaches `mut_elim`, where a feed becomes a `__to_<defer>_<n>` tap on the writer's decision record whose `fire` is the arm's guard conjunction (`Name::defer_tap_field`), and the `Case` is never fanned out. A loop that does not has only the per-arm channels above. Extending one leaves the other's cases uncovered.

The test doc comment now cites that heading for the accumulator-free condition it relies on.

dpmills:
Rebased onto `main` at `357b8615` (one conflict, the `channelize.rs` import list — `PredicateId` from main beside this branch's `unit_expr`), then the review responses on top as `c38d67f8`.

Two claims not flagged in review, fixed in the same commit because the correction at `try_extract_fanout_feed` contradicted them. The module doc and `DeferError::PartialFeedCaseUnsupported`'s own doc both said a source-less conditional feed is rejected. The gated one-shot lift handles a guard-only one; only a scrutinee / pattern feed is rejected, which is what the code at the `guard_only` check does and what `mutability.md` already recorded. Both predate this PR.

PR #172, opened 2026-09-10
Read both sides at an invariant position however the walk reached it

dpmills:
A mutable `Map` seeded with a one-entry literal did not compile, in any of its three uses — read alone, a keyed write per loop position, and the same write in a transaction. The seed's key type is that key's own singleton, and `box`'s scheme shares one variable between its domain and the sum's single candidate, so the key domain reaches that variable along a bound chain. Meeting both sides was gated on `fallback_allowed`, which the shape collapse needs and a narrowing does not, so the variable read the singleton where the walk entered it structurally and the bare present-key domain where it arrived along the chain — and a data domain's invariance rejected the position against itself. `compact_go` now carries whether the walk is inside a data function's domain, propagated where `parents` is, and the merge fires throughout an invariant position.

Scoped to an invariant position rather than ungated, for two measured reasons. Reading the opposite side at every negative variable doubles the walk per level of type nesting — `def by_key(c, f): groupby(c, f)` applied at two types goes from 1.4s to 10s, where this costs 1.5s. And it reaches a function-typed parameter, a compute domain, where a suspended discharge puts two data domains together whose refinement sets differ only in the discharged binder's spelling; `data_domains_disagree` compares those structurally, so one domain reached twice reads as two that disagree. Both are recorded beside the rule in [type-inference.md](src/ccl/design/type-inference.md#an-invariant-position-reads-both-sides-however-the-walk-reached-it).

The second commit is unrelated to the solver. `docs/chl-spec.md` said refinements "are not parsed yet, so every use is a parse error today" and that refinement annotations are "not writable"; both are stale — `output_annotation_ref1` has been running `def itsa_nine(a) => {Int where _ == 9}: 9` since it landed. It also pins the two ways a `Map` key type no key satisfies is reported: the plain annotation names what the program wrote, while the `Mut` form panics at the post-inference boundary for a program that is merely ill-typed.

### DIFF
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -365,25 +365,33 @@ and the marker then claims the remainder is exhaustive. No program observes the 
 a default arm being compiled from the `Case` (`test_default_arm_under_a_record_field`); the
 reading itself is pinned by `a_settled_negative_position_closes_an_open_child_demand`.
 
-**The merge is gated as if it were the collapse, and one variable therefore has two readings
-at one polarity.** `fallback_allowed` answers both, so a variable entered as a position reads
-the merge while the same variable reached through another variable's bound chain reads the
-demand side alone. Only the collapse needs that gate: a choice does not propagate along
-subtyping edges, and a narrowing does. What the difference leaves open is a mutable `Map`
-seeded with a one-entry literal — `box`'s instantiated domain reaches the key variable
-through a chain and keeps the bare reading, which invariance then rejects against the
-singleton the seed establishes (`a_one_entry_seed_does_not_reach_a_mutable_map`).
+##### An invariant position reads both sides however the walk reached it
 
-Ungating the merge closes that shape and costs two things it does not pay for. Reading the
-opposite side at every negative variable compacts it in full there, so the walk doubles per
-level of type nesting — `def by_key(c, f): groupby(c, f)` applied at two types goes from 1.4s
-to 10s. And where a discharge is suspended on the chain, two data domains meet whose
-refinement sets differ only in the spelling of the discharged binder: `{[0, 2] | 𝑥 == 0}`
-against `{[0, 2] | 𝑥 == __arg}`. [`data_domains_disagree`] compares those sets structurally,
-so it reads one domain reached twice as two domains that disagree, and the position has no
-common answer — which `higher_order_dependent_application_discharges_the_binder` reports.
-Forcing the discharge on both contributions before they meet is the prerequisite for letting
-the merge follow a chain.
+Elsewhere the merge fires only at a position, which `fallback_allowed` answers. That gate
+belongs to the collapse: a collapse is a choice, and a choice does not propagate along
+subtyping edges. Applying it to the merge as well gives one variable **two readings at one
+polarity** — the merge where the walk entered it structurally, the demand side alone where it
+arrived along another variable's bound chain — and inside an invariant position there is no
+variance left to tell those readings apart, so the invariance check rejects the position
+against itself.
+
+A mutable `Map` seeded with a one-entry literal is the shape that reaches it. The seed's key
+type is that key's own singleton, and `box`'s scheme shares one variable between its domain
+and the sum's single candidate, so the key domain arrives at that variable through a chain:
+entered as a position it reads the singleton, reached along the chain it reads the bare
+present-key domain, and the two are the collection the program wrote and a wider one.
+`a_one_entry_seed_reads_as_a_map` and its two keyed-write siblings are the three uses.
+
+**Scoped to an invariant position, not ungated.** Reading the opposite side at *every*
+negative variable compacts it in full there, so the walk doubles per level of type nesting —
+`def by_key(c, f): groupby(c, f)` applied at two types goes from 1.4s to 10s, where the
+scoped read costs 1.5s. It also reaches a function-typed **parameter**, which is a compute
+domain: there a suspended discharge puts two data domains together whose refinement sets
+differ only in the discharged binder's spelling, `{[0, 2] | 𝑥 == 0}` against
+`{[0, 2] | 𝑥 == __arg}`, and `data_domains_disagree` compares those sets structurally, so one
+domain reached twice reads as two that disagree
+(`higher_order_dependent_application_discharges_the_binder`). Identifying those two spellings
+is what letting the merge follow *every* chain would need first.
 
 **The gated merge pays a doubling of its own.** The gate confines the merge to the entered
 position, and a structural child *is* an entered position — `compact_go` resets `parents` to
--- a/src/ccl/design/type-inference.md
+++ b/src/ccl/design/type-inference.md
@@ -365,6 +365,23 @@ and the marker then claims the remainder is exhaustive. No program observes the 
 a default arm being compiled from the `Case` (`test_default_arm_under_a_record_field`); the
 reading itself is pinned by `a_settled_negative_position_closes_an_open_child_demand`.
 
+**The gated merge pays a doubling of its own.** The gate confines the merge to the entered
+position, and a structural child *is* an entered position — `compact_go` takes a fresh
+`Position` at every one — so the merge re-opens one nesting level down and the walk doubles per
+contravariant flip. Measured in debug, median of three runs at equal test counts: the whole
+of `tests/type_check.rs` goes from 0.83s to 1.29s, and `test_groupby_key_relation_is_per_occurrence`
+— `def by_key(c, f): groupby(c, f)` at two types, the same program as above — from 0.56s to
+0.94s. `tests/compilation_pipeline` is flat, its time being execution rather than inference.
+
+What is absent is a **result memo**. `CompactState` carries `in_process`, which prunes cycles
+and caches nothing — it is removed again as each variable's walk returns — so a position
+reached twice is compacted twice. A `(uid, pol)` key would not be a correct memo: a position's
+reading also depends on `subst_acc` and on the refinement scope `st.scope` holds, so two
+entries at one variable and polarity are not interchangeable. Hash consing the rendered
+contribution keys on what was produced instead, and sidesteps that.
+
+**Asking the other question.** Because the collapse answers "what must this position be", a caller that needs "what actually reached it" has to suppress the collapse — `compact_type_polarity_only`, the polarity-correct walk alone. The distinction is not academic: an upper bound deposited on a never-inhabited position (the trait-requirement sweep does exactly this) makes the ordinary resolve report a type. [The unobservable-arm pin](#an-unobservable-arm-payload-is-pinned-to-what-its-uses-require) is the caller that must not confuse the two, since a demand is precisely what an unreachable arm can acquire.
+
 ##### An invariant position reads both sides however the walk reached it
 
 Elsewhere the merge fires only at a position, which `fallback_allowed` answers. That gate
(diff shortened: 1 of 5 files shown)


src/ccl/infer/solver/compact.rs
```diff
@@ -2141,12 +2151,23 @@ fn compact_go(
...
+            // **Inside an invariant position it reads both sides however the walk
+            // arrived**, where elsewhere it reads them only at a position. A data
+            // domain reached along a bound chain is the same domain as one entered
+            // structurally, and invariance leaves no variance to tell the two readings
+            // apart — so gating this one on `allow_fallback` gives one variable two
+            // answers at one polarity, which the invariance check then rejects against
+            // itself (`src/ccl/design/type-inference.md`, "An invariant position reads
+            // both sides however the walk reached it"). `allow_fallback` still gates
+            // the shape collapse, which is a choice rather than a narrowing.
+            let read_opposite =
```
groundlar: **Important:** the `invariant` disjunct escapes `st.collapse`. `allow_fallback` is `st.collapse && parents.is_none()`; the new disjunct checks neither half, so it fires on walks where the collapse is suppressed *walk-wide*. That contradicts `compact_type_polarity_only`'s own rustdoc verbatim: "Suppressing it walk-wide rather than at the entry is the same thing: `fallback_allowed` already confines the collapse to the entered position, so disabling it there disables it everywhere." `type-inference.md` ("Asking the other question") states the same guarantee for the caller, `value_reaches`, whose job is not to mistake a demand for a value.

Unreachable today: `value_reaches` runs on arm payloads, and a payload that *is* a function is already not `Type::Infer(_)`, so the top-level answer doesn't move. I instrumented the exact condition (`!pol && invariant && !allow_fallback && !st.collapse && !opposite_bounds.is_empty()`) and ran the full suite — zero hits. So it is a latent invariant break, which is the same kind this PR fixes.

Behavior-preserving fix, and it also lets the two rules be named separately (the comment above distinguishes them; the expression doesn't):

```rust
// The shape collapse: a choice, only at the entered position.
let collapse = allow_fallback && no_concrete;
// The merge: a narrowing, at a negative position and throughout an invariant one.
let merge = !pol && (allow_fallback || (st.collapse && invariant));
let read_opposite = collapse || merge;
```

src/ccl/infer/solver/compact.rs
```diff
@@ -1756,11 +1756,18 @@ fn fun_kind_correspondence(fun_kind: &crate::ccl::ty::FunKind, acc: &Subst) -> S
...
+/// does not distinguish the two sides. It propagates exactly where `parents` does —
+/// along a variable's bound chain and through a refinement's base, reset at every
+/// other structural child — because it answers a question about the position, and a
+/// structural child is a new one.
 fn compact_go(
     ty: &Type,
     pol: bool,
     subst_acc: &Subst,
     parents: Option<&ParentPath<'_>>,
+    invariant: bool,
```
groundlar: **Important:** `invariant` and `parents` encode one concept and always move together — the doc comment above says so — but that is a prose claim about 12 call sites checked by nobody. Ten read `None, false`; getting one wrong reproduces exactly the failure this PR fixes (one position, two readings), silently, which is the hazard the `compact_go`/`key_go` mirroring note two paragraphs up already warns about.

Bundling makes the claim hold by construction:

```rust
/// Where the walk stands relative to the position it entered.
#[derive(Clone, Copy, Default)]
struct Position<'a> {
    /// Variables whose bound chain the walk is currently following.
    parents: Option<&'a ParentPath<'a>>,
    /// Whether that position is a data function's domain, where variance does not
    /// distinguish the two sides.
    invariant: bool,
}
```

A structural child passes `Position::default()` — one name for "this is a new position" — the bound chain passes `pos.under(uid)`, and the Fun arm `Position::domain_of(fun_kind)`. Fine as a follow-up if you'd rather not grow the diff; worth saying so in the description either way.

src/ccl/infer/solver/compact.rs
```diff
@@ -1756,11 +1756,18 @@ fn fun_kind_correspondence(fun_kind: &crate::ccl::ty::FunKind, acc: &Subst) -> S
 /// polarity flips — has to be mirrored there in the same change. A divergence is
 /// silent, and what it produces is a shared clone whose interior was resolved
 /// against a different use's argument.
+///
+/// `invariant` says the walk is inside a **data function's domain**, where variance
```
groundlar: Nit: "propagates exactly where `parents` does" is accurate about propagation but silent about the introduction site — the Fun arm **overwrites** `invariant` with `fun_kind.resolved().is_data()`, including overwriting an inherited `true`. A reader auditing the threading against the stated rule would expect a pass-through there. One clause fixes it: "…and set at a function's domain from its kind."

Worth stating alongside: `resolved()` answers `Unpinned` for an unnarrowed kind variable, so `is_data()` is `false` and an unpinned data function's domain gets no invariant read. Consistent with how `Display` renders such a kind (`⇒`), but it means the fix is scoped to kinds pinned by compaction time and nothing says that out loud.

src/ccl/infer/solver/compact.rs
```diff
@@ -1622,11 +1622,11 @@ fn compact_type_kind(
         TypeKind::Enumerated(domains) => CompactTypeKind::Enumerated(
```
groundlar, line 1662: Question rather than a finding: a Σ's candidate walks with `invariant: false` here (as does `SubtypesOf`'s bound below). A candidate is the domain σ ranges over, and the commit message's own account of the bug is that "`box`'s scheme shares one variable between its domain and the sum's single candidate" — so if that shared variable is read at `invariant: true` through the domain and `false` through the candidate, that is the same shape one layer up.

It evidently doesn't bite on the three programs here, and the adjacent comment argues covariance of the candidate *set*, which is a different claim from variance of a *candidate*. Flagging per CLAUDE.md on recurring shapes: is the candidate position genuinely covariant, or is this the second instance?

src/ccl/design/type-inference.md
```diff
@@ -359,25 +359,33 @@ and the marker then claims the remainder is exhaustive. No program observes the
...
-opposite side at every negative variable compacts it in full there, so the walk doubles per
-level of type nesting — `def by_key(c, f): groupby(c, f)` applied at two types goes from 1.4s
-to 10s. And where a discharge is suspended on the chain, two data domains meet whose
-refinement sets differ only in the spelling of the discharged binder: `{[0, 2] | 𝑥 == 0}`
-against `{[0, 2] | 𝑥 == __arg}`. [`data_domains_disagree`] compares those sets structurally,
-so it reads one domain reached twice as two domains that disagree, and the position has no
-common answer — which `higher_order_dependent_application_discharges_the_binder` reports.
-Forcing the discharge on both contributions before they meet is the prerequisite for letting
-the merge follow a chain.
+##### An invariant position reads both sides however the walk reached it
```
groundlar, line 385: Promoting the bold lead-in to a citable heading is right. But inserting it mid-section re-parents everything after it, and three paragraphs there belong to the `####`, not to the invariant rule:

- "The gated merge pays a doubling of its own" — explicitly about `parents` being reset at structural children, i.e. the general gate.
- "What is absent is a result memo" — general to `CompactState`.
- "Asking the other question" — about `compact_type_polarity_only`, and the guarantee the `st.collapse` comment above concerns. It now reads as a sub-claim of the rule that breaks it.

Move the `#####` down to open on "The gated merge pays a doubling of its own", or move those three above it.

tests/compilation_pipeline/type_annotations.rs
```diff
@@ -343,3 +343,30 @@ fn type_annotation_naming_a_parameter_is_unbound(#[case] code: &str, #[case] nam
...
+/// annotation is an `Annotation mismatch` against the whole `Σ`, which names the
+/// annotation the program wrote; the mutable one panics at the post-inference boundary as
+/// an invalid tree, which reads as a compiler bug for a program that is simply rejected.
+///
+/// Pinned on both, so the day the `Mut` form reports a user error the pin says so.
+#[test]
+fn a_refined_map_key_no_key_satisfies() {
+    check_compile_error(
+        "m: Map({String where _ != \"\"}, Int) = map([(\"a\", 1), (\"z\", 2)])\nm\n",
+        "Annotation mismatch: annotated as \u{3a3} (\u{3c3} : SubtypesOf({String | __elem != \"\"}))",
```
groundlar: Nit: the repo writes `Σ`/`σ` literally nearly everywhere (`sums.rs:89`, `scalars_collections.rs:57`, the notation section of CLAUDE.md). A raw string reads as the error the compiler actually prints:

```rust
r#"Annotation mismatch: annotated as Σ (σ : SubtypesOf({String | __elem != ""}))"#,
```

groundlar (approved):
Sound fix. The defect is identified precisely — one variable resolving two ways at one polarity inside a position with no variance left to tell the readings apart — and `read_opposite` gains exactly the disjunct `!pol && invariant`, so it is a strict superset of the old behavior. Replacing three `#[should_panic]` cases with value assertions is the right trade: they now say what the programs mean. `./ci.sh fast` and `./ci.sh doc_refs` are green at the tip, and I verified commit 2's three claimed parse positions against the built binary.

Approving with one item I'd fix before merge (the `st.collapse` gate, inline) — it is unreachable today but it is a documented walk-wide guarantee and the fix is one token.

**Test coverage.** The three pipeline tests pin the programs; nothing pins the mechanism, and the mechanism is "however the walk reached it". `a_negative_position_meets_both_sides` covers only the entered-structurally case. A sibling runs in microseconds instead of through the pipeline:

```rust
/// The merge fires at an invariant position **reached along a chain**, not only at one
/// the walk entered — which is how `box`'s scheme reaches a `Map`'s key variable.
#[test]
fn an_invariant_position_reads_both_sides_along_a_chain() {
    let int = || Type::Base(BaseType::Int);
    let inner = fresh_var(0); // the domain's own variable
    let outer = fresh_var(0); // reaches `inner` through a bound chain
    let mut cache = ConstrainCache::new();
    constrain_subtype(&crate::ccl::infer::lit_singleton(&Lit::Int(1)), &inner, &mut cache)
        .expect("the singleton flows in");
    constrain_subtype(&inner, &int(), &mut cache).expect("and meets the demand");
    constrain_subtype(&outer, &inner, &mut cache).expect("the chain");
    let dom = compact_type(&Type::data_fun(outer, int()))
        .term.fun.expect("a function slot").domain;
    assert_eq!(
        dom.refinements.as_ref().map(RefinementSet::len),
        Some(1),
        "the singleton reaches the domain through the chain, not just at the position",
    );
    // `Type::fun` (a compute domain) in the same shape must still read the demand alone.
}
```

The commented last line is the half with no coverage at all: nothing asserts the read does *not* fire outside a data domain. That scope decision is held in place only by `higher_order_dependent_application_discharges_the_binder` failing if you widen it — true, but it fails three inferential steps from the gate.

**Adjacent, not yours.** `sum(m)` after a keyed write panics (`Aggregate expected function tiling, got Scalar(Variants([]))`, `tile_operators/aggregate.rs:130`). Pre-existing — it reproduces with `read_opposite` reverted and with a two-entry seed. Noting it because it falls exactly between the two new tests: one uses `sum` with no write, the other reads through `final_mut_var_value`.

src/ccl/infer/solver/compact.rs
dpmills: Fixed in `Keep the merge under the walk-wide collapse flag`, taking the split as written: `collapse` and `merge` are separate bindings, and `merge`'s invariant disjunct is now `st.collapse && pos.invariant`. The full suite is unchanged, which matches your instrumentation finding zero hits.

src/ccl/infer/solver/compact.rs
dpmills: Done in this PR rather than as a follow-up — `Bundle the walk's position so its two halves cannot drift`.

`Position` carries `parents` and `invariant`. A structural child passes `Position::default()`, a step along a bound chain `Position::under(&path)`, and the `Fun` arm `Position::domain_of(fun_kind)`. Nothing else constructs one, so the "always move together" claim now holds by construction rather than across thirteen call sites.

`key_go` threads no parent path, so the sibling-walk agreement the note two paragraphs up asks for is unaffected.

src/ccl/infer/solver/compact.rs
dpmills: Fixed in the same commit. The doc now says both halves are *set* at a function's domain from its kind rather than passed through, and `Position::domain_of` carries the second point: `resolved()` answers `Unpinned` for a kind variable compaction has not narrowed, so an undecided kind's domain is read like a compute one — the same reading `Display` gives it (`⇒`). The rule reaches only kinds pinned by compaction time, and that is now said out loud.

src/ccl/infer/solver/compact.rs
dpmills: Not a second instance — the candidate position is not covariant either, it is just not the thing the flag names.

`invariant` says one thing: the walk is inside the domain of a data function. The `Fun` arm sets it from that function's kind and nothing else sets it. A candidate is a type in a kind's list, not a function's domain, so `false` is what the position is rather than a gap. Where a candidate's interior contains a data function, the `Fun` arm sets the flag on that function's domain on the way down, so nothing is lost by not setting it at the candidate.

I first read `box_intro`'s comment ("`δ` occurs among the candidates, which is an *invariant* position") and `candidate_in_kind`'s ("a candidate *is* a domain and data domains are invariant") as licensing the flag here, and tried it. Both are constraint-time rules about *membership* — a candidate matches by type equality rather than by a subtyping edge — which is a different mechanism from which side compaction reads at a position. Reverted.

On whether the two readings could still disagree unobserved: `data_domains_disagree` has one caller, `CompactFun::merge`, which reads it on two functions' `domain` slots and only when the merged kind is data. A candidate lands in `CompactTypeKind::Enumerated` and is merged by `CompactTypeKind::merge`, so it never reaches that check. The failure this PR fixes has no consumer on the candidate path.

You are right that the adjacent comment argues covariance of the candidate *set*, which is not a claim about a candidate. I have left it as is rather than rewriting it in this PR — say the word if you would rather it said which question it is answering.

src/ccl/design/type-inference.md
dpmills: Fixed in `Keep the general merge notes under the section they belong to`. The three paragraphs move above the `#####`, so the subsection holds only the rule it names.

`Binder slots` needed the same attention and you did not have to name it: the new heading had swallowed it too, and everything from there to the end of the section with it. It is now a `####` of its own, which is the level its role as one of the enclosing `###`'s blind spots asks for — the sibling of `The collapse happens at the position` rather than a tail of it.

tests/compilation_pipeline/type_annotations.rs
dpmills: Fixed — the expectation is a raw string with `Σ` and `σ` literal.

Both programs moved too, from `\n`-escaped one-liners to the multi-line string form the rest of the file uses. CLAUDE.md rules out the escaped form for embedded source, and these two were new in this PR.

PR #174, opened 2026-09-11
Add `@LoadFrom` so a reload can seed a new variable from a retired one

dpmills:
A reload refuses any version that stops declaring a mutable variable, because the value it holds would be dropped — so renaming one, or changing the units it is kept in, costs the state. `@LoadFrom(x)` decorates a declaration and binds it to the value the predecessor held for `x`, read once at the swap, which is where a migration goes. Retiring, renaming and transforming persisted state become edits a running program survives, and the guard gains the other direction: a load naming a variable nothing holds is refused while the program is still serving. The language rule is [chl-spec.md](docs/chl-spec.md#88-loadfrom); the mechanism is [hot-reload.md](src/ccl/design/hot-reload.md#seeding-a-variable-from-the-value-the-predecessor-held).

## The leaf

`@LoadFrom(x)` and the declaration under it parse as one statement (`Stmt::LoadFrom`), which is what keeps a bare `y: T` a parse error everywhere else, and lower to a `let` bound to `TypedExprNode::Carried` — a leaf holding the source's own spelling. Same shape as `Source`, and for the same reason: the name addresses something the compilation is handed. `scope.rs` walks it as a leaf rather than a `VarRef`, so `uniquify` leaves the spelling alone; inference gives it a fresh variable that the declaration's annotation pins.

## One walk answers which variable a name means

`state_identities` replaces `mutable_variables` and the deleted `declared_state`, assigning declarations and `carried` sites their addresses in one pass so the two cannot disagree about the binding chain at a point. A site resolves outward from its own chain, the way a name resolves in the source; the binding a load seeds pushes no segment, or the innermost candidate would be one no source can name.

## The guard, both ways round

`state_conflicts` now also reports `NoPredecessor`, `Undecided` (declared, but the store was never driven, so it holds nothing) and `CarriedAt` (read at a shape the program does not hold it at). A variable a load takes over is no longer `Dropped`. `LiveProgram::reload` renders the two directions as separate paragraphs, since the remedy for one says nothing about the other.

## Consequences

A loaded collection is a collection: op-conversion builds it with the new `Constant::collection`, which tiles a bindings table as a sealed function, so a comprehension over the binding iterates it and releases it key by key. Tiling is not readable off a `Value`, which is why the call site states it; `src/interpreter/design-operators.md` records both tilings.

A store seeded from a load resumes rather than restarts. `unrecomputable_nodes` generalizes to `nodes_reaching`, `carried_derived_nodes` is the second leaf it is asked about, and both store builders read it into `continues` alongside the variable's own identity.

## Tests

20 cases in `tests/hot_reload/cases.rs`, with the accept and refuse tables in `tests/hot_reload.rs` extended; 7 in `tests/compilation_pipeline/transactions.rs` for the grammar, the placement rules and the cold start; 4 in `src/chl_parser/parser.rs`.


docs/chl-spec.md
```diff
@@ -3126,6 +3138,82 @@ Supporting decisions (same source):
...
+
+**It is a snapshot, not a read.** The value is what the predecessor held at the
+swap, and is a constant from then on — not a dependency on `x` going forward. It
+is therefore not a read of a transactional variable and needs no `with begin():`
+block ([8.3 Reads](#83-reads)).
+
+**A variable may be declared, loaded, or both.** A variable the new version
+declares takes the value it held whether or not anything names it. `@LoadFrom` is
+for the variable it declares under a different name, or retires. Doing both keeps
+the old variable live while seeding a new one from it.
```
groundlar: This section is unclear--what is "declaring without naming"? 

IIUC "both" means e.g. preserving the previous version's declaration _and_ using @LoadFrom, which "seeds" the new one--does "seeding" mean copy semantics?

src/ccl/design/hot-reload.md
```diff
@@ -273,6 +281,82 @@ Dropping the handle is what closes it — `SharedHttpServer` holds the listener
...
+so what makes a unit change on persisted state a declaration.
+
+### Which variable a site addresses
+
+`state_identities` assigns declarations and load sites their addresses in one walk, so the two
+cannot disagree about what the binding chain at a point is. A declaration's address is the chain
+enclosing it (`VarPath`); a site's is read off the same chain and then resolved outward, the way a
+name resolves in the source — the innermost enclosing chain holding a variable of that spelling
+wins. The search starts at the site's chain and does not descend, so a load inside a stateful
+function's body finds the ``a`.`total`` of its own instantiation while the same spelling at the top
```
groundlar: What does "starts at the site's chain and does not descend" mean: "looks at enclosing scopes from  where it's encountered"?

src/ccl/design/hot-reload.md
```diff
@@ -273,6 +281,82 @@ Dropping the handle is what closes it — `SharedHttpServer` holds the listener
...
+level reaches nothing. A diagnostic names the spelling alone, which is what the source contains.
+
+The binding a load seeds is not on the chain. Nothing is ever declared under that binding, and
+pushing its name would make the innermost candidate one no source can name: a predecessor declaring
+the loaded spelling inside an instantiation bound to the load's own target name.
```
groundlar: All the md docs could use a definition clarification pass--"binding a load seeds", "loaded spelling", "load's own target name",  etc are all really easy to confuse. Proposal:

```
@LoadFrom(x) # x is the `loaded var` or `From` var
x_new : Int # x_new is the `declared var` or `To` var, alt "target var" (which is used elsewhere in the doc)
```

Acute concern here: I don't follow "the binding a load seeds...Nothing is ever declared under that binding". Does this mean "nothing is ever declared as `x_new`? The load annotates a declaration. Or maybe this means "nothing is ever declared as `x`", but that's allowable under the "both live and seeded" semantics.

src/ccl/design/hot-reload.md
```diff
@@ -273,6 +281,82 @@ Dropping the handle is what closes it — `SharedHttpServer` holds the listener
...
+
+Resolution reads what the retired version **declared**, not what it currently holds a value for. A
+store the running program never drove holds no value while still being the variable the name means,
+so answering from the values would report such a name as addressing nothing and refuse the version
+for dropping a variable it says where to put.
+
+### A loaded value summarizes positions
+
+Whether a rebuilt store resumes above the positions its seed summarizes or begins at its input is
+`continues`, which each store builder hands to `OpConversionContext::iteration_input`. It is a
```
groundlar, line 367: "positions" means... lexical position? Time domain position?

What is "summarizes" here?

docs/chl-spec.md
```diff
@@ -3126,6 +3138,82 @@ Supporting decisions (same source):
...
+A comprehension over a map binds each value and keeps the keys, so that scales
+every quantity the predecessor held.
+
+**A running program is required.** A source containing `@LoadFrom(x)` is an
+upgrade of a specific predecessor: compiled from nothing, it is an error naming
+`x`, and so is one naming a variable the running program does not hold. There is
+no `@LoadFrom(x, default)` — a default would turn that error back into a silent
+wrong answer.
+
+**It is transitional.** The seeding happens once, so the decorator comes out in
```
groundlar: Two things to soften here, both about durability.

The rule as written is a language rule: `@LoadFrom` is only meaningful against a *live process*. But `docs/design.md` lists durable state as `[Sketched]`, and a durable deploy is precisely "compiled from nothing while holding state" — a process restart reading its store from disk. That is the case `@LoadFrom` is the natural migration primitive for, and this paragraph forbids it there.

Suggest phrasing the constraint as current rather than permanent, e.g. "A running program is **currently** required", plus a sentence noting that the interaction with durable state is TBD and will be settled in its own design phase. Same for the parallel sentence in `hot-reload.md` ("cannot be redeployed into a fresh environment, a new region, or CI"), which reads as a permanent property of the feature rather than of today's in-process `Inheritance`.

Nothing to change in the implementation — just not baking the current mechanism's limits into the spec.

src/interpreter/operator_conversion.rs
```diff
@@ -1042,21 +1071,78 @@ in it has a correspondent",
...
+                (self.declared_extent(&path), self.extent_of(&site.ty))
+                && read_at != declared
+            {
+                out.push(StateConflict::LoadFromAt {
+                    path: path.clone(),
+                    held: declared,
+                    read: read_at,
+                });
+            }
+            taken_over.insert(path);
```
groundlar: **Two anonymous load sites can be silently swapped — there is no `site_moved` analogue here.**

`LoadFromSite::resolve` pairs a site with a declaration **by position** (`index`), and `LoadFromSite` — unlike `MutableVariable` — carries no `site: ContentHash`. Meanwhile, a variable that is retired *and* loaded lands in `taken_over` and so never reaches the `Some(decl)` branch below where `site_moved` runs. Net effect: reordering two anonymous call sites of one stateful function, in the version that loads them, hands each site its neighbour's value and is **accepted**.

This is exactly the failure `StateConflict::Moved` exists to refuse. The PR's own test acknowledges the hazard in prose — "Addressing them both to the first declaration is accepted rather than refused, so the program carries on answering with one site holding the other's value" — without testing the reorder.

Reproduced. This is `a_load_inside_a_duplicated_body_seeds_from_its_own_call_site` with the two call sites swapped:

```rust
#[test]
fn a_reordered_call_site_under_a_load_is_refused() {
    let v1 = indoc! {r#"
        def count_by(items, step) => Int:
            total := 0
            for x in items:
                total := total + step
            total

        count_by([1, 2], 3) * 1000 + count_by([1, 2, 3], 5)
    "#};
    // The same two call sites, swapped, with `total` retired and loaded.
    let v2 = indoc! {r#"
        def count_by(items, step) => Int:
            @LoadFrom(total)
            held: Int
            scaled := held * 10
            for x in items:
                scaled := scaled + step
            scaled

        count_by([1, 2, 3], 5) + count_by([1, 2], 3) * 1000
    "#};
    let mut ctx = GlobalContext::default();
    let mut live = LiveProgram::start(&mut ctx, v1, &no_main).expect("v1 compiles");
    assert_eq!(drive_main_int(&mut ctx, &mut live), 6015);
    live.reload(&mut ctx, v2, &no_main)
        .expect_err("neither site's value belongs to the position it now sits at");
}
```

Actual: accepted, and `drive_main_int` returns **150060** — each site holding the other's value (site-correct is `60150`). For comparison, the equivalent reorder of plain *declarations* is refused with a named diagnostic.

The structural fix mirrors the declaration side: give `LoadFromSite` the content hash of the term it sits in and run a `site_moved`-equivalent before accepting a positional resolution. If that is too large for this PR, the minimum is to refuse `index > 0` resolutions outright — they are only reachable through duplicated instantiations — and say why. Silently swapping persisted state between two addresses is the one outcome this guard's whole rationale says must not happen.

src/interpreter/operator_conversion.rs
```diff
@@ -3129,25 +3393,123 @@ fn mutable_variables(expr: &Expr) -> Vec<MutableVariable<'_>> {
...
+    /// that declares it.
+    ///
+    /// Handed to conversion so that a store is built under the identities the
+    /// guard checked, rather than under a second derivation of them.
+    /// Every store, as its declaring `Transact` with the variables it declares.
+    ///
+    /// Keyed by the declaration rather than by the variable because the questions asked
+    /// of a store — whether it continues a recurrence, where its drive resumes — are
+    /// answered for all of its keys at once.
+    fn stores(&self) -> HashMap<NodeId, Vec<&MutableVariable<'e>>> {
```
groundlar: This doc comment has two stacked lead paragraphs: it opens with the old `mutable_variable_paths` summary ("Every mutable variable's identity, in key order…" / "Handed to conversion so that a store is built under…") and then starts again with its own ("Every store, as its declaring `Transact`…").

The first two paragraphs belong on `variable_paths()` just below, which is currently undocumented. Per the prose rules the conclusion is the first sentence, and there are two of them here.

src/interpreter/operator_conversion.rs
```diff
@@ -3129,25 +3393,123 @@ fn mutable_variables(expr: &Expr) -> Vec<MutableVariable<'_>> {
...
+        out
+    }
+
+    fn variable_paths(&self) -> HashMap<NodeId, Vec<VarPath>> {
+        let mut out: HashMap<NodeId, Vec<VarPath>> = HashMap::new();
+        for v in &self.variables {
+            out.entry(v.declared_by).or_default().push(v.path.clone());
+        }
+        out
+    }
```
groundlar, line 3478: `variable_paths` is `stores` with `.path.clone()` — the same grouping loop written twice.

Flagging per the "recurring shapes" rule rather than asking for a refactor in this PR. The question worth answering is whether `OpConversionContext::var_paths` should hold the `MutableVariable`s and let call sites project, which would remove both the second loop and the second field.

src/interpreter/tile_operators/scalar.rs
```diff
@@ -77,6 +125,12 @@ struct ConstantProducer {
     base: ProducerBase,
     value: Value,
     released: bool,
+    /// The collection tile, built on the first pull that asks for one.
+    ///
+    /// A constant's bindings table cannot change, so deriving the tile per pull
+    /// re-clones every key and value to reach the same answer. Only the obsolete
+    /// guard varies, and that applies to the copy handed out.
+    table: Option<Tile>,
```
groundlar, line 135: The justification overstates what the cache saves. `get_impl` still does `table.clone()` on every pull, and `ColumnValue` is `Vec`-backed (`Ints(Vec<i64>)`, `Strings(Vec<SmolStr>)`, …), so every key and value *is* re-cloned per pull.

What the cache actually saves is the per-pull `Value` → `ColumnValue` unzip and `from_values` conversion. Worth saying that instead, since this comment is the whole justification for the field.

src/interpreter/operator_conversion.rs
```diff
@@ -3025,16 +3150,47 @@ fn writer_sources(expr: &Expr) -> HashMap<NodeId, Vec<&Expr>> {
...
+    nodes_reaching(expr, |node| matches!(node, TypedExprNode::LoadFrom(_)))
+}
+
+/// Every node of `expr` whose value comes from a leaf `is_source` accepts.
+///
+/// One walk for the two questions asked of a term this way — whether it reads a
+/// data source, and whether it reads a loaded value — because they differ only
+/// in the leaf. What makes them the same walk is the propagation: a `Var` answers
+/// for what its binding was bound to, so the property reaches a use through every
+/// binding between, and a term's answer is otherwise its children's.
```
groundlar: "One walk" reads as if the two questions share a traversal. They share a walk *function* but are two separate traversals: `iteration_input` now runs `state_identities`, `unrecomputable_nodes` and `load_from_derived_nodes` over `planned` as three passes, where it previously ran two.

Either say "one walk shape, two traversals", or actually fuse them — the propagation rule is identical, so a single pass computing both sets is a small change.

tests/compilation_pipeline/transactions.rs
```diff
@@ -1285,6 +1285,133 @@ fn await_final_of_an_induction_accumulator_rejected() {
...
+            m: Mut(Int, Txn) := 0
+            for k in [1]:
+                @LoadFrom(n)
+                held: Int
+                with begin():
+                    m := m + held
+            await_final(m)
+        "#},
+        "only assignments",
+    );
```
groundlar, line 1375: Both this and `load_in_a_transaction_block_rejected` assert on the pre-existing catch-all messages in `lower/loops.rs` and `lower/transactions.rs`, neither of which mentions `@LoadFrom`. The behaviour is right, but an author who writes a load in a loop body gets back a list of supported constructs that does not name the one they wrote.

Low cost to add it to both messages, and it would let these tests assert on something specific to the case they cover.

tests/hot_reload/cases.rs
```diff
@@ -1004,6 +1004,991 @@ fn a_reload_may_not_change_the_type_of_held_state() {
...
+    );
+
+    let still_serving = exchange(&mut ctx, move || vec![http_get(port, "/get")]);
+    assert_eq!(still_serving, vec!["bob"], "state intact");
+}
+
+/// Reading a loaded value at a type the running program does not hold it at is
+/// refused, by the comparison a declaration gets.
+#[test]
+fn a_reload_may_not_load_a_variable_at_another_type() {
```
groundlar, line 1142: **Both refusal tests run off the scalar `latest-write` base**, so the collection path is pinned only on the accept side.

`a_reload_may_not_load_a_variable_nothing_holds` uses `LATEST_WRITE_LOADS_A_STRANGER` and this one `LATEST_WRITE_MIGRATED_RETYPED`, both `String`. The collection cases — `a_loaded_collection_is_transformed_into_its_replacement`, `a_loaded_record_valued_collection_is_transformed_into_its_replacement` — only cover acceptance.

I checked all four collection refusals and the implementation is correct today, so this is a coverage gap rather than a bug. Against the `qty: Mut(Map(String, Int), Txn)` base:

| Change | Diagnostic |
|---|---|
| `@LoadFrom(nonesuch)` | `` `@LoadFrom(nonesuch)` has no variable to read `` |
| → `Map(String, String)` | `` reads `qty` as (String -> String) where the running program holds it as (String -> Int) `` |
| → `Map(Int, Int)` | `` (Int -> Int) … (String -> Int) `` |
| `Map(String, {units, lot})` → `Map(String, {units})` | `` (String -> {units: Int}) … (String -> {lot: Int, units: Int}) `` |

The load-bearing property none of these pins: **the bounded `<:` mode does not weaken the shape check.** Every probe above used `held <: …` and was still refused, because `LoadFromAt` compares runtime `Extent`s rather than CCL types. That is the thing most likely to regress silently, and "changing the units it is kept in" is the feature's headline motivation.

Suggest two cases off the existing `qty` fixture — one `NoPredecessor` on a collection, one `LoadFromAt` on a collection's codomain under `<:` — plus the two rows in `hot_reload.rs`'s refuse table.

groundlar (commented):
**PR description carries stale names.** The body still says `TypedExprNode::Carried`, "assigning declarations and `carried` sites their addresses", and `carried_derived_nodes` — all renamed to `LoadFrom` / `load_from_derived_nodes` by 49fcf01b. This repo squash-merges, so the body becomes the commit message on `main` and those names will ship wrong.

docs/chl-spec.md
dpmills: "Whether or not anything names it" meant whether or not a `@LoadFrom` names it, which the sentence never said. And yes — seeding is a copy, taken once at the swap, as "It is a snapshot, not a read" above says of the value. Rewritten so both are on the page:

> **A variable may be declared, loaded, or both.** A variable the new version declares carries its value forward whether or not a `@LoadFrom` names it, which is the ordinary reload. The decorator is for the variable the new version declares under a different name, or retires. A version doing both — declaring `x` and loading from `x` — leaves `x` running on its own value and starts the new variable at a copy of it taken at the swap. The two are separate variables from then on, and neither reads the other.

src/ccl/design/hot-reload.md
dpmills: Yes — enclosing scopes only, from where the site sits. The doc now says which chains are excluded rather than naming the direction:

> Outward means the site's own chain and the chains enclosing it, and never a chain below one of those: a load inside a stateful function's body finds the ``a`.`total`` of its own instantiation, while the same spelling written at the top level reaches nothing, ``a`.`total`` sitting under a binding the top level's chain does not contain.

src/ccl/design/hot-reload.md
dpmills: Definitions added at the top of the section, using the two names the doc already had rather than a third pair:

```python
@LoadFrom(qty)      # `qty` is the loaded spelling
qty_units: Int      # `qty_units` is the target
```

On the acute concern: it meant nothing is declared under **the target**, and the reason is mechanical rather than a rule. A chain segment comes from descending into a binding's *definition*, and a load's definition is the leaf alone — a `LoadFrom` node has no children — so a segment for the target would address nothing. Your second reading is right that declaring `x` again is allowed; that is the "both" case, and it is unaffected, since the target is `x_new`. Rewritten:

> The target contributes no chain segment. A segment comes from descending into a binding's definition, and a load's definition is the leaf alone, so nothing is ever declared under the target and a segment for it would address nothing. Pushing it would also move the search's innermost candidate onto a variable no source can mean: a predecessor that had declared the loaded spelling inside an instantiation bound to the target's own spelling.

src/ccl/design/hot-reload.md
dpmills: Neither, quite: a position is an index into the sequence the store folds — an iteration item or a commit — and it is absolute, so it is not a row offset either. Both facts were in the document ("Positions are absolute, in the store and in the drive alike", and "a store's value at position 𝑝 summarizes every position below 𝑝") but not where this section leans on them. "Summarizes" is that second one: the value at 𝑝 is the fold of every position below it. The section now opens with both and links the first.

docs/chl-spec.md
dpmills: Softened in both, and you are right about which fact the paragraph was baking in.

`docs/chl-spec.md` now heads the rule "A running program is required today" and adds:

> A live process is the only predecessor a version can be handed state from today. Durable state is **[Tentative]** ([design.md](docs/design.md)), and a process restarting from a store on disk is the same migration against a predecessor that is not running; which predecessors a load may name is settled with durable state rather than here.

`hot-reload.md` says the same against the mechanism — the predecessor today is the in-process `Inheritance`, which is what stops a redeploy into a fresh environment, and a durable store would be a second kind of predecessor. No implementation change.

src/interpreter/operator_conversion.rs
dpmills: Reproduced exactly as reported — accepted, 150060 against the site-correct 60150.

The suggested fix does not work as stated, and why it does not is what settles the remedy. A content hash on the load site can never match: a `@LoadFrom` edits the body of every site it sits in, so no site in the loading version hashes equal to one in the version it replaces. That is `site_moved`'s second documented limit — "a reorder that also edits both bodies leaves neither site present, so it reads as two edits and the position decides" — reached by construction rather than by coincidence. The limit's rationale is that hitting it takes an author doing two things at once, and a load cannot lean on that.

So the fallback is the rule. `LoadFromSite::resolve` returns a `LoadTarget`, and a chain declaring the spelling more than once answers `Anonymous` rather than picking a position. `StateConflict::LoadFromAnonymous` refuses it with the remedy `Moved` already gives — bind each call site to a name, which `a_load_inside_an_instantiation_reaches_its_own_variable` shows resolving. `LoadFromSite::index` and the walk's second counter are gone with the resolution that read them, so `resolve` is now four lines: find the innermost chain declaring the spelling, and answer `Anonymous` if it declares it twice.

What it costs: `a_load_inside_a_duplicated_body_seeds_from_its_own_call_site` was an accept test and is now `a_load_inside_a_duplicated_body_is_refused`, sitting beside `a_reordered_call_site_under_a_load_is_refused` — your program, verbatim. A `@LoadFrom` inside a `def` called more than once is refused until the call sites are named. Stated in `docs/chl-spec.md`, in `hot-reload.md`'s guard inventory and "Which variable a site addresses", and in the refuse table in `tests/hot_reload.rs`.

src/interpreter/operator_conversion.rs
dpmills: Split. `stores()` keeps its own two paragraphs, and the `mutable_variable_paths` summary moved onto `variable_paths()`, which was undocumented.

src/interpreter/operator_conversion.rs
dpmills: Collapsed — `variable_paths` projects from `stores()` rather than repeating the grouping loop.

On the field: `MutableVariable<'e>` borrows the tree the walk ran over, and `OpConversionContext` outlives that tree, so holding them would need an owned form of the struct. That is a change worth its own diff, and the projection removes the duplicated loop without it. The second field stays.

src/interpreter/tile_operators/scalar.rs
dpmills: Correct, and the comment was the whole justification, so it is now what the cache actually saves:

> What holding it saves is the `Value` to `ColumnValue` conversion: the unzip into keys and values, and a `from_values` over each. The tile is still cloned per pull, so the keys and values are re-cloned either way. A constant's bindings table cannot change, and the obsolete guard that does applies to the copy handed out.

src/interpreter/operator_conversion.rs
dpmills: Fused. `nodes_reaching` is generic over the number of leaves — `nodes_reaching(expr, [reads_a_source, is_a_load])` returns one set per leaf off one traversal, and each node carries an `[bool; N]` through the same propagation. Both call sites destructure it, so `iteration_input` and `set_var_paths` each run two walks now rather than three.

`unrecomputable_nodes` and `load_from_derived_nodes` are gone; their doc bodies moved onto the leaf predicates, which is where the difference between them lives. The prose links that pointed at `unrecomputable_nodes` for the per-term question now point at `OpConversionContext::recomputable`, which is the thing that answers it.

tests/compilation_pipeline/transactions.rs
dpmills: Added to both. `lower/loops.rs` now ends with "A declaration goes above the loop, `@LoadFrom` included: its value is a snapshot, so a per-iteration one would be the same constant every time", and `lower/transactions.rs` with "A declaration goes above the block, `@LoadFrom` included: its value is the predecessor's, read once at the swap". Both tests assert on that clause instead of on `"only assignments"`.

tests/hot_reload/cases.rs
dpmills: Added, off the `qty` fixture as suggested:

- `a_collection_load_naming_nothing_is_refused` — `@LoadFrom(nonesuch)` against `qty: Mut(Map(String, Int), Txn)`, refused naming the spelling, collection intact afterwards.
- `a_bounded_annotation_does_not_weaken_a_collections_shape_check` — `held <: Map(String, String)` against the same base, with a comprehension that typechecks at `String`. Refused by `LoadFromAt` on the extent, which is the property you named: the bounded mode leaves the domain to be inferred and weakens no shape check.

Two rows added to `tests/hot_reload.rs`'s refuse table — the collection row, and the anonymous-call-site row from the thread above.

src/ccl/design/hot-reload.md
```diff
@@ -310,22 +318,38 @@ so what makes a unit change on persisted state a declaration.
...
-pushing its name would make the innermost candidate one no source can name: a predecessor declaring
-the loaded spelling inside an instantiation bound to the load's own target name.
-
-Declarations sharing a chain and a spelling are told apart by index, which is two anonymous call
-sites of one stateful function. Inlining copies a function body into each call site, so a site
-carries an index from the same walk and addresses the declaration at its own position. A scope
-declaring the spelling once has nothing to tell apart and answers whichever site asks.
+wins. Outward means the site's own chain and the chains enclosing it, and never a chain below one of
+those: a load inside a stateful function's body finds the ``a`.`total`` of its own instantiation,
+while the same spelling written at the top level reaches nothing, ``a`.`total`` sitting under a
```
groundlar: nit: These don't render as expected, maybe sub single quotes for the inner backticks?

src/ccl/design/hot-reload.md
```diff
@@ -406,7 +520,10 @@ Three things decide where a rebuilt store picks up, and only two of them are han
   stateful function in one expression. An edit to either body leaves it alone, and a reload that
   would move a value between them is refused (`swapping_two_anonymous_call_sites_is_refused`). What
   is left unaddressed is narrower: a reorder that edits both bodies at once leaves neither site
-  present to recognise, so the position decides and the values cross.
+  present to recognise, so the position decides and the values cross. A `@LoadFrom` reaches that
+  shape by construction, editing the body of every site it sits in, which is why a load is refused
+  wherever the chain declares its spelling more than once
+  ([Which variable a site addresses](#which-variable-a-site-addresses)).
```
groundlar: **Stale names left by the `nodes_reaching` fusion — one in this file, the rest in the PR description.**

**In this document, line 630** (outside the diff, so this thread is the nearest anchor):

> What separates a term that can be built again from one that cannot is whether it reads a source at all, which is what `unrecomputable_nodes` answers.

`unrecomputable_nodes` no longer exists — `grep -rn unrecomputable_nodes src/ --include=*.rs` is empty after this commit. The Rust doc comments that cited it were repointed at `OpConversionContext::recomputable`, but this prose reference was missed. `./ci.sh doc_refs` does not catch it: it validates `<name>.md` citations and Markdown anchors, not Rust identifiers named in prose. The sentence is about the per-term question, so `recomputable` is the right target here too.

**In the PR description.** This repo squash-merges, so the body becomes the commit message on `main`. It still carries the pre-`49fcf01b` spelling throughout:

- `` `TypedExprNode::Carried` `` → `TypedExprNode::LoadFrom`
- "assigning declarations and `carried` sites their addresses" → load sites
- `` `CarriedAt` `` → `StateConflict::LoadFromAt`
- `` `carried_derived_nodes` is the second leaf it is asked about `` → gone entirely

And this commit made two more paragraphs wrong:

- "`unrecomputable_nodes` generalizes to `nodes_reaching`" — neither `unrecomputable_nodes` nor `load_from_derived_nodes` survives; `nodes_reaching` is now const-generic over a leaf array, with `reads_a_source` and `is_a_load` as the predicates.
- `StateConflict::LoadFromAnonymous` is absent from the body. That is the most consequential behaviour change in the PR — a `@LoadFrom` inside a `def` called more than once is now refused until the call sites are named — and the squash message would not mention it.

Worth a pass over the body before merge; `.claude/skills/pr-description/SKILL.md` has the `gh api` workaround for editing a posted title and body.

groundlar (approved):
Approving. Every thread from the last round is addressed, and the fix for the load-site aliasing is better than what I proposed — the content-hash approach genuinely cannot work here, and refusing an ambiguous chain outright is the right generalization. I verified the stated remedy: with both call sites named, the reload resolves and returns the site-correct value.

`./ci.sh` is green on `b4b1b072` and `hot_reload` passes 96/96.

Two follow-ups, neither blocking: the stale names in the thread above, and a `debug_assert!` on the index-density invariant `LoadFromSite::resolve` now depends on — it reads index 0 as "declared at all" and index 1 as "declared more than once", which holds today only through `state_identities` assigning densely, `bind_store` recording every key, and `carried_keys` dropping only reply taps. Three functions, none of which states it.

src/interpreter/operator_conversion.rs
```diff
@@ -472,6 +482,19 @@ impl Inheritance {
...
+    /// What a name addresses, as against [`Self::mutable_state`], which is what
+    /// a variable currently holds. The two differ for a store the running
+    /// program never drove: it declares its variables and holds no value for
+    /// them.
+    fn declared_paths(&self) -> HashSet<VarPath> {
+        self.stores()
+            .flat_map(StoreReadInfo::carried_keys)
+            .map(|(path, _)| path.clone())
+            .collect()
+    }
```
groundlar: opt: a `debug_assert!` here would pin the index-density invariant `LoadFromSite::resolve` depends on.

`resolve` reads index 0 as "this chain declares the spelling at all" and index 1 as "it declares it more than once". Both readings need the indices dense from 0, and this is the single producer both readers — `state_conflicts` and `load_from_values` — take their `declared` set from.

Density holds today, but through three functions and none of them says so: `state_identities` assigns it (`counts.entry(…).or_insert(0)` then `+= 1`); `bind_store` records every key with `carried: Some(paths[i])`, and `compile_let_binding` calls it unconditionally, so a store under a `let` nothing reads is still built; and `carried_keys` drops only `carried: None`, which is the reply tap alone — a `__to_<base>_<n>` spelling no declaration shares.

What a gap would cost is silent rather than noisy. Given a top-level `total` and two anonymous `count_by` call sites inside one named binding:

| Declaration | `VarPath` |
|---|---|
| top-level `total` | `{chain: [], name: "total", index: 0}` |
| first call site | `{chain: ["a"], name: "total", index: 0}` |
| second call site | `{chain: ["a"], name: "total", index: 1}` |

A `@LoadFrom(total)` in the function body resolves at chain `["a"]` to `Anonymous`, correctly. Drop `["a"].total#0` from this set and `at(0)` misses, the search continues outward, and `[].total#0` — a different variable — answers with no diagnostic.

```rust
fn declared_paths(&self) -> HashSet<VarPath> {
    let out: HashSet<VarPath> = self
        .stores()
        .flat_map(StoreReadInfo::carried_keys)
        .map(|(path, _)| path.clone())
        .collect();
    // `LoadFromSite::resolve` reads index 0 as "this chain declares the spelling"
    // and index 1 as "it declares it more than once". Both readings need the
    // indices dense from 0: `state_identities` assigns them that way, and only a
    // reply tap is dropped on the way here (`carried: None`), whose spelling no
    // declaration shares. A gap would make a load resolve outward past the
    // variable it means.
    #[cfg(debug_assertions)]
    {
        let mut by_spelling: HashMap<(&[String], &str), Vec<usize>> = HashMap::new();
        for p in &out {
            by_spelling
                .entry((p.chain.as_slice(), p.name.as_str()))
                .or_default()
                .push(p.index);
        }
        for ((chain, name), mut indices) in by_spelling {
            indices.sort_unstable();
            debug_assert!(
                indices.iter().copied().eq(0..indices.len()),
                "declared indices for {chain:?}`{name}` are {indices:?}, not dense from 0",
            );
        }
    }
    out
}
```

`by_spelling` borrows `out` and drops at the block's end, so the return is unaffected. Optional — I found no reachable violation today; this guards the drift, e.g. `carried: None` acquiring a second meaning or a `Transact` ever becoming conditionally converted.

src/ccl/design/hot-reload.md
dpmills: Fixed, though not with single quotes — the guard really does print backticks, so they belong in the rendering.

The span was mis-delimited. ``` ``a`.`total`` ``` opens on a run of two backticks and closes on the next run of two, so the content is `a`.`total` *minus* its outer pair, and `VarPath`'s `Display` prints those (`write!(f, "`{binding}`.")` then `write!(f, "`{}`", self.name)`). It now takes the padded form this file already used one paragraph down in the guard-diagnostic sentence: `` `a`.`total` ``, whose content is the address exactly as a diagnostic spells it.

Seven spans had it, so the class is converted rather than the two you anchored on: `hot-reload.md` ×2, `tests/hot_reload/cases.rs` ×3, and `operator_conversion.rs` ×2 (the comment on `VarPath`'s `Display`, and `LoadFromSite::resolve`'s doc).

src/ccl/design/hot-reload.md
dpmills: **The index-density `debug_assert!` is in**, on `Inheritance::declared_paths` (043f99e1).

You named the three functions exactly: `state_identities` numbers a chain's declarations from `0` as it meets them, `bind_store` gives every key that walk addressed a `KeyReadInfo` (`carried: Some(paths[i])`), and `carried_keys` drops only reply taps, which the walk addresses none of — a tap's key is appended past `keys.len()` and carries `None` by construction.

`declared_paths` is where the three meet: it builds the `HashSet<VarPath>` that both readers of `resolve` — the guard and `load_from_values` — are handed, so one assertion covers both.

```rust
debug_assert!(
    paths.iter().all(|p| p.index == 0
        || paths.contains(&VarPath {
            index: p.index - 1,
            ..p.clone()
        })),
    "declaration indices run dense per chain and spelling"
);
```

`debug_assert!` rather than `assert!` per CLAUDE.md, "Code Comments" — the check is a pass over the set, not O(1). The invariant is stated in `declared_paths`'s doc comment naming all three sources, and `resolve`'s inline comment now cites where it is asserted rather than only claiming it.

dpmills: **Both halves done** (cffe100f).

**In the document.** `hot-reload.md` line 630 now points at `OpConversionContext::recomputable` — "Whether a fresh compilation of `term` reproduces the value it holds" — which is the per-term question the sentence asks. A repo-wide grep for `unrecomputable_nodes`, `load_from_derived_nodes` and `carried_derived_nodes` is empty.

**In the body.** Rewritten from the diff rather than patched, so the pre-`49fcf01b` spellings are gone along with the two paragraphs this round made wrong. `StateConflict::LoadFromAnonymous` now has the load-bearing sentence in "What the guard refuses", quoting the diagnostic and the remedy, so the squash message carries it. Every count in "Tests" was re-derived from the diff, and three of them were wrong independently of the rename — the old body said 20 reload cases and 7 in `transactions.rs` where the diff adds **27** and **8**.

The title is unchanged: it is imperative and names what is added, and nothing in it went stale.

**One more found while checking.** `hot-reload.md`, "The one guard: a version must be able to take over the state" opened with "Five things are refused" and listed six — the count went 3 → 5 in this PR while three bullets were added. Now six, which is the six the bullets describe (`Retyped` and `LoadFromAt` share the second one, so seven `StateConflict` variants land in six bullets). Fixed in the same commit.

sortalongo:
Post-merge review. None of this contradicts what shipped; it is all follow-up. Skylar's open threads are left out, as are the items his round already fixed. Line numbers are on the merge commit.

## Semantics

**`continues` ignores the source's domain.** `src/interpreter/operator_conversion.rs:2827` sets `continues` for any store whose seed derives from a load, without checking that the loaded variable's domain matches the target's. A `Mut(V, Txn)` source seeding a loop-carried `Mut(V, extent)` target then resumes above positions the loaded value never summarized. Every `@LoadFrom` test that mentions `Txn` is `Txn` to `Txn`, so nothing covers the crossing. Refusing a domain change in `state_conflicts` would close it.

A case that would show it: v1 declares `total: Mut(Int, Txn) := 0` and commits one write of 5; v2 loads it as `t: Int` and folds `+` over `[1, 2, 3]` starting from `t`. The answer should be 11, and 5 if `continues` suppresses the fold.

**`Undecided` makes a migration depend on whether the predecessor read the variable.** `src/interpreter/operator_conversion.rs:1147` refuses a load naming a variable the retired version declared but never drove. By the spec that variable holds its initialiser, so whether v2 can load `qty` turns on whether anything in v1 read `qty`, which is not a property of the source. Seeding from the declaration's initialiser and dropping the refusal would make it one.

## Spec

**`@` is still listed as absent, in two places.** `docs/chl-spec.md:197` reads "`~`, `@` (no matmul, no decorators)" and `:722` says attempting to use these is a parse error, while §1.9 at `:230` introduces the decorator. Removing `@` from both lists and adding it to the punctuation block would settle it.

**§8.8 claims the exact annotation form works.** `docs/chl-spec.md:3204` says "`held: T` binds at `T`" with no caveat, and `an_exact_annotation_on_a_loaded_collection_names_what_differs` and `an_exact_annotation_on_a_rebuilt_collection_is_refused` refuse it for collections. A one-line status note naming the test would cover the gap the commit left when it moved the limitation out of the spec.

**§8.8 does not say a seeded store resumes.** `a_loaded_collection_is_transformed_into_its_replacement` pins that the new version's writer does not fire, and only `src/ccl/design/hot-reload.md:359` states the rule. It belongs in §8.8 too, linked to that heading.

**§8.8 does not say whether `x` may name a feed.** `docs/chl-spec.md:3144` says only "the mutable variable `x`". A load naming a feed is refused as `NoPredecessor`, and no test pairs the two.

**§8.8 does not cover an `if` or `match` branch.** `docs/chl-spec.md:3160` names the top level and a `def` body and rejects `for` and `with begin():`. The grammar admits a branch body and no test covers it.

## Runtime failure modes

**`Constant::collection` panics after teardown.** `src/interpreter/tile_operators/scalar.rs:71` panics unless the value is a `Value::Function`, and the guard compares extents only, so a representation mismatch aborts the process after the old graph is gone. That is the failure the pre-teardown guard exists to prevent. Checking the representation in `state_conflicts` would keep it inside the guard.

**A comment says reported where the code panics.** `src/interpreter/commit_operator.rs:334` distinguishes a `Diverged` failure as the reported one, and `:920` and `:1344` both panic on it, eight pulls later and with a "never settled" message. Either a distinct variant naming the field, or a corrected comment.

## Structure

**`Constant` is two operators sharing a struct.** `released` (`src/interpreter/tile_operators/scalar.rs:107`) is meaningful only in the scalar mode and `table` (`:108`) only in the collection mode, and `get_impl` needs two `unreachable!`s to restore an invariant the type does not carry. An enum payload would make the two states exclusive by construction.

**`LoadFromAnonymous` names the spelling, not the candidates.** `src/interpreter/operator_conversion.rs:3679` names the spelling where `Moved` names both paths, and both paths are what tells the author which two sites collided.

**`state_identities` runs four times per reload** — `src/interpreter/operator_conversion.rs:660`, `:1107`, `:1245` — against the design doc's "one walk". This is separate from the traversal fusion Skylar raised, which landed. Computing it once in `reload` and threading it would match the doc.

**A third lowering catch-all was missed.** `src/ccl/lower/loops.rs:511` still reports "only assignments, function definitions, `<<` feeds and `yield`" where `:1273` and the transactions arm got the `@LoadFrom` clause.

## Docs

**`src/ccl/design/hot-reload.md:18` is now false.** The opening bullet says a held variable "must be one the new version still declares, at the same type"; it should add "or loads with `@LoadFrom`".

**The guard table has no load rows.** `src/ccl/design/hot-reload.md:664` has no row for a rename carried by a load, and none for the four new refusals.

**The new section cites no test.** `src/ccl/design/hot-reload.md:289`-`:390` names functions and types where the neighbouring sections each pin a test (`:288`, `:521`), and there are 27 new cases to cite.

**A bare section number.** `src/ccl/design/hot-reload.md:293` cites `[chl-spec §8.8]`; CLAUDE.md, "Referencing docs and sections" asks for the heading text instead.

**Five spellings for the predecessor** across the spec and the design doc: "running program" 16 times, "retired version" 9, "the predecessor" 8, "previous version" 2, "the version this source replaces" 1. One name per concept would help here more than most places, since the whole feature is about the relation between two versions.

Last one, outside the repo: the vault's `upgrade-beat.md` still records the `carried(x)` builtin as the decision. Worth recording `@LoadFrom` as the decision and `carried(x)` as the discarded alternative, with the reason that a statement introduces a name and the syntax should mark that.

Happy to talk through the two semantics items if either is a design question rather than a fix.

PR #175, opened 2026-09-11
Implement type-alias statements by substituting the aliased type at lowering

dpmills:
The spec's [type-alias statements](docs/chl-spec.md#67-type-alias-statements) section described a feature nothing implemented: a capitalized `=` lowered its right-hand side as a term, so `Item = {price: Int, cost: Int}` was rejected as type syntax in value position and `n: Item` was an unresolved name. This change reads a capitalized assignment target as a type-alias declaration, lowers its right-hand side once as a type, and substitutes it at every use. A record or refined type becomes nameable in any annotation, and the statement itself is erased — no phase after lowering knows the name.

## Declaration is a forward pass, scoped to a block

`pre_declare_type_aliases` (`src/ccl/lower/stmts.rs`) walks a block's statements in source order before the block is lowered, recording each alias's lowered `Type` on `LoweringContext::type_aliases`; `lower_type_expr`'s `Name` arm substitutes it at a use. The pass runs forward rather than declaring on reach because a block's statements fold right-to-left (all but the `for`-body walker's), so an annotation below an alias is lowered before it. `with_block_type_aliases` snapshots and restores the table around the four block walkers, which is what makes an alias block-scoped; `lower_stmts_recovering` declares the top-level block's directly and collects every rejected alias rather than the first. Walker inventory and ordering argument: [`src/ccl/design/lowering.md`](src/ccl/design/lowering.md).

## The case of the name is the whole discriminator

`type_alias_decl` reads a capitalized target as a type, whatever sits on the right. Its complement, `extract_name_target`, now rejects a capitalized binder for every other statement form and names the alias as the one form that binds such a name; a `def` name and a parameter stay unchecked. `BUILTIN_TYPE_NAMES` refuses an alias over a built-in, so no name resolves both ways.

## Tests and docs

`tests/compilation_pipeline/type_aliases.rs` adds 28 cases pinning interchangeability with the type an alias names, block exit from `def`, loop and transaction bodies, and every rejection above. The spec section is now specification rather than a tentative sketch, its `[Open]` narrowed to parameterisation, and `docs/design.md` gains a sentence on the feature.


src/ccl/lower/stmts.rs
```diff
@@ -46,6 +46,9 @@ pub(super) fn lower_stmts_recovering(
     // textually. (No transactional-registry snapshot here — the top level is the
     // outermost scope, so nothing to restore to.)
     pre_register_txn_decls(stmts, ctx);
+    // Type aliases are declared in the same forward pass, and for the same
+    // reason: an annotation below an alias is lowered before it.
+    errors.extend(pre_declare_type_aliases(stmts, ctx));
```
groundlar, line 52: **Critical — an alias is invisible to the transactional pre-registration pass.**

`pre_register_txn_decls` on line 48 runs *before* this line, and it lowers type annotations: `mut_annotation_parts` (`stmts.rs:1205`) calls `lower_type_expr` on the `Mut` value type. With the alias table still empty that returns `Err("unknown type annotation: Cents")`, so `matches!(…, Some(Ok((_, true))))` is false and the variable is never registered as transactional.

`Mut(Cents, Txn)` is therefore not the type `Mut(Int, Txn)` is, against §6.7's "An alias is accepted and rejected in exactly the positions the type it names is". The failure surfaces phases later, with no mention of the alias:

```
`balance` is not a transactional mutable variable, so it has no commit history to await.
```

or, without `await_final`, a `transact_phase::rewrite_as_of_reads` panic on an unpaired fed-out read.

Failing regression test (the same program with `Mut(Int, Txn)` passes today):

```rust
/// A transactional mutable variable's value type is an ordinary type position,
/// so an alias stands where the type it names does.
#[test]
fn alias_is_the_value_type_of_a_transactional_variable() {
    check_scalar(
        indoc! {r#"
            Cents = Int

            balance: Mut(Cents, Txn) := 0

            for i in [1, 2, 3]:
                with begin():
                    balance := balance + i

            await_final(balance)
        "#},
        Value::Int(6),
    );
}
```

Fix, verified: declare the aliases first.

```suggestion
    errors.extend(pre_declare_type_aliases(stmts, ctx));
    pre_register_txn_decls(stmts, ctx);
```

(dropping the two comment lines above, or moving them up with the call). With the swap the test above passes.

`is_mut_annotation` (`stmts.rs:1301`) survives by luck — it only asks `.is_some()`, and `mut_annotation_parts` returns `Some(Err(_))` — so `Mut`-param `def` registration is unaffected. The defect is confined to the `Txn` discriminator.

src/ccl/lower/stmts.rs
```diff
@@ -229,13 +232,15 @@ pub(super) fn lower_stmts_inner(
...
     // innermost-first. (Mutability is carried by `Type::History` and checked
     // post-inference; introduction-vs-write is decided by lexical scope.)
-    let result = lower_final_stmt(last, rest, outer_bindings, ctx).and_then(|final_expr| {
-        rest.iter()
-            .enumerate()
-            .rev()
-            .try_fold(final_expr, |acc, (i, stmt)| {
-                lower_middle_stmt(stmt, &rest[..i], acc, outer_bindings, ctx, is_top_level)
-            })
+    let result = with_block_type_aliases(stmts, ctx, |ctx| {
```
groundlar, line 236: **Critical (same defect, nested-block site).** `pre_register_txn_decls` on line 227 sits outside this wrapper, so a block-local alias is not declared when the `Mut(V, Txn)` annotation is lowered. Same failure as the top-level site:

```rust
#[test]
fn alias_is_the_value_type_of_a_transactional_variable_in_a_nested_block() {
    check_scalar(
        indoc! {r#"
            def run(xs) => Int:
                Cents = Int
                balance: Mut(Cents, Txn) := 0
                for i in xs:
                    with begin():
                        balance := balance + i
                await_final(balance)

            run([1, 2, 3])
        "#},
        Value::Int(6),
    );
}
```

Fixing this one needs more than a line swap: the alias declaration has to move inside the same scope as, and ahead of, `pre_register_txn_decls`, which means `with_block_type_aliases` wraps the pre-registration too rather than just the fold. See the note on `with_block_type_aliases` itself.

src/ccl/lower/stmts.rs
```diff
@@ -1506,6 +1538,145 @@ fn collect_variant_arms(
...
+/// block's on the way out.
+///
+/// This is what makes an alias block-scoped, so every walker that lowers a block
+/// goes through it. The block's first rejected alias short-circuits `lower`, and
+/// the restore runs on both paths.
+///
+/// The top-level block is the exception: it is the outermost scope, with nothing
+/// to restore to and an error sink that takes every rejected alias rather than the
+/// first ([`lower_stmts_recovering`]).
+pub(super) fn with_block_type_aliases<T>(
```
groundlar, line 1733: **Important — the block-scope shape is now duplicated three ways, unevenly, and that is what makes the `Mut(V, Txn)` bug representable.**

"Entering a block" currently means, depending on which of the five walkers you are in, some subset of: snapshot `transactional_vars`, snapshot `mut_param_fns`, run `pre_register_txn_decls`, snapshot `type_aliases`, run `pre_declare_type_aliases`.

- `lower_stmts_inner` — all five, in a hand-written order
- `lower_for_body_stmts`, `lower_loop_body_chain`, `lower_tx_block_inner` — only the last two
- `lower_stmts_recovering` — three of them inline, with no restore

This wrapper is the right abstraction applied to one of the three tables. The ordering constraint it depends on — *aliases are declared before any pass that lowers a type annotation* — lives nowhere: no comment, no assertion, no type. Per CLAUDE.md's recurring-shape rule this is the third duplication, so it is worth recording even if the refactor is out of scope here.

A single `with_block_scope(stmts, ctx, |ctx| …)` owning the whole snapshot set and fixing the pre-pass order would answer "which tables does entering a block scope?" in one place and make the bug above unrepresentable. Minimum ask for this PR: a `TODO` here naming the constraint the wrapper does not yet enforce.

groundlar (commented):
The design-doc work here is unusually good — `src/ccl/design/lowering.md` names the walker inventory, the ordering argument, and the `BUILTIN_TYPE_NAMES` rationale, and §6.7 now reads as specification. Substitution at `lower_type_expr` is the right shape: nothing downstream of lowering sees the name, and predicate `Rc` sharing across two uses of a refined alias is already sound because `predicate_id_collisions` dedups by pointer (`context.rs:1145`).

One **Critical** finding, at two sites: `pre_register_txn_decls` lowers type annotations and runs before the block's aliases are declared, so `Mut(Alias, Txn)` is not the type `Mut(Int, Txn)` is. That contradicts the invariant both §6.7 and the test file's module doc claim. Repro and a verified fix are in the inline comment on `stmts.rs`.

The rest is maintainability: the block-scope snapshot/restore shape is now duplicated three ways with different subsets at each of the five walkers, which is what made the Critical finding representable; and `BUILTIN_TYPE_NAMES` is a third hand-maintained copy of a set `main` single-sourced one commit ago.

All 28 new tests pass as authored.

src/ccl/lower/stmts.rs
```diff
@@ -1506,6 +1538,145 @@ fn collect_variant_arms(
...
+/// [`BaseType::from_keyword`] resolves, the constructors [`lower_type_application`]
+/// dispatches on, the two heads resolved ahead of it (`Mut` in
+/// [`mut_annotation_parts`], `Feed` on a `def` parameter), and `Mut`'s sequencing
+/// domain `Txn`. A type alias may not rebind one, so an alias name and a built-in
+/// name never both resolve.
+///
+/// A new type constructor is added here in the same change that teaches
+/// [`lower_type_application`] its head; `builtin_type_names_are_refused` pins the
+/// names already listed, and nothing catches an omission.
+const BUILTIN_TYPE_NAMES: &[&str] = &[
```
groundlar: **Important — a third hand-maintained copy of a set the codebase just single-sourced.**

`main` landed "BaseType: single-source keyword spelling via `keyword()`/`from_keyword()`" one commit ago. This const re-lists those five keywords by hand, plus the seven heads `lower_type_application` matches on, plus `Mut`/`Feed`/`Txn` — and `builtin_type_names_are_refused` re-lists all fifteen a third time as a literal. The doc comment says it outright: "nothing catches an omission."

The list is complete today (cross-checked against `BaseType::from_keyword`, `ops.rs:48`, and the `lower_type_application` arms), so this is maintainability rather than a bug. Two cheap ways to close it:

- Derive the base half — `name_type(name).is_some()` already answers for `Int`/`UInt`/`String`/`Bool` — leaving only a `TYPE_CTOR_HEADS` const that `lower_type_application` itself matches over.
- Failing that, have the test iterate `BUILTIN_TYPE_NAMES` directly so it cannot drift from the const.

src/ccl/lower/stmts.rs
```diff
@@ -1506,6 +1538,145 @@ fn collect_variant_arms(
...
+/// invisible to every annotation below it. Declaring the block's aliases up front
+/// puts each one in scope throughout its block, and running in source order is
+/// what holds an alias's right-hand side to the aliases above it — `A = A` and a
+/// chain naming an alias declared later are unresolved names here.
+///
+/// The right-hand side is lowered once, at the declaration, so a refinement
+/// predicate sees the scope the alias statement sees. Every use substitutes that
+/// type.
+///
+/// Returns every alias the block got wrong rather than stopping at the first, so
```
groundlar, line 1663: **Nit — this claim and `with_block_type_aliases`'s disagree.** "Returns every alias the block got wrong rather than stopping at the first … the aliases after it are still declared, and their uses below still resolve" holds only at the top level. `with_block_type_aliases` takes `.next()` and aborts the whole block, so in a nested block a rejected alias costs the block everything, not just its own name. Its own doc notes the short-circuit; this one reads as unconditional. Suggest scoping the sentence to `lower_stmts_recovering`.

src/ccl/lower/loops.rs
```diff
@@ -329,6 +329,20 @@ fn body_scope(
...
+    ctx: &mut LoweringContext,
+) -> Result<Expr, LoweringError> {
+    // A loop body is a block, so it declares its own type aliases.
+    with_block_type_aliases(stmts, ctx, |ctx| {
+        lower_for_body_stmts_scoped(stmts, defer_name, mutation_scope, frame_introduced, ctx)
+    })
+}
+
+/// [`lower_for_body_stmts`] with the body's type aliases already in scope.
+fn lower_for_body_stmts_scoped(
```
groundlar, line 345: **Nit — three near-empty forwarding functions with duplicated six-argument signatures.** `lower_for_body_stmts_scoped`, `lower_loop_body_chain_scoped` and `lower_tx_block_scoped` all exist only to be the closure body of a `with_block_type_aliases` call. A shared block-entry wrapper (see the note on `with_block_type_aliases`) would collapse all three. Fine as-is if the wrapper refactor lands separately.

docs/chl-spec.md
```diff
@@ -2500,35 +2496,51 @@ annotation (`def f(n: Count) => Item:`), a type argument (`List(Priced)`,
...
+  what the statement sees, and nothing more.
+- **A capitalized name is bound only by an alias.** Every other binding form —
+  `:=`, `op=`, `<<=`, a `for` target, a comprehension generator — rejects a
+  capitalized binder and names this one. A `def` name and a parameter name are
+  not yet checked; a capitalized one there binds a value the type language
+  cannot see.
+- **A built-in type name keeps its meaning.** `Int`, `UInt`, `String`, `Bool`,
+  `Unit` and `Txn`, and the constructors `Array`, `Collection`, `Feed`,
+  `FullMap`, `List`, `Map`, `Mut`, `Option` and `Set`, are refused as alias
+  targets, so a name resolves as a built-in or as an alias and never as both.
```
groundlar: **Nit — the claim is stronger than what the code does.** `Unit` and `Txn` resolve as *neither*: `name_type` returns `None` for `Unit` by design (`stmts.rs:1689`) and `Txn` is only ever matched syntactically inside `mut_annotation_parts`, so `x: Unit` is "unknown type annotation" while `Unit = Int` is "built-in type". Reserving both spellings is right; the sentence should say they are reserved rather than that they resolve one of two ways.

tests/compilation_pipeline/type_aliases.rs
```diff
@@ -0,0 +1,484 @@
...
+            x
+        "#},
+        "`Int` is a built-in type and cannot be given another meaning",
+    );
+}
+
+#[test]
+fn builtin_type_names_are_refused() {
+    // Every name `BUILTIN_TYPE_NAMES` holds, each refused as an alias target.
+    for name in [
```
groundlar: **Nit — this literal is the third copy of `BUILTIN_TYPE_NAMES`.** Iterate the const instead, so the test cannot drift from what it pins:

```rust
    for name in cambra::ccl::lower::BUILTIN_TYPE_NAMES {
```

(needs the const exported, or the test moved next to it as a unit test). As written, an omission from the const is an omission from the test by construction — which is exactly the gap the const's own doc comment flags.

tests/compilation_pipeline/type_aliases.rs
```diff
@@ -0,0 +1,484 @@
...
+            total
+        "#},
+        "is capitalized, so it names a type, not a value",
+    );
+}
+
+/// An alias leaves no trace in the lowered program: reading a type name as a term
+/// is an unbound name, not a value.
+#[test]
+fn an_alias_name_is_not_a_value() {
```
groundlar, line 440: **Nit — the needle is the alias name itself**, which appears in the source and in almost any error this program could produce, so the test would pass on a failure that has nothing to do with erasure. Assert on the unbound-variable text instead, so the test pins the claim its name makes:

```rust
        "Unbound variable: 'MyInt'",
```

src/ccl/lower/stmts.rs
dpmills: Confirmed and fixed in 9a1e34f5 — your test reproduced it exactly ("`balance` is not a transactional mutable variable"). Both of your regression tests are now in the suite.

Took the swap you suggested, with the reason stated at the call rather than the two old comment lines.

src/ccl/lower/stmts.rs
dpmills: Fixed in 9a1e34f5. Rather than moving the alias declaration out, `pre_register_txn_decls` moved *in* — it is now the first statement of the `with_block_type_aliases` closure in `lower_stmts_inner`. The wrapper already guarantees that everything in `lower` sees the block's aliases, so ordering it this way makes the site correct by construction instead of by a hand-written sequence.

Your nested-block test is in the suite and passes.

src/ccl/lower/stmts.rs
dpmills: Agreed on the diagnosis, and it is what made both `Mut(V, Txn)` sites representable.

Took more than the minimum ask but less than the full refactor: the constraint is now stated on `with_block_type_aliases` — a block-entry pass that lowers a type annotation belongs inside `lower`, with the concrete failure spelled out — and both offending sites are ordered by construction rather than by comment. Also recorded in `src/ccl/design/lowering.md`.

Left the wider `with_block_scope` out. Unifying all five tables means the loop and transaction walkers would start snapshotting `transactional_vars` and running `pre_register_txn_decls`, which they do not today — a behaviour change with its own blast radius, and not one to fold into the feature PR. Worth its own change; happy to open it.

src/ccl/lower/stmts.rs
dpmills: Took the first option. `BUILTIN_TYPE_NAMES` is gone; `is_builtin_type_name` now reads `BaseType::from_keyword(name).is_some() || RESERVED_TYPE_NAMES.contains(&name)`, so the five base spellings come from their single source and the const carries only the ten names that have no other home (the seven `lower_type_application` heads plus `Mut`/`Feed`/`Txn`).

Also took the second — see the reply on the test.

The remaining gap is the one you name: a constructor added to `lower_type_application` and omitted from `RESERVED_TYPE_NAMES`. The doc comment says that outright rather than the old "nothing catches an omission". I tried asserting it from the fallback arm and it does not work — `Int(3)` reaches the fallback legitimately, so the assertion would fire on user input.

src/ccl/lower/stmts.rs
dpmills: Right, the sentence was unconditional and the behaviour is not. Rewritten to say the aliases after a rejected one are still declared, that only `lower_stmts_recovering` reads past the first because it collects per-statement errors, and that through `with_block_type_aliases` the first error aborts the block.

src/ccl/lower/loops.rs
dpmills: Taking your "fine as-is if the wrapper refactor lands separately" — see the reply on `with_block_type_aliases` for why the unification is a separate change.

The alternative that removes them without the refactor is to inline each body as the closure, which re-indents three large function bodies. That would bury the two `Mut(V, Txn)` fixes in whitespace on a PR under review, so I would rather do it in the change that collapses the wrapper properly.

docs/chl-spec.md
dpmills: Correct — checked both: `name_type` returns `None` for `Unit` by design and `Txn` is only ever matched syntactically in `mut_annotation_parts`, so neither is writable as an annotation.

Reworded to say the type language's own names are reserved, splitting the two cases: the four writable primitives keep their meaning, and `Unit` and `Txn` are reserved without being writable at all — the unit type is written `{}` (§6.6) and `Txn` is read only in a `Mut(V, Txn)` second slot.

tests/compilation_pipeline/type_aliases.rs
dpmills: Done. `RESERVED_TYPE_NAMES` and `is_builtin_type_name` are exported from `cambra::ccl::lower`, and the test iterates `BaseType::keyword` for the base half chained with the const for the rest — the same pair the refusal consults. It also asserts `is_builtin_type_name(name)` per name, so a name listed but not refused fails on the assertion rather than on a missing compile error.

tests/compilation_pipeline/type_aliases.rs
dpmills: Good catch — the needle appeared in the source, so the test would have passed on almost any failure. Now `"Unbound variable: 'MyInt'"`, which is the claim the test name makes.

src/ccl/lower/stmts.rs
```diff
@@ -1506,6 +1538,145 @@ fn collect_variant_arms(
...
+            ));
+            continue;
+        }
+        match lower_type_expr(rhs, ctx) {
+            Ok(ty) => ctx.declare_type_alias(name, ty),
+            Err(inner) => errors.push(LoweringError::unsupported(
+                stmt.span,
+                format!(
+                    "`{name}` is capitalized, so `{name} = …` declares a type alias \
+                     and its right-hand side must be a type: {inner}"
```
groundlar, line 1706: **Nit — this message embeds a raw parser-AST dump.** `{inner}` is often `lower_type_expr`'s catch-all (`stmts.rs:1462`), which formats the CHL AST with `{:?}`. Pre-existing, but this PR makes it far more reachable: every capitalized `=` now routes through `lower_type_expr`, so the common typo reads

```
`Five` is capitalized, so `Five = …` declares a type alias and its right-hand side
must be a type: unsupported type annotation form: Lit(Int(5))
```

and a near-miss on the refinement spelling is worse — `Cents = Int | _ >= 0` renders a four-line `Compare { left: Spanned { span: Span { start: 8, end: 15 }, node: BinOp { … } } }` into the diagnostic.

`a_term_on_the_right_of_a_capitalized_name_is_rejected` pins the prefix, so the dump is unpinned tail. Worth either giving `lower_type_expr`'s catch-all a spelling-based message, or truncating `{inner}` to its first line here. The `|` case in particular deserves the pointer §6.4 already makes — `where` is the separator, `|` is reserved for variant arms:

```
its right-hand side must be a type; a refinement is written `{Int where _ >= 0}`
```

docs/design.md
```diff
@@ -39,7 +39,7 @@ Programs are run on the Cambra engine, which compiles and executes CHL programs.
...
 
 **Refinement types** look like this: `{𝑥: 𝑇 | 𝑝(𝑥)}` — a value of the base type for which a predicate holds. The machinery is implemented inside the checker today — basic refinement inference, constraint solving over refined subtyping, explicit refinement acquisition, and specialization of polymorphic call sites — and built-ins like `groupby` produce refined types internally. What does **not** exist yet is the surface syntax: the CHL spelling is `{T where p(_)}`, which drops the named binder in favour of `_` for the refined value ([docs/chl-spec.md](chl-spec.md#64-refinement-syntax)) and is **[Decided]** but not in the grammar. Verifiable, whole-application contracts expressed through refinements are the driving design direction rather than a current capability. The `{𝑥: 𝑇 | 𝑝(𝑥)}` form above stays the *internal* notation — a dependent codomain refinement has to name the binder it closes over — so it is what compiler output and the CCL design docs show.
```
groundlar: **Stale, three lines below one this PR edits.** "the CHL spelling is `{T where p(_)}` … and is **[Decided]** but not in the grammar" is no longer true: `{T where p}` parses to `ChlExpr::BraceRefinement`, lowers to `Type::Refined` (`stmts.rs:1420-1428`), and this PR ships four passing tests that use it (`type_aliases.rs:101-166`).

Since the paragraph above is already being touched here, suggest correcting this one in the same change — the surface syntax exists; what is still missing is discharging it, e.g. `{Int where _ >= 0}` against a literal `7` fails subsumption (`_ == 9` against the singleton `Int@9` succeeds). That is the accurate version of "rather than a current capability" and it keeps the sentence's point.

groundlar (commented):
Two follow-ups from checking whether a refined alias works (it does — `Cents = {Int where _ >= 0}`): one diagnostics nit this PR widens the blast radius of, and one stale claim in a paragraph adjacent to the one being edited.

groundlar (approved):

src/ccl/lower/stmts.rs
dpmills: Fixed at the catch-all rather than at the alias message, since it is the shared site and every other annotation position reaches it too. `describe_type_form` names the form in the surface's own words, so `Five = 5` now ends "…must be a type: an integer literal is not a type", and a bare `x: 5` gets the same tail with no alias involved.

Took the `|` pointer as its own arm. You are right that it never reached the variant-arm message: comparison binds looser than `|`, so `Int | _ >= 0` arrives as a `Compare` with the `|` nested inside it, not as a `|`-chain. It now says a refinement is written in braces with `where`, with your spelling as the example.

Three tests, and `a_term_on_the_right_of_a_capitalized_name_is_rejected` now pins the message to its end so the tail is no longer unpinned.

docs/design.md
dpmills: Corrected in the same change. The sentence now says the surface syntax exists — it parses, lowers, and is enforced against a body — and that what is missing is discharge over anything but a singleton.

I verified the two examples before writing them rather than taking them from the review: `{Int where _ >= 0}` against the literal `7` fails with "annotated as {Int | __elem >= 0}, but inferred as Int@7", while `def f(a) => {Int where _ == 9}: 9` compiles and `: 8` fails. Same split I ran into writing this PR's refinement tests, which is why they use the `== 9` shape.

PR #177, opened 2026-09-11
Allow keying a collection by a tuple or a record

dpmills:
A tuple or a record could not key a collection: a key domain owes equality, and `Equatable`'s instance table keys on base types. Equality is now read structurally — two products are equatable when they are the same shape and each component is — so a `map`, `set` or `groupby` over a product types and runs, as does `(1, 2) == (3, 4)`.

## Answering a trait off the table

`narrow_product` answers a determined product (`Offered::Product`) without touching the candidate set: it records the product on the obligation, states it as an upper bound on every sibling position, and requires it of each component through its own obligation over a fresh variable. A component still unknown resolves later by the ordinary delivery path. `TraitObligation` gains `operands` (each position's type, recorded by `watch`) and `structural` (the shape settled at); `places_under` skips an answered obligation, whose untouched rows would deposit a base where a product stands. `Trait::is_structural` is `Equatable` alone: ordering two products needs an order on their components, and a record's fields carry none. Reasoning: [type-inference.md](src/ccl/design/type-inference.md#a-product-is-answered-off-the-table).

`ConstrainError::NoTraitInstance`'s `accepted` widens to `Vec<Type>` to name the settled product; `check.rs` decomposes a structural trait on the shared fields.

## Comparing and looking up a product at run time

`compare_records` conjoins `Equals` over the fields both `Records` columns carry and disjoins `NotEquals`. Records are width-subtyped, so skipping a one-sided field is what makes the two operand orders agree (`a_wider_record_compares_on_the_shared_fields`). `CheckedLookup` takes a product key at every input shape: `is_key_tiling` admits a record of scalars, and the key pivots to one column through `scalar_tile_to_column_value`.

## Tests and docs

`tests/type_check.rs` covers shape mismatch, nesting, a failing component, and a wider key that types and is never present. `joins_aggregates_groupby.rs` and `transactions.rs` run product-keyed `map`, `groupby` and checked lookup, and a transactional map keyed by a per-row product. Two `type_annotations.rs` cases now reach the refinement question instead of the missing instance; [3.4 Comparisons](docs/chl-spec.md#34-comparisons) states the reading.


src/ccl/infer/solver/traits.rs
```diff
@@ -585,13 +634,92 @@ impl TraitObligation {
...
+        // set; a product cannot then be the same type, and homogeneity is what makes that a
+        // contradiction rather than a second possibility.
+        if self.candidates.borrow().len() < self.trait_.instances().len() {
+            return self.reject(pos, product);
+        }
+        // A second product answers the same obligation, so the two have to be the same
+        // product: the rows are homogeneous, and a wider one is a subtype, so the
+        // upper-bound edge below admits `(𝐴, 𝐵, 𝐶)` against `(𝐴, 𝐵)` on its own.
+        if let Some(settled) = self.structural.borrow().as_ref() {
+            return if product_fields(settled) == product_fields(product) {
```
sortalongo: Order-sensitive where `check.rs` is not. `product_fields` returns fields in the vector's stored order and `Type::Record` is a plain `Vec` that no construction site sorts, so this `==` rejects two spellings of the same record shape. `check.rs:201` answers the same question with a shared-field intersection, so the two sites disagree on the rule. Compare sorted field sets here.

src/interpreter/binop.rs
```diff
@@ -192,10 +193,67 @@ pub fn apply_binop_column(op: BinOpKind, left: ColumnValue, right: &ColumnValue)
...
+    let rows = left
+        .values()
+        .map(ColumnValue::len)
+        .chain(right.values().map(ColumnValue::len))
+        .min()
+        .unwrap_or(0);
+    let mut acc = BitVec::from_elem(rows, op == CompareKind::Equals);
+    for (field, l) in left {
+        let Some(r) = right.get(&field) else {
+            continue;
```
sortalongo: This skip is the shared-field rule discussed in the summary. Drop it and compare every field. The tile-layer `assert!` already guarantees both columns carry exactly the fields their extent names, so this is defending against a state the runtime declares impossible. If the assertion turns out not to hold on this path, that is the bug to fix rather than absorb here.

src/interpreter/binop.rs
```diff
@@ -192,10 +193,67 @@ pub fn apply_binop_column(op: BinOpKind, left: ColumnValue, right: &ColumnValue)
...
+/// while the other silently answered `equal`.
+///
+/// The length is the shortest column on either side, since a column still filling is a
+/// column whose later positions have not been decided.
+fn compare_records(
+    op: CompareKind,
+    left: HashMap<String, ColumnValue>,
+    right: &HashMap<String, ColumnValue>,
+) -> BitVec {
+    let rows = left
```
sortalongo: The row count is minimised over every column on both sides, including the unshared ones the loop then skips, so a wider operand's short extra column can truncate an answer the shared fields already decided. Take the minimum over the shared fields only.

src/interpreter/binop.rs
```diff
@@ -192,10 +193,67 @@ pub fn apply_binop_column(op: BinOpKind, left: ColumnValue, right: &ColumnValue)
         (BinOpKind::Compare(op), ColumnValue::Bools(l), ColumnValue::Bools(r)) => {
             ColumnValue::Bools(zip_bool_compare(op, l, r))
         }
+        // A product compares **componentwise**, which is the whole of what makes one
```
sortalongo: This says a field one side lacks is a shape error rather than an inequality. The doc comment on `compare_records` directly below says such a field does arrive and is deliberately skipped, and the code skips. Delete this claim.

src/ccl/design/type-inference.md
```diff
@@ -2643,9 +2669,10 @@ A contribution arriving at a position is one of three things, and each has its o
 |---|---|---|
 | a **base** | `Int` | narrows the candidate set |
 | **not determined yet** | a variable, a hole, a `Feed` handle whose payload arrives separately | nothing to say |
-| **determined, and not a base** | a tuple, record, variant, function | rejected — no instance accepts it ([What the tables hold](#what-the-tables-hold)) |
+| a **product** | a tuple, a record | answered componentwise by a structural trait, rejected by every other ([A product is answered off the table](#a-product-is-answered-off-the-table)) |
```
sortalongo: This row makes the table four rows, and the sentence above it still reads "one of three things". Change three to four.

tests/compilation_pipeline/joins_aggregates_groupby.rs
```diff
@@ -853,6 +853,108 @@ fn checked_lookup_answers_presence(#[case] code: &str, #[case] expected: Value)
...
+
+/// A tuple-keyed checked lookup, present and absent.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[case::present("(1, 2)", Value::Int(10))]
+#[case::absent("(9, 9)", Value::Int(0))]
+// A component every key shares: its type is that component's own singleton rather than the
+// join two distinct values would give, so this is the case that reads a refinement at the
+// invariant position.
+#[case::sharing_a_component("(1, 2)", Value::Int(10))]
```
sortalongo: Byte-identical to `#[case::present]` above, and the map in the body has keys `(1, 2)` and `(3, 4)`, which share no component, so the comment describing the case is false. Delete the case and its comment.

tests/compilation_pipeline/joins_aggregates_groupby.rs
```diff
@@ -853,6 +853,108 @@ fn checked_lookup_answers_presence(#[case] code: &str, #[case] expected: Value)
...
+///
+/// Records are width-subtyped, so a value reaching a narrower position keeps its extra
+/// columns at run time. They are not part of the value at that type, and the two operand
+/// orders have to agree about that: reading a one-sided field as a fault aborted the process
+/// in one order while the other answered `equal` without it.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[case::wider_on_the_left("t == (a=1, b=2)")]
+#[case::wider_on_the_right("(a=1, b=2) == t")]
+fn a_wider_record_compares_on_the_shared_fields(#[case] comparison: &str) {
```
sortalongo: This pins the shared-field answer, which the spec text this PR adds forbids. It is also the test that would tell us whether the tiling at the comparison names two fields or three. Remove it with the skip, and replace it with whatever the narrowing question resolves to.

tests/type_check.rs
```diff
@@ -5611,6 +5662,44 @@ fn checked_lookup_on_a_set_is_membership_as_a_value() {
...
+///
+/// The key domain is the key morphism's image, so what a key type has to be is whatever the
+/// morphism produces — and a group-by's key predicate compares the two, which is the
+/// `Equatable` obligation a product now answers componentwise
+/// ([`products_are_equatable_componentwise`]). Nothing else about a keyed collection reads
+/// the key's shape.
+#[test]
+fn a_product_keys_a_collection() {
+    assert_eq!(
+        infer_program("m = map([((1, 2), 10), ((3, 4), 20)])\nm[(1, 2)]?").to_string(),
```
sortalongo: Multi-line CHL programs embedded as a single string with `\n`. CLAUDE.md, "Embedded source in tests", requires `indoc!{r#"…"#}` here. Twelve cases across this file and `joins_aggregates_groupby.rs` do this. Convert them.

tests/type_check.rs
```diff
@@ -300,20 +300,21 @@ fn negation_rejects_an_operand_with_no_instance() {
     );
 }
 
-/// Composites are **not** comparable, and not addable either.
+/// A composite satisfies only a **structural** trait, and only where its components do.
 ///
 /// The tables have no row for a tuple, record or collection — but an absent row is
 /// not by itself a rejection, and for a while it was not one: a composite offers no
```
sortalongo: Narrates the history of the change, and its example `(1, 2) == (3, 4)` is the program this PR makes legal. State the present rule and drop the clause.

src/ccl/infer/check.rs
```diff
@@ -187,6 +188,52 @@ impl Typing for CheckCtx {
...
+            // **A product is determined, so it is not residue.** A structural trait is
+            // answered componentwise, so the operands' components pair and each pair has to
+            // answer the trait in turn. Checked here rather than excused: this rule is what
+            // catches a later pass rewriting the tree into something ill-typed, and a
+            // product is the shape inference answers structurally rather than through the
+            // candidate set (`traits::narrow_product`), so nothing else downstream would.
+            if trait_.is_structural()
+                && operand_types
+                    .iter()
+                    .all(|t| matches!(strip_refinements(t), Type::Tuple(_) | Type::Record(_)))
```
sortalongo: `Tuple(_) | Record(_)` admits a tuple paired with a record. The shared field set is then empty, the loop below runs zero times and the wall returns `Ok`, so this check is more permissive than inference. Require the shapes to agree and raise when they do not.

sortalongo (changes requested):
Requesting changes. The type-level work is right and `Offered::Product` names a distinction the enum was missing, in the one place that classifies a contribution. Three things need to change: a probable false rejection from an order-sensitive shape comparison, a runtime comparison that contradicts the specification text this PR adds, and a set of smaller cleanups.

## The shared-field comparison contradicts this PR's own spec text

`docs/chl-spec.md` gains: "The two operands must be the *same* product: a wider one is a different shape, not a comparison over the fields they share." `compare_records` does the opposite.

The runtime already enforces the strict rule. Every tile a producer emits passes a hard `assert!` in `src/interpreter/tile_operators/mod.rs` that the tile could have come from its tiling, and for a record column `is_compatible_with_extent` requires exactly the fields the extent names. The union arm in the same match documents the contrasting subset rule for tags, so the distinction between the two is deliberate. A narrowing also already exists: pivoting list-literal elements into a column reads only the keys the extent names, so `[(a=1, b=2), (a=3)]` drops `b` today.

The shared-field reading is therefore the outlier. It reasons about a column shape the assertion above it rejects.

The direction to take: a value carries no fields its type does not name. Where a program needs downcasting, that gets an explicit mechanism rather than undeclared data riding along implicitly. Drop the skip, and drop the test that pins its answer.

One question is worth settling before choosing the fix. `a_wider_record_compares_on_the_shared_fields` passes today, so either the tiling at that comparison still names three fields — the annotation narrowed the type but not the tiling — or the assertion is not on that path. If it is the first, the static type and the runtime tiling disagree about which fields the value has, and that disagreement is the defect the skip is papering over. The precedent for where a narrowing belongs is the `Mut(V) <: V` deref arm removed from subtyping as "a coercion in disguise", with the work moved to term elaboration.

## Product equality has more than two readings today

Context for that decision rather than a request on this PR. `Predicate::Record::contains` requires only the predicate's own named fields, so it is a second shared-field rule, and it predates this change. `tile_operators/map.rs` calls it nine lines above a whole-value hash probe, so one key is tested under both rules inside a single loop body. Lookup, hashing, the commit store and per-group aggregation all use whole-value equality, and the groupby sort panics outright when two records have different field sets. Making `==` strict does not by itself reduce this to one relation.

## Otherwise

CI is green across all ten checks, every doc anchor and quoted heading resolves, no ignored or `should_panic` tests were added, and equality is decided structurally throughout the production code with no rendering in the path.

src/ccl/infer/solver/traits.rs
dpmills: Done — `product_shape` sorts the field set, and `narrow_product` compares through it. `check.rs` no longer answers the same question a second way: it pairs on the narrowest shape and rejects two shapes that are not nested, so the two sites now state one rule.

src/interpreter/binop.rs
dpmills: Deleted. The claim was false and the code was right.

src/interpreter/binop.rs
dpmills: Fixed — the row count is the minimum over the shared fields only.

src/ccl/design/type-inference.md
dpmills: Fixed: four.

tests/compilation_pipeline/joins_aggregates_groupby.rs
dpmills: Deleted, case and comment.

tests/type_check.rs
dpmills: Converted — twelve cases across both files.

tests/type_check.rs
dpmills: Rewritten to the present rule; the `(1, 2) == (3, 4)` clause is gone.

src/ccl/infer/check.rs
dpmills: Done, and it caught something. `Tuple(_) | Record(_)` now requires the shapes to be nested — a tuple against a record shares no field, so the loop ran zero times — and the narrowest is what the components pair on.

Making it *strict* equality instead panics the wall on `a_wider_record_compares_on_the_shared_fields`:

```
post-inline produced an invalid tree: [No Equatable instance for BinOp: operand 2 is
{a: Int@1, b: Int@2}, but the only type accepted there is {a: Int@1, b: Int@2, c: Int@9}]
```

**post-inline**, so inference agrees the two operands are one product and `inline_and_beta_reduce` is what breaks it: substituting the argument for the binder drops the parameter's narrowing annotation. Filed, with the mechanism, as the vault issue `type-checker-inlining-drops-a-narrowing-annotation`.

src/interpreter/binop.rs
dpmills: I probed the question you raised, and it is the first branch: the tiling names three fields where the annotation names two. Instrumenting `compare_records` on that test prints

```
PROBE compare_records left=["a", "b", "c"] right=["a", "b"]
```

So the assertion is on the path and passes, because the extent is read off the tiling, which nothing narrowed. Tightening `check.rs` to strict equality then shows the disagreement is not only in the tiling — it is in the typed tree, and it arrives at **inlining**: substituting `(a=1, b=2, c=9)` for `t` drops the parameter's declared type, and the comparison's two operands are two different products from there on.

So the direction you name is right and the lift belongs where `refinement_discharged_by` already asks the analogous question for predicates — a record restriction, as the `restrict(pred)` guard is for an undischarged precondition. That is the vault issue `type-checker-inlining-drops-a-narrowing-annotation`.

What I did not do is drop the skip, because with the narrowing lost in the tree a type-correct program would abort at run time instead. Kept, and its doc comment now says what makes it sound rather than claiming a shared-field rule: typing rejects a comparison between two different products (verified — `(a=1,b=2,c=3) == (a=1,b=2)`, `(1,2) == (a=1,b=2)`, `(1,2) == (1,2,3)` and `(a=1,b=2) == (a=1,z=2)` are all type errors at this head), so an unshared column is one the tiling did not narrow and is not part of the value. The intersection reads the value at its type.

tests/compilation_pipeline/joins_aggregates_groupby.rs
dpmills: Kept, renamed to `a_surplus_column_does_not_reach_the_comparison`, and re-documented. It answered the question: the tiling names three fields, so the static type and the runtime tiling disagree — see the thread on the skip. What it pins now is that a column the comparison's type does not name does not reach the answer, in either operand order.

dpmills:

**The shared-field comparison.** Your question decides it, and the answer is the first branch. Instrumenting `compare_records` on `a_wider_record_compares_on_the_shared_fields` prints

```
PROBE compare_records left=["a", "b", "c"] right=["a", "b"]
```

so the assertion is on that path and passes: the extent is read off the tiling, and the annotation narrowed the type without narrowing the tiling.

Tightening `check.rs` to strict shape equality then shows it is worse than a tiling gap. The wall reports

```
post-inline produced an invalid tree: [No Equatable instance for BinOp: operand 2 is
{a: Int@1, b: Int@2}, but the only type accepted there is {a: Int@1, b: Int@2, c: Int@9}]
```

at **post-inline**, so inference agrees the two operands are one product and substitution is what breaks it: `inline_and_beta_reduce` replaces `t` with the argument and drops the parameter's declared type. `inline.rs` already owns the analogous question for predicates — `refinement_discharged_by` asks whether the argument entails what the parameter demands, asserts when it does not, and names the `restrict(pred)` lift for the undischarged case. Width is the same question in the field-set dimension, with no assert and no lift. Filed as the vault issue `type-checker-inlining-drops-a-narrowing-annotation`, which also records what retires both intersection rules.

So the skip stays, because dropping it while the tree carries the wider type aborts a program that type-checks. What changed is that it is no longer a shared-field rule: the spec text this PR adds is true at the type level — `(a=1,b=2,c=3) == (a=1,b=2)`, `(1,2) == (a=1,b=2)`, `(1,2) == (1,2,3)` and `(a=1,b=2) == (a=1,z=2)` are all type errors at this head — and the doc comment now says the intersection is the product the type names, with an unshared column being one the tiling did not narrow.

`check.rs` pairs on the narrowest shape and rejects two shapes that are not nested, so the rule there and the rule in `compare_records` are one rule rather than an intersection that is undefined for a tuple against a record.

**Product equality's other readings.** Noted rather than acted on, as you framed it. `Predicate::Record::contains` is untouched here.

sortalongo (approved):
Approving.

The response answered the blocking question with a probe rather than an argument, and it disproved the premise I reviewed on. I claimed the tile assertion guarantees a record column carries exactly the fields its extent names, so the skip defended against a state that could not arise. The assertion passes precisely because the tiling is the wide one: the annotation narrowed the type and nothing narrowed the tiling.

Tracking that to `inline_and_beta_reduce` dropping the parameter's declared type is the more useful result, and it is the right diagnosis. A value outliving the type that narrowed it is a different defect from a tiling gap, it sits in a pass whose job is not to change types, and `refinement_discharged_by` is the right precedent for what the fix looks like. Filing it separately is the correct scope call.

Accepting both deferrals. Removing the skip while the tree still carries the wider type would fail a program that type-checks, so it stays until the annotation survives inlining. The type level is strict now and the specification text this PR adds is true at this head, which is what I actually cared about.

One concession worth naming so it is on the record rather than in a comment. The post-inference check exists to catch a later pass rewriting the tree into something ill-typed, inlining is such a pass, and the new nesting rule is calibrated to admit its output. That is disclosed at the site, and it is acceptable while the issue is open and prioritised. It does mean this check no longer reports a width disagreement arriving from any other source.

Two follow-ups, neither blocking and neither needing another review pass.

`product_shape` now sorts the field set, and nothing exercises it. A program comparing two record spellings whose fields are written in different order would both demonstrate the original defect and pin the fix.

The doc comment on the renamed surplus-column test says both operands are `{a: Int, b: Int}`. That holds before inlining and not after, and what happens after is the content of the issue just filed. Worth a sentence so the test does not read as though the tiling were the only disagreement.

A smaller note: the solver requires equal shapes and the checker requires nesting, so those remain two rules with one documented reason, rather than one rule.

PR #178, opened 2026-09-11
Partially implement SMT-based refinement checks

octalsrc:
This PR adds dependencies on the [z3 SMT solver](https://github.com/Z3Prover/z3) and on the [easy_smt](https://crates.io/crates/easy-smt) Rust crate.

It implements SMT-based refinement checking for an initially small subset of the Cambra language. Examples of the subset are found in `tests/compilation_pipeline/type_annotations.rs`.


src/ccl/infer/solver/constrain.rs
```diff
@@ -314,17 +329,52 @@ impl ConstrainCache {
 /// (see Parreaux 2020 §3.4).
 pub type ExtrudeCache = HashMap<(InferVarId, bool), Rc<InferVar>>;
 
+/// Special empty scope that signals a post-planning check, during
+/// which SMT fallback should not be used.
+pub struct SkipSmtScope;
+
+impl ScopeEnv for SkipSmtScope {
```
dpmills: This is the temporary hack part, right? If so, put a big comment on it saying so, and also saying that we shouldn't add to it so that claude doesn't start hanging more functionality off it

octalsrc:
Some major points to note about the changes:

- `close_let_type` now substitutes the rhs of a `let` into the refinement. It skips `:=` lets for now. Previously, `close_let_type` was a no-op.
- The post-inference refinement check performed by `inline.rs` has been weakened, since there are now several cases where lhs and rhs are still semantically equal but no-longer structurally equal, and we can't currently do SMT checks post-inference. Two tests that expected the stronger `inline.rs` check to fail have been gated behind `--features strong-post-inference-tests` for now.
-  Type context information from `InferCtx` is now threaded into `constrain_subtype` (`constrain_subtype_with`) so that in-scope variables referenced in a refinement comparison can be constrained according to their own context-tracked subtypes.
- After the type inference phase, a pass strips all nested refinements out of type slots in refinement bodies, except those on data functions.


And a minor point:

- The inspector fixtures needed to be refreshed because node id counts shifted (due to additional refinement bodies in the examples?)

src/ccl/inline.rs
```diff
@@ -596,7 +598,29 @@ fn refinement_discharged_by(arg_ty: &Type, param_ty: &Type) -> bool {
     if let Some(value) = arg_ty.mut_value_type() {
         supplied.extend(value.refinements());
     }
-    demanded.iter().all(|d| supplied.contains(&d))
+
+    // We can't semantically compare refinements in this phase, so we
```
dpmills: **[T1] This is the one hack in the PR that costs silent wrong answers rather than a bad message.**

This returns `true` whenever the *supplied* set is non-empty, whatever it supplies — so `Int@7` now discharges a demand of `Int@5`. The assert it feeds still says (unchanged, a few lines up at ~385):

> A hard assert, not `debug_assert`: this reads a live post-inference data path, and a release build proceeding past it would miscompile to wrong results rather than fail.

The justification — "We can't semantically compare refinements in this phase" — is not true, and *this PR* is what makes it untrue. I replaced the whole block with:

```rust
if demanded.iter().all(|d| supplied.contains(&d)) { return true; }
let supplied_owned: Vec<Refinement> = supplied.iter().map(|r| (*r).clone()).collect();
smt_sub(arg_ty.mut_value_type().unwrap_or(arg_ty).peel_refinements(),
        &supplied_owned, demanded, &NoScope).unwrap_or(false)
```

and the **full suite passes, including both `strong-post-inference-tests` tests re-enabled**. So the feature flag and its two `#[cfg]`s come out entirely.

One open decision, not a blocker: `.unwrap_or(false)` turns a solver/process error into the hard assert. A `Process` error probably wants its own message rather than "does not entail". Everything else is settled.

Also: the doc comment above this function still ends "Anything subtler is exactly what should trip the assert and get a real `restrict` lift" — true again after the fix, false without it. And the `Mut`-stamp paragraph just above becomes redundant if the `Type::refinements()` History arm goes (see `src/ccl/ty.rs`).

src/ccl/inline.rs
```diff
@@ -1139,6 +1163,7 @@ mod tests {
     /// carries a different refinement than the parameter demands still asserts.
     #[test]
     #[should_panic(expected = "does not entail")]
+    #[cfg(feature = "strong-post-inference-tests")]
```
dpmills: **[T1]** These two tests are hidden behind a feature no CI step, no `ci.sh` gate, and no clippy config builds — so they are permanently dead, and the regression they pin is now unguarded. I enabled the feature and confirmed both fail against this branch.

The second one's own doc comment names exactly what has happened:

> Pins the guard that keeps `refinement_discharged_by` from being weakened to "any refined param is fine" — the shape this arrives in is a monomorphization key that cannot tell two call sites' refinements apart, which lets one literal's singleton reach the param slot of a call made at another.

With the `smt_sub` fix on `inline.rs` both pass, so the right move is to delete the `#[cfg]`s and the feature.

If for some reason the weakening has to stay: use `#[ignore]` rather than `#[cfg(feature = ...)]`. An ignored test still compiles and still shows up in test output; a feature nothing builds does neither.

Cargo.toml
```diff
@@ -37,6 +37,12 @@ test-helpers = []
 # Enable when bisecting a miscompile: `cargo test --features deep-typecheck`.
 deep-typecheck = []
 
+# Run tests that expect refinement disequalities after the type
+# inference phase. The checks for those disequalities are currently
+# disabled, because the addition of non-trivial refinements makes the
+# equality check much more brittle than before.
+strong-post-inference-tests = []
```
dpmills: **[T1]** This feature is built by nothing — not CI, not any `ci.sh` gate, not any of the three clippy configurations (`--all-targets` does not enable non-default features). CI's own comment on the reversed-refinement-order step states the principle:

> an unrun knob would rot exactly as an uncompiled feature does

With the `smt_sub` fix in `inline.rs`, both gated tests pass and this can be deleted outright.

src/ccl/infer/solver/constrain.rs
```diff
@@ -1619,13 +1704,39 @@ fn constrain_go_impl(
...
                 // resolves to a concrete base lacking those refinements.
                 let demanded = Type::refined(rbase.clone(), deficit);
-                constrain_go(lbase, &demanded, sl, sr, cache)
+                constrain_go(lbase, &demanded, sl, sr, cache, scope)
+            } else if scope.is_skip_smt() {
+                Err(ConstrainError::Mismatch {
+                    lhs: lhs.clone(),
+                    rhs: rhs.clone(),
+                })
+            } else if super::smt::smt_sub(
```
dpmills, line 1725: **[T1] This is the real cause of the `Unsupported` diagnostic regression, and the fix is here rather than at the scope.**

After `lambda_elim` every refinement predicate is point-free — `__elem ▷ ((id, 1 ▷ const) ▷ zip ≫ eq)` — and `Encode::expr` reads only `Var` / `Apply(Proj)` / `BinOp` / literals. So **no** post-`lambda_elim` deficit is ever encodable, and each one currently turns a decided mismatch into "outside the supported SMT encoding".

Replacing the `.map_err(...)?` with:

```rust
.or_else(|error| match error {
    // A predicate the encoder cannot read leaves the deficit undecided,
    // which is the pre-SMT answer: a structural mismatch.
    SmtError::Encoding { .. } => Ok(false),
    error => Err(ConstrainError::SmtError {
        lhs: lhs.clone(), rhs: rhs.clone(), error: Box::new(error),
    }),
})?
```

leaves the whole suite green except the two groupby cases, which then fail *because they report the original message again*:

```
post-planning produced an invalid tree: [Type mismatch for Apply: expected {Int | __elem ▷ (...)}, found Int]
```

Falling back to mismatch is exactly what `main` did before this PR: incomplete, never unsound, never worse. It also takes `InferError::Unsupported` off the user-facing path entirely, which is worth doing before anyone ships a diagnostic that reads like a compiler apology.

tests/compilation_pipeline/joins_aggregates_groupby.rs
```diff
@@ -197,7 +197,7 @@ fn test_groupby(#[case] code: &str, #[case] expected: Tile) {
 // rejects the collection domain here, while the keyed-collection branches above reach
 // `post-planning` and an `Apply` mismatch on the same program. So the pin is on the
 // post-pass tree check rather than on either message.
-#[should_panic(expected = "produced an invalid tree: [Type mismatch")]
+#[should_panic(expected = "produced an invalid tree: [Unsupported")]
```
dpmills: **[T1]** Revert to `"produced an invalid tree: [Type mismatch"` once the `SmtError::Encoding → Ok(false)` change lands on `constrain.rs` — I verified the original message comes back verbatim.

Worth noting what this pin was masking: the program here is a genuinely ill-typed one (the known `Int` vs `Int@1` invariant-data-domain defect the comment above describes), and the PR changed it to report an encoder limitation instead of the type error. The comment right above still explains the `expected Int, found Int@1` diagnostic that the new pin no longer produces.

src/ccl/ty.rs
```diff
@@ -2448,6 +2448,7 @@ impl Type {
     pub fn refinements(&self) -> &[Refinement] {
         match self {
             Type::Refinement(_, refinements) => refinements.as_slice(),
+            Type::History { value, .. } => value.refinements(),
```
dpmills: **[T1] Unexercised, and it breaks a law ~30 call sites rely on.**

I deleted this arm and the full suite stayed green — nothing covers it.

`peel_refinements()` did not gain a matching arm, so `refined(peel(t), refinements(t)) == t` no longer holds for a `Mut(V, D)` with refined `V`. `channelize::join_refinements` is the concrete shape: it builds `Type::refined(a.peel_refinements(), ...)` out of `a.refinements()`, which now yields `Refinement(History{value: {Int|p}}, {p})` — `p` at two levels.

The motivating caller is `Encode::assume_refinements`/`sort` on a `Mut`-typed binder. Peel the history locally in `Encode::path_type` and the shared accessor's contract stays intact. (The doc comment just above — "empty for an unrefined type" — was also not updated either way.)

src/ccl/infer/solver/constrain.rs
```diff
@@ -2245,12 +2356,13 @@ mod tests {
             .is_err()
         );
 
-        // Behind a data function, neither refinement direction relates.
+        // Behind a data function, neither refinement direction relates. The
+        // acquisition edge rejects without naming `DataDomainMismatch`: the
```
dpmills: **[T1] The assertion did not need weakening, and the comment explaining it describes an unreachable path.**

`constrain_subtype` passes `SkipSmtScope`, so the deficit returns `Mismatch` *before* the fallback and the caller relabels it `DataDomainMismatch`. I restored the original

```rust
matches!(
    constrain_subtype(&bare, &refined, &mut ConstrainCache::new()),
    Err(ConstrainError::DataDomainMismatch { .. })
)
```

and it passes on this branch as-is.

The design-doc paragraph naming this test was deleted in the same change (`type-inference.md`, the "Both directions is also all it takes… `a_data_domain_relates_only_to_itself` pins all four directions plus the reflexive case" block). Nothing it said stopped being true — please restore it with the assertion.

src/ccl/infer/context.rs
```diff
@@ -169,6 +195,10 @@ impl InferCtx {
         }
     }
 
+    pub fn empty() -> Self {
```
dpmills: **[T1]** `InferCtx::empty()` has zero call sites anywhere in the repo. It is also what carries the `pub(super)` → `pub` widening on `InferCtx` and the `pub use context::InferCtx` re-export in `infer/mod.rs` — all three can go together.

src/ccl/infer/solver/smt.rs
```diff
@@ -0,0 +1,1161 @@
...
+}
+
+/// The whole exchange: declare the query's constants, assert it, ask.
+fn run(
+    ctx: &mut Context,
+    decls: Vec<(String, Sort)>,
+    formula: SExpr,
+) -> Result<Response, SmtError> {
+    for (sym, sort) in decls {
+        // println!("Declaring {} as {:?}...", &sym, &sort);
```
dpmills: **[T1]** Two commented-out `println!`s (here and at `run`'s `ctx.assert`). If they are worth keeping, `log::debug!` — `inline.rs` already pulls `log` in this PR.

src/ccl/infer/solver/smt.rs
```diff
@@ -0,0 +1,1161 @@
...
+/// fast.
+#[derive(Debug, Clone)]
+pub enum Sort {
+    Bool,
+    Int,
+    /// Represented as `Int` where `_ >= 0`.
+    UInt,
+    /// Represented as `Int` where `_ >= 0 && _ < n`.
+    UIntRange(usize),
+    /// A completely uninterpreted sort.
```
dpmills: **[T1] `String` here is not an uninterpreted sort.**

No `declare-sort` is ever emitted — `Sort::UI(s) => (ctx.atom(s), None)` just uses the atom as a sort name. It works only because `String` is z3's **built-in** string-theory sort. Verified directly:

```
$ printf '(declare-const s String)\n(assert (= (str.len s) 2))\n(check-sat)\n' | z3 -in
sat
```

So the doc is wrong, and the variant is not usable for a genuinely uninterpreted sort as written (that would need the `declare-sort` command). The two-line doc correction is the T1 part; renaming the variant can wait.

Related, and cheap to note here: `the_solver_will_error_on_type_errors` asserts on z3's literal wording (`"Sorts String and Int are incompatible"`), which pins the suite to a solver's diagnostic text.

src/ccl/infer/solver/smt.rs
```diff
@@ -0,0 +1,1161 @@
...
+///
+/// A lookup rather than an enumeration, because that is the whole of what the
+/// encoding needs: it meets a free name while translating a predicate and asks
+/// what the scope binds it at. Enumerating instead would declare every binder in
+/// scope, and a binder nothing mentions can still change the answer — a
+/// contradictory one would prove the entailment outright.
+///
+/// The implementors are the two places a query is raised from: inference's
+/// lexical scope (`InferCtx`'s `ScopeStack`, `src/ccl/infer/context.rs`) and the
+/// tree's binders (`TreeScope`, `src/ccl/inline.rs`). [`NoScope`] is the empty
```
dpmills: **[T1]** `TreeScope` does not exist anywhere in the repo — `grep -rn TreeScope --include=*.rs` returns only this line. `inline.rs` does not call `smt_sub` at all in this PR.

With the `smt_sub` fix on `inline.rs`, this becomes `NoScope` and the sentence is nearly right already. `doc_refs` cannot catch this because it is a backticked path in prose, not a link.

src/ccl/design/type-inference.md
```diff
@@ -773,7 +810,100 @@ Singletons are *not* erased after inference. They are ordinary refinements and r
...
+runs on its own refinements too: `x = 2; y = x ^+ 1` reaches `y == 3` by declaring `y`, meeting
+`x` inside `y`'s predicate, and declaring `x` with its own. The binder is declared before its
+refinements are translated, so binders that reference each other terminate.
+
+A lookup and not an enumeration. Declaring every binder in scope would let one nothing mentions
+change the answer — a contradictory binder proves the entailment outright — and it costs a
+declaration per binder on a query that names two. A product's fields are not enumerated either,
+for the same reason.
+
+Three environments implement it:
```
dpmills: **[T1] This paragraph — the one carrying the soundness argument — contradicts the code on all three counts.**

- "Three environments implement it": there are three in `src/`, but they are `NoScope`, `SkipSmtScope`, and `ScopeStack`. `SkipSmtScope` — the one actually wired to `constrain_subtype` — is not mentioned.
- "`NoScope` … what plain `constrain_subtype` supplies": `constrain_subtype` supplies `SkipSmtScope`, whose entire point is the opposite behaviour. `NoScope` is what `constrain_subtype_under` supplies.
- "**`inline`** passes `TreeScope`": neither exists.

The `TreeScope` bullet is a description of the fix this PR should have shipped for `refinement_discharged_by`, so if that lands it becomes true modulo the name. The other two are just wrong today.

.github/workflows/ci.yml
```diff
@@ -107,6 +107,24 @@ jobs:
...
 
+      # The refinement-entailment fallback shells out to a `z3` subprocess
+      # (`easy_smt::ContextBuilder::with_z3_defaults` in
+      # `src/ccl/infer/solver/smt.rs`), so without the binary on PATH every
+      # query fails and the solver slot poisons. Pinned by release and verified
+      # by hash, the same discipline as the elan install above, and for a
+      # sharper reason: z3's answers decide refinement subtyping, so a version
+      # floating with the runner image would move what the type checker accepts
+      # without a commit saying so.
+      - name: Setup Z3
```
dpmills, line 118: **[T2] The CI side is right; the local side has no counterpart.**

Without z3 on PATH, 18 integration tests plus lib tests fail with:

```
called `Result::unwrap()` on an `Err` value: Process { message: "No such file or directory (os error 2)" }
```

Nothing in that names z3. `./ci.sh fast` — the documented inner-loop gate — is red for every contributor who has not read this file, and no README/docs change accompanies the new toolchain requirement.

Minimum before merge: a `ci.sh` guard copying the `lake` precedent already in the file ("skips loudly with no `lake` on PATH; fails instead under CI"), plus a `SmtError::Process` panic message that names z3 and the install step. Both cheap, and this is the biggest onboarding cliff in the PR.

src/ccl/infer/mod.rs
```diff
@@ -333,6 +335,31 @@ pub(super) fn map_constrain_err(err: ConstrainError, ctx_label: &str) -> InferEr
             domains: domains.iter().map(coalesce_for_error).collect(),
             origin: ctx_label.to_string(),
         },
+        ConstrainError::SmtError { lhs, rhs, error } => match error.as_ref() {
```
dpmills, line 348: **[T2] Fine to ship panicking — but say why, because three of four variants abort the compiler.**

- `SolverReportedUnknown` panicking is a bet that the encoded fragment never yields `unknown`. z3 can answer `unknown` for reasons outside this encoder's control.
- `SolverError` panicking crashes the compiler on what the variant's own doc calls "a misencoding on this side" — and `the_solver_will_error_on_type_errors` demonstrates a reachable one (base `String`, predicate `__elem == 5`). The encoder does not type-check its own output before sending it.
- The existing comment ("For now, treat UNKNOWN as a bug. We may later need…") is the right instinct but too small and in the wrong voice for what it is guarding.

The TODO this wants: state that all three are deliberate prototype tripwires, and name the intended policy — most likely `unknown` → undecided → structural mismatch, i.e. the same shape as the `SmtError::Encoding` handling on `constrain.rs`.

House style, while you are here: the new comments in this arm and in `inline.rs` use "we"/"we'll"/"For now". CLAUDE.md: *"State what the code does, in the present tense. No tour, no `we`, no second person."*

src/ccl/infer/solver/smt.rs
```diff
@@ -0,0 +1,1161 @@
+//! Semantic refinement subtyping, discharged to an SMT solver.
```
dpmills, line 1: **[T2] The largest missing comment in the PR: the fragment is defined over surface syntax, and half the pipeline is not surface syntax.**

`Encode::expr` reads `Var` / `Apply(Proj)` / `BinOp` / literals. After `lambda_elim` every predicate is point-free (`▷`, `≫`, `const`, `zip`), so **no** refinement predicate at or after that pass is encodable, and every post-`lambda_elim` check silently falls back to structural matching.

That is a real bound on the feature's reach and it is documented nowhere. It belongs here in the module doc, and as a line in the design doc's "Not yet implemented" bullet — which today claims only that non-LIA fragments are out, implying that LIA ones are in wherever they appear.

src/ccl/infer/solver/smt.rs
```diff
@@ -0,0 +1,1161 @@
...
+    /// The type a value has given `name` here, or `None` when this scope does not
+    /// bind it or nothing has settled it.
+    ///
+    /// Owned, because during emission the slot a binder holds is an inference
+    /// variable and answering means resolving it — there is no settled `Type` in
+    /// the scope to hand out a borrow of.
+    fn binder_type(&self, name: &Name) -> Option<Type>;
+    /// Flag indicating whether a subtype comparison using this
+    /// environment should skip the SMT fallback.
+    fn is_skip_smt(&self) -> bool;
```
dpmills, line 183: **[T2]** This is caller policy squatting on the scope trait. `ScopeEnv`'s own doc a few lines up says lookup "is the whole of what the encoding needs"; `smt.rs` never calls `is_skip_smt` (only `constrain.rs` does); and every implementor — including the test helper `One` — writes a boilerplate `false`.

It is also what produced two near-identical empty scopes in two modules (`NoScope` here, `SkipSmtScope` in `constrain.rs`) differing only in this flag, which is what made the `constrain_subtype` / `constrain_subtype_under` inconsistency easy to miss.

Keep it for now, but the trait doc should say it is a temporary policy channel and name the shape it wants: an argument on `constrain_subtype_in`, or `Option<&dyn ScopeEnv>`.

src/ccl/infer/strip.rs
```diff
@@ -0,0 +1,153 @@
+//! Post-coalesce phase: drop the refinements riding the type slots *inside* a
```
dpmills, line 1: **[T2] The module doc is candid that this erases rather than fixes — it should be explicit that that is a stopgap.**

> it is the copy a substitution leaves stale

The defect is that substitution does not reach interior predicate type slots; this pass deletes the evidence. That is a reasonable trade to ship, but it should read as a decision with the real fix named, not as an observation — otherwise it gets re-derived from scratch later. (CLAUDE.md's "bug fixes that reveal structural problems" case.)

Worth folding into the same comment: `compiled_refinements` is a `Vec<Refinement>` with `contains` on every insert and every query, and `Refinement`'s `PartialEq` is structural over whole predicate terms — O(nodes × slots × |compiled| × predicate size) per compile. Unmeasured, but there is no result memo anywhere in the SMT path either: every deficit is a fresh z3 subprocess round-trip with no caching of `(base, lhs, rhs)`.

src/ccl/lambda_elim.rs
```diff
@@ -1798,7 +1846,7 @@ fn elim_lambdas_impl(ctx: &mut ElimContext, expr: Expr) -> Result<Expr, LambdaEl
         debug_typecheck(e);
         #[cfg(debug_assertions)]
         assert!(
-            original_ty.without_pi_names() == e.ty.without_pi_names(),
+            reclosed_let_ty || original_ty.without_pi_names() == e.ty.without_pi_names(),
```
dpmills, line 1866: **[T2]** This disables the invariant for the new case rather than restating it. Fine to ship; the comment should say that the strong form is "compare `original_ty` against the discharge re-applied" and why that was not done.

Separate open question worth recording next to the new `Let` arm: it splices `def.clone_preserving_ids()` — the *eliminated* definition — into the node's type, so that type now quotes a different term than a structurally-equal predicate elsewhere on the tree quotes. That is the split `tests/predicate_sharing.rs` guards against. Reasoned about, or incidental?

src/ccl/infer/context.rs
```diff
@@ -247,7 +277,7 @@ impl InferCtx {
                 // The result is discarded because a fresh variable cannot conflict
                 // with its first upper bound; a genuine mismatch surfaces later,
                 // when a value flows in and fails against this bound.
-                let _ = constrain_subtype(&v, &bound, &mut ConstrainCache::new());
+                let _ = constrain_subtype_in(&v, &bound, &mut ConstrainCache::new(), &NoScope);
```
dpmills, line 282: **[T2]** `let _ = ...` discards the result here, and a `SolverError` permanently poisons the thread's solver slot (by design — `Slot::Poisoned` is documented as permanent). So an error swallowed at this site leaves every later query failing, and the first one that reaches `map_constrain_err` panics attributing the failure to an unrelated comparison.

One comment at the discard site is enough for now.

tests/compilation_pipeline/type_annotations.rs
```diff
@@ -343,3 +346,285 @@ fn type_annotation_naming_a_parameter_is_unbound(#[case] code: &str, #[case] nam
 fn refined_add() {
     check_scalar("z: {Int where _ == 1 ^+ 3} = 1 ^+ 3\n()", Value::Unit)
 }
+
+#[test]
```
dpmills, line 363: **[T3] One mechanical pass over the ~20 new tests below.**

- **`indoc!`.** Every new program here is a bare `"\ndef foo…"` string starting at column 0, and `collection_product1` escapes its quotes (`\"foo\"`) instead of using `r#""#`. CLAUDE.md is explicit: *"Write multi-line programs with `indoc! {r#"…"#}` … **never** as a single string with embedded `\n`"* — precisely because an off-side-rule language's indentation is invisible in an escaped string.
- **Names.** `refined_add_let`, `refined_add_let2`, `refined_add_let22`, `refined_add_let3`, `refined_add_let4`, `external_scope1`–`4`, `records1`–`5`, and a `collection_int2` with no `collection_int1`. Sequence numbers rather than claims, and most carry no doc comment while the existing tests in this file do. The `smt.rs` unit tests in this same PR are named properly — same treatment here.
- **`collection_int3` and `collection_int4` pin the same thing.** Identical programs but for the annotation bound (`_ <= 10` vs `_ <= 5`); one should be *accepted* once aggregates are supported and one should stay rejected, yet both assert only `", but inferred as Int"`. Neither test can tell them apart, and nothing records which one flips on a fix. A one-line comment on each is enough.
- `records4` (`x == x` on a record at top level) exercises `Equatable`, not annotations or refinements — probably belongs elsewhere.

tests/compilation_pipeline/transactions.rs
```diff
@@ -85,6 +85,29 @@ fn check_compile_error(code: &str, needle: &str) {
     "#},
     40
 )]
+// // Refined transaction bodies
```
dpmills: **[T3]** Two commented-out `#[case]` blocks with no note on what blocks them. Either a TODO naming the blocker, or delete — CLAUDE.md's no-dead-code rule.

tests/type_check.rs
```diff
@@ -4363,45 +4365,17 @@ fn the_alpha_variant_join_is_arrival_order_independent() {
...
-/// This also blocks binding a multi-argument function's parameters rather than
-/// substituting them away — see `src/ccl/design/ir.md`, "A substituted
-/// parameter's site rides its projection". The two are one fix.
 #[test]
-fn let_binder_escaping_into_a_result_refinement_is_ill_scoped() {
-    let errors = infer_program_err(indoc! {r#"
-        def g(n: Int):
-            m = n ^+ 1
-            m ^+ 1
+fn let_binder_does_not_escape_into_result_refinement() {
```
dpmills, line 4396: **[T3]** Two things on the converted test:

- The assertion is a substring probe over `Display` (`contains("n_arg") && !contains("m_let")`), with the program's binders renamed specifically to make the probe work and no doc comment saying so. The design doc states what the type should be — `{Int | __elem == (n ^+ 1) ^+ 1}` — so assert that instead; it is a stronger pin and it survives a `Display` change legibly.
- The deleted `TODO(refinement-let-scope)` recorded two open concerns that do not look addressed: *"nested `let`s compose, so the predicate grows multiplicatively and wants a size bound"*, and *"a binder that cannot be eliminated should widen the type to the unrefined base rather than error"*. They should land somewhere rather than vanish with the test.

Also, the next test's doc comment still opens "The contrast that isolates **the defect above**" — the defect above is gone.

tests/programs/main.rs
```diff
@@ -37,7 +37,6 @@ mod nonneg_inventory;
 mod polymorphic;
 mod prefix_lines;
 mod reachability;
-mod refinement;
```
dpmills, line 40: **[T3]** `tests/programs/refinement/` is deleted outright rather than converted.

Its program was `no_zero_no_one_div(4, 2)` against `{ {Int where _ != 1} where _ != 0}`, pinning the compile error. Under this PR's SMT fallback, `2` satisfies both predicates — so the program should now *compile*. That is a free showcase for the feature, and deleting it loses the flattening-diagnostic coverage too.

tests/compilation_pipeline/type_annotations.rs
```diff
@@ -343,3 +347,374 @@ fn type_annotation_naming_a_parameter_is_unbound(#[case] code: &str, #[case] nam
...
+/// fragment, so the deficit falls back to the structural matching the fallback
+/// exists to get past, and a well-typed program reports an internal invariant
+/// failure.
+///
+/// Widening the encoding past surface syntax retires this pin, as would checking
+/// against the pre-elimination predicate. Once it passes, the program evaluates
+/// `4 // 2`. The rejecting counterpart is `tests/programs/refinement/`.
+#[test]
+#[should_panic(expected = "produced an invalid tree: [Type mismatch")]
+fn an_argument_entailing_two_written_levels_is_rejected_after_lambda_elim() {
```
dpmills: **This pin is honest and finding it was good work — but "two written levels" is not what triggers the class, and the class is a regression in failure mode.**

A single ordinary refinement reaches it:

```
def f(a: Int, b: {Int where _ >= 1}):
    a ^+ b
f(1, 5)
```

```
post-lambda-elim produced an invalid tree:
  [Type mismatch for type of (1, 5): expected {Int | __elem >= 1}, found Int@5]
```

`5 >= 1`. That is a correct program, and it aborts the compiler with an internal invariant failure and no source span.

On the same base (`dbe079211`), the identical program gives:

```
Error: type inference error
   ╭─[<test>:3:1]
 3 │ f(1, 5)
   │ ───┬───
   │    ╰───── Type mismatch for Apply: expected {Int | __elem >= 1}, found Int@5
```

So the trade is: inference gets the verdict right, and the failure mode goes from a spanned user diagnostic to `produced an invalid tree`. For a program that should compile, that is worse than the wrong answer it replaced.

I probed the boundary. It is not arity and not nesting — it is whether the refined parameter slot survives to the post-pass check:

| program | result |
|---|---|
| `def f(a: Int, b: {Int where _ >= 1}): a ^+ b` / `f(1, 5)` | **panic** |
| `def f(a: Int, b: {Int where _ >= 1}): b` / `f(1, 5)` | ok |
| `def f(b: {Int where _ >= 1}): b ^+ 1` / `f(5)` | ok |
| `def f(a: Int, b: {Int where _ >= 1}, c: Int): b` / `f(1, 5, 2)` | ok |
| `def f(a, b: {Int where _ >= 1}, m: Mut(Int)): m := b` | ok |
| `no_zero_no_one_div(4, 2)` (this pin) | **panic** |

Two suggestions, neither large:

1. **Re-scope the TODO** so it names what actually distinguishes the failing shapes. "Two written refinement levels" sends the next reader to the nesting, which is not involved. Adding the two-argument `_ >= 1` case beside this one would make the class visible at the pin.
2. **Consider whether the post-pass check should report a diagnostic rather than an invalid tree here.** That check is a compiler-bug tripwire, and it is firing on a well-typed program because a *different pass* can prove what it cannot. The inconsistency between passes is the fact; the tree is not broken. If the fragment gap ships as-is — which is defensible — the honest failure is "the post-pass check cannot re-derive a semantic entailment", a named limitation with a real message, rather than an internal invariant failure.

To be explicit: nothing I asked for caused this, and it is intrinsic to enabling semantic entailment at inference while the encoder cannot read point-free predicates. Shipping it is a reasonable call. But the PR description should say that a class of valid programs now aborts where they previously reported a type error.

ci.sh
```diff
@@ -270,6 +290,10 @@ ci_all() {
   # Before `ci_test`, so the oracle binary exists by the time the suite runs:
   # the differential tests skip themselves without it, and a skip in the middle
   # of the gate is the failure mode this step exists to remove.
+  # Before `ci_test`, for the reason `ci_solver`'s own comment gives.
+  # shellcheck disable=SC2310
+  # intentional: || captures failure without exiting
+  ci_solver || failed="${failed} solver"
```
dpmills, line 293: **The new gate stole `ci_formal`'s comment block.**

```sh
  # Before `ci_test`, so the oracle binary exists by the time the suite runs:
  # the differential tests skip themselves without it, and a skip in the middle
  # of the gate is the failure mode this step exists to remove.
  # Before `ci_test`, for the reason `ci_solver`'s own comment gives.
  ci_solver || failed="${failed} solver"
  ...
  ci_formal || failed="${failed} formal"
```

`ci_solver` was inserted between `ci_formal`'s explanation and `ci_formal`. The oracle-binary paragraph now reads as a justification for the solver check, and `ci_formal` has none. Move the three-line block back down onto `ci_formal`.

(`ci_solver` itself is correct — verified exit 1 with the install message and exit 0 with z3 on PATH.)

dpmills (commented):
Re-reviewed against `3d8f8a8d5`. **All 23 comments addressed**, and the gates are green: full suite with z3 on PATH, all three clippy configurations, `cargo fmt --check`, `doc_refs`, `shellcheck`, `fixtures`. No leftovers — `strong-post-inference-tests`, `TreeScope`, `Sort::UI`, and `InferCtx::empty` are gone from the tree.

Three landed stronger than what I proposed:

- **`refinement_discharged_by`** — my sketch used `.unwrap_or(false)`, which would have silently turned a missing solver into "precondition not discharged". Splitting it into `Encoding → false` and `panic!("{error}")` for the rest, with a `Display for SmtError` whose `Process` wording carries the install step, is the right call. Both previously-dead tests pass.
- **`map_constrain_err`** — making `SmtError::Encoding` `unreachable!` is better than leaving a diagnostic for a state that cannot arise. Verified it holds: `ConstrainError::SmtError` has exactly one construction site and it filters `Encoding` out. `InferError::Unsupported` pre-existed with other users, so no dead variant.
- **`Type::refinements()`** — the peel moved to `Encode::path_type` with a comment citing the law it preserves and the caller that depends on it, rather than just deleting the arm.

Two resolved by the rebase rather than by edits, correctly: `a_groupby_over_a_singleton_element_literal` is no longer a `#[should_panic]` at all, because #153 on the new main fixed the underlying `Int` vs `Int@1` defect and turned it into a positive `Tile` assertion. The `SmtError::Encoding → Ok(false)` change still stands on its own merits.

`ci_solver` behaves both ways — exit 1 with the install message, exit 0 with z3 present.

Two follow-ups below, one substantive and one mechanical. Neither is caused by the response; the first is intrinsic to the feature and is a judgement call about what ships.

dpmills (approved):

PR #179, opened 2026-09-11
Pin `txn_multi_read`'s answer by reading `pool` with `await_final`

dpmills:
`txn_multi_read` observed `pool` with a bare read fed out of a block that does not write it, an [as-of read at an arbitrary commit position](docs/chl-spec.md#83-reads) the seed satisfies, so the corpus pinned `Function [ 100 ]` as a wrong answer. The program now reads `await_final(pool)`, the only terminal read, and `expect_scalar` pins `Function [ () -> 40 ]`.


tests/programs/txn_multi_read/program.cambra
```diff
@@ -4,6 +4,7 @@ for r in [10, 20, 30]:
     with begin():
         if pool > r:
             pool := pool - r
-with begin():
-    out << pool
+# `await_final` is the only terminal read; a bare `pool` fed out of a block
+# would sample an arbitrary commit position and pin no value.
```
groundlar, line 8: Should this warn / inform the user if they forget `await_final`? This is an easy mistake, something like `NondeterministicRead` is probably an error class we want?

Fine for now, food for followup issue

groundlar (approved):

tests/programs/txn_multi_read/program.cambra
dpmills: Maybe? Something to think about as we add more concurrency stuff later

PR #180, opened 2026-09-11
Parse `a -> b` as the two-tuple `(a, b)`

dpmills:
The map literal `[k -> v, …]` did not parse, so a program spelling a map stopped at `found '->'`. `->` now has a precedence level that builds the `Expr::Tuple` the parenthesised spelling builds, so a map literal, a map comprehension, and `for k -> v in m` are an ordinary list, comprehension, and tuple target; nothing below the parser tells `1 -> 2` from `(1, 2)`. `map([1 -> 10, 2 -> 20])` runs to its entries (`test_map_entry_pairs`).

## Where the pair binds

Between `<<` and the ternary ([2.3 Expression precedence](docs/chl-spec.md#23-expression-precedence)): `m << k -> v` feeds the entry `(k, v)`, and `k -> v if c else w` pairs `k` with the whole conditional. It does not chain, so `a -> b -> c` is a parse error. The ternary's else-branch recurses into `ternary` rather than `expr`, so `k if hot else base -> price` pairs the conditional with `price`, and the ternary still chains right.

## One node, two spellings

A function type miswritten `T -> U` reaches `lower_type_expr`'s `ChlExpr::Tuple` arm, whose message now names `=>` as well as `{T, U}`. `expression()`'s doc comment points at the spec table instead of copying the ladder; that table and `design-chl-parser.md` gain the `=>` row they omitted.

## Checking the advance

`expect_compile_error_past` also asserts the cleared construct is absent, since a needle from the program's own text matches any error on that line. `nonneg_inventory` now stops at `requires Transaction`, and `storefront` v0 and v1 at the membership `in` of `SKU`'s predicate.


src/chl_parser/parser.rs
```diff
@@ -907,32 +894,61 @@ where
...
 
         // ---- Ternary `then if cond else else_` ----------------------
-        let ternary = bool_or
+        // The else-branch is a **ternary**, not the whole expression: that is what chains
+        // (`a if c else b if d else e`, right-associative) while leaving anything looser to
+        // the levels above. Recursing into `expr` here instead put `pair` inside the else,
+        // so `k if hot else base -> price` read as a conditional between a scalar and a
+        // pair rather than as the entry whose key is the conditional — the grouping
+        // `docs/chl-spec.md`, "2.3 Expression precedence" states.
+        let ternary = recursive(|ternary| {
```
sortalongo: Narrowing the else-branch from `expr` to `ternary` changes two existing parses: `a if c else m << 2` regroups, and `a if c else \x -> x` becomes an error. Nothing pins either. Add a test asserting `1 if c else m << 2` is a `Feed` whose target is the conditional.

src/chl_parser/parser.rs
```diff
@@ -907,32 +894,61 @@ where
...
+        // `a -> b` is the two-tuple `(a, b)` (`docs/chl-spec.md`, "2.4 Atoms"), so
+        // it builds the same `Expr::Tuple` the parenthesised spelling does and
+        // nothing below the parser tells the two apart: a map literal is an
+        // ordinary list of pairs, and `for k -> v in m` an ordinary tuple target.
+        //
+        // It binds tighter than `<<`, so `m << k -> v` feeds the entry `(k, v)`,
+        // and looser than the ternary, so `k -> v if c else w` pairs `k` with the
+        // whole conditional. Neither operand reaches the lambda level, so a lambda
+        // on either side takes parentheses. Not chainable, like `feed`: nothing
+        // associates a third component, so `a -> b -> c` is a parse error.
```
sortalongo: Non-associativity is not enforced, it falls out of `or_not` leaving the second arrow unconsumed, so `[a -> b -> c]` reports "expected `,` or `]`" at the arrow instead of naming the chain. Collect `ternary` with `.separated_by(just(Token::Arrow))` and `try_map` an error saying `->` does not chain.

src/chl_parser/parser.rs
```diff
@@ -907,32 +894,61 @@ where
...
+        // ---- Pair `k -> v` (single, not chainable) ------------------
+        //
+        // `a -> b` is the two-tuple `(a, b)` (`docs/chl-spec.md`, "2.4 Atoms"), so
+        // it builds the same `Expr::Tuple` the parenthesised spelling does and
+        // nothing below the parser tells the two apart: a map literal is an
+        // ordinary list of pairs, and `for k -> v in m` an ordinary tuple target.
+        //
+        // It binds tighter than `<<`, so `m << k -> v` feeds the entry `(k, v)`,
+        // and looser than the ternary, so `k -> v if c else w` pairs `k` with the
+        // whole conditional. Neither operand reaches the lambda level, so a lambda
```
sortalongo: Both operands being `ternary` makes `["inc" -> \x -> x + 1]` a parse error while the identical AST spells fine as `[("inc", \x -> x + 1)]`. A map of functions is plausible here. Consider making the value operand admit a lambda.

src/ccl/lower/stmts.rs
```diff
@@ -1508,11 +1508,14 @@ pub(super) fn lower_type_expr(
...
-        // A parenthesised comma list `(T, U)` is a *term* product; the tuple
-        // *type* is written with braces `{T, U}` (`docs/chl-spec.md`).
+        // A parenthesised comma list `(T, U)` is a *term* product; the tuple *type* is
+        // written with braces `{T, U}` (`docs/chl-spec.md`). `->` builds the same node, so
+        // a function type written with the wrong arrow arrives here too and the message
+        // names both — nothing below the parser tells `1 -> 2` from `(1, 2)`.
         ChlExpr::Tuple(_) => Err(LoweringError::unsupported(
             annotation.span,
-            "a tuple type is written with braces: `{T, U}`",
+            "a tuple type is written with braces: `{T, U}`, and a function type with \
```
sortalongo: This message has no test, and it does not fire for the commonest miswriting: `def f(x: Int) -> Int:` dies earlier in `def_stmt`, which wants `=>` before the `:`. Add a lowering test on `Handler = Request -> Response` asserting the message names `=>`.

tests/compilation_pipeline/joins_aggregates_groupby.rs
```diff
@@ -294,6 +294,18 @@ fn test_map(#[case] code: &str, #[case] expected: Vec<(Value, Value)>) {
...
 
+// The entry pair `k -> v` is the two-tuple `(k, v)` (`docs/chl-spec.md`, "2.4
+// Atoms"), so a map literal and a map comprehension reach `map` as the list of
+// pairs the parenthesised spelling builds — nothing below the parser reads the
+// two spellings differently.
+#[rstest]
+#[timeout(Duration::from_secs(30))]
+#[case::literal("map([1 -> 10, 2 -> 20])", vec![(Value::Int(1), Value::Int(10)), (Value::Int(2), Value::Int(20))])]
+#[case::comprehension("map([k -> k * 10 for k in [1, 2]])", vec![(Value::Int(1), Value::Int(10)), (Value::Int(2), Value::Int(20))])]
+fn test_map_entry_pairs(#[case] code: &str, #[case] expected: Vec<(Value, Value)>) {
```
sortalongo: `test_map_entry_pairs` has a body character-identical to `test_map` above it. Add the two cases to `test_map` and delete this function.

tests/programs/storefront/mod.rs
```diff
@@ -92,29 +89,25 @@
...
-//! pins is the blocker behind them:
+//! The variant tags, the refinement braces and the `->` entry pairs of the
+//! `catalog` and `inventory` literals all parse now (`docs/chl-spec.md`,
+//! "3.15 Variant constructors", "6.4 Refinement syntax" and "2.4 Atoms"), so both
+//! files pin the same blocker behind them: `SKU`'s predicate
+//! `_ in catalog.keys()` parses only as far as `in`, which is no expression
+//! operator.  The parser recovers past each statement, so the same run also
+//! reports the other unshipped surfaces listed above — `import`, `static assert`,
+//! `requires`, `with begin():` in value position, and the bare annotation
+//! `orders: Feed(…)`.
```
sortalongo: Nothing asserts this claim. Parser recovery stops well before `/stats` in both files, so the absence of `found '->'` says nothing about those lines. Drop the claim and rely on `entry_pair_rollup_from_the_storefront`, which does test the shape.

tests/programs/common/mod.rs
```diff
@@ -117,6 +117,25 @@ pub fn expect_scalar(source: &str, expected: &str) {
 /// Fires for panics from any pipeline stage (lower, infer, lambda_elim,
 /// operator_conversion, …) — not just lowering.
 pub fn expect_compile_error(source: &str, needle: &str) {
+    let _ = compile_error_message(source, needle);
+}
+
+/// As [`expect_compile_error`], and additionally that the report does **not** mention
+/// `cleared` — the construct this program used to stop at.
```
sortalongo: Worth saying here that `needle` may match the echoed source line, so `cleared` is the load-bearing half. Otherwise a later caller reads the needle as a located assertion. Also, "used to stop at" narrates history.

src/chl_parser/parser.rs
```diff
@@ -907,32 +894,61 @@ where
...
+                            else_expr: Box::new(else_expr),
+                        },
+                    ),
+                })
+        })
+        .boxed();
+
+        // ---- Pair `k -> v` (single, not chainable) ------------------
+        //
+        // `a -> b` is the two-tuple `(a, b)` (`docs/chl-spec.md`, "2.4 Atoms"), so
```
sortalongo: This fact is stated in about eleven places across the spec, the design doc, this comment, the lowering comment, and four test doc comments. Keep the spec as the definition site and cut this to the citation.

src/chl_parser/parser.rs
```diff
@@ -145,23 +145,10 @@ type PErr<'src> = extra::Err<Rich<'src, Token, Span>>;
...
-/// 10. `&`
-/// 11. `++`
-/// 12. `+`, `-`
-/// 13. `*`, `//`
-/// 14. unary `-`
-/// 15. postfix: call `f(...)`, subscript `x[...]`, attribute `x.name` / `x.0`
-/// 16. atom: literal, name, parenthesised, list, record, brace type, comprehension
+/// The precedence ladder is `docs/chl-spec.md`, "2.3 Expression precedence", and the
+/// combinators below are that table read bottom-up — each level built from the one tighter
+/// than it. Stated there and not restated here: a third copy of one ladder is a third thing
```
sortalongo: These two sentences argue for the deletion the change just made. The citation in the sentence before them is the whole content. Delete them.

src/chl_parser/design-chl-parser.md
```diff
@@ -82,21 +82,23 @@ The grammar is precedence-climbed for binary operators. From lowest to
...
-10. `&`
-11. `++`
-12. `+`, `-`
-13. `*`, `//`
-14. unary `-`
-15. postfix: call `f(…)`, subscript `x[…]`, attribute `x.name` / `x.0`
-16. atom: literal, name, parenthesised, list, record, brace type, comprehension
+2. `=>` (function type)
+3. `<<` (feed)
+4. `->` (pair)
```
sortalongo: This hand-renumbered ladder is the second copy the new comment at `parser.rs:150` objects to, so the copy count went from three to two, not to one. Replace the list with a link to the spec table.

docs/chl-spec.md
```diff
@@ -367,24 +368,31 @@ noted:
...
+| 11 | `^` | logical xor |
+| 12 | `&` | logical and |
+| 13 | `++` | collection union |
+| 14 | `+` `-` | additive |
+| 15 | `*` `//` | multiplicative |
+| 16 | unary `-` | prefix |
+| 17 | postfix: `f(args)`, `x[i]`, `x.attr`, `x.0` | left-fold |
+| 18 | atom | literal, name, `(...)`, `[...]`, `{...}`, comprehension |
+
+The pair binds tighter than the feed and looser than the ternary, so
```
sortalongo: This paragraph restates the table three lines above it, and the same two examples appear in the design doc, the parser comment, and three test doc comments. Delete it; the table row already says "pair; non-associative".

docs/chl-spec.md
```diff
@@ -3380,13 +3399,16 @@ with parser-level support that lowering rejects:
...
-  follows the map-literal decision above: `[k -> v for …]`
-  (**[Decided]** as surface, unimplemented; the north-star
-  `storefront` `/stats` rollup uses it).
+  unit syntax, [2.4 Atoms](#24-atoms)). The map literal `[k -> v, …]` parses; reading it
+  as a `Map`
+  without `map([…])` remains **[Decided]**. Earlier plans to spell the entries
+  `[k: v, …]`, `[k=v, …]`, or Unicode `[k ↦ v, …]` are superseded by the
+  map-literal decision.
+- **Map comprehensions** — `[k -> v for …]` parses as a comprehension of pairs
+  ([3.12 Comprehensions](#312-comprehensions)); reading the result as a `Map` follows the map-literal decision
```
sortalongo: Over 100 columns, as is line 3421. The edit also left both bullets ragged. Re-wrap them.

sortalongo (approved):
Approving. The syntax choice is sound. `->` collides with nothing: the lambda commits on `\` before any arrow is seen and its binders are bare identifiers, `match` arms use `case`, `for` headers use `in`, and the function type is `=>`. Binding the pair between `<<` and the ternary gives it the same binding as the comma it desugars to, which is the right reference point. Inline notes below; none blocks the merge.

Three notes not tied to a changed line:

- The ternary rewrite is a real behavior change beyond the stated purpose. `a if c else m << 2` regrouped from `a if c else (m << 2)` to `(a if c else m) << 2`, and `a if c else \x -> x` is now a parse error. Both match the ladder the spec always documented, so this is a fix, but the PR body should say the else-branch narrowed.
- `src/chl_parser/parser.rs:188` is now stale. It says a ternary else-branch and an `<<` right operand "recurse into the whole `expr`"; after this change they recurse into `ternary` and `pair`.
- Design question worth recording rather than answering here: erasing `a -> b` to `Expr::Tuple` leaves a map literal with no syntactic identity. The still-open read of `[k -> v, …]` as a `Map` without `map([…])` can then only be type-directed, and a duplicate-key check on a map literal cannot be syntactic. Worth one line in "3.11 List, tuple, record literals" saying that is a decision rather than a consequence.

On padding: the two-tuple identity and the binding rules are each restated five or more times across the spec, the design doc, inline comments and test doc comments. Same shape as #185.

src/chl_parser/parser.rs
dpmills: Added, plus the second parse you name. `a_ternary_left_of_a_feed_is_the_whole_target` asserts `1 if c else m << 2` is a `Feed` whose target is the conditional, and `a_lambda_else_branch_is_refused` asserts `a if c else \x -> x` is an error.

src/chl_parser/parser.rs
dpmills: Done. `pair` collects its tail with `.repeated()` and reports through `validate`: `a -> b -> c` now says "`->` does not chain: a pair has two components. Parenthesise to nest one pair inside another."

`try_map` does not work here — a failure backtracks out of the whole expression production and the top-level `choice` reports "expected expression" at the second arrow, with the custom error dropped. `validate` emits and recovers, so the arrow is what the diagnostic names. Pinned by `a_pair_does_not_chain`.

src/chl_parser/parser.rs
dpmills: Taken — the value operand admits a lambda, so `["inc" -> \x -> x + 1]` parses. `lambda` moves above `pair` to be reachable there. The key stays at `ternary`: a lambda on that side takes the pair's own `->` as its binder terminator. The spec's precedence row and "2.4 Atoms" both say so, and `a_pair_value_admits_a_lambda` pins it.

src/ccl/lower/stmts.rs
dpmills: Added both: `a_pair_arrow_in_type_position_names_both_spellings` for the message, and `a_pair_arrow_in_a_def_return_annotation_is_a_parse_error` for the miswriting that never reaches it — `def_stmt` reports `found '->', expected '=>'` before lowering sees the annotation.

tests/compilation_pipeline/joins_aggregates_groupby.rs
dpmills: Folded in as `arrow_literal` and `arrow_comprehension`; `test_map_entry_pairs` is gone. The comment above `test_map` now says what the two spellings share.

tests/programs/storefront/mod.rs
dpmills: Dropped. The module doc now points at `entry_pair_rollup_from_the_storefront` for the rollup's shape and says recovery stops before those lines, so this run says nothing about them.

tests/programs/common/mod.rs
dpmills: Both done — the doc says `cleared` is the load-bearing half and why, and "used to stop at" is gone.

src/chl_parser/parser.rs
dpmills: Cut. What is left is the citation plus the two facts that are about this code rather than about the language: that nothing below the parser tells `k -> v` from `(k, v)`, and why the value admits a lambda where the key cannot.

src/chl_parser/parser.rs
dpmills: Deleted.

src/chl_parser/design-chl-parser.md
dpmills: Replaced with the link. `**`'s straddle of the unary minus stays, as a parser fact: it is why that level shares a `recursive` layer with unary `-` rather than sitting in its own.

docs/chl-spec.md
dpmills: Deleted. The lambda rule it also carried moves onto the table row, which is the one copy.

docs/chl-spec.md
dpmills: Re-wrapped, both bullets.

dpmills:

**The ternary rewrite.** Both regroupings are now pinned rather than only stated: `a_ternary_left_of_a_feed_is_the_whole_target` for `1 if c else m << 2`, and `a_lambda_else_branch_is_refused` for `a if c else \x -> x`. The description says the else-branch narrowed.

**`parser.rs:188`.** Corrected — the arm-body list drops the ternary and `<<`, neither of which recurses into `expr` any more.

**The map literal's syntactic identity.** Recorded as a decision in the spec rather than left to be inferred, and the `->` value operand now admits a lambda, so `["inc" -> \x -> x + 1]` parses.

**Padding.** The ladder now has one copy: `design-chl-parser.md`'s hand-renumbered list and the spec paragraph that restated the table are both links to it, and the lambda rule rides the table row rather than a second paragraph.

PR #181, opened 2026-09-11
Add `**` exponentiation and require its exponent to be non-negative

dpmills:
CHL has no exponentiation operator, and a negative exponent has no integer value. `**` adds one and demands a non-negative exponent in the type, so `2 ** -1` is a compile error rather than a runtime division by zero. `ArithmeticKind::Pow` runs from a new `StarStar` token through lowering, `Trait::Exponentiable`, and an interpreter `IntPow` by squaring.

## Which exponents compile

Lowering routes `**` through `pow_with_checked_exponent` (`src/ccl/lower/exprs.rs`), binding the exponent to a fresh `__pow_exp` annotated `{Int | __elem >= 0}`. Inference discharges it, so a literal passes on its own refinement and an unrefined `Int` does not: `2 ** -1`, `0 ** -1`, `2 ** -64`, and `2 ** e` for `e = 0 - 1` all fail on `__elem >= 0` (`tests/compilation_pipeline/scalars_collections.rs`). An annotated parameter `{Int where _ >= 0}` carries the proof into its body. `**` answers a bare `Int`, so its own result cannot be another exponent; right-associativity is pinned in the parser tests. A `String` base is a missing `Exponentiable` instance (`tests/type_check.rs`).

`TODO(check-time-smt)` bounds this: `Check` decides refinements structurally, so the demand cannot yet live in `**`'s signature. The binding costs one thing: a `**` inside a refinement predicate becomes a `Let`, leaving the SMT fragment as a term rather than an unencodable operator.

## Precedence and rendering

`**` straddles unary minus: tighter than the one on its left (`-2 ** 2` is `-(2 ** 2)`), looser than the one on its right (`2 ** -1` parses). The two share a parser layer. `symbolic.rs` gains `binop_assoc`: `**` is the only operator grouping to the right. Spec: [2.3](docs/chl-spec.md#23-expression-precedence), [3.3](docs/chl-spec.md#33-arithmetic-and-logical-operators).


src/interpreter/binop.rs
```diff
@@ -26,6 +26,60 @@ pub enum ArithmeticKind {
...
+/// reciprocal.
+trait IntPow: Copy + MulAssign {
+    /// The multiplicative identity, which seeds [`Self::raised`] and is what
+    /// `a ** 0` yields for every `a`.
+    const ONE: Self;
+
+    /// `self` raised to a non-negative `exponent`, by squaring, so the cost is
+    /// logarithmic in `exponent` and overflow arrives through the same `*` that
+    /// [`ArithmeticKind::Mul`] uses.
+    fn raised(mut self, mut exponent: u64) -> Self {
```
sortalongo: Blocker. `acc *= self` and `self *= self` are plain multiplications, and the release profile sets no `overflow-checks`. `2 ** 64` panics in debug and yields `0` in release. Use `wrapping_mul` so both profiles agree.

src/interpreter/binop.rs
```diff
@@ -26,6 +26,60 @@ pub enum ArithmeticKind {
...
     Mul,
     FloorDiv,
+    Pow,
+}
+
+/// Integer exponentiation, the scalar behind [`ArithmeticKind::Pow`].
+///
+/// A separate trait because `**` has no `*Assign` operator to bound
+/// [`zip_arithmetic`]'s element type by, and because the two element types
+/// differ on negative exponents: `usize` has none, and `i64`'s take the
```
sortalongo: This says `i64` exponents "take the reciprocal", in the present tense. They no longer can. The comment at line 69 states the same rule as history ("the reciprocal it used to compute"). Replace both with the current invariant: the exponent is non-negative by typing, so no reciprocal case exists.

src/ccl/ops.rs
```diff
@@ -94,6 +94,13 @@ pub enum ArithmeticKind {
     Mul,
     /// Floor division (`//`).
     FloorDiv,
+    /// Exponentiation (`**`).
+    ///
+    /// A negative exponent takes the reciprocal through the same division `//`
```
sortalongo: This documents the opposite of what ships, and cites the spec section this PR rewrites. A negative exponent is now a type error, not `1 // (a ** n)`. Replace the paragraph with the non-negativity invariant and cite `pow_with_checked_exponent`.

docs/chl-spec.md
```diff
@@ -383,19 +389,26 @@ noted:
 | 14 | `+` `-` | additive |
 | 15 | `*` `//` | multiplicative |
 | 16 | unary `-` | prefix |
-| 17 | postfix: `f(args)`, `x[i]`, `x.attr`, `x.0` | left-fold |
-| 18 | atom | literal, name, `(...)`, `[...]`, `{...}`, comprehension |
+| 17 | `**` | exponentiation; *right*-associative |
```
sortalongo: This table row places `**` one level tighter than unary `-`, which asserts that `2 ** -1` does not parse. The prose at line 401 asserts the opposite, and the parser agrees with the prose. Correct the table.

docs/chl-spec.md
```diff
@@ -383,19 +389,26 @@ noted:
...
 The pair binds tighter than the feed and looser than the ternary, so
 `m << k -> v` feeds the entry `(k, v)` and `k -> v if c else w` pairs `k` with
 the whole conditional. Neither of its operands reaches the lambda, so a lambda
 on either side is parenthesised.
 
+`**` binds tighter than the unary `-` on its left and looser than the one
+on its right: `-2 ** 2` is `-(2 ** 2)`, and `2 ** -1` parses without
+parentheses — and is then rejected for its exponent's sign, which is a
+typing rule and not a grammatical one. `**` groups to the right, so
+`2 ** 3 ** 2` is `2 ** (3 ** 2)`.
```
sortalongo: `2 ** 3 ** 2` does not compile. A `**` result is a bare `Int`, which no exponent position accepts, so the associativity rule is illustrated with a rejected program. State that a `**` chain is rejected by the exponent's typing rule.

src/ccl/lower/exprs.rs
```diff
@@ -568,13 +568,70 @@ pub(super) fn lower_binop(
...
+/// fallback. An annotation is narrowed by inference, so `Check` compares the refinement
+/// against itself and matches. Fold this back into the operator's signature once `Check`
+/// raises its own queries.
+///
+/// The binding costs one thing worth naming: a `**` written *inside* a refinement predicate
+/// makes that predicate a `Let`, so it leaves the SMT fragment as a term rather than as an
+/// unencodable operator. `**` in a predicate was already undecidable — SMT-LIB has no
+/// integer exponentiation — so nothing that worked stops working, but the reason a reader
+/// meets changes, and folding this back into the signature restores it.
+fn pow_with_checked_exponent(
```
sortalongo: Blocker. The annotation minted here is what `[2 ** x for x in [1, 2, 3]]` trips: inference discharges it, the post-inference wall does not see the discharge, and the wall panics rather than reporting a type error. A `for` loop with the same body is fine. Carry the comprehension source's element refinement into the binder's post-inference type.

src/ccl/lower/exprs.rs
```diff
@@ -568,13 +568,70 @@ pub(super) fn lower_binop(
...
     Ok(Expr::binop(left_expr, kind, right_expr))
 }
 
+/// `a ** b` with the exponent bound through a **non-negative annotation**.
+///
+/// `**` requires a non-negative exponent: a reciprocal has no integer value, so rather than
+/// give `a ** -n` one, the exponent has to carry `{Int | __elem >= 0}` and a program that
+/// cannot show it is rejected where it is written.
+///
+/// TODO(check-time-smt): stated as an annotated binding rather than in the operator's own
```
sortalongo: Two notes. The TODO names check-time SMT as the precondition, but `TraitInstance.args` is `&'static [BaseType]`, so a trait row cannot carry a predicate either; the `Exponentiable` doc already says so. Name both preconditions. Separately, consider whether this demand can ride the exponent node's `user_annotation` rather than a synthesized binding — the binder is what puts a `Let` inside refinement predicates, which is the one consequence the PR documents.

tests/compilation_pipeline/scalars_collections.rs
```diff
@@ -129,10 +129,54 @@ fn test_collection_param_consumed(#[case] code: &str, #[case] expected: Value) {
...
+/// The demand is **strict** — an `Int` that carries no such refinement is rejected, not
+/// admitted — so what compiles is what a program can show.
+#[rstest]
+#[timeout(Duration::from_secs(10))]
+#[case::literal("2 ** -1")]
+#[case::zero_base("0 ** -1")]
+// Large enough that the old reciprocal overflowed `i64` on its way to dividing by zero.
+#[case::overflowing("2 ** -64")]
+// Not a literal, so nothing bounds it: an unrefined `Int` cannot show itself non-negative.
+#[case::unrefined("e = 0 - 1\n2 ** e")]
```
sortalongo: This case and the one at line 175 embed multi-line CHL with `\n`. CLAUDE.md requires `indoc! {r#"…"#}` for programs with line structure.

src/ccl/symbolic.rs
```diff
@@ -247,10 +251,25 @@ fn fmt_inner(expr: &Expr, opts: &SymbolicOpts) -> (Precedence, String) {
...
             let op_prec = binop_prec(op);
             let sym = op.sym();
-            // Left at same prec is fine (left-associative).
-            let l = fmt(left, op_prec, opts);
-            // Right needs one level tighter to avoid right-association.
-            let r = fmt(right, op_prec.next_highest(), opts);
+            // The side the operator groups towards keeps a child of equal
+            // precedence unparenthesised; the other side parenthesises one, so
+            // the rendering re-parses with the grouping it was built from.
+            let (l_prec, r_prec) = match binop_assoc(op) {
```
sortalongo: `Associativity::Right` computes `l_prec`, and the next statement overwrites it for `Pow`, so the table contributes one live behavior while the straddle is a hardcoded arm beside it. Associativity and the straddle are one fact split in two. Give `binop_prec` a left/right binding-power pair per operator and drop `Associativity`.

sortalongo (approved):
Approving so this is not gated on review. **Two items below block merge — please do not merge until both are fixed.**

The core invariant holds. I could not construct a path that puts a negative integer into an exponent: probes covered a subtraction result, an unannotated parameter, a loop accumulator, a `Mut(Int, Txn)` store read, a match payload, and a collection element, and every one is rejected. Discharge is semantic rather than syntactic and correct in both directions. The parser matches the spec on the straddle and on right-associativity.

### Blocker 1: a comprehension exponent panics the compiler

`[2 ** x for x in [1, 2, 3]]` panics with `post-inference produced an invalid tree: [Type mismatch for Apply: expected {Int | __elem >= 0}, found Int]`. Inference accepts the program, and the post-inference wall then treats its own rejection as a compiler bug rather than a user error. The same program written as a `for` loop compiles and runs.

The root cause predates this PR: an annotated function called from a comprehension panics identically with no `**` involved. What changes here is reachability. The panic is now reachable from a program containing no refinement syntax at all, which makes it a shipping defect for this operator.

Fix: carry the comprehension source's element refinement into the binder's post-inference type so `check_pre_channelize` sees what inference discharged.

### Blocker 2: debug and release disagree on overflow

`IntPow::raised` uses plain multiplication, and `[profile.release]` sets no `overflow-checks`. So `2 ** 64` panics in debug and evaluates to `0` in release, and `2 ** 63` wraps to `i64::MIN`. #186 landed six days ago to stop debug and release being two different compilers; this reintroduces that class in the interpreter.

Fix: use `wrapping_mul` in `raised`.

### Not blocking

Inline below. The largest is that `ArithmeticKind::Pow`'s doc comment and two comments in `binop.rs` state the reciprocal rule this PR removes, so the one place that defines the operator tells the next reader the opposite of the invariant.

Test gaps worth closing: `0 ** 0` is untested although the spec claims `a ** 0` is `1` for every base; the case named `overflowing` is rejected at type-check, so it exercises nothing that `2 ** -1` does not; `2 ** 0` is duplicated with the same expected value.

src/interpreter/binop.rs
dpmills: `wrapping_mul`, via an `IntPow::mul_wrapping` the two impls supply, so `2 ** 64` is `0` in both profiles. Pinned by `exponentiation_wraps`, which also takes `2 ** 63` to `i64::MIN`.

The sign `debug_assert!` became an `assert!` in the same spirit: `unsigned_abs` turns a negative exponent into a large positive one and answers, so a release build past it is a wrong number rather than a crash.

`zip_arithmetic`'s `+ - * //` are still the plain operators and still diverge, so `**` now agrees across profiles and `*` does not. Filed as the vault issue `interpreter-integer-arithmetic-divergences`, together with `//` truncating where the spec says floor; PR 182 extracts the kernel those two rules would be stated in.

src/interpreter/binop.rs
dpmills: Both replaced with the invariant. The trait doc says the exponent is non-negative by typing and that the two element types differ only in whether that has to be stated; `raised` says what it does now (wrapping) and why.

src/ccl/ops.rs
dpmills: Rewritten to the non-negativity invariant, citing `pow_with_checked_exponent`.

docs/chl-spec.md
dpmills: Corrected — the row says `**` straddles the unary `-` at 16 rather than sitting under it, and points at the paragraph.

docs/chl-spec.md
dpmills: Rewritten. The paragraph now states that `2 ** 3 ** 2` parses as `2 ** (3 ** 2)` and does not compile, because a `**` answers a bare `Int` and no exponent position takes one — and that `(2 ** 3) ** 2` does, the grouping being what the rejection is about rather than the associativity. `an_exponent_not_shown_non_negative_is_rejected` gains a `chain` case.

src/ccl/lower/exprs.rs
dpmills: Fixed, and it reached further than `**`. Three changes:

1. **`compact_go` takes a discharged demand's refinements at a positive position.** It read the opposite side there only for the shape collapse, so `[1, 2, 3]`'s three singletons met at a bare `Int` and the `{Int | __elem >= 0}` each of them discharged separately had nothing left in the type to show for it. The demand survived solving, so it holds of the value; it unions on only where the value described itself with none, a value carrying its own predicate having been what discharged the demand.
2. **`Check` raises its own semantic queries** — which this TODO named as the precondition. `CheckCtx` implements `ScopeEnv` off the binder scope `Typing::scoped` already handed it and discarded.
3. **An undecided deficit means different things at the two derivations.** A point-free predicate is outside the encoded fragment, and a pass-boundary re-derivation reading that as a mismatch reports the encoder's limit rather than anything about the tree. `refinements_that_survive_to_post_planning_check_are_rejected` pinned that defect and is now the passing `a_point_free_refinement_is_trusted_at_a_pass_boundary`.

`[2 ** x for x in [1, 2, 3]]` evaluates to `[2, 4, 8]`; `[2 ** x for x in [1, 2, -3]]` is still a type error. Both pinned.

src/ccl/lower/exprs.rs
dpmills: Both preconditions named, and the second half taken.

The demand now rides the exponent node's `user_annotation`, so there is no binder and no `Let` inside a refinement predicate — the one consequence the PR documented is gone. `TypedExpr::user_annotation` is a live inference input (`emit_node` reconciles it and `api.rs` clears it), so nothing new was needed for it.

The TODO is renamed `TODO(pow-signature)` and says what is still missing: `Check` now raises its own queries, so that half is discharged, but `TraitInstance.args` is `&'static [BaseType]` and an `Exponentiable` row carries no predicate. Folding the demand into the signature waits on a trait row that can.

tests/compilation_pipeline/scalars_collections.rs
dpmills: Converted. `2 ** 0` is also deduplicated — `test_arithmetic`'s copy is gone, the pow block owns it — `0 ** 0` is added, and the `overflowing` case is replaced: it tested nothing `2 ** -1` did not, and the overflow it was named for is now `exponentiation_wraps`.

src/ccl/symbolic.rs
dpmills: Done. `binop_prec` and `binop_assoc` collapse into one `binop_binding` returning `Binding { own, left, right }`; `Associativity` is gone. The side an operator groups towards is the side whose threshold equals its own level, and `**`'s straddle is the same pair stating two different levels — which one number plus an associativity cannot say between them, which is why the straddle was a hardcoded arm.

dpmills:
Changes pushed as a commit on top, so the threads still point at what you read. Both blockers are closed.

**Blocker 1** reached further than `**`, as you said. The trace: `[2 ** x for x in [1, 2, 3]]` was accepted by inference — `[2 ** x for x in [1, 2, -3]]` is a proper type error, so the discharge is happening per element — and the mismatch was at the comprehension's element read. `[1, 2, 3]` coalesces to `[0, 2] ⤇ Int`, the join dropping the three singletons, while the mapped lambda's domain carries `{Int | __elem >= 0}`. Three changes, each stating a rule rather than patching the case:

1. **A positive position takes a discharged demand's refinements.** `compact_go` read the opposite side at a positive position only for the shape collapse. A demand that survived solving is a fact about the value, so it unions on — and only where the value described itself with none, a value carrying its own predicate having been what discharged the demand. Without that restriction every position with a demand grows a predicate the re-derivation cannot reproduce; with it, six tests that were failing pass again.
2. **`Check` raises its own semantic queries.** This is the `TODO(check-time-smt)`'s own precondition, and the plumbing was already there: `Typing::scoped` hands `CheckCtx` the binder's type and it was discarding it. `CheckCtx` now implements `ScopeEnv`, so `{Int | __elem == t + 3}` is dischargeable against what `t` is known to be.
3. **An undecided deficit means different things at the two derivations.** A point-free predicate is outside the encoded fragment and the encoder reads neither side, so a pass-boundary re-derivation reporting a mismatch reports its own limit. `refinements_that_survive_to_post_planning_check_are_rejected` pinned exactly that and is now the passing `a_point_free_refinement_is_trusted_at_a_pass_boundary`.

`[2 ** x for x in [1, 2, 3]]` evaluates to `[2, 4, 8]`.

**Blocker 2.** `wrapping_mul`, and the sign check is an unconditional `assert!` for the same reason — `unsigned_abs` would answer rather than crash. `zip_arithmetic`'s `+ - * //` still diverge between profiles, so `**` agrees and `*` does not; that class, with `//` truncating where the spec says floor, is the vault issue `interpreter-integer-arithmetic-divergences`. PR 182 extracts the kernel those rules would be stated in.

**Thread 7's second half** is taken: the demand rides the exponent node's `user_annotation`, so there is no binder and no `Let` inside a refinement predicate.

**Test gaps** all closed — `0 ** 0`, the duplicate `2 ** 0` removed, and `overflowing` replaced by `exponentiation_wraps`, which pins `2 ** 64` at `0` and `2 ** 63` at `i64::MIN`.

Fixtures: `arithmetic` and `defer_lift` renumber, one class — Check's queries clone predicate terms and a clone freshens ids. Both trees are otherwise identical, 107 and 149 nodes before and after, with no label, type or tiling string changed.

src/ccl/lower/exprs.rs
dpmills: Backed out in "Leave pass-boundary checking as main has it". The fix changed three general mechanisms: `compact_go` carrying a discharged demand's refinements onto a positive position, `Check` raising its own semantic queries, and a pass-boundary re-derivation trusting a predicate the encoder cannot read. Those belong with the loose-check fix happening separately rather than in this PR, and the three depend on each other, so none of them stays.

`[2 ** x for x in [1, 2, 3]]` panics at the post-inference check again, pinned as a defect by `a_comprehension_exponent_reaches_the_wall`. The same panic reproduces with no `**`, from an annotated function called in a comprehension. Leaving this thread open until that fix lands.

PR #182, opened 2026-09-11
Fold closed scalar computations so a collection literal's elements are constants

dpmills:
Op conversion builds a collection literal's binding table out of its elements, so an element that is still an application has no value to put there, and it rejects `sum([500 * one_dollar])`. A new planning pass replaces a closed scalar computation with the literal it computes: a `Var` an enclosing `Let` binds to a literal, a binop over two literals, `neg` or `not_fn` over one. The walk is bottom-up, so one pass folds a whole chain.

## What does not fold

`src/ccl/planning/const_fold.rs` states each bound: a compound constant, whose substitution would copy nodes and need minted `NodeId`s; a negative `//`, where the spec's floor division and the runtime's truncation disagree; and an operation with no result. Arithmetic is duplicated from `apply_binop_column` because `ccl` may not depend on `interpreter`.

## Tests, docs and fixtures

New pipeline cases cover a folded element, a chain, a shadowing binder and two elements the fold declines, plus boolean operators with a per-element operand that the fold leaves for run time. Expectations that pinned an operator chain over literals now pin the literal, and the `arithmetic` and `polymorphic` snapshots shrink where a folded subtree collapses to one `Lit`. [`optimization.md`](src/ccl/design/optimization.md#planning-cclplanning) now lists all six planning rewrites in run order, three of which it had omitted.


src/ccl/planning/const_fold.rs
```diff
@@ -0,0 +1,274 @@
...
+        // CCL's `let` is non-recursive: the binding scopes over `body` and not over
+        // `bound_expr` (`scope::for_each_scoped_item` states every such rule). The bound
+        // expression therefore folds in the enclosing environment.
+        // **A definition the body's type discharges does not fold.** Where the binder is
+        // free in the body's type, the recorded type of this `Let` was closed over the
+        // definition as it stands (`close_let_type` in `src/ccl/infer/check.rs`), and the
+        // post-planning wall re-runs that discharge over whatever `bound_expr` is by
+        // then. Folding it would put two spellings of one value either side of a
+        // structural comparison.
+        let discharged = crate::ccl::subst::type_free_vars(&body.ty).contains(&binding.name);
```
sortalongo: The `discharged` guard skips the whole `bound_expr` subtree rather than the definition node, so an unrelated downstream `^+` stops an otherwise-foldable element compiling. `y = sum([1 + 2, 3])` with `z = y ^+ 1` fails with "constant folding did not reduce this one", while the same program with `z = y + 1` returns 7. The same expression was rejected before this pass, so this narrows a gap rather than opening one, and the module doc states the bound at lines 34 to 36. Nothing pins it. Add a test.

src/ccl/planning/const_fold.rs
```diff
@@ -0,0 +1,274 @@
...
+    match (op, left, right) {
+        // `^+` computes the sum `+` computes; they differ in the type the result takes,
+        // and the node keeps its recorded type through the fold.
+        (BinOpKind::Arithmetic(op), Lit::Int(l), Lit::Int(r)) => Some(Lit::Int(match op {
+            ArithmeticKind::Add | ArithmeticKind::AddRefined => l.checked_add(*r)?,
+            ArithmeticKind::Sub => l.checked_sub(*r)?,
+            ArithmeticKind::Mul => l.checked_mul(*r)?,
+            // Only where floor division and truncation coincide — see the module
+            // docs. The bound also excludes a zero divisor.
+            ArithmeticKind::FloorDiv if *l >= 0 && *r > 0 => l / r,
```
sortalongo: Declining here is right. The divergence underneath is a live wrong answer: `(0 - 7) // 2` returns -3 at run time where `docs/chl-spec.md`, "3.3 Arithmetic and logical operators" says -4. File it as its own change rather than leaving it described only here.

tests/compilation_pipeline/scalars_collections.rs
```diff
@@ -232,6 +232,70 @@ fn a_re_keying_constructor_rejects_an_empty_literal(#[case] code: &str) {
...
+        xs = [1, 2]
+        sum([sum(xs), 3])"});
+}
+
+/// The second way the fold declines: both operands are literals, and folding would answer
+/// a question `docs/chl-spec.md`, "3.3 Arithmetic and logical operators" (floor division)
+/// and the runtime (truncation toward zero) disagree on. Settling that disagreement is
+/// what makes this element foldable.
+#[test]
+#[should_panic(expected = "constant folding did not reduce this one")]
```
sortalongo: Two of the three non-folds the description names have no test. Nothing covers a compound constant or an operation with no result; grepping `tests/` for overflow and for division by zero returns nothing. Add a `#[should_panic]` case for each, beside this one.

src/ccl/planning/const_fold.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// The longest string concatenation this pass will build.
+///
+/// Folding trades compile-time memory for run-time work, and concatenation is where that
+/// trade turns over: the result is as large as both operands together, it is stored in
+/// `Consts`, and it is cloned at every occurrence the substitution reaches.
+const MAX_FOLDED_STRING: usize = 64 * 1024;
+
+fn eval_binop(op: BinOpKind, left: &Lit, right: &Lit) -> Option<Lit> {
```
sortalongo: The duplication the description explains is avoidable. `ccl` and `interpreter` are modules of one crate, so a third module both import keeps the layering and removes the second implementation. The shared piece is the scalar kernel that `zip_arithmetic` maps over a column and this applies to one pair. The `//` disagreement is what a second implementation already costs.

src/ccl/planning/const_fold.rs
```diff
@@ -0,0 +1,274 @@
...
+            ArithmeticKind::Add | ArithmeticKind::AddRefined => l.checked_add(*r)?,
+            ArithmeticKind::Sub => l.checked_sub(*r)?,
+            ArithmeticKind::Mul => l.checked_mul(*r)?,
+            // Only where floor division and truncation coincide — see the module
+            // docs. The bound also excludes a zero divisor.
+            ArithmeticKind::FloorDiv if *l >= 0 && *r > 0 => l / r,
+            ArithmeticKind::FloorDiv => return None,
+            // The runtime raises by squaring through the same `*` that `Mul` uses
+            // (`IntPow::raised`), so folding an exponentiation that overflows would
+            // answer where the runtime does not. A negative exponent is the reciprocal
```
sortalongo: "A negative exponent is the reciprocal `1 // (a ** n)`" is unreachable at this head. The exponent is non-negative by type and `IntPow::int_pow` asserts it. Say that `u32::try_from` guards overflow.

src/ccl/planning/const_fold.rs
```diff
@@ -0,0 +1,274 @@
...
+//! arithmetic case goes through a guard or a `checked_*`, and declining leaves the
+//! application in place for the runtime to evaluate as it would have.
+//!
+//! **Types.** A refinement predicate is a shared `Rc` whose structural identity is what
+//! matches a producer's contract against its consumer's
+//! (`src/ccl/design/type-inference.md`, "Sharing is an invariant, not an optimization
+//! detail"), so folding one occurrence of a predicate and not another would break the
+//! match. The folded node keeps its recorded type.
+//!
+//! **A definition the body's type discharges.** The post-planning wall compares recorded
```
sortalongo: This bullet repeats the one at line 29, same exclusion and same `close_let_type` mechanism. Delete it and move the `{Int | __elem == 4}` example into the first.

tests/compilation_pipeline/joins_aggregates_groupby.rs
```diff
@@ -354,10 +354,15 @@ fn a_bare_groupby_tail_is_driven() {
     "1:{Int | __elem ▷ ((id, 1 ▷ const) ▷ zip ≫ eq)}",
     Tile::Scalar(ColumnValue::Ints(vec![1]))
 )]
-#[case("1 + 2", "(1, 2) ▷ add:Int", Tile::Scalar(ColumnValue::Ints(vec![3])))]
+// Closed arithmetic reaches op conversion as the literal it computes, a whole chain of
+// it in one planning pass (`src/ccl/planning/const_fold.rs`). What these two used to pin
```
sortalongo: This narrates the change rather than the code, and only parses against the deleted expectation. State what the two cases pin now.

src/interpreter/operator_conversion.rs
```diff
@@ -2666,9 +2669,17 @@ fn expr_to_value(expr: &Expr) -> Result<Value, ConversionError> {
             tag: FieldKey::Name(tag.as_str().into()),
             inner: Box::new(expr_to_value(payload)?),
         }),
+        // The element may well *be* a constant and still arrive here: what reaches this
```
sortalongo: The exclusion list appears in this comment and again in the error string within eleven lines, and a third time in `const_fold.rs`'s module doc. Shorten the error and leave the enumeration to the module doc.

sortalongo (approved):
Approving. The pass is careful where it matters. Every integer operation goes through a `checked_*`, `//` folds only where floor division and truncation coincide, and concatenation is capped. The exclusions are an allowlist, so a new builtin defaults to unfolded. Nothing is restamped and no `NodeId` is minted, which is why the fixture shrink is a clean subtree collapse rather than a renumbering.

Nothing below blocks the merge. The two worth acting on are the untested non-folds and the floor-division divergence, which this PR correctly declines to fold around but leaves described rather than filed.

Two notes not tied to a line:

- The description should quote the `ConversionError::Unsupported` message it rewrites, before and after, per the requirement added in #213. This repo squash-merges, so this body becomes the commit message.
- The fixture diff has a second class the description does not name: `polymorphic`'s node ids renumber. `web/CLAUDE.md` asks for every diff class in the message.

src/ccl/planning/const_fold.rs
dpmills: Pinned, both halves: `a_downstream_refined_sum_stops_an_element_folding` for the `^+` program and `an_unrefined_downstream_sum_leaves_the_element_foldable` for the same program under `+`, which answers 7.

src/ccl/planning/const_fold.rs
dpmills: Filed as the vault issue `interpreter-integer-arithmetic-divergences`, together with the debug/release overflow split that `**` just fixed for one operator. The module doc and the test that pins the non-fold both cite it, and the doc now states the wrong answer rather than only the disagreement: `(0 - 7) // 2` gives -3 where the spec says -4.

tests/compilation_pipeline/scalars_collections.rs
dpmills: Both added beside it: `an_element_with_no_result_is_rejected` (an overflowing `*`, and `// 0`) and `an_element_built_from_a_compound_constant_is_rejected`.

src/ccl/planning/const_fold.rs
dpmills: Extracted. `src/scalar_ops.rs` holds the operator enums and the vectorized kernel, below both `ccl` and `interpreter`, so the layering holds without a second copy.

The vectorized form is the primitive and `const_fold` calls it on a one-element column, rather than a scalar kernel the column maps — the ~15% note is on the column loop, and folding one pair is on no hot path. `eval_binop` is now a guard and a delegation: `runtime_answers` is the exclusion list, which is the whole of what the pass decides, and the value comes from the kernel the `BinOp` operator will run.

The operator mapping collapses too. `impl From<ccl::BinOpKind> for scalar_ops::BinOpKind` is the one place the two spellings meet, and `builtin_to_binop` is the `Builtin` unwrap around it rather than a second enum-to-enum copy.

src/ccl/planning/const_fold.rs
dpmills: Rewritten, and the arm changed with it. `**` now folds where it declined: the runtime raises by squaring through `wrapping_mul`, so a result past `i64` is a wrapped value rather than no value and the fold answers the same one. What the guard excludes is a negative exponent, which the kernel asserts on.

src/ccl/planning/const_fold.rs
dpmills: Merged into the first, `{Int | __elem == 4}` example and all.

tests/compilation_pipeline/joins_aggregates_groupby.rs
dpmills: Rewritten to what they pin: a symbolic form with no operator left in it. The cases below keep `add`, `eq` and `gt` because a binder is among their operands.

src/interpreter/operator_conversion.rs
dpmills: Shortened to the citation; the enumeration is the module doc's.

dpmills:

**The second implementation is gone.** `src/scalar_ops.rs` holds the operator enums and the vectorized element-wise semantics, below both `ccl` and `interpreter`, which keeps the layering `builtin_to_binop` states without a second copy. The vectorized form is the primitive and `const_fold` calls it on a one-element column rather than the reverse — the ~15% note is on the column loop, and folding one pair is on no hot path. `eval_binop` is a guard plus a delegation now: `runtime_answers` is the exclusion list, which is the whole of what the pass decides.

The operator mapping collapsed with it. `impl From<ccl::BinOpKind> for scalar_ops::BinOpKind` is the one place the two spellings meet, and `builtin_to_binop` is the `Builtin` unwrap around it.

One behavior change falls out: `**` folds where it declined. The runtime raises by squaring through `wrapping_mul` (PR 181), so a result past `i64` is a wrapped value rather than no value, and the fold answers the same one.

**The floor-division divergence** is filed as the vault issue `interpreter-integer-arithmetic-divergences`, together with the debug/release overflow split in `zip_arithmetic`. One issue because one kernel decides both.

**The untested non-folds** are covered: `an_element_with_no_result_is_rejected` (an overflowing `*`, `// 0`) and `an_element_built_from_a_compound_constant_is_rejected`, plus `a_downstream_refined_sum_stops_an_element_folding` and its `+` counterpart for the `discharged` guard.

**The description** now quotes the `ConversionError::Unsupported` message before and after, and names both fixture classes — the subtree collapse and the handful of new high ids, which is not a wholesale renumbering: in `arithmetic` 73 of 107 ids survive unchanged, in `polymorphic` 100 of 115.

PR #183, opened 2026-09-11
Take an empty collection's types from its use site, and add `empty_map()`

dpmills:
`emit_list` fixed `[]`'s element type to `unit` at emission, so an annotation naming another element type met a mismatch, and no term spelled an empty map. `[]` now emits the fresh element variable a non-empty literal gets, a pin chooses it only where nothing demands a type, and `empty_map()` names the empty map and set: `m: Mut(Map(String, Int), Txn) := empty_map()` seeds a store its writers fill.

## Where the element type is chosen

Zero elements is the join rule at zero: `emit_list` drops its early return, so the element variable stays open. `pin_empty_list_elements` runs at the top of `coalesce_pass`, as its own walk over term nodes before `coalesce_node` starts — not inside that walk, whose `Apply` arm descends into `function` before `argument` and so reads a comprehension's lambda parameter before the literal. Per node, `pin_empty_list_element` returns when `value_reaches` the codomain, else equates it with `payload_pin`'s choice (`Unit` when the uses require nothing) through `constrain_subtype` both ways, recorded on the variable.

## The empty map states its own sum

`Builtin::EmptyMap` lowers at `Type::map_of(Type::Hole, Type::Hole)`, both parameters left to the annotation; `Set(K)` is `Map(K, unit)`, so one term answers both. `map([])` and `set([])` stay rejected — [The empty collection has no key morphism](src/ccl/design/collections.md#the-empty-collection-has-no-key-morphism). `operator_conversion` makes it a producer: `Constant::collection` at the extent read from the type.

## Tests and docs

`empty_list_takes_its_element_type_from_the_use_site` covers annotation, operator read, and `++` join; `a_re_keying_constructor_rejects_an_empty_literal` pins the rejection; `an_empty_map_seeds_a_transactional_store` asserts entries only writers put there. [The empty literal names no element type](src/ccl/design/collections.md#the-empty-literal-names-no-element-type) and [3.11 List, tuple, record literals](docs/chl-spec.md#311-list-tuple-record-literals) state the rule.


src/ccl/ops.rs
```diff
@@ -273,6 +273,19 @@ pub enum Builtin {
     /// exactly — `box(5)` is `Σ (σ : [5]). σ`, not `Σ (σ : [Int]). σ`. Retaining the
     /// alternatives rather than joining past them is the whole service.
     Box,
+    /// `empty_map : Σ (σ : SubtypesOf(𝐾)). σ ⤇ 𝑉` — the collection with no entries.
```
sortalongo: Mark this `[Interim]` and name what retires it.

`empty_map()` asserts the witness kind `SubtypesOf` because `TypeKind` has no variable to leave it open. That absence is justified on `TypeKind::refuses` by a premise the empty literal breaks: the kind is decidable from the classified type, but zero rows witness no kind. A type-kind variable is what retires this term.

Suggest `TODO(witness-kind-var)` here, citing `src/ccl/design/collections.md`, "The empty collection has no key morphism".

src/ccl/design/collections.md
```diff
@@ -78,6 +78,62 @@ witness, and each is the unboxed form of the sum below it in the list. Crossing
...
+`Case` arm's payload takes
+([type-inference.md, An unobservable arm payload is pinned to what its uses
+require](type-inference.md#an-unobservable-arm-payload-is-pinned-to-what-its-uses-require)).
+
+Emptiness is the premise the pin rests on, not a stand-in for "no value reached the element type".
+A non-empty literal whose elements are themselves undetermined — `\x -> [x, x]`, never called —
+has a value-free element type too, and there the variable is a type parameter the program left
+ambiguous; pinning it would accept a program that has no type.
+
+### The empty collection has no key morphism
```
sortalongo: Retitle this `### The empty collection has no key morphism [Interim]` and state the retiring condition in the section: a type-kind variable, which means lifting the assumption stated on `TypeKind::refuses` in `src/ccl/ty.rs`.

docs/chl-spec.md
```diff
@@ -1093,7 +1118,7 @@ is the unit type, written `{}` (§6.6). `[]` is the empty list.
 > | finite map | `[k -> v, …]` — a collection of entry pairs (§2.4); `[ … ]` is the collection delimiter |
 > | `[1, 2, 3]` — list | same spelling, but shared across collection types: the literal can denote an `Array`, `List`, or `Set`, disambiguated by annotation or usage, with `list([…])` / `set([…])` constructors for explicitness (**[Tentative]** — §6.3) |
 > | empty record / unit | `()`, the unit value: with no fields there is nothing to tell a record from a tuple, so an empty record, an empty tuple, and unit coincide (§3.1, §6.6) |
-> | `[]` — empty list | same spelling; the empty-**map** spelling is **[Open]** (§2.4) |
+> | `[]` — empty list | same spelling; the empty **map** is `empty_map()` ([3.11 List, tuple, record literals](#311-list-tuple-record-literals)) |
```
sortalongo: This row moves the empty-map spelling from `[Open]` to settled. It is not settled. It rests on `TypeKind` carrying no variable, and on the `Set`/`Map` distinction that is still open. Mark it `[Tentative]`.

src/ccl/infer/solve.rs
```diff
@@ -820,8 +820,71 @@ fn pin_unobservable_arm_payload(p: &Pattern) -> bool {
...
+    if value_reaches(codomain) {
+        return;
+    }
+    let chosen = payload_pin(codomain);
+    let mut cache = ConstrainCache::new();
+    let pinned = constrain_subtype(&chosen, codomain, &mut cache)
+        .and_then(|()| constrain_subtype(codomain, &chosen, &mut cache));
+    assert!(
+        pinned.is_ok(),
+        "pinning an empty list's element variable cannot fail: its only bounds are what \
```
sortalongo: This assertion is reachable, so a type error becomes a compiler panic.

`payload_flow_target` takes the first upper bound that resolves concretely. The premise that the remaining bounds all still accept the choice holds for an unreachable `Case` arm payload, which is where the rule came from. It does not hold for `[]`, which is ordinary reachable code that any number of use sites can constrain. `xs = []` read once at `Int` and once at `String` returns `Err` here.

Return the `Result` and push a `LocatedInferError` onto `ctx.errors` instead of asserting. A test with two conflicting use sites would pin it.

src/ccl/infer/solve.rs
```diff
@@ -897,6 +960,20 @@ pub(super) fn coalesce_pass(expr: &mut Expr) -> Vec<LocatedInferError> {
...
     };
+    // **Every empty literal's element is chosen before the walk starts.** The choice is
+    // recorded on the element *variable*, and that variable also stands in the binder slots
+    // the literal feeds — a comprehension's lambda parameter, a `for` target — so a read
+    // that precedes the pin resolves one occurrence against the unpinned graph and leaves
+    // it unresolved at the wall. The walk cannot reach the literal first in general: the
+    // `Apply` arm descends into `function` before `argument` on purpose, so a comprehension
+    // over an inline `[]` reads the lambda's parameter before it ever sees the list.
+    //
+    // Here rather than at emission because a type asserted before the constraints arrive is
```
sortalongo: Thirteen comment lines above one call, restating what `pin_empty_list_element`'s own doc comment already says. Cut to the sentence about `Apply` descending into `function` before `argument`, which is the only fact not stated elsewhere.

src/ccl/infer/solve.rs
```diff
@@ -820,8 +820,71 @@ fn pin_unobservable_arm_payload(p: &Pattern) -> bool {
...
+/// The empty literal denotes the function with no positions, so nothing can read
+/// its codomain, and the type language has no uninhabited type to name that with
+/// (`docs/chl-spec.md`, "6.6 The empty product is unit"). A type is chosen here on
+/// the rule an unreachable arm's payload takes: whatever the position's uses
+/// require, and `Unit` when they require nothing
+/// (`src/ccl/design/type-inference.md`, "An unobservable arm payload is pinned to
+/// what its uses require").
+///
+/// **Emptiness is the premise, not a shortcut for it.** A non-empty literal whose
+/// elements are themselves undetermined — `\x -> [x, x]`, never called — has a
```
sortalongo: This argument and its `\x -> [x, x]` example also appear in `src/ccl/design/collections.md`, "The empty literal names no element type". Delete it here and cite the heading.

docs/chl-spec.md
```diff
@@ -1082,6 +1084,29 @@ form; the literal parses, and `map([…])` is what reads it as a `Map`
...
+empty set**: it takes no arguments, and its key and value types come from the
+annotation on what it seeds.
+
+```python
+cart: Mut(Map(String, Int), Txn) := empty_map()
+prices: Map(String, Int) = empty_map()
+seen: Set(String) = empty_map()
+```
+
+`Set(K)` is `Map(K, unit)` ([6.3 Direction: collection types [Decided]](#63-direction-collection-types-decided)), so one
```
sortalongo: 120 columns. Prose in a `.md` wraps at 100.

sortalongo (approved):
Approving, with two things to fix before merge.

The mechanism is right. The element type becomes an ordinary inference variable the normal solver resolves, the pin reuses `payload_pin` rather than inventing a second rule, and no test is ignored or pinned to panic.

**Mark `empty_map()` as `[Interim]`.** What separates a list from a map is the `TypeKind` classifying the sum's witness, `UIntRanges` against `SubtypesOf(K)`, and `TypeKind` has no variable. Its own doc comment says that absence is deliberate, on the premise that kind membership "is decidable from the classified type, so the only thing inference can be missing is that type". The empty literal is where that premise fails: `[]` gets a classified type, so its kind is decided as `UIntRanges`, but zero rows witness no kind. `empty_map()` exists to assert the witness kind the type language cannot leave open. That is a stopgap for a missing kind variable rather than a permanent term, and the marker should say so and name what retires it.

The name is a second, smaller concession. `empty_map()` answers a `Set(String)` annotation too, which works only because `Set(K)` is `Map(K, unit)`. That rests on the already-open `Set`/`Map` distinction, whose candidate fix is a nominal type constructor. A nominal split would force `empty_map()` and `empty_set()`.

**The pin's assertion is reachable.** Two conflicting use sites for one `[]` panic the compiler instead of reporting a type error. Detail inline.

The remaining comments are minor.

src/ccl/ops.rs
dpmills: Marked, with the retiring condition stated: the term asserts the witness kind because `TypeKind` has no variable, and the absence is justified on `TypeKind::refuses` by a premise `[]` breaks — membership is decidable from the classified type, but `[]` has a classified type and zero rows witness no kind. `TODO(witness-kind-var)`, citing the heading.

src/ccl/design/collections.md
dpmills: Retitled, and the section states both concessions: the type-kind variable that retires the term, and the name, which answers a `Set` annotation only because `Set(K)` is `Map(K, unit)` — a nominal constructor would force `empty_set()` apart from it.

docs/chl-spec.md
dpmills: Marked `[Tentative]`, with both reasons in the row.

src/ccl/infer/solve.rs
dpmills: Fixed — it pushes a `LocatedInferError` and the walk continues. `an_empty_literal_read_at_two_types_is_a_type_error` pins it, with two annotations rather than two operator reads: a trait obligation conflicts at emission, where an annotation's upper bound waits for a lower bound an empty literal never supplies. The report is a type inference error blamed on `xs = []`.

src/ccl/infer/solve.rs
dpmills: Cut to that sentence.

src/ccl/infer/solve.rs
dpmills: Deleted, and the heading cited.

docs/chl-spec.md
dpmills: Re-wrapped.

dpmills:

Both items done: `empty_map()` is `[Interim]` in all three places, with the type-kind variable named as what retires it and the `Set`/`Map` concession recorded alongside; and the pin reports instead of asserting, pinned by `an_empty_literal_read_at_two_types_is_a_type_error`.

One thing that test turned up. Two conflicting *operator* reads never reach the pin — `x + 1` and `y == "a"` over one `xs = []` conflict at emission as a trait rejection, with a good user-facing message. It takes two conflicting annotations, because an annotation's upper bound waits for a lower bound an empty literal never supplies. That is what the test uses.

PR #185, opened 2026-09-12
Compare the statement shapes a discharge can capture, and rename the predicate walk for terms

dpmills:
The type-blind structural comparison behind `Refinement`'s `PartialEq` has two further callers, and the widest asks it about terms rather than predicates: `Subst::eq_modulo_ty_slots` compares a discharge's captured argument, any expression the program wrote. It had no arm for the statement shapes, so a captured body carrying a mutable accumulator compared unequal to a structural copy of itself — the reflexivity loss its fallthrough `debug_assert!` names. This adds those arms, renames the pair to `eq_term_modulo_ty_slots` / `hash_term_modulo_ty_slots`, and asserts at predicate-install sites that a predicate holds none of them.

## The new arms

`MutDecl` and `For` bind over their bodies as `Let` does, so each pairs its binder and compares references by position; `MutWrite`'s written name resolves through that pairing like a `Feed` name; `Begin` recurses. `hash_term_modulo_ty_slots` gains the matching scoping, without which two terms the new `eq` equates would hash apart.

## What the two guards check

`debug_assert_predicate_shape` checks at `Refinement::born` and `MemoStore::point_at` that a predicate holds none of those shapes, reaching a cast's target because the comparison does. `eq_has_arm` classifies every `TypedExprNode` variant with no `_`, so a variant added later stops the build instead of reaching the catch-all and comparing unequal to its own copy. Why `LetRec` and `Transact` are left out is on that catch-all arm in `src/ccl/ty.rs`, and nowhere else.

## Tests and docs

Four cases in `src/ccl/ty.rs`: two pin `eq` and hash on α-variants of a mutable-accumulator loop and of a keyed write under a `Begin` transaction block, two pin that the predicate-shape guard fires, at the top level and one refinement deep; [Interior term binders stay named, and compare by position](src/ccl/design/type-inference.md#interior-term-binders-stay-named-and-compare-by-position) records the interior-binder rule; the rest of the diff is the rename, `formal`'s doc comments included.


src/ccl/design/type-inference.md
```diff
@@ -1411,11 +1411,26 @@ reference it (`subst::codomain_depends_on`) and strips it elsewhere.
...
 parameter stays a name. That binder sits inside the refinement rather than outside it, so the
 compared terms carry it
-and the correspondence is local: `eq_refinement_predicate` threads a pairing and compares a
-reference to it by its binder's position, and `hash_refinement_predicate` hashes it by position so
+and the correspondence is local: `eq_term_modulo_ty_slots` threads a pairing and compares a
+reference to it by its binder's position, and `hash_term_modulo_ty_slots` hashes it by position so
 the `Eq`/`Hash` contract survives. The name stays the stored form, so the predicate is still a term
 planning can compile.
 
+`eq_term_modulo_ty_slots` answers for terms, not only for predicates. `Refinement`'s `PartialEq`
```
sortalongo: rm cruft

src/ccl/ty.rs
```diff
@@ -3175,6 +3175,70 @@ pub const REFINEMENT_BINDER: &str = "__elem";
 /// ([`TraitInstance`](crate::ccl::infer::solver::traits::TraitInstance)).
 pub type RefinementTemplate = fn(&[TypedExpr]) -> TypedExpr;
 
+/// A refinement predicate is a **boolean expression over
+/// [`crate::ccl::REFINEMENT_BINDER`]**, so it holds none of the statement shapes a
+/// program body is built from: a mutable variable introduction or write, a loop, a
+/// statement sequence, or a recurrence carrier (`LetRec`, `Transact`).
```
sortalongo: `Begin` is the transaction block, not a statement sequence. `ExprStmt` is the statement chain, and it is on the permitted list. Name the node types here instead of describing them.

src/ccl/ty.rs
```diff
@@ -3175,6 +3175,70 @@ pub const REFINEMENT_BINDER: &str = "__elem";
...
+///
+/// What makes it a restriction rather than a preference: the mutability-elimination
```
sortalongo: Minor. This paragraph argues for the design. State the restriction and where it is asserted, and cut the rest.

src/ccl/ty.rs
```diff
@@ -3244,23 +3309,32 @@ impl PartialEq for Refinement {
...
+/// of [`hash_term_modulo_ty_slots`]. It compares node shape, scalar leaves
+/// (operators, builtins, literals, names, tags), binder names, and child `Expr`s
+/// pairwise, but never the embedded `Type`s (`ty` slots, annotations, binding
+/// types). Those are inference metadata, and copies of one term legitimately differ
+/// in them (freshened for a specialization, rebuilt by a discharge, pinned at
+/// different use types). So this is a structural relation chosen for its context,
+/// not an approximation of a finer one — there is no plan to make it a derived `==`
+/// (which would wrongly compare those slots).
+///
+/// **Three callers, one relation, and the domain is the widest of the three.**
```
sortalongo: Minor. Announce sentence. Delete it.

src/ccl/ty.rs
```diff
@@ -3471,8 +3545,52 @@ pub(crate) fn mentions_witness(ty: &Type, binder: &WitnessId) -> bool {
...
+        | N::Compose(_)
+        | N::Copair(_)
+        | N::DisjointJoin(_)
+        | N::ExprStmt { .. }
+        | N::Feed { .. }
+        | N::Define { .. }
+        | N::Defer
+        | N::Begin { .. }
+        | N::Error => true,
+        // The two a term this walk compares never holds; the catch-all says why.
```
sortalongo: Minor. Broken sentence.

src/ccl/ty.rs
```diff
@@ -3660,44 +3834,61 @@ fn eq_refinement_predicate_go(
...
-            // shape, and reporting `false` makes a predicate compare unequal to a
-            // structural copy of itself — reflexivity, quietly lost. It surfaces far away
-            // as a type mismatch whose two sides print identically, since the predicate is
-            // the one part of a type `Display` does not show.
+            // shape, and reporting `false` makes a term compare unequal to a structural
+            // copy of itself — reflexivity, quietly lost. It surfaces far away as a type
+            // mismatch whose two sides print identically, since a predicate is the one
+            // part of a type `Display` does not show.
+            //
+            // `LetRec` and `Transact` are the shapes deliberately left out. A
```
sortalongo, line 3842: Keep this comment, and make it the only place that explains why `LetRec` and `Transact` have no arm. The design-doc paragraph and the PR description section say the same thing.

src/ccl/ty.rs
```diff
@@ -3660,44 +3834,61 @@ fn eq_refinement_predicate_go(
...
-/// last, exactly as [`eq_refinement_predicate`] carries its pairing; a binder's
+/// last, exactly as [`eq_term_modulo_ty_slots`] carries its pairing; a binder's
 /// own name is never hashed, so hash stays coarser than `eq` and the
 /// `Eq`/`Hash` contract holds.
-fn hash_refinement_predicate<H: std::hash::Hasher>(
+///
+/// The generic walk covers the two shapes `eq` has no arm for, hashing their children
+/// in the enclosing scope. That direction is the safe one: `eq` answering `false` makes
+/// it finer, and a hash coarser than `eq` is what the contract asks for. Neither is
+/// reachable — this hashes predicates only, through [`Refinement`]'s `Hash`.
```
sortalongo: The test helper `term_hash` runs this on non-predicates, so this sentence is not true. Delete it. The coarser-hash point above it stands on its own.

src/ccl/ty.rs
```diff
@@ -4347,6 +4566,143 @@ mod tests {
...
+        // loop's source makes a different term.
+        assert!(!eq_term_modulo_ty_slots(&a, &build("total", "x", "ys")));
+    }
+
+    /// The keyed write and the statement sequence, the two arms
+    /// [`alpha_variant_terms_compare_and_hash_alike`] does not reach. A keyed write is the
+    /// shape `transact_phase` normalizes away, so it exists in exactly the window this walk
+    /// runs in, and its key is a term the comparison reads.
+    #[test]
+    fn a_keyed_write_under_a_sequence_compares_its_key() {
```
sortalongo: `Begin` is not a sequence. Rename the test.

src/ccl/design/type-inference.md
```diff
@@ -1411,11 +1411,26 @@ reference it (`subst::codomain_depends_on`) and strips it elsewhere.
...
+asks it about a predicate, where it is the same-restriction relation, and `planning::groupby`'s
+pointful-site match asks it about two collection paths drawn from one. `Subst::eq_modulo_ty_slots`
+asks it about a discharge's captured argument, which is any expression the program wrote, a
+function body included. Its domain is the widest of the three, so every node a program builds has
+an arm — the mutable-variable introduction and the loop among them, each binding over its body the
+way a `let` does.
+
+`LetRec` and `Transact` have no arm. A predicate never acquires one, because the phases that mint
```
sortalongo: The absence of these two arms does not need justifying in the design doc. Drop this paragraph. The comment on the catch-all arm in `ty.rs` is the one place to say it.

sortalongo (approved):
Approving. The code, the tests, and the rename are fine. The comments around the new arms say more than they need to. Inline notes below; none blocks the merge.

Two notes that are not tied to a line:

- "statement shape" is new vocabulary. It is in the PR title, a panic message, two test names, and the design doc, and nothing defines it. Define it once or say "statement node".
- The PR description section on why `LetRec` and `Transact` have no arm can be one clause. The comment on the catch-all arm covers it, and the description becomes the commit message on squash.

src/ccl/ty.rs
dpmills: Named instead of described: the doc now says it holds no `MutDecl`, `MutWrite`, `For`, `Begin`, `LetRec` or `Transact` — the six `offending` matches, and `ExprStmt` is not among them.

src/ccl/ty.rs
dpmills: Cut to one sentence: where the assert runs, and the one fact that makes the placement necessary (the mutability-elimination phases do not rewrite inside a predicate, so the report would otherwise come from a wall two passes later). The `inline` example and the cast-target paragraph are gone — the latter is already the inline comment at the code that does the descent.

src/ccl/ty.rs
dpmills: Deleted.

src/ccl/ty.rs
dpmills: Fixed: "The two shapes a term this walk compares never holds".

src/ccl/ty.rs
dpmills: Kept, and it is now the only place. The two design-doc paragraphs are gone, and the PR description no longer has that section.

src/ccl/ty.rs
dpmills: Deleted. You are right — `term_hash` runs it on the terms the α-invariance tests build, so "this hashes predicates only" was false.

src/ccl/ty.rs
dpmills: Renamed to `a_keyed_write_under_a_transaction_block_compares_its_key`, and its doc says "the `Begin` transaction block" rather than "the statement sequence".

src/ccl/design/type-inference.md
dpmills: Dropped, along with the `LetRec`/`Transact` paragraph below it. Both restated the `ty.rs` doc comments; the catch-all comment and `eq_term_modulo_ty_slots`'s doc are the one place each now. I also reflowed the paragraph above, where an earlier edit had left a 22-column line mid-paragraph.

src/ccl/design/type-inference.md
dpmills: Dropped.

dpmills:
Rebased onto the new `main` (`03736b652`), and all nine items are in `2da08571d` — replies on each thread.

The PR description's `LetRec`/`Transact` section is gone too, so the catch-all comment in `ty.rs` is the only place that explains it.

PR #186, opened 2026-09-14
Enforce three debug-gated compiler checks in every build

dpmills:
Three compiler checks ran only under `debug_assertions`, each pinned by a `#[should_panic]` test carrying no `cfg` of its own — so each test was red under release, where the check does not fire, and release was a second compiler accepting trees debug rejects.

All three become unconditional: the record-time closure check (`enforce_bound_scope` and its gap set), `TagMap::from_arms`' duplicate-tag scan, and planning's `__pair`-binder check on a compiled predicate term. Each site records what it costs.

Release-profile failures go from four to one. The remaining one, `committed_fixtures_match_fresh_dumps`, is a separate pre-existing defect: debug and release mint different numbers of `NodeId`s.


dpmills:

- #186 ◀

groundlar (approved):

PR #187, opened 2026-09-14
Take a zip's binder from whichever operand names one

dpmills:
`zip_pair_ty` read the pair's binder off its first operand alone, so a zip whose dependent operand is the second one built an unnamed function whose codomain still references a binder — the shape `subst::open_codomain` rejects, one line after the `Apply` arm builds the type. The pair now takes its binder from either operand, under two assertions checked where the type is built; its doc comment carries why the binder is read off the operands where the kind is declared. A correlated filter beside a correlated body consequently reaches planning and is refused there over a surviving `__pair` binder, pinned in `tests/compilation_pipeline/comprehensions.rs`.

Two neighbouring gaps are pinned rather than fixed. A refined function operand still drops the pair to `Type::Hole`, which `a_zip_reads_through_a_refined_operand` asserts against under `#[should_panic]`. The comment in `src/ccl/simplify.rs` names two sites that drop a binder the same way; an assertion in each fires on nothing in the suite, so both are latent.


dpmills:

- #187 ◀

src/ccl/lambda_elim.rs
```diff
@@ -238,6 +238,16 @@ pub(crate) fn substitute(expr: Expr, name: &Name, replacement: &Expr) -> Expr {
...
+///
+/// **The binder is either operand's.** Both are morphisms over the one domain the pair
+/// shares, so either may be the dependent one, and the pair's codomain holds both of
+/// theirs: a binder the second alone names still has a reference in that codomain and so
+/// still has to ride it. Reading only the first leaves an unnamed function whose codomain
+/// references a binder, and [`crate::ccl::subst::open_codomain`] then has no binder to open
+/// that reference at: it debug-asserts on the shape one line later in the `Apply` arm, and
+/// returns the codomain unresolved where assertions are compiled out. A correlated filter
+/// beside a correlated body produces that shape. Where both name one, they name the same
+/// one: each is this elimination's own `param`.
```
groundlar: **Design question — why is the binder read off an operand at all?**

The paragraph directly above this one argues that the *kind* must be **declared** rather than read off `f`, because reading it off `id` is what flattens a collection's columns to capabilities. The binder now has the same failure history — this PR is the bug report — and the fix keeps reading it off, plus an assertion that the two readings agree.

Every elimination-arm caller has the binder in hand: it is `param`, as the new assertion's own message says. Threading it as an argument, exactly as `fun_kind` already is, makes the rule unfalsifiable and deletes the need for the assertion:

```rust
pub(crate) fn zip_pair_ty(f: &Expr, g: &Expr, binder: Option<&Name>, fun_kind: &FunKind) -> Type
```

The counter-argument is presumably that the two `simplify.rs` callers have no binder to pass. If so, that belongs in this doc — and it is also how you'd surface the issue I raise on `simplify.rs:915`, because those callers would have to say `None` out loud.

Per CLAUDE.md's "when in doubt, ask": this is a non-trivial call about where a fact lives, and the paragraph above already decided the analogous question the other way.

src/ccl/lambda_elim.rs
```diff
@@ -238,6 +238,16 @@ pub(crate) fn substitute(expr: Expr, name: &Name, replacement: &Expr) -> Expr {
...
 pub(crate) fn zip_pair_ty(f: &Expr, g: &Expr, fun_kind: &FunKind) -> Type {
     match (&f.ty, &g.ty) {
```
groundlar, line 254: **Nit (pre-existing, flagging because the fix is inside this match).** This matches bare `Type::Fun`, so a refined function operand `{(k: A) ⇒ B | p}` falls through to `Type::Hole` and loses both the binder and the kind.

The codebase already named this hazard for kinds — `Type::fun_kind` (`src/ccl/ty.rs:2247`) peels refinements precisely because "a refined function (`{Fun | p}`) still carries a kind, and a match on the bare type drops it". Out of scope here; noting it so it isn't found a third time.

src/ccl/lambda_elim.rs
```diff
@@ -247,10 +257,19 @@ pub(crate) fn zip_pair_ty(f: &Expr, g: &Expr, fun_kind: &FunKind) -> Type {
...
+                name: g_name,
+                codomain: c,
+                ..
+            },
         ) => {
+            assert!(
+                !matches!((name, g_name), (Some(x), Some(y)) if x != y),
+                "a zip's operands are morphisms over one domain, so a binder either of them \
+                 names is that domain's: {name:?} and {g_name:?}",
+            );
```
groundlar, line 272: **Nit — `assert!` vs `debug_assert!`, and the message wording.**

*On the choice:* the neighbouring binder machinery uses `debug_assert!` (`subst::open_codomain`, `subst::close_pi_binder`). This one is always-on and panics the compiler in release on an invariant that is structurally guaranteed — `elim_lambda_kinded`'s `Apply` and `Compose` arms recurse on both operands with the same `param`, and `result_ty` only ever names the Pi `param`, so it cannot fire. `assert_ne!` at line 730 is precedent for the other choice, so it's a judgment call; just worth making deliberately.

*On the message:* it parses correctly as written — `names` is the verb ("a binder **that** either of them names"), and `that domain's` elides a second `binder`. But it garden-paths: "a binder either of them names" reads as a noun pile until you back up and re-parse. And the trailing `: {name:?} and {g_name:?}` dumps two values without saying what they are or that they conflict — the message names the invariant but never states the violation.

```rust
"a zip's operands are morphisms over one domain, so a binder either operand \
 names is that domain's binder; these name two: {name:?} and {g_name:?}",
```

Spelling out the elided `binder` after `that domain's` is what unsticks the parse; `those names` would break it, since `names` has to stay the verb.

src/ccl/lambda_elim.rs
```diff
@@ -247,10 +257,19 @@ pub(crate) fn zip_pair_ty(f: &Expr, g: &Expr, fun_kind: &FunKind) -> Type {
...
             let codomain = Type::Tuple(vec![*b.clone(), *c.clone()]);
```
groundlar, line 273: **Assert the postcondition, not only the precondition.**

The new `assert!` states that `f` and `g` don't name *different* binders. That holds by construction at every in-tree caller (both elimination arms recurse at the same `param`; both `simplify.rs` callers pass `name: None`), so it documents rather than catches.

What actually went wrong is an **output** shape: an unnamed function whose codomain references an enclosing function. Today that is caught one call later, in `subst::open_codomain`, under a `debug_assert`. Asserting it where the type is built names the invariant where it lives — and CLAUDE.md calls out "pass output shapes" as exactly where a defensive check earns its keep:

```rust
let codomain = Type::Tuple(vec![*b.clone(), *c.clone()]);
debug_assert!(
    name.is_some()
        || g_name.is_some()
        || !crate::ccl::subst::references_enclosing_function(&codomain),
    "a pair with no binder carries a codomain referencing one: the index has no \
     binder to open at, so nothing downstream can resolve it",
);
```

I ran this against the full suite: **zero failures**. It is a free, always-true-today check that would have caught the original bug here rather than in `subst`. It is also the invariant `src/ccl/design/type-inference.md`, "A binder reference is stored in one of two forms" already states.

src/ccl/lambda_elim.rs
```diff
@@ -1892,6 +1912,37 @@ mod tests {
         Expr::var(s)
     }
 
+    /// A zip whose **second** operand is the dependent one keeps the binder.
+    ///
+    /// The pair's codomain holds both operands' codomains, so a reference the second
+    /// carries is in the pair's codomain too, and an unnamed function whose codomain
+    /// references a binder is what `subst::open_codomain` rejects — the `Apply` arm calls it
+    /// on this very type one line after building it.
```
groundlar, line 1935: **A CHL-level test is available** — `check_compile_error` (`tests/compilation_pipeline/helpers.rs:133`) exists for exactly this case, and its doc says so: "Use this for negative tests and **for program features that have only been partially implemented**." It `catch_unwind`s the pipeline and asserts a substring, which is what you need here, since planning *panics* rather than returning `Err` (so the usual `compile_program(…).is_err()` pattern doesn't apply). Already used at `comprehensions.rs:225`, `scalars_collections.rs:239`, `joins_aggregates_groupby.rs:305`.

```rust
#[test]
fn correlated_filter_beside_correlated_body_reaches_the_pair_wall() {
    check_compile_error(
        "sum([sum([v * r for v in [1, 2, 3] if v > r]) for r in [1, 2]])",
        "a `__pair` binder survived into a compiled predicate term",
    );
}
```

I ran it both ways:

- **post-fix**: passes.
- **pre-fix**: fails with `expected panic to contain "a __pair binder survived…"; got: an unnamed function's codomain references it: the index has no binder to open at`.

That pins both halves of the commit message's story — the `subst` panic is gone, and the wall reached is the stated one — at the CHL level, which the unit test cannot do. I checked whether any nearby program becomes a *positive* test instead (ten variants: correlated body without filter, correlated filter without body, triple nesting, `and`-ed predicate); every one that reaches the bug hits the same `__pair` wall, and the rest never reach it. So this is the strongest CHL test available today.

It pins the next wall's message, which changes when correlated comprehensions land — but that's the point: at that moment this test converts into the positive one. Keep the unit test too; it isolates the rule.

src/ccl/lambda_elim.rs
```diff
@@ -1892,6 +1912,37 @@ mod tests {
...
+            ),
+            FunKind::Compute,
+        );
+        let f = var("f").with_ty(Type::fun(int.clone(), int.clone()));
+        let g = var("g").with_ty(dependent);
+        let paired = zip_pair_ty(&f, &g, &FunKind::Compute);
+        assert!(
+            matches!(&paired, Type::Fun { name: Some(n), .. } if n == &binder),
+            "the pair binds what its second operand names, got {paired}",
+        );
```
groundlar, line 1945: **The test asserts the name, not the property the name is for.**

The failure mode is `open_codomain` having no binder to open at, and the index depth is the other half of that. A change that set the name right but broke the depth — a `Tuple` that *did* count as a codomain crossing, say — would still pass this assertion. Two more lines pin the actual downstream property:

```rust
let Type::Fun { codomain, .. } = &paired else { panic!("a pair is a function") };
let opened = crate::ccl::subst::open_codomain(&paired, codomain);
assert!(
    !crate::ccl::subst::references_enclosing_function(&opened),
    "the pair's binder opens its second operand's reference, got {opened}",
);
```

src/ccl/simplify.rs
```diff
@@ -906,11 +906,12 @@ fn try_exponential_beta(expr: &mut Expr) -> bool {
...
-            // makes `id`'s kind matter: `zip_pair_ty` reads the *first* operand's
-            // kind, so a bare `Compute` here would propagate out through the zip
-            // to every consumer of the rewritten chain — `id` is the unit of
-            // composition and has no kind of its own to lend.
+            // `mint_kind` (see `try_pairwise_in_compose`). `id` is the unit of
+            // composition and has no kind of its own to lend, and a bare `Compute`
+            // here reaches the tuple type the zip built-in is applied at. The kind
+            // the pair carries is declared rather than read off an operand:
+            // `zip_pair_ty` takes it as an argument, and reads only the domain off
+            // the first.
```
groundlar: **This sentence is inaccurate as of this commit.** `zip_pair_ty` now reads the domain off the first **and the binder off either**.

```
// `zip_pair_ty` takes it as an argument, and reads the domain off the
// first, the binder off either.
```

src/ccl/simplify.rs
```diff
@@ -906,11 +906,12 @@ fn try_exponential_beta(expr: &mut Expr) -> bool {
...
             let g_dom = g.ty.domain();
```
groundlar, line 875: **The same shape, unfixed, two sites down — and this PR edits the comment directly above one of them.**

Both mint `name: None` around a codomain read raw off an operand:

- **`simplify.rs:932`**, just below: `zip_pair(id_node, g, mint_kind).with_ty(zip_ty)`. The `.with_ty(zip_ty)` **overrides** the type `zip_pair_ty` just computed, so this PR's fix does not reach this site at all. `zip_ty` is `mint(a, Tuple([a, c]))` with `name: None` and `c = g.ty.codomain()` taken verbatim — a dependent `g` produces exactly the shape this PR is fixing.
- **`simplify.rs:1167`** (`arm_ty` in the zip-distribution rewrite): same for both legs, then zips two unnamed functions.

I instrumented both sites with an assertion that no operand arrives with `name: Some(_)` and ran the whole suite: **neither fires**. So this is latent, not a live bug, and I would not widen this PR to fix it.

But it is the third instance of "drop the binder, keep the codomain", and CLAUDE.md is explicit that a recurring shape gets flagged when noticed rather than left for the fourth duplication. A sentence in the PR description, or a `TODO` carrying the reasoning, is enough. (Threading the binder into `zip_pair_ty` — see my comment on its doc — would force both sites to state what they mean.)

groundlar (approved):
LGTM after comments

src/ccl/lambda_elim.rs
dpmills: Keeping the read, and the doc now carries the derivation.

The kind and the binder are not the same question. `elim_lambda_impl` mints a Pi exactly when `codomain_depends_on(param, &body_ty)` (`src/ccl/lambda_elim.rs:721`), so the pair has to name a binder exactly when its own codomain references one — and the pair's codomain is the tuple of its operands'. That dependence is a fact only the operands carry. A caller holds the *name*; it does not hold "is this pair dependent?". The kind is the mirror image: no operand carries it, because `id` has none to lend.

So declaring the binder means naming every pair an elimination builds. The third caller shows it: `lambda_elim.rs:1421`, the outer-binder `Case` arm, zips `id` (typed by `Type::fun`, unnamed by construction) with the payload morphism. I instrumented it with an assertion that neither operand arrives named and ran the full suite — it never fires. Under `binder: Option<&Name>` that pair becomes `(param: A) ⇒ (A, Pᵢ)` with nothing in the codomain referencing `param`, which is the rule at line 721 run backwards.

The other reading — thread `param` but still consult the operands for dependence — adds an argument carrying no information the operands lack, and the assertion is still what checks it.

The doc paragraph is rewritten to say this: the rule, where it comes from, and why a caller cannot supply it. Your other four suggestions are all in: the message rewrite, the postcondition `debug_assert!`, the CHL test, and the opened-codomain assertion.

src/ccl/lambda_elim.rs
dpmills: Confirmed and recorded rather than fixed — the doc's `Hole` paragraph now names it:

> A **refined** function operand takes that path as well, the match being on the bare [`Type::Fun`]: [`Type::fun_kind`] peels refinements because a refined function still carries a kind, and a binder rides one the same way.

It also connects to your `simplify.rs:915` note. The `.with_ty(zip_ty)` override there computes its type from `Type::domain` / `Type::codomain`, which *do* peel, so it produces a concrete type exactly where `zip_pair_ty` returns `Hole`. Removing the override needs this match fixed first.

src/ccl/lambda_elim.rs
dpmills: Message taken verbatim.

Keeping `assert!`. The house rule is `assert!` for an O(1) check and `debug_assert!` only when the check itself costs something; this one is a `Name` comparison, and `assert_ne!` at line 731 is the neighbour that matches. The postcondition check you suggest below goes the other way for the same reason — `references_enclosing_function` is a full recursive scan over the type, so it is a `debug_assert!`.

src/ccl/lambda_elim.rs
dpmills: Added as written. Full gate green.

src/ccl/lambda_elim.rs
dpmills: Added, as `a_correlated_filter_beside_a_correlated_body_is_refused_for_its_pair_binder` in `tests/compilation_pipeline/comprehensions.rs`. Reproduced both directions here: pre-fix it fails on `an unnamed function's codomain references it`, post-fix it passes on the `__pair` refusal. The unit test stays.

Two corrections to the citations. `check_compile_error` is used in `type_aliases.rs`; the three sites you name (`comprehensions.rs:225`, `scalars_collections.rs:239`, `joins_aggregates_groupby.rs:305`) are `#[should_panic(expected = …)]` over `check_scalar` / `run_pipeline`. Either works under `catch_unwind` semantics; I took `check_compile_error` for the failure message.

The name drops `wall`: the skill calls it out as a metaphor that already names three different mechanisms in this repo, and a test function name is where that spreads furthest.

src/ccl/lambda_elim.rs
dpmills: Added, plus one line in front of it:

```rust
assert!(
    crate::ccl::subst::references_enclosing_function(codomain),
    "the fixture's second operand is the dependent one, got {codomain}",
);
```

`pi_kinded` names unconditionally, so without this the opened assertion holds vacuously for a fixture that turned out not to be dependent — the same gap you are pointing at, one level down.

src/ccl/simplify.rs
dpmills: Corrected to your wording: `zip_pair_ty` takes the kind as an argument, and reads the domain off the first, the binder off either.

src/ccl/simplify.rs
dpmills: Not widening the PR, and I reproduced your measurement independently — an assertion at both sites that no operand arrives with `name: Some(_)`, full suite, neither fires.

Flagged at the exponential-beta site rather than in the PR body alone, because the shape is visible there:

```rust
// `mint` drops the binder: a dependent `g` carries its binder reference
// in `g_cod`, and `name: None` leaves nothing to open it at — the shape
// `subst::open_codomain` rejects. It reaches the zip below because
// `.with_ty` overrides the type `zip_pair_ty` computed, which does carry
// the binder. `arm_ty` in the zip-distribution rewrite has the same
// shape.
```

I left the "neither is reached by a named operand today" half out of the comment: it is a coverage fact, and a comment is the wrong place for one to rot. It is in the PR description instead.

Worth noting for whoever closes it: the override is not purely redundant. `zip_ty` is built from `Type::domain` / `Type::codomain`, which peel refinements, so it produces a concrete type exactly where `zip_pair_ty`'s bare `Type::Fun` match returns `Hole` — the nit you raise on `lambda_elim.rs:252`. Deleting the override needs that match fixed first, which is why the two land together or not at all.

src/ccl/lambda_elim.rs
dpmills: Replaced the doc paragraph with a test — a comment recording a gap reports the same thing whether the gap is open, closed, or gone, and this one fails silently besides (`Hole` is also what an operand of no known shape yields).

`a_zip_reads_through_a_refined_operand` in `src/ccl/lambda_elim.rs` feeds `{(k: Int) ⇒ {Int | k} | p}` as the second operand and asserts the pair binds `k`, under `#[should_panic]`. Today it panics on `got _` — the pair is `Hole`, so the domain and codomain go with the binder. Closing the gap turns it red.

The doc's `Hole` sentence now states the match rather than its consequences: "Returns `Type::Hole` unless both arguments match a bare `Type::Fun`".

PR #188, opened 2026-09-14
Defer the unreadable-input correspondence to the first path that reads it

dpmills:
`unreadable_inputs` diffed two planned trees up front, and reads the result only for a variable this version adds to a loop that already carries one — which most reloads have none of. Building it on first use takes a reload of `v1_single_line.cambra` from 2.67s to 1.83s in release, the step itself from 855ms to 11ms.


dpmills:

- #188 ◀

groundlar (approved):

PR #189, opened 2026-09-14
Advance producer state once per delivery round instead of once per pull

dpmills:
`TileProducer::get` is a read, but the stores, `Aggregate` and `FanOut` advanced their state on whichever pull reached them first, so a tile depended on how often it had been asked for and by which fan-out branch — an answer nothing downstream can cache. This introduces the **delivery round**, the span between two `Scheduler::check_for_notifications` calls, and moves every self-driven advance onto that boundary: a producer steps on the round's first pull and serves the same tile for the rest of it. [`design-operators.md`, "The read contract"](src/interpreter/design-operators.md#the-read-contract) specifies the result.

Read `src/interpreter/scheduler.rs` first: `open_round` / `current_round` over a thread-local counter — a round is a span with no `get` in flight on the thread, so it belongs to the thread rather than to a scheduler, which is what `check_shared_state.py` now excepts. `check_for_notifications` opens the round and takes the deferred queue before polling sources.

`RoundCache` (`src/interpreter/tile_operators/mod.rs`) freezes a round's answer for `CommitProducer`, `InductionStoreProducer`, `AggregateProducer` and the non-cyclic `FanOut`; `TileProducer::get` asserts the contract in debug builds.

`fanout.rs` carries the rest: a cyclic fan-out is now an explicit delay element, serving the recurrence branch the previous round's tile and readers the round's own (`serving` / `pending`, the latter exposed as `latest_tile`, was `cached_tile`), and asking the scheduler for the next round when a lap is outstanding.

Tests drive the alternation: a `Driven` harness in `commit_operator.rs`, a `pull` helper in `tests/compilation_pipeline/helpers.rs`.


dpmills:

- #189 ◀

dpmills:
Superseded by #201. The round counter was a global stand-in for a per-producer dirty bit that did not exist — `Consumer::notify` carries no payload and every `subscribe` forwarded the downstream consumer past its own operator, so nothing in the graph could know whether its input had anything new.

Measured on the asset-cart demo (release, 48 whole-program tests): ~25-31s with no mechanism, ~3-5s with this PR's round cache, ~4s with #201's notification gate and no round cache at all. The gate recovers the same win by a local signal, with no thread-local and no global clock.

Also measured while investigating: the `RoundCache` on `CommitProducer` and `InductionStoreProducer` protected a case that never arises — neither is ever repeat-pulled, even with the fan-out's cache disabled — and across the suite no producer ever answered a same-round repeat differently without a release in between (~60,000 repeats, 0 violations). The idempotence assert here also keys on the producer's own `obsolete_guard` rather than a release to its input, so it reports false positives when a downstream release travels through.

PR #190, opened 2026-09-14
Accept an `if` guard in a `for` header and fold it into the body

dpmills:
A `for` header takes an `if`, as a comprehension does: `for x in xs if p:` runs the body only where `p` holds. The guard is a fact about the body rather than about the iteration, so the loop sweeps the same extent either way; that is what separates it from a comprehension's `if`. The parser keeps the guard in its own `Stmt::For` field, and lowering folds it back into the body it guards through `guarded_body` in `src/ccl/lower/loops.rs`, called from each of the three `for` lowering sites. Everything that holds of a written `if` then holds of a header guard, including the refusal of a `with begin():` transaction under one. Specified in `docs/chl-spec.md`, "4.6 `for` — iteration".


dpmills:

- #190 ◀

dpmills:
Closing: the header guard's only caller is one line of `v1.cambra`, which does not compile. On `main` the change is sugar — `guarded_body` wraps the body in a `ChlStmt::If` and no pass downstream distinguishes the two spellings — against a `guard` field on `Stmt::For` and a spec paragraph warning that a header `if` means something different from a comprehension's. Re-extractable from the demo branch if `v1.cambra` compiles and still wants the header spelling.

PR #191, opened 2026-09-14
Build every `Type::History` through a constructor and give it a position binder

dpmills:
`Type::History` spells the arrow `domain ⤇ value` across two independent fields, so nothing owns its binding discipline the way `Type::pi_kinded` owns a `Type::Fun`'s codomain, and a feed's value cannot depend on its own position. This routes every construction through `Type::feed`, `Type::mutable`, and `Type::history`, giving that discipline one owner to move to, then adds an `Option<Name>` position binder to `Type::History`, bound in `value` and set only by the new `Type::history_pi`. The binder is for a feed whose contribution varies by position, and is inert today: nothing calls `history_pi`, every existing construction takes a non-binding constructor, so the field is `None` everywhere.

The constructors take `(domain, value)`, which is `as_feed`'s return order rather than the `Mut(value, domain)` rendering, so the argument order at each converted site is what to check; `Type::feed`'s doc records the choice. `Type::history_pi` closes `value` with `close_pi_binder` exactly as `pi_kinded` closes a codomain. Every `Type::History` pattern gains a `..`; hashing, spec keys, and substitution are where a set name becomes live work.


src/ccl/ty.rs
```diff
@@ -2246,6 +2261,64 @@ impl Type {
...
+            history_kind,
+        }
+    }
+
+    /// A history whose **value depends on its position** — `feed((name: domain) ⤇ value)`.
+    ///
+    /// Closes `value` over `name` exactly as [`Type::pi_kinded`] closes a codomain, so a
+    /// history built here cannot carry a free name for the binder it provides. This is the
+    /// only constructor that sets [`Type::History`]'s `name`.
+    pub fn history_pi(
```
sortalongo: `close_pi_binder` rewrites free references to `name` into a `PiBound` index, but `PiWalk` descends a history's `value` at the same depth (`src/ccl/subst.rs:1933`), and its own doc at `src/ccl/subst.rs:1872` says only a `Fun` codomain crossing deepens. The index this mints therefore resolves to the nearest enclosing `Fun` binder, and `open_codomain` would open it as that function's. Nothing calls `history_pi`, so no current program is affected.

Drop it from this PR and add it in the change that teaches `PiWalk`, `Display` and the opening sites to cross a history's binder.

src/ccl/ty.rs
```diff
@@ -2246,6 +2261,64 @@ impl Type {
...
+    pub fn history(domain: Self, value: Self, history_kind: HistoryKind) -> Self {
+        Type::History {
+            name: None,
+            value: Box::new(value),
+            domain: Box::new(domain),
+            history_kind,
+        }
+    }
+
+    /// A history whose **value depends on its position** — `feed((name: domain) ⤇ value)`.
```
sortalongo: The rendering in this doc is not the one the code produces. `Display`'s `History` arm ignores `name` and prints `feed(domain ⤇ value)`. Drop the rendering from the doc.

src/ccl/ty.rs
```diff
@@ -2246,6 +2261,64 @@ impl Type {
...
+    /// Argument order matches [`Type::feed`] and [`Type::as_feed`] rather than the
+    /// rendering, so the two constructors do not disagree about which slot comes first.
+    pub fn mutable(domain: Self, value: Self) -> Self {
+        Self::history(domain, value, HistoryKind::Overwrite)
+    }
+
+    /// A history at an explicit [`HistoryKind`], for a rebuild carrying the kind it
+    /// replaces.
+    ///
+    /// **The one place a history is built.** A history denotes the arrow `domain ⤇ value`,
```
sortalongo: `history_pi` writes its own `Type::History` literal twenty lines below, so there are two build sites rather than one. Have `history_pi` delegate to `history` and set the name.

src/ccl/subst.rs
```diff
@@ -1452,11 +1452,12 @@ impl Subst {
                 value,
                 domain,
                 history_kind,
-            } => Type::History {
-                value: Box::new(self.apply_type(value)),
-                domain: Box::new(self.apply_type(domain)),
-                history_kind: *history_kind,
-            },
+                ..
+            } => Type::history(
```
sortalongo: This destructures with `..` and rebuilds through `Type::history`, which hardcodes `name: None`, so the rebuild erases a set binder. The `Fun` arm above preserves its binder and descends through `under_binder_ty`. Ten call sites across nine files route through `Type::history`, so every rebuild among them erases it the same way.

Give the rebuild path a constructor that carries `name` through, so erasing it is not the default.

src/ccl/infer/solver/constrain.rs
```diff
@@ -2885,14 +2886,10 @@ mod tests {
         );
     }
 
-    // --- Mutable handles (`Type::History { kind: Overwrite }`) ---
+    // --- Mutable handles (`Type::History { kind: Overwrite, .. }`) ---
 
     fn mut_ty(value: Type, domain: Type) -> Type {
```
sortalongo: `mut_ty(value, domain)` forwards to `Type::mutable(domain, value)`, so two argument orders for one shape sit on adjacent lines. Reorder `mut_ty`'s parameters to match the constructor.

sortalongo (approved):
Approving. The constructor funnel is the right change and it is behavior-preserving: `Display` is unchanged, `Type`'s serde goes through `Display`, and no node ids move, so leaving the golden fixtures untouched is correct.

One thing to settle before this merges, inline on `history_pi`. It closes a binder that no traversal crosses, so the index it mints resolves against the nearest enclosing `Fun`. Nothing calls it today, which is why this is an approval rather than a request for changes, but it should not land as it stands.

Two findings with no line in the diff:

- `Display for Type`'s `History` arm at `src/ccl/ty.rs:2076` ignores `name` and prints `feed(domain ⤇ value)`. Every pattern in the tree now carries `..`, so a set binder is dropped silently rather than breaking the build. A `debug_assert!` naming the invariant in the readers that ignore the field would keep that loud.
- `src/ccl/design/ir.md:215` still spells the variant with three fields, and `mutability.md:300` repeats that spelling. The new field exists only in rustdoc.

On the shape of the change: the funnel collapses about twenty copies of one struct literal and clearly earns its place. The `name` field and `history_pi` are a different case, since nothing in the tree is working around their absence today. Landing the funnel now and adding the binder in the change that teaches the traversals to cross it would keep each half reviewable.

src/ccl/infer/solver/constrain.rs
dpmills: Reordered to `(domain, value)` and flipped the three call sites, matching `feed_ty` on the same page. The helper gains a doc line naming why the constructors' order differs from the `Mut(value, domain)` rendering.

src/ccl/subst.rs
dpmills: Void for this PR: the position binder is dropped, so `Type::history` has no `name` to erase.

Kept for the change that adds it back, with one distinction the caller survey turned up. `Type::without_pi_names`'s `History` arm also rebuilds through `Type::history`, and dropping the binder there is what that function does — its `Fun` arm sets `name: None` deliberately. So the carrying constructor is for the substitution and freshening paths, not for every caller of `Type::history`.

src/ccl/ty.rs
dpmills: Void: `history_pi` is dropped. `Type::history` is the only code that writes the literal, and `Type::feed` and `Type::mutable` delegate to it.

src/ccl/ty.rs
dpmills: Void: the doc went with the constructor. `Display`'s `History` arm is untouched by this PR.

src/ccl/ty.rs
dpmills: Dropped, as asked.

The index is as you describe: `close_pi_binder`'s doc states that an index counts the `Fun` codomains crossed, and `PiWalk`'s `History` arm descends `value` at the same depth.

For whoever lands it, the only consumer anywhere in the repo is `510ba09ac` on the `demo-map-cart` line, which abstracts a feed contribution over the channel's position. It calls `history_pi` once in `emit_let` and introduces `Type::history_stream`, reassembling the arrow as `pi_kinded(p, domain, value, FunKind::Data(None))` — which is what makes the field mean anything. That branch does not deepen at a `History` either, and its `Subst::apply_type` still rebuilds through `Type::history`, so both gaps are unbuilt there too and land with it.

dpmills:
The position binder is dropped and the five inline threads are answered. That voids both findings with no line in the diff: `Display`'s `History` arm has no `name` to ignore, and `ir.md`'s three-field spelling and `mutability.md`'s are correct again.

`ir.md` gains one sentence naming the three constructors, since the funnel invariant otherwise lived only in rustdoc.

The binder is used — by `510ba09ac` on the `demo-map-cart` line, and by nothing else in the repo — so it lands with that change, which is also where `PiWalk` and `Subst::apply_type` have to learn to cross a history. Detail is on the `history_pi` thread.

Rebased onto `ac332ec77` (#218) and force-pushed; the title and body now describe the funnel alone.

PR #193, opened 2026-09-14
Give `IterateExtent` an input edge per source it iterates, and attribute source nodes only to reads that denote them

groundlar:
An `IterateExtent` over a data-source domain is the operator the scheduler wakes when that source produces, yet it stated no inputs, so the inspector's graph drew it unconnected to the source it waits on. It now states one `share` edge per distinct source its extent reaches. Stating that edge needs the source node minted, and minting previously made that expression a provenance parent, which for an iteration is planning's marker or a store's recurrence read — nowhere the program writes a source. So a read that mints is now separated from a read that attributes, and the graph gains the edge without widening where the program names `stdin()`.

## Reads that mint, denotations that attribute

`Boundaries::source_reads` carries `SourceRead { expr, denotes }`: `record_source_read` records a denoting read (the `Source(name)` arm), the new `record_source_iteration` one that only mints. `materialize_sources` takes parents from the denoting reads, falling back to the first read so a source nothing denotes still gets a row. `IterateExtent` holds no `NodeId`, so it reads its expression from the ambient recording (`provenance::currently_named`). The doc comments on `SourceRead` and `materialize_sources` carry the reasoning.

## One walk over an extent's sources

`Extent::for_each_source` replaces three separate recursions: the subscribe handles, `subscribe_to_iteration_action`, and the retire walk — which missed `Union` arms, leaving a dead producer's release record pinning the source's agreement. `visit_inputs` drives the recording too, so the reads recorded and the edges stated are one list.

## What the tests pin

`source_accumulator` (stdin into a `:=` accumulator) is the gallery's one program where a store's trigger iterates a source: two `IterateExtent`s, one source node, one span. `source_shared`'s golden compares spans by site rather than by count now, and its web fixture carries the new edges. The `IterateExtent` row in [`design-operators.md`](src/interpreter/design-operators.md#operator-identity-and-the-graph-the-inspector-reads) records them.


dpmills:
Why do we need the source to be a node in the graph at all? Can we just treat IterateExtent as the root?

groundlar:
> Why do we need the source to be a node in the graph at all? Can we just treat IterateExtent as the root?

Done!

web/src/wireValidate.ts
```diff
@@ -105,7 +105,7 @@ const ALLOWED_NATURE = ["expansion", "machinery"];
 
 // An operator node's `role`, and an operator input edge's `kind` — the two
 // closed vocabularies of the operator pane.
-const ALLOWED_ROLE = ["operator", "source", "sink"];
+const ALLOWED_ROLE = ["operator", "sink"];
```
dpmills: Do we still need "sink"? Isn't everything just operators now?

dpmills (approved):

web/src/wireValidate.ts
groundlar: Today sink nodes aren't built when the graph construction runs. Prototyping changing this as a separate PR and will update here or followup PR.

PR #195, opened 2026-09-14
Give an application one read view of its function in the post-inference check

dpmills:
The post-inference check looked through a feed handle in `CheckCtx::as_function` alone, so the argument edge and the binder discharge of an application still received the raw `Feed`. A comprehension over a channel applies it, so `sum([x for x in out])` reached `constrain_argument` holding a handle rather than a function and tripped its assertion. `Type::read_view` names what a handle denotes when read — a channel as the data function `domain ⤇ value`, a mutable variable as its value — and `CheckCtx::apply` takes it once, then hands that one function to all three consumers.

Tests cover the bare and the mapped read, with the contribution fed inside and outside a transaction. A filtered comprehension is pinned to its current `channelize` failure instead; the test's comment locates it.


src/ccl/infer/check.rs
```diff
@@ -421,6 +421,14 @@ impl Typing for CheckCtx {
         function: &Type,
         at: &dyn Fn() -> String,
     ) -> Result<(), LocatedInferError> {
+        // **A feed handle's read view is the function being applied**
+        // ([`Type::feed_read_view`]). [`Typing::as_function`] takes that view one line
+        // above this in [`Typing::apply`], so a handle arriving here unpeeled leaves the
+        // two disagreeing about whether a channel is a function: the domain lookup below
+        // finds none, and the assertion reports a caller that passed a domain. A
+        // comprehension over a channel is the program that reaches it.
+        let view = function.feed_read_view();
```
sortalongo: This peels `Feed` only. `as_function` peels refinements and loops `mut_value_type()`, so a `Mut`-typed collection in function position destructures there and not here.

Suggest peeling once in `Typing::apply` and passing the same view to `as_function`, `constrain_argument`, and the discharge match below, instead of peeling per consumer.

Does #197 add another peel in function position? If so, the shared peel belongs in this PR first.

src/ccl/ty.rs
```diff
@@ -2570,6 +2570,18 @@ impl Type {
         }
     }
 
+    /// The **read view** of the feed channel this denotes — the data function
+    /// `domain ⤇ value` — or `None` if it is not one.
+    ///
+    /// A read view is kind-specific: a feed channel reads as the whole collection it has
+    /// accumulated, where [`Type::mut_value_type`] derefs a mutable variable to one value.
+    /// So this is the function a read of a channel applies. [`Type::as_feed`] answers the
+    /// same question as the two halves, for a caller that needs them apart.
```
sortalongo: This paragraph restates the docs on `as_feed` and `mut_value_type` a few lines up. Keep the one-line summary and delete it.

src/ccl/infer/check.rs
```diff
@@ -421,6 +421,14 @@ impl Typing for CheckCtx {
         function: &Type,
         at: &dyn Fn() -> String,
     ) -> Result<(), LocatedInferError> {
+        // **A feed handle's read view is the function being applied**
+        // ([`Type::feed_read_view`]). [`Typing::as_function`] takes that view one line
+        // above this in [`Typing::apply`], so a handle arriving here unpeeled leaves the
+        // two disagreeing about whether a channel is a function: the domain lookup below
+        // finds none, and the assertion reports a caller that passed a domain. A
+        // comprehension over a channel is the program that reaches it.
```
sortalongo: This argues for the change and describes the counterfactual; the assert below already states the invariant. "one line above this" rots on the next insertion, and intra-doc links do not render in a `//` comment.

One line is enough: `// A feed handle applies as its read view, as as_function destructures it.`

src/ccl/design/type-inference.md
```diff
@@ -1125,7 +1125,7 @@ The typing rules (`infer_simple_sub::emit_defer` / `emit_feed` / `emit_define`):
...
-2. **`Feed(a) <: 𝑇`** for non-feed `𝑇` ⇒ `a <: 𝑇` — transparent read (`sum(d)`, `d + 1`, `x <<= y` chains discharge through the handle). The post-inference structural check mirrors this: `CheckCtx::as_function` peels `Feed` like outer refinement tags.
+2. **`Feed(a) <: 𝑇`** for non-feed `𝑇` ⇒ `a <: 𝑇` — transparent read (`sum(d)`, `d + 1`, `x <<= y` chains discharge through the handle). The post-inference structural check mirrors this, at both of the sites an application reaches: `CheckCtx::as_function` peels `Feed` like outer refinement tags, and `constrain_argument` takes the same read view (`Type::feed_read_view`) before looking for a domain. A comprehension over a channel applies it, so a handle unpeeled at either site leaves them disagreeing about whether a channel is a function.
```
sortalongo: The last sentence repeats the argument from the `constrain_argument` comment. The preceding clause already names both sites. Delete it.

tests/compilation_pipeline/transactions.rs
```diff
@@ -561,6 +561,78 @@ fn progress_feed_grant_deny() {
...
+///
+/// `Typing::apply` peels the handle for `Typing::as_function` and then passes
+/// `constrain_argument` the function it was given, so before `Type::feed_read_view` was
+/// taken there too the two disagreed about whether a channel is a function and the
+/// argument check asserted on a caller that had passed one.
+///
+/// An aggregate applied to the channel directly (`sum(out)`, as
+/// [`await_final_and_as_of_read_the_same_mut_var`] reads it) consumes the collection
+/// without applying it, which is why the suite held no program that reached this. The
+/// transaction is beside the point — a bare feeding loop reaches it too.
```
sortalongo: History narration: "before `Type::feed_read_view` was taken there too" and "which is why the suite held no program that reached this". Keep the first paragraph only.

tests/compilation_pipeline/transactions.rs
```diff
@@ -561,6 +561,78 @@ fn progress_feed_grant_deny() {
...
+#[rstest]
```
sortalongo: The doc says the transaction is beside the point, and one of the three tests has no `with begin():`. Move them to `feeds_cases.rs`.

tests/compilation_pipeline/transactions.rs
```diff
@@ -561,6 +561,78 @@ fn progress_feed_grant_deny() {
...
+/// domain is the channel's, so a `ChanDom` sits inside the refinement's predicate —
+/// where channelize's erasure never reaches, because `Type::walk_children_mut` visits a
+/// `Type::Refinement`'s base and not its predicate. The post-channelize check then reports
+/// the residue at `__elem`.
+///
+/// The unfiltered reads beside it place it: what the filter meets is the channel domain
+/// surviving into a predicate, not anything about applying a channel.
+#[test]
+#[should_panic(expected = "feed handle type survived channelize")]
+fn a_filtered_comprehension_over_a_feed_channel_is_not_reachable() {
```
sortalongo: The name says the program is not reachable, but it is reached and fails in `channelize`. Rename to `..._fails_at_channelize`. `Value::Int(2)` below is dead under `should_panic`.

src/ccl/ty.rs
```diff
@@ -2570,6 +2570,18 @@ impl Type {
...
+    pub fn feed_read_view(&self) -> Option<Type> {
```
sortalongo: `data_fun` gives this a `Fun` with no Σ, so `function.sum()` in `constrain_argument` is `None` by construction and a peeled channel always gets an empty binder list. That is correct only if a channel domain carries no Σ. State that invariant here, or read the domain from `as_feed()` directly.

sortalongo (approved):
Approving. The first two inline comments (the peel in `constrain_argument`, the `feed_read_view` doc) are worth taking before merge; the rest are nits.

Three sites now destructure a handle to the function it is, each with its own policy: `as_function`, `constrain_argument`, and `mut_elim::fun_parts`. None share a helper. The fix suggested on `constrain_argument` removes one.

PR description: the last sentence reports a doc edit the diff already shows. Drop it.

src/ccl/infer/check.rs
dpmills: Taken, as suggested: `CheckCtx::apply` takes the view once and hands the same function to `as_function`, `constrain_argument`, and the discharge match. The policy itself is now named — `Type::read_view` answers for every handle (a mutable variable as its value, a channel as the collection it has accumulated, anything else as itself), so `as_function`'s `Fun` and `Feed` arms collapse into one and the discharge reaches a named Pi behind a handle, which peeling refinements alone did not.

#197 does not add a peel here: it adds `read_through` at Emit's function position, which is the solver-side deref of a value operand, not a structural destructuring.

`mut_elim::fun_parts` stays on its own peel. It panics on a feed channel deliberately (`"letrec phase: loop source is not a function"`), and the shared view would admit one — so converting it would delete an assertion rather than share a policy. Two of the three sites share now.

src/ccl/infer/check.rs
dpmills: Gone. `constrain_argument` is back to its pre-PR body; the one comment that remains is on `apply`, where the view is taken.

src/ccl/ty.rs
dpmills: Rewritten for `Type::read_view`. The kind-specific sentence is what the generalization *is* now rather than a restatement, and it is one sentence.

src/ccl/ty.rs
dpmills: Stated, on `read_view`: `Type::History` has no binder slot, so a handle cannot carry a Σ, and a consumer reading binders off a channel's view finds none because there are none — not because the view dropped them.

src/ccl/design/type-inference.md
dpmills: Deleted. The clause now names both sites through `CheckCtx::apply`, which is where the view is taken.

tests/compilation_pipeline/transactions.rs
dpmills: Both sentences gone; the first paragraph is the whole doc.

tests/compilation_pipeline/transactions.rs
dpmills: Moved to `tests/compilation_pipeline/feeds_cases.rs`, all three.

tests/compilation_pipeline/transactions.rs
dpmills: Renamed to `a_filtered_comprehension_over_a_feed_channel_fails_at_channelize`, and it now uses `check_compile_error`, which takes the expected message instead of a value — so nothing is dead and the residue it stops on is pinned by name (``channel domain chan(out) survived channelize at `__elem` ``).

PR #196, opened 2026-09-14
Erase channel types from refinement predicates in `channelize`

dpmills:
`channelize` erased channel types from type structure alone, so a `ChanDom` sitting inside a refinement predicate survived the pass and the strict wall rejected the program — the shape a filtered comprehension over a channel always builds, since the filter refines the source domain and that domain is the channel's. `erase_chan_domains_in_predicates` erases there, and `erase_chan_domains_in_slot` pairs the two halves so every slot the pass reaches (a node's type, a user annotation, a binder's declared type) gets both; the debug residue check grows the same reach. Filtered reads of a channel now compile and evaluate, including a `let`-bound one and a filter naming an outer `let`, so the case that pinned the channelize error pins a value instead.

## Gating the eraser

The rebuilding walk runs only when the read-only `predicates_hold_type` finds a channel type. `PredMemo::rebuild` clones a predicate before its callback can report whether anything changed, and the clone draws fresh node ids even when discarded, which would move the golden inspector fixtures for every program with nothing to erase. The general fix is a probe on the shared term inside `walk_refined_predicates_mut` / `PredMemo::rebuild`, before the clone; deferred here, and until it lands every pass with a conditional predicate rewrite re-derives this pre-scan locally.

## Follow-ups

- `mut_elim::erase_mut_in_type` and its post-condition `contains_mut_type` (`src/ccl/mut_elim.rs`) have the same predicate blind spot for `Overwrite` histories — the twin of what this fixes for `Append`. Filed.
- `c = [x for x in out]`, unfiltered and `let`-bound over a channel, still panics with ``type residue survived channelize on let binding `__mono` : (chan(out) ⤇ Int)``. That residue is in type structure rather than a predicate; filed separately.


sortalongo (changes requested):
The diagnosis is right, the layer is right, and the two new test cases pass and fail without the fix. The erasure stops one slot short: binder-declared types get no predicate erasure, and a let-bound filtered comprehension still panics at the post-channelize wall. The per-call `PredMemo::default()` breaks the one-memo-per-pass contract. The pre-scan is a channelize-local workaround for a limitation of the shared rebuild helper, and the pass's own post-condition still skips predicates, so the eraser and its checker disagree.

Line numbers are at the PR head.

### Findings

**Critical — `src/ccl/channelize.rs:854-859`.** The `walk_binders_mut` closure erases `b.ty` and `b.user_annotation` with `erase_chan_domains_in_type` only. `c = [x for x in out if x > 1]` then `sum(c)` panics at head with ``channel domain chan(out) survived channelize at `__elem` ``. Fix: call `erase_chan_domains_in_predicates` on both binder slots and add the let-bound case to the rstest.

**Critical — `src/ccl/channelize.rs:812`.** `PredMemo::default()` is built once per type slot per node. `src/ccl/ccl_utils.rs:1808-1814` names this as the bug the type exists to prevent: a predicate `Rc` shared by two residue-bearing slots is rebuilt into two distinct terms. Fix: build one `PredMemo<()>` in `run` at `src/ccl/channelize.rs:618` and thread it through `erase_chan_domains` beside `map` and `kinds`.

**Important — `src/ccl/channelize.rs:766-798` (the pre-scan).** `predicates_hold_channel_types` re-walks every predicate to decide whether to enter `walk_refined_predicates_mut`. The reason is real: `PredMemo::rebuild` clones the predicate before the callback runs, and `TypedExpr::clone` draws a fresh id through `provenance::copy_id` (`src/ccl/provenance.rs:2026-2033`) even when the copy is discarded on a `false` report. That is a limitation of the shared helper, and every pass with a conditional predicate rewrite will re-derive this pre-scan. Fix: move the gate into `walk_refined_predicates_mut` / `PredMemo::rebuild` as a probe on the shared term that runs before the clone, and pass channelize's `in_expr` as that probe. If deferred, name the recurring shape in the PR body.

**Important — `src/ccl/channelize.rs:823`.** The rebuild callback returns `true` unconditionally, so once inside the walk every visited predicate is reallocated even when the erasure found nothing in it. Fix: give `erase_chan_domains_in_type` a `bool` return, thread it out of `go`, and return it, the shape `src/ccl/simplify.rs:171-180` uses.

**Important — `src/ccl/channelize.rs:124-129` and `918-924`.** `has_type_residue`'s doc claims predicates carry no residue and are not inspected, and `assert_no_type_residue` says walking them would need cycle guards. This PR ships that guarded walk. Fix: have `assert_no_type_residue` call `predicates_hold_channel_types` and correct both comments.

**Important — `src/ccl/mut_elim.rs:213-225`.** `erase_mut_in_type` and its post-condition `contains_mut_type` (`194-205`) share the predicate blind spot for `Overwrite` histories. Fix: note it in the PR body and file a follow-up.

**Nit — `tests/compilation_pipeline/transactions.rs`, case `names_an_outer_let`.** The case string carries an embedded `\n` (`"n = 1\nsum(...)"`), against CLAUDE.md's `indoc!` rule. Fix: put `n = 1` in the shared program and keep each case a one-liner.

**Nit — `src/ccl/channelize.rs:806-815`.** The `erase_chan_domains_in_predicates` doc stacks three claims in one sentence with a content-bearing em-dash aside. Fix: one claim per sentence.

**Pre-existing, not this PR — `src/ccl/channelize.rs:940`.** `c = [x for x in out]` (unfiltered, let-bound) panics at head and at the parent: ``type residue survived channelize on let binding `__mono` : (chan(out) ⤇ Int)``. Separate issue.

### Verification

- `cargo test -q a_filtered_comprehension_reads_a_feed_channel` at head: 2 passed.
- Same test at the #195 head with only the test file restored from this PR: 2 failed with ``channel domain chan(out) survived channelize at `__elem` ``. The fix is load-bearing.
- At head: `inspector_goldens` 12 passed, `predicate_sharing` 6 passed, `compilation_pipeline` 955 passed, `--lib channelize` 12 passed.
- `c = [x for x in out if x > 1]` / `sum(c)` as a standalone program: panics at head and parent.

### Next steps

1. Erase predicates in the binder closure; add the let-bound case to the rstest.
2. Thread one `PredMemo` from `run`.
3. Return a real changed bit from the callback; either move the pre-clone probe into the shared walker or name the recurring shape in the PR body.
4. Point `assert_no_type_residue` at the predicate walk; fix the two stale comments.
5. Note the `mut_elim` twin; file the `__mono` binder failure separately.

dpmills:
All five findings taken; the sixth is deferred with the shape named in the PR body.

**Binder slots.** `erase_chan_domains_in_slot` is now the one entry point for a type slot — structure then predicates — and the `walk_binders_mut` closure goes through it for both `b.ty` and `b.user_annotation`, as do a node's own type and annotation. `c = [x for x in out if x > 1]` then `sum(c)` is pinned as `a_let_bound_filtered_comprehension_reads_a_feed_channel`; reverting just the binder line fails it with ``type residue survived channelize on let binding `c` : ({[0, 1] | __elem ▷ out ▷ (λ x : Int → x > 1)} ⤇ Int)``.

**One memo per pass.** Built in `run` and threaded through `erase_chan_domains` beside `map` and `kinds`.

**The changed bit.** `erase_chan_domains_in_type` returns whether it rewrote anything, `go` folds it over the predicate's slots and children, and the callback returns that.

**The eraser and its checker.** `predicates_hold_type` takes the test it applies. The gate asks for a channel type (`is_channel_type`); `has_type_residue` asks for anything channelization must have removed (`is_type_residue`, which is the channel types plus `Hole` / `Infer`) and now walks predicates, so `assert_no_type_residue` covers what the eraser is answerable for. That is what the `let`-bound program trips above — the checker reports it, not a downstream wall. Both stale comments are rewritten.

**The pre-scan.** Deferred, and named in the PR body as a recurring shape: the probe belongs on the shared term inside `walk_refined_predicates_mut` / `PredMemo::rebuild`, before the clone, and every pass with a conditional predicate rewrite will otherwise re-derive it locally. The `TODO(discarded-copy)` on `PredMemo::rebuild` is the same cost from the other end.

**`mut_elim` twin** and the **`__mono`** failure: both filed, and both named in the PR body. The `__mono` residue is in type structure rather than a predicate, and this change does not move it.

**The `\n` case.** Gone — `n = 1` is in the shared program and both cases are one-liners. The `let`-bound program is its own test rather than a third case, since two lines of its own content is what `indoc!` is for and the slot it exercises is a different one.

Full `./ci.sh` is green at this head, fixtures included.

sortalongo (approved):
All five findings verified fixed at `6f789193`, by reading the code rather than the summary. The binder fix is load-bearing: reverting only `src/ccl/channelize.rs:896,898` to `erase_chan_domains_in_type` fails `a_let_bound_filtered_comprehension_reads_a_feed_channel` with ``type residue survived channelize on let binding `c` : ({[0, 1] | __elem ▷ out ▷ (λ x : Int → x > 1)} ⤇ Int)``, while the other two cases still pass, so the new test discriminates the binder path. One `PredMemo` at `:635`, threaded from the single `erase_chan_domains` entry. Real changed bit folded through `go` at `:845-850`. `has_type_residue` at `:154-164` reaches predicates through the parameterized `predicates_hold_type`. Approving; the deferred pre-scan is named in the body, which is what I asked for.

Three things to fold in, none of which need another round.

**The new entry point is not applied recursively — `src/ccl/channelize.rs:846`.** Inside a predicate, `go` calls `erase_chan_domains_in_type`, so a refinement sitting in a predicate's own type slot keeps its channel domain; `predicates_hold_type`'s `in_type` has the same limit, so `assert_no_type_residue` cannot see the miss. Both shared walkers hand the callback the means to re-enter for exactly this (`src/ccl/ccl_utils.rs:1729-1731`, `:2140-2143`). Fix: call `erase_chan_domains_in_slot` there.

**The two follow-ups are not in the tracker.** No issue file has been added to the vault since 2026-09-11, so the `mut_elim` twin and the `__mono` residue live only in this PR body, which becomes the squash-merge commit message. Fix: file both before merging.

**Two doc claims are wider than the code.** `src/ccl/channelize.rs:854-856` says every slot the pass reaches, but `Expr::walk_type_slots_mut` (`src/ccl/expr.rs:1946-1963`) enumerates six slots to this pass's hand-listed four, adding `Cast { target }` and `Transact { domain }`. And `a_let_bound_filtered_comprehension_reads_a_feed_channel`'s doc credits the strict wall for a failure the debug `assert_no_type_residue` (`channelize.rs:979`) reports, a distinction this file's own docs draw. Fix: narrow both to what holds.

Tests run at head: `channelize` 12+2 passed, `a_filtered_comprehension_reads_a_feed_channel` 2 passed, `a_let_bound_filtered_comprehension` 1 passed, `predicate_sharing` 6 passed, `inspector_goldens` 12 passed.

dpmills:
All three folded in, on top of the reviewed commit and rebased onto main.

**The new entry point is applied recursively.** `go` calls `erase_chan_domains_in_slot`, taking the memo `walk_refined_predicates_mut` hands the callback. The checker follows: `predicates_hold_type`'s inner walk re-enters through a new `slot_holds_type`, which is `structure_holds_type` (the reach of `erase_chan_domains_in_type`) plus `predicates_hold_type`, so the two trios mirror each other pairwise and `has_type_residue` is now one call to the slot-level checker.

No program reaches the recursive arm today, and it is worth saying where it goes instead of leaving that as an untested branch. I instrumented the nested call to report when it fires and ran the whole suite: zero hits. Both programs that would build the shape — a filter whose predicate contains a filtered read, and the same read `let`-bound and referenced from a filter — are rejected before channelize, by the scope invariant at `src/ccl/infer_var.rs:482`:

```
open bound recorded on ?117: `out` is free in the upper bound `{chan(out) | __elem ▷ out ▷ (λ y : Int → y > 1)}` but is neither in the holder's telescope [x, __elem] nor discharged by the edge's substitutions
```

What does compile is a filter over an already-filtered read, `sum([x for x in [y for y in out if y > 0] if x > 1])`, so that is pinned as a third case, `over_a_filtered_read`. It exercises the gate change without reaching the recursive arm.

**The two follow-ups.** Both issue files were written on 2026-09-15 but never committed, which is why the tracker check found nothing — they existed only as untracked files in the vault working copy. Committed now as `60b478d`, and the PR body names both by file stem so the squash-merge message points at them: `mutability-mut-erasure-skips-predicates` and `defer-let-bound-channel-read-residue`.

**The two doc claims.** `erase_chan_domains_in_slot` no longer claims every slot the pass reaches; it states what it is, and that `has_type_residue` checks the same pair. The `let`-bound test credits `assert_no_type_residue`. A third claim in the same class was still standing: `assert_no_type_residue`'s own doc said refinement predicates are checked by the strict `typecheck` instead and that walking them here would need cycle guards — this PR ships that walk. Rewritten.

`./ci.sh fast`, the release and lib-only clippy passes, `cargo doc`, and `doc_refs` are green locally; the rest is on the PR's own CI run, `fixtures` included.

PR #200, opened 2026-09-15
Retire `stream` and `arrow` as second names for `Type::Fun`, and spell a feed's read view `⤇`

dpmills:
`stream` and `arrow` were each a second name for a `Type::Fun`: `stream` for the one a `Type::History` denotes, `arrow` for a function type in general. Neither had a definition, and an undefined metaphor drifts and then collides (`.claude/skills/spec-prose/SKILL.md`, "Invented vocabulary"). Each such use now names the mechanism instead: a `Type::Fun`, a function type, a collection, or the handle's read view, which the solver already uses.

A feed's read view is built `FunKind::Data` at every site that constructs it, and `Display` writes `feed(domain ⤇ value)`, but twenty-one comments and doc passages spelled it with the compute glyph `⇒` — including `docs/mutability-walkthrough.md`, which stated the rendered form and a sample render. Those now say `⤇`. The `⇒` of an inference rule is logical implication and is untouched.

The change is comments, doc comments, design-doc prose, a few panic and assert message strings, one test name, and two test-local bindings in `src/ccl/subst.rs`; no executable code moves. Both words keep their other senses: `stream` for a sequence of values over positions (`commit_stream`, `StoreValueStream`, `TcpStream`, `FinalOrDefault`'s argument) and `arrow` for a glyph or token (`FunKind::arrow()`, `Token::Arrow`, diagram arrows).


sortalongo (approved):
**Verdict**: Approve. Every fix below is one word or one re-wrap, so no second round is
needed. Findings 1 through 5 are the ones to take before merging: four comments the PR
rewrote still carry the compute glyph on a data function, which is the defect the PR
exists to remove.

**What the PR does**

- Replaces `stream` and `arrow` as names for a `Type::Fun` with `collection`, `read view`, and
  `function type`, in comments, doc prose, panic strings, one test name, two test locals.
- Flips 21 renderings of a feed's read view from `⇒` to `⤇`.
- Moves no executable code.

**Findings**

1. `src/ccl/channelize.rs:609-610` — the edit leaves "erase each `Feed`-kind history to its bare /
   bare `Fun`", duplicating "bare". Delete one.

2. `src/ccl/ty.rs:1184`, `src/ccl/ty.rs:1192`, `src/ccl/channelize.rs:1960`,
   `src/ccl/mut_elim.rs:955`, `src/ccl/mut_elim.rs:963` — each names a data function and spells it
   `⇒`, inside a comment the PR edited elsewhere: `transact_phase::history_ty` builds
   `data_fun(Txn, value)`, `copair_type` returns `Type::data_fun`, and `ccl_utils::history_ty`
   returns `data_fun`. Spell all five `⤇`.

3. `docs/mutability-walkthrough.md:628` — "each `Feed` history to its stream `Fun`" is the phrase
   the PR converted in `channelize.rs:37`, left in a doc the PR edited. Write "to its bare `Fun`".

4. `docs/mutability-walkthrough.md:635`, `:681`, and lines 70, 161, 187, 539, 544, 550, 551, 678 —
   every history in this doc renders `⇒`. Lines 635 and 681 are the erased feed channel `copair_type`
   builds as a data function. The other eight render `:=` histories (`cnt`, `pool`, `__commits`,
   `to_out_*`, `__hist`), which `history_ty` also builds as `data_fun`, the same object finding 2
   covers. Write `⤇` at all ten. On line 161 flip only the inner `𝐼 ⇒ 𝑉`; the outer arrow is
   `get_prev_seq`'s own compute type.

5. `docs/chl-spec.md:2928`, `src/ccl/planning/loops.rs:508`, `src/ccl/infer/schemes.rs:101`,
   `src/ccl/ops.rs:354`, `src/ccl/ccl_utils.rs:828` — `stream` still names a function type: the
   chl-spec line is the mirror of the sentence converted at `design/mutability.md:26`, `loops.rs`
   says "stream type (𝐼 ⇒ V)" in a file the PR converted elsewhere, and `ccl_utils` has a local
   `refined_stream` holding a `Type`. Convert the first four; report the identifier.

6. `src/ccl/infer/check.rs:365`, `src/ccl/infer/solve.rs:464`,
   `src/ccl/infer/solver/constrain.rs:1569`, `src/ccl/ty.rs:1201`, `:1838`, `:2560`,
   `tests/compilation_pipeline/feeds_cases.rs:420` — deleting "stream" leaves "its whole
   `domain ⤇ value`", an adjective with no noun. Restore one: "its whole channel `domain ⤇ value`".

7. `tests/type_check.rs:3066` — the rewritten line keeps `(α → γ)`, the term arrow, for a function
   type. Write `(α ⤇ γ)`.

8. `src/ccl/simplify.rs:1967`, `src/ccl/subst.rs:2579` — `simplify_exponential_eta_carries_the_curry_arrow`
   and `a_data_compose_is_not_downgraded_to_a_compute_arrow` keep `arrow` while their own comments
   and locals converted. Rename both, or convert no identifiers and say so in the body.

9. `src/ccl/design/mutability.md:419` — the substitution grows a line already past the 100-column
   wrap to 139 characters. Re-wrap the paragraph.

10. `src/ccl/channelize.rs:817` — "viewed as the `domain ⤇ value` it states" is longer than the
    phrase it replaces and pushes an em-dash to the head of the next line. Write "viewed as its
    `domain ⤇ value`" and re-flow.

**Checked and clean**

- All 21 glyph flips sit on data functions. Verified against the constructors:
  `solve.rs:517`, `constrain.rs:1581`, `constrain.rs:1604`, `channelize.rs:741`,
  `ccl_utils::history_ty`, `transact_phase::history_ty`. None is a capability or an unpinned kind
  variable, and none is wrong.
- `⇒` is kept where it means logical implication or a compute function: `constrain.rs:1664`,
  `subst.rs:1076`, `ty.rs:457-468`.
- `read view` is the codebase's existing name, not a coinage: it predates the PR in `solve.rs`,
  `coalesce.rs`, and `type-inference.md:1129`. `collection` and `function type` are existing names
  too.
- `arrow` survives only in the glyph and token senses: `lexer.rs:122`, `FunKind::arrow()`,
  `design.md:247`.
- PR body matches the diff: 21 glyph sites across 20 lines, plus the panic strings, test name, and
  two `subst.rs` locals it lists. The body is within length.
- CI: one check, `check` / CI, SUCCESS.
- `./ci.sh doc_refs` is unaffected: no heading is renamed and no citation text changes.
- Edited Markdown re-wraps under 100 columns except finding 9; `ir.md` and `lowering.md` paragraphs
  stay soft-wrapped as they already were.

dpmills:
All ten taken, in `bd7bad413`, with `restrict` and `filter_values` following in `c9a228503`. The full `./ci.sh` is green.

1. Deleted the duplicate `bare`.
2. All five now `⤇`. `planning/loops.rs:508` carried the same defect — the tap's type is the erased channel `copair_type` builds as a data function — so that one flipped too, alongside its finding-5 word.
3. `to its bare `Fun``, matching the `channelize.rs` phrasing.
4. Done, and extended: the walkthrough had ten more histories rendering `⇒` that the rule reaches — the second `__commits` (`:547`), the two `__hist` bindings (`:592`, `:668`), `balance` in the mermaid node (`:170`) and the ASCII box (`:198`), `get_prev_txn`'s inner `𝐼 ⤇ {time, write}` (`:162`, outer arrow kept), the mutable variable's extent (`:90`), and the store's `Txn ⤇ {key: value}` (`:771`). Twenty flips in that file. A `Transact` writer's `[reads]⇒[writes]` footprint is untouched — not a function type.
5. First four converted. On the identifier: converted rather than reported. `refined_stream` holds a `Type`, so the PR's own rule reaches it the same way it reached the two `subst.rs` test locals.
6. Restored a noun at all seven. Not the same noun everywhere: `read view` where the sentence's subject is already the channel (`check.rs`, `constrain.rs`, `ty.rs` ×3), `channel` in `solve.rs:464` where the clause before it says "read view", and `collection` in `feeds_cases.rs` where the claim is about the type's shape.
7. `(α ⤇ γ)`.
8. Renamed: `simplify_exponential_eta_carries_the_curry_type` and `a_data_compose_is_not_downgraded_to_a_compute_function`. No other references.
9. Re-wrapped the paragraph. Its first line was 106 columns before this PR, so the whole paragraph moved rather than just the substituted line.
10. `viewed as its `domain ⤇ value``, re-flowed. Three more comments had the same dangling em-dash after a re-flow (`check.rs:365`, `solve.rs:464`, the walkthrough caption at `:202`, whose ASCII alignment is preserved) — fixed in place.

Five more `stream`-as-`Type::Fun` sites outside the lists, in files the PR already edits: `ccl_utils.rs:310`/`:313` (a decision history and its eliminated collection), `:356`/`:363` (a tap), `channelize.rs:1545` (an accumulator binding), `design/mutability.md:52` (`get_prev_txn`'s `𝑤`), and the walkthrough's read law (`:115`), diagram caption (`:202`) and reply channel (`:632`).

**`restrict` and `filter_values` (`c9a228503`).** These filter a collection, so every rendering of their types now spells the collection slots `⤇` — the upstream, the narrowed result, and `restrict`'s predicate, which `fn_of_bare_predicate` declares `FunKind::Data` and which `design/optimization.md` already rendered `⤇`. Eight files: `ops.rs`, `ccl_utils.rs` (including two `expect` strings), `simplify.rs`, `operator_conversion.rs`, `design/optimization.md`, `design/type-inference.md`, `interpreter/design-operators.md`, `operational-semantics/lowering.md`. `iterate`'s result goes with them — `make_iterate` builds `Type::data_fun`, and `make_restrict` rides that kind through `fun_like`, which is what makes a restrict chain's collections data in the first place.

Three slots deliberately keep `⇒`, each checkable:

- A transformer's own arrow. Its domain is a function type, which is why it is applied to its upstream rather than composed with it (`ops.rs:295-297`).
- `iterate`'s predicate: every call site passes `trivially_true_predicate`, which builds `Type::fun`.
- `map_filter`'s predicate. Its kind comes from whatever `map_filter_site` found, so nothing in the constructor settles it.

`make_restrict`'s comment said the kind riding through from `upstream` means "restricting a capability yields a capability". `make_iterate` calls that a kind error — every chain led by an iteration source inherits `Data`. The comment now gives the reason `fun_like` is used instead of restating a case that cannot arise.

Two code sites disagree with these docs, both in the `Type::fun`-hardcodes-`Compute` cluster rather than anything this PR touches: `trivially_true_predicate` (`ccl_utils.rs:709`) builds the predicate `Compute` while `fn_of_bare_predicate` declares it `Data`, and `lambda_elim.rs:1193` types `filter_values`'s filtered collection `Type::fun`.

Still deferred: `infer/schemes.rs` writes every scheme's function type with the term arrow `→` (`∀α β. ((α → β), β) → β`), the same shape finding 7 corrected in `tests/type_check.rs`. Converting that file is a sweep across every scheme, not a one-word fix.

PR #201, opened 2026-09-15
Gate a `Memo`'s input pull on a one-bit input-notified flag

dpmills:
Every pull of a `Memo` walked its whole input subtree, even with nothing new upstream to merge, so a read cost the size of the graph rather than the amount of new data. `Notified` is a one-bit flag that a producer's `subscribe` installs on its own input and that a pull clears: while it is clear and the cache is non-empty, the cache is the answer and no pull goes below. `Memo` is the one operator wired to it. On another branch's asset-cart demo, the 48 whole-program tests in release go from ~25-31s to ~4s.

## The gate

`ProducerBase::new` keeps the `Notified::Always` default, so every other producer still reads on every pull; `ProducerBase::listening` takes a live flag. Its `Rc<Cell<bool>>` is allowlisted in `check_shared_state.py`. Two cases stay ungated, both documented at `MemoProducer::get_impl`: an empty cache, which is not yet an answer, and a drained input, left to the existing release-only short-circuit so its debug path stays a release-contract probe.

## Delivery becomes load-bearing

An operator holding a cumulative cache re-reads only after a delivery, so a loop that pulls without delivering gets one answer repeated. `Scheduler::pull_laps` is the deliver-then-pull alternation `src/main.rs` runs, exposed for tests, and `read_initial_scalar` now takes a `Scheduler` so it delivers between pulls. `InductionDriverProducer` requests a wakeup when it closes its source — the one transition nothing else re-arms, and the one a sibling reader left holding a pre-final cache depends on.

## Tests

`pull_laps` replaces the hand-rolled pull loops. Twelve test sites passed `&mut Scheduler::new()` inline at `subscribe` and dropped the scheduler, along with every queued wakeup, on the next line; they now hold a named `sched` that outlives the subscription. Those tests passed before only because no pull depended on delivery. New coverage: the gate and its empty-cache exemption in `fanout.rs`, and both pull orders of two memo'd readers over one cycle.


sortalongo (approved):
Approving. The design is right: a per-producer dirty bit replaces the global round clock #189 tried, the change is confined to `Memo`, and the win is large. Local run: `compilation_pipeline` 950 passed, 0 failed.

Items below, in order of weight. No test exercises the gate (T1; a draft is in the appendix). Two stale-read scenarios are not ruled out (C1, C2); each needs a fix or a sentence saying why it cannot happen. Two docs now state the old policy (C3, D1). The explanatory prose is written seven times, and the sentence copied most often is inaccurate (R3).

## Correctness

Each item is a scenario the PR does not rule out. None is reproduced by a test yet.

**C1. A sibling reader can hold a pre-final cache after the recurrence closes.** The recurrence advances on the pull, not on the wakeup. The driver re-arms only while not done (`src/interpreter/commit_operator.rs:3053`). In one walk, reader B's memo pulls first and clears its flag. Reader A's memo then pulls and drives the store to done. Nothing re-arms. B's flag stays clear and its cache is non-empty, so every later pull returns the pre-final tile. Before this PR, B re-pulled and saw the final state.
- Fix: request one wakeup on the `done` transition, or add a test with two readers on the cyclic fan that shows the order does not matter.

**C2. An init-seed read can be gated after a non-empty undecided pull.** `read_initial_scalar` (`commit_operator.rs:340-344`) pulls the same producer up to eight times with no delivery. The empty-cache carve-out at `fanout.rs:891` covers a cache that stays empty. It does not cover the `continue` paths at `:363` and `:366`, where the first pull filled the cache with an undecided tile. The next seven pulls return that tile unchanged.
- Fix: thread the scheduler through and deliver between pulls, or state why an undecided seed tile cannot reach a memo.

**C3. The gate runs in debug and skips the release-contract probe.** The drained shortcut at `fanout.rs:876` is release-only for a stated reason: a memo sits above most scalar producers, so skipping the upstream pull in debug shields them from the release-contract check (`fanout.rs:867-875`). The new gate skips the same pull in every build. The comment says "Unlike the drained check above, this holds in every build" and does not reconcile the two.
- Fix: put the gate under the same `cfg!(debug_assertions)` guard, or rewrite the comment at `:867-875` and the `Memo` row in `src/interpreter/design-operators.md` to the reduced coverage.

**C4 (nit). A widening guard can be answered from a narrower cache.** No live caller passes a non-universal guard (`main.rs:155`, `sinks.rs:135`, `commit_operator.rs:338`).
- Fix: `debug_assert!` the guard is universal at the gate.

## Tests

**T1. No test asserts the gate.** The diff adds no `#[test]`. The one memo unit test (`fanout.rs:1029`) builds `ProducerBase::new`, which is `Notified::Always`, so it never reaches the gate. The PR's evidence is a suite count and a wall-clock number, neither of which distinguishes "pulled less" from "happened to pass".
- Fix: a counting input under a memo. Assert the second pull does not read the input, a notification re-enables the read, and an empty cache is never gated. Draft in the appendix.

**T2 (nit). One test still pulls without delivering.** `commit_operator.rs:6056` pulls eight times in a loop. The PR added `let mut sched` at `:6050` and no `check_for_notifications`.
- Fix: alternate there as at the other sites.

## Design docs

**D1. Two docs state the old policy.** `src/interpreter/design-operators.md`, "Tile Operators", `Memo` row: "Only release builds then skip the upstream pull." `src/interpreter/scheduler.rs:12-22` states the driver calls `check_for_notifications` between pulls but not that a pull without delivery may now return a stale cumulative answer. The PR touches no `.md`.
- Fix: update the `Memo` row. Add one sentence to the scheduler module doc naming the consequence.

## Redundancy and documentation

**R1. The `Rc<Cell<bool>>` justification is written three times.** `src/interpreter/tile_operators/mod.rs:261` (type doc), `mod.rs:276` (`shared-state-ok` on `Flag`, which cites the type doc and then restates it), `.github/scripts/shared-state/check_shared_state.py:87` (eight lines restating both).
- Fix: keep the type doc. The `Flag` comment becomes `// shared-state-ok: see the type doc.` The Python entry becomes one line naming the type.

**R2. "Installed on the operator's own input" is written twice.** `mod.rs:307` (doc on `Notified::consumer`) and `fanout.rs:817` (call-site comment in `Memo::subscribe`, same sentence, stress italics on `*this*`).
- Fix: delete the call-site comment.

**R3. "The cycle re-arms through the wakeup queue, so an undelivered lap does not run" is written four times, and it is wrong.** `commit_operator.rs:4477`, `tests/compilation_pipeline/helpers.rs:67-73`, `tests/compilation_pipeline/sources_incremental.rs:596`, and the PR body. A lap runs on any pull that reaches the store (`commit_operator.rs:6053`, "Pulling the reader drives the cycle"). Delivery is what makes something re-pull.
- Fix: one corrected sentence in `helpers.rs`. Delete the other two comments. Correct the body.

**R4. The `get_impl` gate comment restates itself and the code.** `fanout.rs:879-890`, thirteen lines for a two-line `if`. Line 879 and line 886 make the same claim about a cumulative cache. The bold heading restates the condition. Bold is stress formatting, which CLAUDE.md's prose rules exclude.
- Fix: two sentences. The cache is cumulative, so when nothing has notified since the last pull it is the answer. An empty cache is not yet an answer, because a demand read outside the delivery loop arrives before any notification, so it pulls.

**R5. The field doc repeats the type doc.** `mod.rs:330` is `mod.rs:253` again.
- Fix: delete the field doc.

**R6. History in a doc comment.** `mod.rs:272-273`: "what every producer had before this existed."
- Fix: delete the clause.

**R7. Three names for one thing.** "latch" (PR body, `mod.rs:276`), `Notified` (type), "flag" (variant, constructor, ten comment uses).
- Fix: "flag" everywhere in prose. Drop "latch".

## Structure

**S1. The same two lines are added at eighteen test sites.** `sched.check_for_notifications(); producer.get(…)` in `commit_operator.rs` (8), `sources_incremental.rs` (5), `transactions.rs` (5). `pull_to_terminal` (`commit_operator.rs:4471`) already wraps the to-terminal case. The fixed-count loops have no helper.
- Fix: one test helper that delivers then pulls, with a lap count. Use it at every site.

**S2. The scheduler-lifetime change is load-bearing and unstated.** Twelve sites change `&mut Scheduler::new()` to a named `sched` that outlives `subscribe`. Before this PR those tests dropped the scheduler, and every wakeup with it, right after subscribing. They passed because a pull did not depend on delivery. The PR fixes that harness bug in passing and does not say so.
- Fix: one sentence in the body.

## Appendix: drafted gate test

Style follows `fanout.rs`'s `mod tests`. Not written to the repo.

```rust
/// A memo whose input has said nothing since its last pull answers from the cache
/// without reading the input. A notification re-enables the read. The tile is
/// incomplete (`LessThanEq(0)`) so `upstream_drained` stays false and the drained
/// shortcut does not mask the gate.
#[test]
fn a_memo_reads_its_input_only_after_its_input_notifies() {
    use crate::interpreter::tiling::Predicate;
    use crate::interpreter::{ColumnValue, Consumer, Value};
    use bit_set::BitSet;
    use std::cell::{Cell, RefCell};
    use std::rc::Rc;

    /// One row on the first pull, empty afterwards. Counts pulls and keeps the
    /// consumer the memo installed so the test can fire it.
    struct Counting {
        base: OperatorBase,
        pulls: Rc<Cell<usize>>,
        consumer: Rc<RefCell<Option<Box<dyn Consumer>>>>,
    }
    struct CountingProducer {
        base: ProducerBase,
        pulls: Rc<Cell<usize>>,
    }

    impl TileOperator for Counting {
        impl_operator_base!();
        fn visit_inputs(&self, _visit: &mut dyn FnMut(InputEdgeSpec<'_>)) {}
        fn subscribe(
            &mut self,
            _intent_guard: TileGuard,
            consumer: Box<dyn Consumer>,
            _scheduler: &mut Scheduler,
        ) -> Box<dyn TileProducer> {
            *self.consumer.borrow_mut() = Some(consumer);
            Box::new(CountingProducer {
                base: ProducerBase::new(CountingProducer::alloc_id(), self.tiling()),
                pulls: self.pulls.clone(),
            })
        }
    }

    impl TileProducer for CountingProducer {
        impl_producer_base!();
        fn get_impl(&mut self, _projection_guard: TileGuard) -> Tile {
            self.pulls.set(self.pulls.get() + 1);
            if self.pulls.get() > 1 {
                return self.tiling().empty_tile();
            }
            Tile::SealedFunction {
                domain: ColumnValue::UInts(vec![0]),
                codomain: Box::new(Tile::Scalar(ColumnValue::Ints(vec![7]))),
                domain_predicate: Predicate::LessThanEq(Value::UInt(0)),
                deleted: BitSet::new(),
            }
        }
        fn release_impl(&mut self, _obsolete_guard: TileGuard) {}
    }

    let tiling = Tiling::SealedFunction {
        domain: Extent::Base(BaseType::UInt),
        codomain: Box::new(Tiling::Scalar(Extent::Base(BaseType::Int))),
    };
    let pulls = Rc::new(Cell::new(0usize));
    let consumer_slot: Rc<RefCell<Option<Box<dyn Consumer>>>> = Rc::new(RefCell::new(None));
    let mut memo = Memo::new(Box::new(Counting {
        base: OperatorBase::new(tiling.clone()),
        pulls: pulls.clone(),
        consumer: consumer_slot.clone(),
    }));

    let mut sched = Scheduler::new();
    let mut producer = memo.subscribe(tiling.universal_guard(), Box::new(|| {}), &mut sched);

    let first = producer.get(tiling.universal_guard());
    assert_eq!(pulls.get(), 1, "the first pull reads its input");
    assert!(!first.is_empty(), "the first pull fills the cache");

    let second = producer.get(tiling.universal_guard());
    assert_eq!(second, first, "the gated pull answers the tile it just answered");
    assert_eq!(pulls.get(), 1, "no notification since the last pull, so no read");

    consumer_slot.borrow_mut().as_mut().unwrap().notify();
    let third = producer.get(tiling.universal_guard());
    assert_eq!(third, first, "the input had nothing new");
    assert_eq!(pulls.get(), 2, "a notification re-enables the read");
}
```

A companion with an input that returns empty on pull 1 and the row on pull 2, asserting two reads after two ungated pulls, pins the empty-cache carve-out at `fanout.rs:891`.

dpmills:
All sixteen taken. Two of the four correctness items did not reproduce; both are closed at their source anyway, and I say below what the evidence was.

## Correctness

**C1 — sibling holding a pre-final cache.** The driver now requests one wakeup on its `done` transition, under the existing `!source_fully_released` latch, so it fires exactly once.

I could not construct the interleaving. With two memo'd `StoreDenseRead`s on one cycle fan, a laggard pulled to a non-empty pre-final cache (`[1]`) and then left alone while the leader converged answered `[1, 3, 6]` on its next bare pull — the fan's wakeups during the leader's laps re-flag every branch. So the test is the second option you offered: `both_memo_readers_of_one_cycle_reach_the_final_state`, an `rstest` over both pull orders, pinning that the order is not the runtime's to choose. The wakeup closes the gap by construction rather than by measurement.

**C2 — init-seed read.** `read_initial_scalar` takes `&mut Scheduler` and delivers before every pull past the first. Both call sites already have the scheduler in scope, and `subscribe` is not inside a `get`, so the delivery meets `check_for_notifications`' stated precondition. This is right independent of the gate: a bound spent on pulls with nothing delivered between is a bound spent on steps the runtime does not take.

**C3 — the debug probe.** Taken as the first option, in the form that keeps one rule rather than two: the gate exempts a drained input and leaves it to the release-only rule above, so a drained input is still pulled in debug and the probe is intact. Expressed as `&& !self.upstream_drained`, which is dead in release (that path has already returned) and load-bearing in debug. The `Memo` row in `src/interpreter/design-operators.md` states the policy, including the exemption and why it exists.

**C4 — widening guard.** `debug_assert!` at the gate, inside the branch that answers from the cache, so it only fires when the shortcut is actually taken.

## Tests

**T1.** Three of them, from your draft: `a_memo_reads_its_input_only_after_its_input_notifies` (gated pull does not read; a notification re-enables it) and `a_memo_with_an_empty_cache_reads_its_input_unnotified` (the carve-out). Each fails with the corresponding condition removed — I checked both. The double is a `Scripted` operator that hands over one tile per pull, counts pulls, and keeps the consumer the memo installed.

**T2.** That loop and every other now go through the helper below.

## Structure

**S1.** `Scheduler::pull_laps(scheduler, producer, laps, done)` — deliver, then pull, up to `laps` times, stopping on a predicate. It lives in the library behind `#[cfg(any(test, feature = "test-helpers"))]`, which is what lets the unit tests and the integration crate share one helper rather than one per crate. Every hand-rolled pull loop uses it, `pull_to_terminal` and the pipeline harness included. The `check_for_notifications` calls that remain in tests are single deliveries after adding source data, plus the two-reader test, where the interleaving is the subject.

**S2.** In the body.

## Docs and prose

**D1** the `Memo` row and the `WakeupQueue` doc; **R1** the `Flag` comment is now `// shared-state-ok: see the type doc.` and the Python entry one line; **R2** the call-site comment deleted; **R4** the gate comment is two sentences and the bold is gone; **R5** the field doc deleted; **R6** the history clause deleted; **R7** "flag" throughout, "latch" gone from prose.

**R3.** You are right that the sentence was wrong. It now exists once, on `pull_laps`, and says what is actually true: a lap runs on any pull that reaches the store, so a loop that only pulls is not stalled by itself — what it loses is the read, because a memo answers from its cache and the pull never reaches the store. Corrected in the PR body too.

Full `./ci.sh` is green at this head.

PR #203, opened 2026-09-15
Give a curried tile any number of domain levels so correlated comprehensions nest to any depth

dpmills:
Correlated comprehensions nested more than two deep had nowhere to land: `Tile::CurriedFunction` and `Tiling::CurriedFunction` held exactly two domain levels (`domain1`, `domain2`) with one offsets array between them, so `Product` emitted a curried tile but accepted only a sealed one as its outer stream, and `MapAggregate` folded straight from two levels to a `SealedFunction`. This replaces the fixed pair with `domains: Vec<_>`, one offsets column per level above the innermost, a structured (boxed) codomain, and one set of logically-removed positions per level. `[[[… for v in …] for r in …] for q in …]` compiles at any depth, aggregated at every level or at none.

## One level per pairing, one per fold

`Product` appends a level to whatever function arrives — a stream of rows, or a function already grouped by earlier pairings — which is what makes it closed under its own output and the depth unbounded. That closure is what the depth rests on, and only `Product` has it: a source that is the same set for every row nests without limit, while `IterateRowCollection`, which reads a different collection per row, still takes a `SealedFunction` and adds one level. `MapAggregate` collapses the innermost level, leaving a sealed function where one domain remains and a curried one where more do. `MapExtractAggregate` takes either shape and changes only the codomain; `Uncurry` and `MapResult` still require exactly two levels, and now say so (`src/interpreter/design-operators.md`, "A correlated inner comprehension").

## Accumulators key on a path

A key repeats across its siblings' groups, so below the outermost level a value no longer identifies an element. `MapAggregateProducer` keys accumulators on the element's path (`Tile::curried_paths_at`) and rebuilds the levels from the sorted paths (`build_curried_from_paths`); releases still read `path[0]`, the outermost level a guard names.

## A guard names a level

`Domain(p)` under 𝑛 `Codomain` wrappers references level 𝑛 (`guard_level` / `guard_at_level`). The old hard-coded single wrapper was right only at depth two: `Tile::to_guard` answers with the innermost level's keys, so at depth three it was handing them out against level 1's domain.

## A removal names its level

`deleted` stayed a single bitset over the flat innermost entries while `domains` and `offsets` went per-level, so a group ordinal and an entry index were one type indexing different things — which is how a group ordinal twice reached the innermost set. `Deleted` stores a set per level and every producer says which it marks. What gets removed does not change: each producer still marks the innermost level, a group going by way of its entries.

## Pruning propagates upward

Offsets are strictly ascending, so a group holding nothing cannot be represented; `retain_levels` drops every ancestor left with no surviving descendant. Boxing the codomain lets a fold's `Aggregation` sit under a curried tile, which gives `Tiling::codomain()` an answer for `CurriedFunction` and `Tile::retain` an `Aggregation` arm.

## Tests and invariants

`tests/compilation_pipeline/comprehensions.rs` covers depths two through four, an innermost source whose domain comes from a map rather than an extent, an inner comprehension reading nothing outer, and three levels with no aggregate at all. `validate_tile` checks the generalized shape: at least two levels, one offsets column between each adjacent pair, key uniqueness within a group rather than across a level, and `deleted` bounded by the innermost level.


dpmills:
This PR was marked merged by GitHub during a stack restructure — its head briefly became an ancestor of its base — and nothing from it reached `main`. The work it carried is now #210, merged with the empty-group invariant so the representation lands as one statement and no intermediate shape is written twice.

PR #206, opened 2026-09-15
Report the sub-comparison when a domain mismatch's two sides render identically

dpmills:
A `DataDomainMismatch` whose two domains print the same reports `expected σ, found σ`: one type shown twice, because a witness's binder identity is not something `Display` carries. `ConstrainError::DataDomainMismatch` now carries the comparison that found the pair unequal as `cause`, with `lhs` and `rhs` boxed to keep the variant off `clippy::result_large_err`, and `map_constrain_err` reports that cause in place of the inequality where the two sides render alike. The cause's own sides differ structurally, so `identical_rendering_hint` quotes the `WitnessRef` that separates them instead of telling the reader the types do not disagree. `a_domain_mismatch_whose_sides_render_alike_reports_its_cause` covers it.


sortalongo (approved):
Do a cruft-removal pass before merging pls.

dpmills:
Cruft pass done, in `3029124ce`, on top of a rebase onto the new `main`.

The reason the inequality carries the comparison that found it was written three times — the `cause` field doc, the reporting site in `map_constrain_err`, and the test doc. The field doc keeps it. The other two now state what their own code does and point there, which took the reporting comment from seven lines to four and the test doc from nine to six. Gone with them: a bold stress heading, a pair of sentences arguing the case ("the reader is left to distrust the rendering rather than read it"), and a restatement of the `expected σ, found σ` rendering the summary paragraph already gives.

Also de-italicized `*render alike*` on the field doc, per CLAUDE.md's no-stress-italics rule.

`a_domain_mismatch_whose_sides_render_alike_reports_its_cause` still passes.

PR #210, opened 2026-09-15
Collapse SealedFunction and CurriedFunction into one nesting Function tiling

dpmills:
A collection tiled one of two ways — `SealedFunction` for a single key level, `CurriedFunction` for exactly two — and the two-level form stored its groups behind strictly ascending offsets, so a key whose group held nothing had no representation. A per-group filter that removed a group's last entry dropped the key along with it, and the group vanished from the result. Both variants become one recursive `Tiling::Function { keys, values }` over a `Tile::Function { row_starts, keys, values, domain_predicate, deleted }`, Compressed-Sparse-Row at every level with **non-decreasing** `row_starts`. A group a filter empties keeps its key and folds to the aggregate's identity, nesting is no longer capped at two levels, and a `Record` may sit between two of them.

## One node per level

`row_starts` names where each ambient row's run of `keys` begins and `values` is a tile over those keys, so `CurriedFunction`'s `domain1`/`offsets`/`domain2`/`codomain` are the same columns split across two nodes, and a top-level collection is one node whose `row_starts` is `[0]`. `deleted` rides each node rather than the tile, so a removed group and a removed entry are bits at different levels instead of one bit read two ways. `Tile::len` becomes `Tile::rows`, and `validate_tile` checks the pairing the shape rests on: a scalar is one entry per row, a record is its fields over the same rows, a collection is one run of keys per row.

## An empty group is not a removed key

`Tile::retain_keys` masks a collection's keys within each row, so a row keeps its identity and its group shrinks — to nothing where every key goes. `retain_curried_function_partial_mask_prunes_empty_group` is now `retain_keys_partial_mask_empties_the_groups_it_clears`. A key that is *gone* is instead marked in `deleted` at its own level and taken, with its subtree, by `Tile::compact`.

`tests/programs/groupby_filtered_rollup` adds a sale whose `qty` the filter rejects, in a region holding no other sale, and its expectation goes from

```
Function [ "east" -> 150, "south" -> 75, "west" -> 200 ]
```

to

```
Function [ "east" -> 150, "north" -> 0, "south" -> 75, "west" -> 200 ]
```

## Aggregation folds the innermost level

`MapAggregate` collapses the **innermost** collection and leaves the levels above it intact (`fold_innermost`), where it used to take a `CurriedFunction` and fold its one inner level. It keys accumulators by an element's path rather than by its key value, a key below the outermost level repeating across siblings, and marks an element terminal where the input's `domain_predicate` names that element's outermost ancestor, rather than reading the predicate as one flag for the whole tile. `AggregateKind::accumulate`, `extract` and `initial_accumulator` take a `Tile` in place of a `ColumnValue`, which is what lets `Sole` yield a collection-valued element.

## A fan-in takes its pairing depth from the call site

Two arms that are projections of one grouped row are identical below the grouping, so with a single tiling variant their shapes no longer say whether the pair belongs at the group or at the row; the sealed-versus-curried split had named it. `fan_in_at` and `fan_in_named_at` take the depth, which op-conversion reads off the ambient input (`level_count`) or, for a product former with no input, off components that must agree (`leaf_ambient`, `debug_assert`ed). `a_binop_pairs_two_projections_at_the_grouped_row` pins it on `sum([sum([s.amount + s.qty for s in g]) for g in groupby(sales, \r -> r.region)])`.

## A guard nests exactly as the tiling does

`FunctionGuard::Domain` names a collection's own keys and `Codomain` steps one level in, so the arm that read `Codomain(Domain(_))` as a `CurriedFunction`'s inner domain is the general recursion now. `Tile::to_guard` names only what the **open** keys hold: a codomain guard is read against every row of the values it wraps, so naming a settled group's inner keys released that key value under a sibling still growing. `TileGuard::flatten_or` unions two arms that name one place, so a guard's shape no longer depends on how it was assembled.

## Rendering

A tiling prints one arrow per level, and an aggregation prints its accumulator tiling rather than an extent:

| Before | After |
|---|---|
| `SF({[0, 3]} → Int)` | `Fn({[0, 3]} → Int)` |
| `CF({[0, 3]} → Int → Bool)` | `Fn({[0, 3]} → Int → Bool)` |
| `agg(Sum, Int)` | `agg(Sum, Scalar(Int))` |

The three inspector wire fixtures under `web/src/__fixtures__/` are re-blessed for both.

## Docs and tests

`src/interpreter/design-operators.md` carries the reasoning — the `Tiling` and `Tile` tables, the empty-versus-removed rule, the guard-as-level-reference rule, and rewritten `MapResult`, `MapFilter` and `Uncurry` rows. `docs/mutability-walkthrough.md`, `docs/operational-semantics/deprecation.md` and `src/ccl/design/optimization.md` take the new names. The suite is green at 3042 tests.


src/interpreter/tiling/tile.rs
```diff
@@ -24,56 +24,50 @@ pub enum Tile {
...
+    /// key count. A [`Self::Scalar`] is one entry per row and a [`Self::Record`] is one
+    /// sub-tile per field at the same `R`; a collection groups instead of pairing, which is
+    /// what `row_starts` records.
+    ///
+    /// Stored Compressed-Sparse-Row-wise: `domain` is every row's keys run together, and
+    /// `codomain` is a tile over those keys. A chain of collections nests one node per level,
+    /// holding the same columns a flat level list would — and unlike a flat list it can say
+    /// where a record sits between two levels, which is what lets records and collections
+    /// nest freely.
+    Function {
```
sortalongo: Name this `DataFunction`. A collection level is a data function in the type vocabulary, the `⤇` arrow, so `Function` reads as the compute arrow and collides with it.

src/interpreter/tiling/tile.rs
```diff
@@ -363,151 +339,208 @@ impl Tile {
...
     ///
-    /// After retention the physical arrays are compact and `deleted` is cleared —
-    /// every surviving entry is considered live.  Use [`Tile::compact`] to build
-    /// the mask automatically from the current `deleted` set.
-    pub fn retain(&mut self, mask: &BitVec) {
+    /// The mask is over this tile's own rows, so a scalar keeps the entries it names, a
+    /// record keeps them in every field, and a collection keeps those rows' whole groups.
+    pub fn retain_rows(&mut self, mask: &BitVec) {
         match self {
             Tile::Scalar(cv) => cv.retain(mask),
```
sortalongo: Release-path panic. `valid_over` blesses a scalar column that is empty at any row count, and its own comment says why that state is real: a binop over a lagging operand builds the record before both columns arrive. This arm then hands that column to `ColumnValue::retain`, which opens with an unconditional `assert_eq!` on length. `to_guard` and `retain_keys` both reach it, and `select_rows` has the same defect via `select_indices`. Make both arms a no-op on an empty column, as `merged_column` already does.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+        })
+        .collect();
+    if r_rows.is_empty() {
+        return left.select_rows(&l_rows);
+    }
+    if l_rows.is_empty() {
+        return right.select_rows(&r_rows);
+    }
+    let mut taken = left.select_rows(&l_rows);
+    taken.merge_part(right.select_rows(&r_rows), false);
```
sortalongo: `merge_part` implements the whole-value fold here, not a row concatenation, and the interleaving in `order` is discarded. Merging two collections whose values are aggregations returns an accumulator with the wrong row count, so keys read each other's values. Debug catches it; release returns it. Select into the concatenation in `order`'s own sequence instead.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+                            keys.extend((l_from..l_to).map(RowSource::Left));
+                            keys.extend((r_from..r_to).map(RowSource::Right));
+                        }
+                    }
+                }
+            }
+            // A key a merge names sits in one of the two key columns, so picking it is a
+            // gather out of the two run together. A key both sides delivered is one key,
+            // and the two spellings are equal, so either serves.
+            let mut combined = l_domain.clone();
```
sortalongo: This clones and gathers both domain columns on every merge, so an incrementally delivered collection costs the size of what it already holds per pull rather than the size of the delta. The aggregation and store arms below already take a no-overlap fast path; the collection arm should too.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+    }
+
+    /// Whether a level sits here, or inside a record here.
+    ///
+    /// A [`Self::Record`] is one sub-tile per field over the *same* rows, so it introduces
+    /// no dimension of its own and a collection inside one is a level of whatever holds the
+    /// record. A [`Self::Aggregation`] and a [`Self::Store`] carry their own completeness,
+    /// so what they hold is answered by their own guard rather than by descending into it.
+    ///
+    /// Two questions about a codomain read alike and part here. *Does the chain continue?*
```
sortalongo: Keep the record-ends-the-chain rule here and cite it from the other seven places that restate it: lines 650, 751, 803, 843, 1878, 1911, and `design-operators.md` 128. The `MapAggregate` example appears verbatim in two of them. `tiling_kind.rs:202` already cites rather than restates, which is the shape to copy.

src/interpreter/tiling/tile.rs
```diff
@@ -242,12 +226,57 @@ impl Tile {
         }
     }
 
-    /// Merge the contents of `other` into `self`.  Requires the two tiles to be compatible (i.e. non-overlapping)
+    /// Merge the contents of `other` into `self`. Requires the two tiles to be compatible
+    /// (i.e. non-overlapping).
+    ///
+    /// A merge extends a value with a part of it that arrived later. At the top level that
```
sortalongo: Keep the merge-extends rule here and cut it from `merge_part`, `merged_rows`, the test doc at 2058, and `design-operators.md` 67. Six statements of one fact.

src/interpreter/tiling/tile.rs
```diff
@@ -24,56 +24,50 @@ pub enum Tile {
...
+    /// sub-tile per field at the same `R`; a collection groups instead of pairing, which is
+    /// what `row_starts` records.
+    ///
+    /// Stored Compressed-Sparse-Row-wise: `domain` is every row's keys run together, and
+    /// `codomain` is a tile over those keys. A chain of collections nests one node per level,
+    /// holding the same columns a flat level list would — and unlike a flat list it can say
+    /// where a record sits between two levels, which is what lets records and collections
+    /// nest freely.
+    Function {
+        /// Where each row's run of keys begins in `domain`, one entry per row. **Non-decreasing**:
```
sortalongo: The `row_starts` rule, two equal starts being an empty group, is also at 1324, 2300, and `design-operators.md` 82. Keep this one. Same pattern for the vectorized-over-rows invariant at 29, restated at 1278, 1300, and `tiling_kind.rs` 26, where line 27 here and line 26 there are byte-identical.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+        );
+        assert_eq!(
+            tile.values_at(2).rows(),
+            3,
+            "`C` stands over `B`'s three keys"
+        );
+    }
+
+    /// A delivery that re-states a key two levels down grows the innermost group, and the
+    /// levels above it keep their keys: the collapse runs at the level that repeated.
```
sortalongo: "the collapse" names a pass that no longer exists. Same at 1577 ("re-collapse"), 2082, and 2103. Say what `merge` does now. A comment has to make sense from the current version of the code alone.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+                vec![Value::UInt(100), Value::UInt(200)],
+                vec![Value::Int(10), Value::Int(20)]
+            ),
+            "both deliveries' entries sit in the one group, over the one scalar value"
+        );
+    }
+
+    /// The other half of that rule: a repeat that re-states the scalar beside the grown
+    /// collection has delivered one position twice, and there is no answer for which value
+    /// stands. `join_rows` says so rather than picking.
```
sortalongo: `join_rows` is a dangling identifier. The reporter is `merged_column` at 1221.

src/interpreter/design-operators.md
```diff
@@ -57,17 +56,33 @@ A `Tile` holds the actual data. Its shape mirrors its `Tiling`:
...
+A merge **extends** a collection, so a key it already holds is a group that grew rather than a
+second key. The two sides' key runs are matched per row and the matched groups merged
+(`merged_rows`), so the level below gains the new elements and nothing is appended and then
+repaired. A record between the two levels moves with the keys it stands under, its collection
+fields by their runs and its scalar fields by holding one value per key. A regrown key that
+re-states a scalar has delivered a position twice, which the release contract forbids and
+`merged_column` reports. A key grows only where what it holds can (`Tile::holds_a_level`):
+under a scalar a repeated key stays a repeat, for `validate_tile` to report.
+
+Tiles representing collections (`Function`) support logical deletes by storing a `BitSet` of deleted values.  These are set by filteriing operator like `Restrict` and compacted away by
```
sortalongo: 184 columns on a line this PR edits; rewrap at 100. It also carries a pre-existing typo worth taking while you are here: "filteriing operator like" should be "filtering operators like". Line 82 calls `row_starts` "offsets", which is the deleted field's name, and line 133 names a "regrown-key fixup" that line 69 says does not exist.

src/ccl/aggregate.rs
```diff
@@ -82,7 +94,17 @@ impl AggregateKind {
...
+        if self.is_partial() {
+            return accumulator.empty_at_no_rows();
+        }
+        Tile::Scalar(self.scalar_seed(&accumulator.extent()))
+    }
+
+    fn scalar_seed(&self, accumulator_extent: &Extent) -> ColumnValue {
         match (self, accumulator_extent) {
             (AggregateKind::Sum, Extent::Base(BaseType::Int)) => ColumnValue::Ints(vec![0]),
             (AggregateKind::Max, Extent::Base(BaseType::Int)) => ColumnValue::Ints(vec![i64::MIN]),
```
sortalongo: `Max` seeds `i64::MIN` and reports itself total, so an emptied group answers with the sentinel. Over `UInt` the seed is `0` and over `String` it is the empty string, both legitimate values, so nothing distinguishes an empty group from a real answer.

Marking `Max` partial is not the fix: presence for a partial fold is carried in accumulator length, extraction emits a key only where its terminal flag is set, so the key would vanish instead. That is the bug this PR exists to fix.

The root problem is that `is_partial` answers two questions with one boolean. Its own doc says it decides both whether the fold can fault and whether presence is out of band. `max` splits them: its merge never faults, but it has no valid empty answer. The agreed direction is that every aggregate returns an option with `none` for empty, and `or_else` at the call site chooses. The seed predates this PR, so this is not yours to fix here, but the PR is what makes it reachable.

src/interpreter/operator_conversion.rs
```diff
@@ -1495,6 +1496,11 @@ fn convert_impl_inner(
             if as_builtin(function) == Some(Builtin::Zip) =>
         {
             let input = expect_input(input, "zip")?;
+            // The arms are each applied to `input`, so the levels they run over are its —
+            // which is what the pair sits under. Their own shapes cannot say: two arms that
+            // agree below the ambient (two projections of one grouped row, say) look
+            // pairable all the way down.
+            let ambient = level_count(input.tiling());
```
sortalongo: The arms are what get paired, and an arm that folds a level has fewer levels than the input, so this over-counts. A tuple of two aggregates over a group-by trips `FanIn::new_at`. Both arms were the one-level variant before, which is why it held. Cap the input's count by the arms' own counts.

sortalongo (approved):
Approving the direction. Depth as data rather than baked into the variant is the right call, and the unit tests are good — three levels, a record between levels, and per-level deletion are all covered. The naming pass in the last commit is a real improvement.

Approving so this stops blocking the fifteen PRs above it, not because the list below is optional. The four correctness items should land before merge.

**Before merge**

1. An emptied group under `max` reports `i64::MIN`, and under `UInt` reports `0`, which is indistinguishable from a real maximum. Keeping the key is what makes the seed reachable as an answer. See the aggregate comment below.
2. A release-path panic in `retain_rows` on a shape `valid_over` blesses.
3. `gathered_rows` combines aggregations through the whole-value fold, which returns the wrong rows.
4. The zip site reads the fan-in depth off the input rather than the arms.

**Decision to record.** Every aggregate returns an option, with `none` for the empty collection, and the caller chooses the empty behaviour explicitly with `or_else`. This covers `sum` as well: zero is sometimes the answer and sometimes not, and the option forces the choice. It also fixes `sole`, where an empty group currently makes the key vanish rather than reporting anything. I am writing this up as a Direction note under "7.1 Aggregates" in `docs/chl-spec.md`. Note `or_else` does not exist yet anywhere in the spec or the implementation, and `sole` and `drain` have no surface spelling today.

**Naming.** `Tile::Function` should be `DataFunction`. A collection level is a data function in the type vocabulary, so `Function` collides with the compute arrow.

**Documentation.** The prose says the same things many times over. Twenty percent of the added lines are comments, and `tile.rs` carries 472 of them. The worst case is stated eight times. Details inline; the pattern is to keep one statement where the rule lives and cite it from the rest. `tiling_kind.rs` already does this correctly and is the model.

**Status.** The head has no CI run — the last run on this branch predates the final commit — and the branch now conflicts with main. Both want fixing before this lands regardless of review.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+            // A record's fields stand over the rows the record does, so what it holds is
+            // what its fields hold — each read the same way, because a `Codomain` arm is
+            // read against every row whatever sits at it.
+            Tile::Record(fields) => TileGuard::Record(
+                fields
+                    .iter()
+                    .map(|(name, field)| (name.clone(), field.held_guard()))
+                    .collect(),
+            ),
+            Tile::Function {
```
groundlar: Below the top level this names every key the level holds and never reads that level's `domain_predicate`, so it releases a middle key whose group is still open. `to_guard` applies "a key is released once what it holds is complete" at the top only; `held_guard` drops the rule at every level beneath.

Repro: `0 ⤇ 10 ⤇ 100 ↦ 1`, every predicate `False`. `to_guard()` is `Codomain(Or[Domain({10}), Codomain(Domain({100}))])`, and a later delivery `0 ⤇ 10 ⤇ 101 ↦ 2` is `contains_guarded` by it. After `remove_guarded` + `compact`, key 10 is gone and `101` with it. Debug trips the obsolete-guard assert in `tile_operators/mod.rs:441`; release drops the element.

```rust
let chain = |c, v| Tile::function(ColumnValue::from_uints(vec![0]), Box::new(Tile::grouped(
    ColumnValue::UInts(vec![0]), ColumnValue::from_uints(vec![10]), Box::new(Tile::grouped(
        ColumnValue::UInts(vec![0]), ColumnValue::from_uints(vec![c]),
        Box::new(Tile::Scalar(ColumnValue::Ints(vec![v]))), Predicate::False, BitSet::new())),
    Predicate::False, BitSet::new())), Predicate::False, BitSet::new());
assert!(!chain(101, 2).contains_guarded(&chain(100, 1).to_guard()));
```

Name a middle level's keys only where its predicate calls them whole, and name an open key's contents through `Codomain`, as `to_guard` does at the top.

src/interpreter/tiling/tile.rs
```diff
@@ -664,32 +647,49 @@ impl Tile {
...
+                // keys would release the same key value under a sibling still growing.
+                let open = BitVec::from_fn(domain.len(), |key| {
+                    !domain_predicate.contains(&domain.index_at(key))
+                });
+                if !open.any() {
+                    return TileGuard::Function(FunctionGuard::Domain(domain_predicate.clone()));
+                }
+                let mut open_values = (**codomain).clone();
+                open_values.retain_rows(&open);
+                TileGuard::flatten_or(vec![
```
groundlar: Every nested tile with a partial outer predicate now yields `Or[Domain(𝑝), Codomain(…)]`, and `Or & Or` distributes (`guard.rs:120`) into a `Domain & Codomain` pair, which `FunctionGuard::intersect` answers with `todo!()` (`guard.rs:265`). `FanOutBranch::release_impl` intersects its live subscribers' guards, so two consumers of one incrementally delivered nested collection panic on release.

`g.intersect(&g)` on the three-level fixture from the `held_guard` comment reproduces it. `main` had the same `todo!` for `CurriedFunction` only; this widens it to every nested shape and every record between levels. Returning the empty guard for the mixed pair is sound, since releasing less is always safe.

src/interpreter/tiling/tile.rs
```diff
@@ -704,93 +704,1442 @@ impl Tile {
...
+    /// a key the producer promises is releasable even before it arrives. A
+    /// [`FunctionGuard::Codomain`] arm cannot carry that, because it is read against every
+    /// row of the values it wraps — a promised key would be released under a row that
+    /// already holds it, and that row's value would go with it.
+    fn held_guard(&self) -> TileGuard {
+        match self {
+            // A record's fields stand over the rows the record does, so what it holds is
+            // what its fields hold — each read the same way, because a `Codomain` arm is
+            // read against every row whatever sits at it.
+            Tile::Record(fields) => TileGuard::Record(
```
groundlar: A `Codomain` guard is read against every row of the values it wraps, present and future, so what it carries must discriminate by key. A scalar field's `Scalar(true)` does not: for `K ⤇ {n: Int, xs: J ⤇ Int}` with open keys, the guard releases `n` for keys that have not arrived yet.

The representation has the matching gap. A column is empty or one entry per row, so a delivery that regrows key 0 (no `n`) and adds key 1 (with its `n`) cannot be built: `record_between_levels(&[(0, None, &[(200, 20)]), (1, Some(8), &[(300, 30)])], …)` fails `Tile::grouped`'s validation at line 741. `a_record_between_two_levels_holds_its_keys` pins `n: Scalar(true)` as intended.

No CHL program reaches this yet, but the doc's "a scalar field beside a still-growing collection holds the key" does not survive a release. Either a record's scalar fields stay out of the codomain arm until their key is whole, or the column needs per-row presence.

src/interpreter/tile_operators/fan.rs
```diff
@@ -304,41 +232,31 @@ impl TileProducer for FanInProducer {
...
-                        output_domain = Some(domain);
-                    }
-                    codomains.push(*codomain);
+                    // Lift the values out, leaving the chain of keys behind: what stays is
+                    // the output's own shape, and every input has to agree on it.
+                    codomains.push(std::mem::replace(
+                        filtered.values_at_mut(self.depth),
+                        Tile::Record(HashMap::new()),
+                    ));
+                    debug_assert!(
```
groundlar: Alignment below the top level is now a `debug_assert!`, where the curried branch hard-asserted `domain1`, `offsets` and `domain2`. Presence is intersected on the outermost keys only (lines 207–245), so a mismatch at an inner level passes in release and the output takes input 0's layout with input 1's values under it — record fields at different row counts.

That mismatch is reachable in principle: `MapResult`'s withhold path (`map.rs:390-436`) now runs on nested inputs and drops innermost keys one by one, so one arm can trail the other at the inner level (`[[(x, f(x)) for x in v] for v in g]` with `f` still streaming). Reasoned from the code, not reproduced. Intersect presence at every level up to `depth`, or keep the hard assert.

src/interpreter/tile_operators/lookup.rs
```diff
@@ -92,7 +92,7 @@ impl CheckedLookup {
             .get(&tuple_field(0))
             .ok_or_else(|| "`lookup?`'s input rows have no collection field".to_string())?;
         match collection {
-            Tiling::SealedFunction { .. } | Tiling::Scalar(Extent::Function { .. }) => {}
+            Tiling::Function { .. } | Tiling::Scalar(Extent::Function { .. }) => {}
```
groundlar: `Tiling::Function { .. }` now matches a collection of collections, so the `Err` below — which says exactly that shape is unsupported — no longer fires, and the failure moves to the runtime `assert!(!codomain.holds_a_level())` in `answer_for`. `[[or_zero(m[r.b]?) for r in v] for v in g]` panics on the first `get` rather than at conversion.

The same construction-to-runtime shift, from `SealedFunction` patterns becoming `Function { .. }`, is in `answer_tiling` (128–136), `VariantWrap::new` (`scalar.rs:203` builds a one-level tiling while `get_impl` keeps every level), `ExtractFinal`'s `source_codomain_tiling` (`extract_final.rs:90`), `MapDomain`/`Restrict`, the `union.rs` constructors, and `MapFilter::new` (`combinators.rs:736`, whose doc says two levels and whose check admits one). A `!codomain.holds_a_level()` guard or an exact depth check in each `new` restores the old rejection.

src/interpreter/tiling/tile.rs
```diff
@@ -24,56 +24,50 @@ pub enum Tile {
...
-        domain2: ColumnValue,
-        /// Flattened codomain values; supports vectorized transformations.
-        codomain: ColumnValue,
-        /// The region of `domain1` for which no new elements will ever be seen — the same
-        /// statement [`Self::SealedFunction`]'s makes, and covering each key in the region
-        /// together with its whole `domain2`/`codomain` list.
+        /// The region of `domain` whose values are complete: no new element will be seen under
+        /// them, at any depth. Covering every key says the key set itself is closed, which is
+        /// what makes [`Tile::is_terminal`] read this as one flag.
         domain_predicate: Predicate,
```
groundlar: What this means at an inner level is unwritten, and producers disagree: `nest_levels` writes `False`, `MapResult`'s nested branch (`map.rs:376`) and `Converse` (`combinators.rs:124`) write `True` even for a partial group, `regroup_rows` copies the node's own, and `FanIn` keeps input 0's. Nothing reads an inner predicate today (`to_guard` goes through `held_guard`; `is_terminal` and aggregation read the top), so there is no live bug. An inner level's keys repeat across rows, so a predicate over key values cannot name one row's group. Define it per row or drop it below the top.

src/interpreter/design-operators.md
```diff
@@ -57,17 +56,33 @@ A `Tile` holds the actual data. Its shape mirrors its `Tiling`:
...
+stateful operators like `Memo` and aggregation. A `Function` carries **one set per domain
+level**, so a bit names a position in that level's own column and a removed group and a removed
+entry are different bits rather than one bit read two ways. Every producer today removes at the
+innermost level, a group going by way of its entries, which is where `Tile::retain_rows` reads.
+
+An empty group and a removed one are different tiles. The offsets are non-decreasing, so two equal
+starts are a key whose group holds nothing, and an aggregate over it folds to the aggregate's
+identity; a removed key is marked in `deleted` at its own level and taken, with its subtree, by
+`Tile::compact`. `Tile::retain_rows` produces the first and never the second.
```
groundlar: `retain_rows` drops whole rows; `retain_keys` is what leaves a key with an empty group. Line 80 has the same mix-up.

Other leftovers in the same vein:
- `fan_in` is named at lines 176, 255, 256 and 454; the function is `fan_in_at`.
- `helpers.rs:49-56`: the `apply_function_tile` doc lists `Tile::Function` twice, and "`Function` (which always produces `ColumnValue::Variants`)" describes no current behaviour.
- The retired names survive in `map.rs` (`f_domain2`, `f_domain2_extent`, `new_domain2`), the `to_guard` doc ("For Collection:"), `docs/mutability-walkthrough.md`'s table header (`Collection`, a third name beside `Function` and the proposed `DataFunction`), and `tests/programs/common/mod.rs:279`, which now says `Function` tiles panic one line after documenting that they are handled — it means multi-level ones.

src/interpreter/tiling/tiling_kind.rs
```diff
@@ -228,14 +267,17 @@ impl fmt::Display for Tiling {
...
+                write!(f, "Fn({domain:?} → ")?;
+                let mut node = codomain.as_ref();
+                while let Tiling::Function { domain, codomain } = node {
+                    write!(f, "{domain:?} → ")?;
+                    node = codomain;
+                }
+                write!(f, "{node})")
             }
             Tiling::Aggregation { kind, accumulator } => {
                 write!(f, "agg({kind:?}, {accumulator:?})")
```
groundlar: `{accumulator:?}` Debug-dumps the boxed `Tiling`; `{accumulator}` matches the rest of this `Display`.

src/interpreter/tile_operators/scalar.rs
```diff
@@ -896,17 +912,17 @@ mod tests {
...
         else {
-            panic!("expected SealedFunction, got {out:?}");
+            panic!("expected Function, got {out:?}");
         };
         assert_eq!(domain, ColumnValue::from_uints(vec![]));
         assert_eq!(
             *codomain,
             Tile::Scalar(ColumnValue::Ints(vec![])),
-            "the empty codomain is shaped by the declared arm extent (Int), not \
+            "the empty values is shaped by the declared arm extent (Int), not \
```
groundlar: The rename left this message ungrammatical ("the empty values is shaped"); `panic!("expected a Record values, …")` at line 1075 is the same.

src/interpreter/tiling/tile.rs
groundlar: The other half of this arm. `holds_a_level` looks through a record, so for `K ⤇ {agg: Aggregation, xs: L ⤇ Int}` the keys match and `Both` rows reach the aggregation at any key count, where `assert_eq!(order.len(), 1)` at line 1172 fires. Its message, "reaches here only at the whole value", is false.

Confirmed with two keys, both regrown by the right side: `assertion left == right failed … left: 2 right: 1`. With one regrown key it passes by coincidence. Selecting into the concatenation fixes the no-overlap path but not this one; a `Both` row needs the aggregate's `accumulate` at that row.

groundlar (approved):

src/interpreter/tiling/tile.rs
dpmills: Renamed to `Tile::DataFunction` and `Tiling::DataFunction`, with `data_function` and `is_data_function`.

src/interpreter/tiling/tile.rs
dpmills: Fixed: `retain_rows` and `select_rows` leave an empty column as it is.

src/interpreter/tiling/tile.rs
dpmills: Fixed both halves. Under a collection an aggregation's rows append rather than fold (`merge_part` now respects `whole`), and `gathered_rows` picks from the concatenation in `order`'s sequence. A row both sides deliver, groundlar's case, folds the two contributions by the aggregate's `accumulate`, as a whole aggregation's halves do. Covered by `merging_new_keys_over_aggregations_keeps_one_accumulator_each`, `merging_interleaved_rows_over_aggregations_keeps_each_with_its_key` and `a_regrown_key_folds_the_aggregation_beside_it`.

src/interpreter/tiling/tile.rs
dpmills: Fixed: where no key can match (rows under a collection, or a whole level whose values hold no level) `merge_part` appends in place (`append_keys`), so a delivery costs its own size. A whole level whose values hold a level still matches keys.

src/interpreter/tiling/tile.rs
dpmills: Done: the rule and the `MapAggregate` example live in `holds_a_level`'s doc, and the other sites cite it.

src/interpreter/tiling/tile.rs
dpmills: Done: the rule lives in `Tile::merge`'s doc, and the other sites cite it.

src/interpreter/tiling/tile.rs
dpmills: Done: the `row_starts` field states the empty-group rule and `Tile::DataFunction` states the vectorized-over-rows invariant; the other sites cite them, and `Tiling::DataFunction` no longer repeats the line.

src/interpreter/tiling/tile.rs
dpmills: Reworded all four to describe the key matching `merge` does.

src/interpreter/tiling/tile.rs
dpmills: Fixed.

src/interpreter/design-operators.md
dpmills: Rewrapped, fixed the typo, `offsets` is now `row_starts`, and the fixup sentence is gone.

src/ccl/aggregate.rs
dpmills: Left as is here; the fix is the option-returning aggregates in your Direction note.

src/interpreter/operator_conversion.rs
dpmills: Fixed here by capping the level at the arms' own counts; `a_tuple_of_folds_pairs_at_the_group` covers two aggregates over a group-by. The cap still over-counts where every arm keeps a level, so #239 replaces it with a level the AST around the zip sets (`OpConversionContext::level`).

src/interpreter/tiling/tile.rs
dpmills: Fixed in #241: a level whose values hold a level names no keys of its own, and a held key is named by its whole path.

src/interpreter/tiling/tile.rs
dpmills: #241 implements the meet, qualifying the codomain's predicates by the domain's keys (`Predicate::Qualified`), and the union as a `TileGuard::Or`, rather than answering the empty guard.

src/interpreter/tiling/tile.rs
dpmills: Fixed in #241: a record's scalar field is released with its key rather than named in the `Codomain` arm.

src/interpreter/tile_operators/fan.rs
dpmills: The check is a hard `assert!` again here. #239 aligns the arms over whole paths at every level above the pair.

src/interpreter/tile_operators/lookup.rs
dpmills: Restored the rejection in `CheckedLookup::paired`, `Restrict::new` and `MapFilter::new` (exactly two levels), and fixed `MapFilter`'s two runtime messages, which said the opposite of what they check. The others stay open: PRs above build `VariantWrap`, `ExtractFinal` and the union constructors over nested inputs, `MapDomain` already handles one, and #198 widens `answer_tiling` on purpose.

src/interpreter/tiling/tile.rs
dpmills: #241 defines it per path: a level states completeness beneath the path that reaches it (`Predicate::Qualified`).

src/interpreter/design-operators.md
dpmills: Fixed: `retain_keys` at both lines, `fan_in_at` throughout, the `apply_function_tile` doc, the `map.rs` names, the `to_guard` doc, the walkthrough's table header and `common/mod.rs`.

src/interpreter/tiling/tiling_kind.rs
dpmills: Fixed; the fixtures render `agg(Sum, Int)` again.

src/interpreter/tile_operators/scalar.rs
dpmills: Fixed both.

dpmills:
Of the four before-merge items, the release-path panic, the aggregation merge and the zip depth are fixed (see the threads); the `Max` seed stays for the option-returning aggregates. The branch is rebased onto `main`, and CI is green on the head.

PR #211, opened 2026-09-15
Share one domain-level append between `Product` and `IterateRowCollection`

dpmills:
`Product` and `IterateRowCollection` each build a `CurriedFunction` one level deeper than the function tile they are given, and each spelled the step out itself: read a `SealedFunction` as the one-level case, push the new keys and starts, replace the codomain, and carry `deleted` and the `domain_predicate` through. `Tile::append_level` names that step, and `Tiling::append_level` is the same step on the static shape; both operators now build on them. The invariants the step rests on — a group is whole when it is appended, removals stay at the level that named them — are stated once on the method rather than twice in operator comments.

`append_level` takes a closure returning a `DomainLevel` (`starts`, `keys`, `codomain`) built from the codomain the tile arrives with, because the new level is made out of that codomain: `Product` repeats each row's value across the group it opens, and `IterateRowCollection` opens the row's own bindings. `Product`'s case for an inner domain not yet delivered moves out to `unpaired_rows`. The `Deleted::at_level(0, …)` promotion and the `domain_predicate` union are the ones both operators already performed; `Product`'s row count now comes from the codomain column rather than the innermost domain, which the curried tile's vectorization makes equal.

Seven unit tests cover the two methods, and the level append is written up under [`design-operators.md`, "Tile"](src/interpreter/design-operators.md#tile).


dpmills:
Closing: `Tile::append_level` and `Tiling::append_level` now land in #197, which introduces `IterateRowCollection` as their first caller, and #198 adds `Product` as the second. Writing each producer against the shared method from the start removes the duplicate-then-extract step this PR existed to undo — nothing here is lost.

PR #214, opened 2026-09-16
Run each CI gate as its own job, and cache the Lean build

dpmills:
A run took the sum of its gates because they shared one job, and the Lean gate
rebuilt from scratch every time because nothing cached it.

Each gate becomes a job. `check` — the context the "CI" ruleset requires —
now only collects their results, so splitting, renaming, or adding a gate needs
no ruleset edit. Shared Rust setup moves to a local composite action rather
than being copied into four jobs; its rust-cache saves only from main, because
rust-cache keys on the job id and a saving branch would store one entry per
gate and start evicting at the repo cache limit.

Two caches cover the Lean model. `elan-init` records the toolchain pin and
fetches nothing, so elan downloaded ~150MB inside the gate itself; that is now
keyed on the pin. `formal/.lake` is keyed on every input the build has, with
a restore-keys prefix holding the toolchain fixed so editing one theorem reuses
the other modules through lake's own per-module traces.

The `Clean rustpython-parser cache` step goes: `rustpython` left Cargo.lock
with the custom parser, so it cleaned a package that no longer exists.


.github/workflows/ci.yml
```diff
@@ -21,14 +21,30 @@ permissions:
...
 jobs:
-  check:
+  # The checks that need no build toolchain, plus the code-changed flag every
+  # other job gates on. Nothing here depends on the flag: a docs-only PR is
+  # exactly when a section anchor rots, so the doc-reference check has to run
+  # on it.
+  repo_checks:
     runs-on: ubuntu-latest
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
```
groundlar: nit: clearer as something like `is_code_change` since it's no longer consumed as part of the `filter` step

groundlar (approved):

.github/workflows/ci.yml
dpmills: Renamed to `is_code_change` in d6d83a9. The paths-filter `code:` key keeps its name, since that is what defines `steps.filter.outputs.code`; only the job output that the seven gates read is renamed.

PR #218, opened 2026-09-16
Make node ids independent of `debug_assertions`, and gate that configuration in CI

dpmills:
Two debug-only checks minted node ids, and `NodeId` is a process-global mint counter, so the same program took different ids in a debug build and a release one. The inspector's golden fixtures (committed `/api/snapshot` payloads under `web/src/__fixtures__/`) record those ids, so the corpus was debug-only: 17 of the 29 programs in `tests/programs/` dumped differently under the two builds, now 0. `provenance::run_debug_check` seals the provenance record around a check — a clone keeps its origin's id, a construction takes `NodeId::PLACEHOLDER` — under the precondition its doc comment states. `assert_reads_stable` and `verify_narrowing_is_complete` run under it, each resolving types; `lambda_elim`'s pre-elimination render builds a single node and takes `throwaway` directly.

`cargo test --release --test inspector_goldens committed_fixtures_match_fresh_dumps`, before:

```
fixture drift: arithmetic.snapshot.json differs from a fresh dump of tests/programs/udf_closure/program.cambra. First difference at $.paneLinks[1].edges[0][1]:
  committed: 203
  fresh:     182
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out
```

and after:

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 11 filtered out
```

## The new gate

Nothing ran the suite with `debug_assertions` off — `ci_test` runs it on, and `ci_clippy_release` lints that configuration without running it. `ci.sh test_no_assertions` runs it, under the new `profile.no-assertions` (`dev` codegen with the flag off): 18s against the release profile's 91s. Its CI job keys its own `rust-cache` entry, `debug-assertions` being a codegen flag rather than a runtime one. Without the sealing half the gate fails on `committed_fixtures_match_fresh_dumps`, which is what couples the two halves.

## Fixtures

Re-blessed in one class, id renumbering: the id map is bijective at every id-bearing wire position (`node_id`, `root`, `subscribed`, `id`, and the `edges` pairs, per `src/inspector_model/wire.rs`) and no id increases.


src/ccl/provenance.rs
```diff
@@ -2018,6 +2018,51 @@ pub(crate) fn preserving_ids<R>(f: impl FnOnce() -> R) -> R {
...
+/// `assert_unique_node_ids` backstops a placeholder that escapes regardless.
+///
+/// A check that only reads needs no scope and loses nothing by having one. Reach
+/// for this wherever the check *resolves* a type: resolution forces suspended
+/// substitutions at each refinement leaf, which rebuilds the predicate term.
+///
+/// [`TypedExpr::clone_preserving_ids`]: crate::ccl::expr::TypedExpr::clone_preserving_ids
+/// [`TypedExpr::throwaway`]: crate::ccl::expr::TypedExpr::throwaway
+#[cfg(debug_assertions)]
+pub(crate) fn run_debug_check<R>(f: impl FnOnce() -> R) -> R {
```
sortalongo: `SEALED` is decremented by hand after `f()`, so a panic inside `f` leaves it raised. `f` is an assertion, so panicking is its failure mode, not an edge case. If the unwind is caught, every later `TypedExpr::new` on that thread silently returns a `PLACEHOLDER` throwaway. `preserve_ids()` on the line above already returns a guard that unwinds correctly. Give `SEALED` the same guard treatment.

src/ccl/provenance.rs
```diff
@@ -2018,6 +2018,51 @@ pub(crate) fn preserving_ids<R>(f: impl FnOnce() -> R) -> R {
     f()
 }
 
+thread_local! {
+    /// Depth of the enclosing [`run_debug_check`] scopes.
+    static SEALED: std::cell::Cell<u32> = const { std::cell::Cell::new(0) };
```
sortalongo: `run_debug_check` is `#[cfg(debug_assertions)]` but `SEALED` and `record_sealed` are not, so nothing can raise the flag in a release build while `TypedExpr::new` still reads a thread-local on every node it constructs. Gate the pair and let `record_sealed` be a `const false` in release so the branch folds away.

src/ccl/infer/solve.rs
```diff
@@ -906,9 +906,13 @@ pub(super) fn coalesce_pass(expr: &mut Expr) -> Vec<LocatedInferError> {
     // With the whole graph in its final state, re-check every read the walk
     // performed (skipped on the error path — a failed program's reads are
     // not expected to be stable).
+    // Under `preserving_ids` because the re-resolution rebuilds predicate terms it
+    // then drops: minting for them would advance the process-global id counter in
+    // debug builds alone (see `provenance::preserving_ids`, "Containment: a
```
sortalongo: The cited heading does not exist. `preserving_ids` has no "Containment: a discarded diagnostic region" section, and the string appears nowhere in the repo. Because the citation names a Rust item rather than a `.md` file, `doc_refs` does not check it, so it rots silently. State the fact inline instead.

.github/workflows/ci.yml
```diff
@@ -182,6 +182,23 @@ jobs:
...
+  test_no_assertions:
+    needs: repo_checks
+    if: ${{ !cancelled() && needs.repo_checks.outputs.is_code_change == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: ./.github/actions/rust
+
+      - name: Test (no debug assertions)
+        run: ./ci.sh test_no_assertions
```
sortalongo: The `test` job runs under `CAMBRA_PROVENANCE_GATE=1` and this one does not, so the no-assertions configuration is checked less than the debug one. Deliberate for cost, or an oversight?

sortalongo (approved):
Approving. Sealing the record around a debug check is the right shape: it puts the rule at the one place that mints rather than at each check, and the new gate is coupled to the fix because it fails without the sealing half.

I verified the fixture claim independently rather than trusting the diff. Walking the five snapshot pairs in lockstep: structure identical at every position, every non-id scalar unchanged including the `type` strings and span offsets, and the id map injective, strictly monotone, and non-increasing in each file. Distinct id counts match per file (149/149, 127/127, 115/115, 107/107, 76/76). All 1685 differing lines differ only in an integer literal. Ids are not consistent across files, which is expected, since each program is dumped in its own process. The re-bless is pure renumbering.

Four findings, none blocking.

.github/workflows/ci.yml
dpmills: Deliberate for cost.  The overlap doesn't seem important, and the extra build time fits into the longer runtime of the provenance gate tests so this doesn't increase overall CI runtime

src/ccl/provenance.rs
dpmills: Done in 4945000. `SEALED` now enters through a guard, and the guard is shared with `PRESERVING_IDS`: `ScopeDepth` is the trait on `LocalKey<Cell<usize>>` that both counters enter through, so one `Drop` lowers both on an unwind. The duplication was the shape that produced this bug, so collapsing it seemed better than writing a second guard beside the first.

`panic_inside_a_debug_check_unwinds_without_leaving_the_record_sealed` pins it — the mint inside `f` takes `PLACEHOLDER`, and the mint after the caught unwind does not. It fails against the hand-written decrement.

src/ccl/provenance.rs
dpmills: Done in 4945000, as described: `SEALED` and the `debug_assertions` arm of `record_sealed` are gated, and the other arm is a `const fn` returning `false`, so `TypedExpr::new`'s throwaway branch folds away with assertions off.

src/ccl/infer/solve.rs
dpmills: Done in 4945000. The comment states it inline and names `run_debug_check`, which is what the call site uses — the old text said `preserving_ids` as well as citing a heading that was never written.

The sibling call site in `infer/mod.rs` had the same defect in the other direction: it called its check "a pure read of the graph" beside a seal that is there precisely because `resolve_var_type` rebuilds predicate terms. Fixed in the same commit.

PR #230, opened 2026-09-18
Compile a nested for accumulator by folding the inner loop to a sum

dpmills:
A `for` nested inside another was rejected at lowering whenever the inner body wrote a mutable accumulator, leaving the whole class unspellable in CHL — summing a list of lists, an inner loop correlated with the outer element. Lowering now rewrites an inner `for` that only accumulates into the accumulation it denotes, `for y in s: total += 𝑒` becoming `total += sum([𝑒 for y in s])`, which the comprehension path already compiles. `total := 0; for xs in [[1, 2], [3, 4]]: for x in xs: total += x` previously failed to compile and now answers 10.

## The fold

`fold_inner_accumulation_loops` (`src/ccl/lower/loops.rs`) runs over a `for` body before the accumulator scan, at both entry points in `src/ccl/lower/stmts.rs`, so the enclosing loop sees an ordinary top-level write. It recurses innermost-first, which makes nesting unbounded, and a single-branch `if` around the write contributes its condition as a comprehension guard. Reasoning: the function's doc comment, and `src/interpreter/design-operators.md`, "A correlated inner comprehension" for the outer/inner pairing.

## What still refuses

`*=`, `//=` and `:=` keep their `for` and meet the existing reject, no aggregate stating their law; a body reading the accumulator it writes is a scan, rejected by its own message. The reject's text changes from

> mutation of `total` is nested inside an inner `for` in this for-loop body; nested-loop mutation is not yet supported (a conditional `if p: total += …` write is supported — only an inner `for` is not).  Move the mutation to the outer loop body, or rewrite using a generator expression.

to

> `total` is written inside an inner `for` by an operation no aggregate states.  An inner loop accumulating with `+=` or `-=` folds to a `sum`; `*=`, `//=` and `:=` do not.  Rewrite the inner loop as a generator expression, or move the mutation to the outer loop body.

## Collection-valued source elements

`decode_source_positioned` (`src/interpreter/commit_operator.rs`) decoded a writer's source column through a local `source_value_at` that handled scalars and records and yielded no pairs for anything else, so a loop whose elements are collections ran zero iterations and answered 0, and an element record holding a collection panicked. It now calls the shared `materialize_collections`, and `source_value_at` is deleted. Both defects predate this change and reproduce with no inner `for` at all.

## Tests and docs

`tests/compilation_pipeline/mutability.rs` covers six nested shapes and the three rejects; the suite passes 3148 tests. A separate test is `#[ignore]`d for another pre-existing defect: a binding declared between the two loops mis-states `Zip`'s ambient depth, writing the fold by hand asserts identically, and inlining the binding answers 66. `docs/chl-spec.md` states the rule and the example under "4.6 `for` — iteration", and narrows "12. Reserved for future work" to the shapes that stay refused.


dpmills:
Superseded: folding an inner `for` into the accumulation it denotes is not the approach we're taking. Compiling a nested `for` as a nested recurrence — a recurrence per position of the loop around it — replaces it, and handles the general case rather than the subset where the inner loop only accumulates.

The branch stays as a record of the attempt. `fold_inner_accumulation_loops` exists nowhere else: not on main, not in the branches below this one.

